PR #4, opened 2021-07-30
Fix to address the issues in using params from group_vars/inventory

praveenramoorthy:
Fix to address the issues in using params from group_vars/inventory files.

- Password and Username by default will be from global values, it will be overridden if username and password is specified in particular task.
- Handled an exception
- Removed the default ("true") for "validate_certs". It is already set to true through inheritance from uri module. Setting it to true by default at nd, makes it part of params and this in-turn makes it always override the global setting(ansible_httpapi_validate_certs). This makes users to set it false explicitly for each task in playbook.

### DIFF
--- a/plugins/module_utils/nd.py
+++ b/plugins/module_utils/nd.py
@@ -90,8 +90,8 @@ def nd_argument_spec():
         password=dict(type='str', required=False, no_log=True, fallback=(env_fallback, ['ND_PASSWORD', 'ANSIBLE_NET_PASSWORD'])),
         output_level=dict(type='str', default='normal', choices=['debug', 'info', 'normal'], fallback=(env_fallback, ['ND_OUTPUT_LEVEL'])),
         timeout=dict(type='int', default=30, fallback=(env_fallback, ['ND_TIMEOUT'])),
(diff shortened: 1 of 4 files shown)


plugins/module_utils/nd.py
```diff
@@ -92,7 +92,7 @@ def nd_argument_spec():
         timeout=dict(type='int', default=30, fallback=(env_fallback, ['ND_TIMEOUT'])),
         use_proxy=dict(type='bool', default=True, fallback=(env_fallback, ['ND_USE_PROXY'])),
         use_ssl=dict(type='bool', default=True, fallback=(env_fallback, ['ND_USE_SSL'])),
-        validate_certs=dict(type='bool', default=True, fallback=(env_fallback, ['ND_VALIDATE_CERTS'])),
+        validate_certs=dict(type='bool', fallback=(env_fallback, ['ND_VALIDATE_CERTS'])),
```
lhercot, line 95: Why disable cert verification by default?

lhercot (changes requested):

lhercot (approved):
LGTM

plugins/module_utils/nd.py
lhercot: Understood. We applied this to other attributes too so we can address all of it.

PR #16, opened 2022-04-05
[bugfix] allow 'local' to be specified as login domain

xinyuezhao:
- remove condition self.params.get('login_domain') != 'local' to allow 'local' being set as login domain. 
- set DefaultAuth as default value for login_domain


lhercot (approved):
LGTM

PR #41, opened 2023-03-15
[bugfix] ensure python2 compatible methods are used for the bytes io …

akinross:
…object

### DIFF
--- a/plugins/httpapi/nd.py
+++ b/plugins/httpapi/nd.py
@@ -238,7 +238,7 @@ class HttpApi(HttpApiBase):
                             remote_path="/temp",
                         )
         """
-        if not data:
+        if data is None:
             data = dict()
 
         self.error = None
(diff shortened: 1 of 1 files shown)


akinross:
fix proposal for issue raised on ansible-mso repository: https://github.com/CiscoDevNet/ansible-mso/issues/336

shrsr (approved):
LGTM

plugins/httpapi/nd.py
```diff
@@ -261,7 +269,24 @@ def send_file_request(self, method, path, file=None, data=dict(), remote_path=No
                 self.nd.fail_json(msg="Cannot use requests_toolbelt MultipartEncoder() because requests_toolbelt module is not available")
             mp_encoder = MultipartEncoder(fields=fields)
             multiheader = {"Content-Type": mp_encoder.content_type, "Accept": "*/*", "Accept-Encoding": "gzip, deflate, br"}
-            response, rdata = self.connection.send(path, mp_encoder.to_string(), method=method, headers=multiheader)
+
+            if sys.version <= "3":
```
akinross, line 273: I think we should avoid using this conditional because the python documentation states the following:

sys.version[¶](https://docs.python.org/2/library/sys.html#sys.version)
A string containing the version number of the Python interpreter plus additional information on the build number and compiler used. This string is displayed when the interactive interpreter is started. Do not extract version information out of it, rather, use [version_info](https://docs.python.org/2/library/sys.html#sys.version_info) and the functions provided by the [platform](https://docs.python.org/2/library/platform.html#module-platform) module.

So would propose to use
```
if sys.version_info.major == 2

**Review by akinross — COMMENTED** [review]
see comment

File: plugins/httpapi/nd.py [hunk 2]
```diff
@@ -261,7 +269,24 @@ def send_file_request(self, method, path, file=None, data=dict(), remote_path=No
                 self.nd.fail_json(msg="Cannot use requests_toolbelt MultipartEncoder() because requests_toolbelt module is not available")
```
akinross, line 269: This error needs to be raised differently because `nd` is not part of the connection plugin. What type of error should we raise here?

lhercot (approved):
LGTM

PR #43, opened 2023-03-17
Added default encoding - lantin-1 to send_file_request method

sajagana:
Note:

Added default encoding - lantin-1 to send_file_request method to handle NDO backup upload.


lhercot:
I think we need to revisit this because I think there might be a case that one of the file we upload (in another service than backup) is not a latin-1 encoded file. This would break that part.

lhercot:
I will be closing this as I think this is mute as we are moving to ND backups vs NDO backups. We can reopen it in future if needed.

PR #53, opened 2023-09-25
Added nd_service.py and nd_service_instance.py modules to manage ND services

sajagana:
Note:

Local service package import is not supported and the module supported on ND version 2.3+

### DIFF
--- a/plugins/module_utils/nd.py
+++ b/plugins/module_utils/nd.py
@@ -22,6 +22,7 @@ from ansible.module_utils.six.moves import filterfalse
 from ansible.module_utils.six.moves.urllib.parse import urlencode
 from ansible.module_utils._text import to_native, to_text
 from ansible.module_utils.connection import Connection
+from ansible_collections.cisco.nd.plugins.module_utils.constants import ALLOWED_STATES_TO_APPEND_SENT_AND_PROPOSED
 
 
 def sanitize_dict(dict_to_sanitize, keys=None, values=None, recursive=True, remove_none_values=True):
@@ -400,20 +401,7 @@ class NDModule(object):
     def exit_json(self, **kwargs):
         """Custom written method to exit from module."""
 
-        if self.params.get("state") in (
-            "absent",
-            "present",
-            "upload",
-            "restore",
-            "download",
-            "move",
-            "backup",
-            "enable",
-            "disable",
-            "restart",
-            "delete",
-            "update",
-        ):
+        if self.params.get("state") in ALLOWED_STATES_TO_APPEND_SENT_AND_PROPOSED:
             if self.params.get("output_level") in ("debug", "info"):
                 self.result["previous"] = self.previous
             # FIXME: Modified header only works for PATCH
@@ -430,20 +418,7 @@ class NDModule(object):
             self.result["url"] = self.url
             self.result["httpapi_logs"] = self.httpapi_logs
 
-            if self.params.get("state") in (
-                "absent",
-                "present",
-                "upload",
-                "restore",
-                "download",
-                "move",
-                "backup",
-                "enable",
-                "disable",
-                "restart",
-                "delete",
-                "update",
-            ):
+            if self.params.get("state") in ALLOWED_STATES_TO_APPEND_SENT_AND_PROPOSED:
                 self.result["sent"] = self.sent
                 self.result["proposed"] = self.proposed
 
@@ -461,20 +436,7 @@ class NDModule(object):
     def fail_json(self, msg, **kwargs):
         """Custom written method to return info on failure."""
 
-        if self.params.get("state") in (
-            "absent",
-            "present",
-            "upload",
-            "restore",
-            "download",
-            "move",
-            "backup",
-            "enable",
-            "disable",
-            "restart",
-            "delete",
-            "update",
-        ):
+        if self.params.get("state") in ALLOWED_STATES_TO_APPEND_SENT_AND_PROPOSED:
             if self.params.get("output_level") in ("debug", "info"):
                 self.result["previous"] = self.previous
             # FIXME: Modified header only works for PATCH
@@ -492,20 +454,7 @@ class NDModule(object):
                 self.result["url"] = self.url
                 self.result["httpapi_logs"] = self.httpapi_logs
 
-            if self.params.get("state") in (
-                "absent",
-                "present",
-                "upload",
-                "restore",
-                "download",
-                "move",
-                "backup",
-                "enable",
-                "disable",
-                "restart",
-                "delete",
-                "update",
-            ):
+            if self.params.get("state") in ALLOWED_STATES_TO_APPEND_SENT_AND_PROPOSED:
                 self.result["sent"] = self.sent
                 self.result["proposed"] = self.proposed
 
--- a/plugins/module_utils/nd.py
+++ b/plugins/module_utils/nd.py
@@ -187,6 +187,7 @@ class NDModule(object):
         # nd_rest output
         self.jsondata = None
         self.error = dict(code=None, message=None, info=None)
+        self.ignore_not_found_error = False
 
         # info output
         self.previous = dict()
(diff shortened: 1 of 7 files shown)


plugins/modules/nd_service_instance.py
```diff
@@ -0,0 +1,184 @@
...
+    name = nd.params.get("name")
+    target_version = nd.params.get("target_version")
+    instance_name = nd.params.get("instance_name")
+    state = nd.params.get("state")
+
+    base_path = "/nexus/infra/api/firmware/v1"
+    instance_path = "{0}/serviceinstances".format(base_path)
+
+    # To avoid object not found error while querying the object
+    nd.ignore_not_found_error = True
```
akinross, line 133: same as other module

plugins/modules/nd_service_instance.py
```diff
@@ -0,0 +1,184 @@
...
+
+    # Delete part
+    if state == "delete":
+        service_object = nd.query_obj("/nexus/infra/api/firmware/v1/services/{0}:{1}".format(name, target_version))
+        if service_object:
+            nd.existing = service_object
+        else:
+            nd.previous = nd.existing = {}
+            nd.exit_json()
```
akinross, line 142: is it needed that we exit_json in multiple locations in the module, why define for each state the logic that you need and follow flow till the end?

plugins/modules/nd_service_instance.py
```diff
@@ -0,0 +1,184 @@
...
+    # To avoid object not found error while querying the object
+    nd.ignore_not_found_error = True
+
+    # Delete part
+    if state == "delete":
+        service_object = nd.query_obj("/nexus/infra/api/firmware/v1/services/{0}:{1}".format(name, target_version))
+        if service_object:
+            nd.existing = service_object
+        else:
+            nd.previous = nd.existing = {}
```
akinross, line 141: don't think we need this, since the init function of the ND module already sets this to {}

akinross (changes requested):

plugins/modules/nd_service.py
```diff
@@ -0,0 +1,161 @@
...
+        nd.proposed = payload
+        if not module.check_mode:
+            nd.existing = [nd.request(base_path, method="POST", data=payload)]
+    else:
+        # Absent part
+        if not nd.existing:
+            nd.previous = nd.existing = {}
+        elif len(nd.existing) == 1:
+            nd.previous = nd.existing
+            if not module.check_mode:
```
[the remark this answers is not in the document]
sajagana: When it is in check mode:
1. `current` will be holding the existing object data
2. `proposed` will be empty with HTTP method GET (when it is check mode)
3. The `changed` will be `false`, need to fix it in the nd.py

plugins/module_utils/nd.py
```diff
@@ -398,7 +400,20 @@ def sanitize(self, updates, collate=False, required=None, unwanted=None):
...
+            "upload",
+            "restore",
+            "download",
+            "move",
+            "backup",
+            "enable",
+            "disable",
+            "restart",
+            "delete",
+            "update",
```
akinross, line 415: should we define constant for this an reuse?

plugins/modules/nd_service.py
```diff
@@ -0,0 +1,146 @@
...
+    import_id = nd.params.get("import_id")
+    state = nd.params.get("state")
+
+    base_path = "/nexus/infra/api/firmware/v1/servicepackageimports"
+    if import_id:
+        # Query a object with meta id
+        service_package = nd.query_obj("{0}/{1}".format(base_path, import_id), ignore_not_found_error=True)
+        if service_package:
+            nd.existing = service_package
+    elif state == "absent" or state == "query":
```
akinross, line 116: why do we not do this for present state?

plugins/modules/nd_service_instance.py
```diff
@@ -0,0 +1,184 @@
...
+
+    nd.previous = nd.existing
+
+    unwanted = [
+        "metadata",
+        "status",
+        ["spec", "serviceReference"],
+    ]
+
+    if state == "enable" or (state == "update" and nd.existing):
```
akinross, line 157: should we consider some error or warning handling for scenario's where we want to do an action but we don't execute code because the object doesn't exist in nd?

- state == "update" and not nd.existing
- state == "restart" and not nd.existing
- state == "disable" and not nd.existing

plugins/modules/nd_service_instance.py
```diff
@@ -0,0 +1,184 @@
...
+- Sabari Jaganathan (@sajagana)
+options:
+  name:
+    description:
+    - The name of the Service.
+    aliases: [ service_name ]
+    type: str
+  instance_name:
+    description:
+    - The name of the Service Instance Profile.
```
akinross, line 29:
```suggestion
    - The name of the Service Instance.
```

akinross: should this be just service instance, because I see no references to profile anywhere int the rest of the code or docs

akinross (changes requested):

plugins/modules/nd_service.py
sajagana: Because some of the service package installations may fail due to Internet connectivity or some other permission issues. So we should allow a user to reinstall the service package from the same remote location.

akinross (approved):
LGTM

gmicol (approved):

plugins/modules/nd_service.py
```diff
@@ -0,0 +1,160 @@
...
+    state: absent
+
+- name: Query all service packages with import_url
+  cisco.nd.nd_service:
+    import_url: "https://nd_service.cisco.com/cisco-terraform-v0.1.16.aci"
+    state: query
+  register: query_reseult_import_url
+
+# When the query_reseult_import_url has more than one match
+- name: Remove all query_reseult_import_url result service packages with import_id
```
shrsr, line 80: Remove all service packages with import_id when the query_result_import_url has more than one match

plugins/modules/nd_service.py
```diff
@@ -0,0 +1,160 @@
...
+
+- name: Remove a service package with import_id
+  cisco.nd.nd_service:
+    import_id: "{{ query_result_import_id.current.metadata.id }}"
+    state: absent
+
+# When the import_url has exactly one match
+- name: Remove a service package with import_url
+  cisco.nd.nd_service:
+    import_url: "http://173.36.219.254/cisco-terraform-v0.1.15.aci"
```
shrsr, line 70: Please change the address in the URL

plugins/modules/nd_service.py
```diff
@@ -0,0 +1,160 @@
...
+    type: str
+  import_id:
+    description:
+    - The ID of the imported Service Package.
+    aliases: [ id ]
+    type: str
+  state:
+    description:
+    - Use C(present) for importing a Service Package. The C(present) state is not idempotent.
+    - Use C(query) for listing all Service Packages.
```
shrsr, line 35: - Use C(query) for listing all the Service Packages.

tests/integration/targets/nd_service/tasks/main.yml
```diff
@@ -0,0 +1,296 @@
...
+    <<: *nd_info
+    import_id: "{{ item.metadata.id }}"
+    state: absent
+  loop: "{{ query_all_packages_env_cleanup.current }}"
+
+# Test part
+- name: Import cisco-terraform v0.1.16 from the remote location with check mode
+  cisco.nd.nd_service: &cm_imp_v16
+    <<: *nd_info
+    import_url: "http://173.36.219.254/cisco-terraform-v0.1.16.aci"
```
shrsr, line 31: I think it would be best to move all the IPs in the URL used in the test files to the inventory

plugins/modules/nd_service_instance.py
```diff
@@ -0,0 +1,192 @@
...
+    type: str
+    default: default
+  target_version:
+    description:
+    - The target version of the Service.
+    type: str
+  state:
+    description:
+    - Use C(enable) for enabling a Service Instance.
+    - Use C(query) for listing all Service Instance.
```
shrsr, line 40: - Use C(query) for listing all Service Instances.

shrsr (changes requested):

akinross (approved):
LGTM

shrsr (approved):
LGTM

gmicol (approved):

lhercot (approved):
LGTM

PR #74, opened 2024-05-13
Added modules for multi cluster configuration

anvitha-jain:
Multi-cluster configuration on ND requires two modules nd_federation and nd_federation_member
-> It needs nd_federation to create a Federation before adding Federation Member/attach multi-cluster setup (nd_federation_member)


plugins/modules/nd_federation.py
```diff
@@ -0,0 +1,109 @@
...
+
+    nd.previous = nd.existing
+
+    if state == "query":
+        nd.exit_json()
+    elif state == "absent":
+        if nd.existing:
+            if not module.check_mode:
+                nd.request(federation_path, method="DELETE")
+            nd.existing = {}
```
gmicol, line 90: I'm confused by the assigned value for `nd.existing`. if the `check_mode` is true, then the `nd.existing` should equal the `nd.previous` and `nd.proposed` should be the proposed changes. In your case, here, the `nd.existing` is empty because state is absent but it's whenever the `check_mode` is true or not which is not the way we've been doing for Ansible. I Know we might need to make changes to the nd.py module itself.

plugins/modules/nd_federation.py
```diff
@@ -0,0 +1,109 @@
...
+            },
+        }
+
+        nd.sanitize(payload, collate=True)
+
+        if not module.check_mode:
+            nd.request(path, method="POST", data=payload)
+            time.sleep(10)
+
+    nd.existing = nd.proposed
```
gmicol, line 104: same comment as before and it also applies to `nd_federation_member`: this is only the case if check_mode is false.
Also why don't you put this line under the check condition for `state == "present"` ?

tests/integration/targets/nd_federation/tasks/main.yml
```diff
@@ -0,0 +1,105 @@
...
+- name: Add ND federation again
+  cisco.nd.nd_federation:
+    <<: *add_federation
+    state: present
+  register: add_federation_again
+
+- name: Assertion check for adding ND federation 
+  ansible.builtin.assert:
+    that:
+      - cm_add_federation is changed
```
gmicol, line 49: what is the value of the `current` in the case of `check_mode: True` ? related to the previous comments.

gmicol (changes requested):

tests/integration/targets/nd_federation/tasks/main.yml
anvitha-jain: added the check for check mode

plugins/modules/nd_federation.py
anvitha-jain: fixed nd.existing = nd.proposed

plugins/modules/nd_federation.py
anvitha-jain: nd.existing is added on purpose to support DELETE api.
Without nd.exisiting delete idempotancy won't work

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,151 @@
...
+ANSIBLE_METADATA = {"metadata_version": "1.1", "status": ["preview"], "supported_by": "community"}
+
+DOCUMENTATION = r"""
+---
+module: nd_federation_member
+version_added: "0.3.0"
+short_description:
+    - Setup multi-cluster configuration on Cisco Nexus Dashboard (ND).
+description:
+    - Connects to another Nexus Dashboard (ND) cluster for a single pane of glass view into all clusters’ sites and services.
```
akinross, line 22: Same sentence as in federation module.

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,151 @@
...
+    - M(cisco.nd.nd_federation_member) can only be used with python 3.7 and higher.
+author:
+    - Anvitha Jain (@anvjain)
+options:
+  cluster:
```
akinross, line 27: Have you considered exposing this as list of members input?

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,151 @@
...
+from ansible_collections.cisco.nd.plugins.module_utils.nd import NDModule, nd_argument_spec
+
+
+def main():
+    argument_spec = nd_argument_spec()
+    argument_spec.update(
+        cluster=dict(type="str", aliases=["cluster_ip", "hostname", "ip_address", "federation_member"]),
+        cluster_username=dict(type="str"),
+        cluster_password=dict(type="str", no_log=True),
+        cluster_login_domain=dict(type="str", default="DefaultAuth"),
```
akinross, line 84: Why prepend everything with `cluster_` instead of name, username, password, login_domain?

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,151 @@
...
+    default: present
+    choices: [ absent, present, query ]
+extends_documentation_fragment: cisco.nd.modules
+notes:
+- The C(federation) must exist before using this module in your playbook.
+  The M(cisco.aci.nd_federation) module can be used for this.
+"""
+
+EXAMPLES = r"""
+- name: Setup multi-cluster configuration
```
akinross, line 63: missing examples for absent and query

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,151 @@
...
+    - Connects to another Nexus Dashboard (ND) cluster for a single pane of glass view into all clusters’ sites and services.
+    - M(cisco.nd.nd_federation_member) can only be used with python 3.7 and higher.
+author:
+    - Anvitha Jain (@anvjain)
+options:
+  cluster:
+    description:
+      - The IP address of the cluster.
+    type: str
+    required: true
```
akinross, line 31: does not match arg spec

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,151 @@
...
+    aliases: [ cluster_ip, hostname, ip_address, federation_member ]
+  cluster_username:
+    description:
+      - The username for the cluster.
+    type: str
+    required: true
```
akinross, line 37: does not match arg spec

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,151 @@
...
+  cluster_password:
+    description:
+      - The password for the cluster.
+    type: str
+    required: true
```
akinross, line 42: does not match arg spec

plugins/modules/nd_federation.py
```diff
@@ -0,0 +1,110 @@
...
+    - Connects to another Nexus Dashboard (ND) federation for a single pane of glass view into all federations’ sites and services.
+    - M(cisco.nd.nd_federation) can only be used with python 3.7 and higher.
+author:
+    - Anvitha Jain (@anvjain)
+options:
+  name:
+    description:
+      - The name of the federation.
+    type: str
+    required: true
```
akinross, line 32: does not match arg spec

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,151 @@
...
+    cluster_username = nd.params.get("cluster_username")
+    cluster_password = nd.params.get("cluster_password")
+    if cluster_password is not None:
+        # Make password base64 encoded
+        cluster_password = (base64.b64encode(str.encode(cluster_password))).decode("utf-8")
+    cluster_login_domain = nd.params.get("cluster_login_domain")
+    state = nd.params.get("state")
+
+    path = "/nexus/api/federation/v4/members"
+    cluster_path = path
```
akinross, line 109: why use two variables? can we overwrite the on when found?

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,151 @@
...
+
+    cluster_obj = nd.query_obj(path, ignore_not_found_error=True).get("items")
+
+    if cluster_obj:
+        cluster_info = next((cluster_dict for cluster_dict in cluster_obj if cluster_dict.get("spec").get("host") == cluster), None)
+        if cluster_info:
+            cluster_path = "{0}/{1}".format(path, cluster_info.get("status").get("memberID"))
+            nd.existing = cluster_info
```
akinross, line 117: same filter question as in federation module

plugins/modules/nd_federation_member.py
akinross, line 84: Do we need the default or can we exclude from payload when login domain is not provided?

tests/integration/targets/nd_federation/tasks/main.yml
```diff
@@ -0,0 +1,105 @@
...
+  cisco.nd.nd_federation:
+    <<: *nd_info
+    name: 'lh-dmz1-pod1-ndo-v402'
+    state: query
+  register: query_federation
+
+- name: Assertion check for quering ND federation 
+  ansible.builtin.assert:
+    that:
+      - query_all_federations is not changed
```
akinross, line 73: should we add another federation and should we check expected length (amount of federations)?

tests/integration/targets/nd_federation_member/tasks/main.yml
```diff
@@ -0,0 +1,133 @@
...
+  cisco.nd.nd_federation_member:
+    <<: *nd_info
+    cluster: '173.36.219.33'
+    state: query
+  register: query_federation_member
+
+- name: Assertion check for querying ND federation members
+  ansible.builtin.assert:
+    that:
+    - query_all_federation_members is not changed
```
akinross, line 91: should we add another member and should we check expected length (amount of member)?

tests/integration/targets/nd_federation_member/tasks/main.yml
```diff
@@ -0,0 +1,133 @@
...
+    cluster: 'lh-dmz1-pod1-ndo-v402'
+    state: absent
+  ignore_errors: true
+  register: rm_local_federation_member
+
+- name: Assertion check for removing ND federation members
+  ansible.builtin.assert:
+    that:
+    - rm_local_federation_member is not changed
+    # - rm_local_federation_member.msg == "Remove all non local federation members to remove this member"
```
akinross, line 109: Why is this commented out? Can it be removed or do we need to add error handling some how?

akinross (changes requested):

plugins/modules/nd_federation.py
```diff
@@ -0,0 +1,110 @@
...
+            "spec": {
+                "name": name,
+            },
+        }
+
+        nd.sanitize(payload, collate=True)
+
+        if not module.check_mode:
+            nd.request(path, method="POST", data=payload)
+            time.sleep(10)
```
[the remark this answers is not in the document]
anvitha-jain: Idempotency doesn't work if don't wait until the object is created

plugins/modules/nd_federation_member.py
anvitha-jain: we need default, it messes idempotancy if removed

tests/integration/targets/nd_federation_member/tasks/main.yml
anvitha-jain: we need to add error handling.

plugins/modules/nd_federation_member.py
anvitha-jain: username, password, login_domain are same as ND parameters, if not prepended it might get confused with similar keys.

akinross: why exactly, cant't you exclude from the the compare?

akinross: ok makes sense

plugins/modules/nd_federation.py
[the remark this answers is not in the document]
akinross: should we then query in a loop (x amount of times) until it is created? because what now if it is done after 2 seconds (waste 8 seconds) or after 15 second (same issue as not waiting)

plugins/modules/nd_federation.py
```diff
@@ -0,0 +1,110 @@
...
+  state:
+    description:
+      - The state of the federation configuration.
+    type: str
+    default: present
+    choices: [ absent, present, query ]
+extends_documentation_fragment: cisco.nd.modules
+"""
+
+EXAMPLES = r"""
```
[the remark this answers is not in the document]
anvitha-jain: added

plugins/modules/nd_federation_member.py
anvitha-jain: changed

plugins/modules/nd_federation.py
anvitha-jain: changed

plugins/modules/nd_federation_member.py
anvitha-jain: changed

plugins/modules/nd_federation_member.py
anvitha-jain: changed

plugins/modules/nd_federation_member.py
anvitha-jain: added

plugins/modules/nd_federation.py
```diff
@@ -0,0 +1,110 @@
...
+    nd = NDModule(module)
+    name = nd.params.get("name")
+    state = nd.params.get("state")
+
+    path = "/nexus/api/federation/v4/federations"
+    federation_obj = nd.query_obj(path, ignore_not_found_error=True).get("items")
+
+    if federation_obj:
+        federation_info = next((federation_dict for federation_dict in federation_obj if federation_dict.get("spec").get("name") == name), None)
+        if federation_info:
```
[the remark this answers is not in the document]
anvitha-jain: As discussed, there's provision for adding just one federation and one member. GET will error out without a POST. And on UI you cannot create a local cluster/member and the POST only helps in association of federation member/multicluster config.

plugins/modules/nd_federation_member.py
anvitha-jain: It only lets me add one. so based on current UI flow, cluster is not exposed as list.

plugins/modules/nd_federation_member.py
anvitha-jain: answered above

plugins/modules/nd_federation_member.py
anvitha-jain: removed

tests/integration/targets/nd_federation/tasks/main.yml
anvitha-jain: added

tests/integration/targets/nd_federation_member/tasks/main.yml
anvitha-jain: added

plugins/modules/nd_federation.py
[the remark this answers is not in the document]
anvitha-jain: yup added and changed

plugins/modules/nd_federation_member.py
anvitha-jain: changed

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,280 @@
...
+
+
+def main():
+    argument_spec = nd_argument_spec()
+    argument_spec.update(
+        clusters=dict(
+            type="list",
+            elements="dict",
+            options=dict(
+                cluster_hostname=dict(type="str", aliases=["cluster_ip", "hostname", "ip_address", "federation_member"]),
```
akinross, line 121: I am confused about your cluster_hostname. Is this not a required parameter? I see it is required for query, and present. So whenever you specify a cluster in clusters the cluster_hostname is required I think.

This would simplify the logic in lines 146-151

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,280 @@
...
+    elif state == "present":
+        remove_member_list = []
+        add_member_list = []
+        payload_dict = {}
+        if len(federation_member_obj) <= 1:
+            # if there are no members or just a local member, add all members from users.
+            add_member_list = clusters
+        else:
+            # Remove exisiting members not specified by the users.
+            member_exists = False
```
akinross, line 212: Do we need this boolean flag or could we use next() function like you do above in line 197? same applies for the add equivalent in line 223

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,280 @@
...
+        if add_member_list or remove_member_list:
+            if remove_member_list:
+                for member in remove_member_list:
+                    cluster_member_path = "{0}/{1}".format(member_path, member.get("status").get("memberID"))
+                    nd.request(cluster_member_path, method="DELETE")
+
+                    try:
+                        payload_dict["DELETE"].append(cluster_member_path)
+                    except:
+                        payload_dict["DELETE"] = [cluster_member_path]
```
akinross, line 241: Why do the try/except here and not initialise the payload dict with "DELETE" key and empty list? Same applies for POST.

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,280 @@
...
+                    nd.sanitize(payload, collate=True)
+
+                    if not module.check_mode:
+                        # If federation does not exist, create a new federation
+                        if not federation_obj:
+                            nd.request(federation_path, method="POST", data={"spec": {"name": local_cluster_name}})
+                        nd.request(member_path, method="POST", data=payload)
+
+            if not module.check_mode:
+                nd.existing = nd.query_obj(member_path, ignore_not_found_error=True).get("items")
```
akinross, line 270: why do we do another query here?

akinross (changes requested):

akinross:
rebase the code, then CI sanity check should pass again

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,280 @@
...
+                        cluster_member_path = "{0}/{1}".format(member_path, member_host.get("status").get("memberID"))
+                        nd.request(cluster_member_path, method="DELETE")
+
+                # Remove the federation if there are no more members.
+                if len(nd.query_obj(member_path, ignore_not_found_error=True).get("items")) == 1:
+                    federation_info = next(
+                        (federation_dict for federation_dict in federation_obj if federation_dict.get("spec").get("name") == local_cluster_name), None
+                    )
+                    if federation_info:
+                        nd.request("{0}/{1}".format(federation_path, federation_info.get("status").get("federationID")), method="DELETE")
```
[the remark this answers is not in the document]
anvitha-jain: federation_info already set in line 164. -> loop removed

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,280 @@
...
+    else:
+        nd.existing = federation_member_obj
+
+    nd.previous = nd.existing
+
+    if state == "query":
+        nd.exit_json()
+    elif state == "absent":
+        if nd.existing:
+            if not module.check_mode:
```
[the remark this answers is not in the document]
anvitha-jain: changed

plugins/modules/nd_federation_member.py
anvitha-jain: I am trying to send a list of members in the payload, which is not allowed by the API, so i have to make one API call for each member in the list -> due to which  nd.existing is usually the last list member sent. which causes issue for idempotency. so, making a query here

plugins/modules/nd_federation_member.py
anvitha-jain: yes, and modified

plugins/modules/nd_federation_member.py
anvitha-jain: changed

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,280 @@
...
+                    nd.fail_json(msg="'cluster_hostname' is required when quering a specific federation member.")
+            elif state == "present":
+                if not (cluster.get("cluster_hostname") and cluster.get("cluster_username") and cluster.get("cluster_password")):
+                    nd.fail_json(msg="'cluster_hostname', 'cluster_username' and 'cluster_password' are required when state is present.")
+
+    federation_path = "/nexus/api/federation/v4/federations"
+    member_path = "/nexus/api/federation/v4/members"
+
+    # GET local cluster name
+    local_cluster_name = nd.query_obj("/nexus/infra/api/platform/v1/clusters").get("items")[0].get("spec").get("name")
```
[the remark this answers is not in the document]
anvitha-jain: added check

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,268 @@
...
+__metaclass__ = type
+
+ANSIBLE_METADATA = {"metadata_version": "1.1", "status": ["preview"], "supported_by": "community"}
+
+DOCUMENTATION = r"""
+---
+module: nd_federation_member
+version_added: "0.3.0"
+short_description:
+    - Setup multi-cluster configuration on Cisco Nexus Dashboard (ND).
```
akinross, line 21: This fixes your CI issue

```suggestion
short_description: Setup multi-cluster configuration on Cisco Nexus Dashboard (ND).
```

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,268 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2024, Anvitha Jain (@anvjain) <anvjain@cisco.com>
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+import base64
+import time
```
akinross, line 10: remove time is not used

plugins/modules/nd_federation_member.py
akinross, line 21: fix indentation in the docs

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,268 @@
...
+- name: Remove a federation member
+  cisco.nd.nd_federation_member:
+    host: nd
+    username: admin
+    password: SomeSecretPassword
+    cluster: 172.37.20.15
+    state: absent
+  delegate_to: localhost
+"""
+RETURN = r"""
```
akinross, line 107:
```suggestion

RETURN = r"""
```

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,268 @@
...
+                    # Use next() to check if the member exists in user-specified clusters
+                    member_exists = next((True for user_member_host in clusters if user_member_host.get("cluster_hostname") == member_host), False)
+                    if not member_exists:
+                        remove_member_list.append(existing_member_hosts)
+
+            # Add members specified by the users.
+            for user_member_host in clusters:
+                user_host = user_member_host.get("cluster_hostname")
+                # Use next() to check if the user-specified member exists in existing members
+                found = next((True for existing_member_hosts in federation_member_obj if existing_member_hosts.get("spec").get("host") == user_host), False)
```
akinross, line 222:
```suggestion
                # Use next() to check if the user-specified member exists in existing members
                found = next((True for existing_member_hosts in federation_member_obj if existing_member_hosts.get("spec").get("host") == user_member_host.get("cluster_hostname")), False)
```

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,268 @@
...
+        else:
+
+            # Remove existing members not specified by the users.
+            for existing_member_hosts in federation_member_obj:
+                member_host = existing_member_hosts.get("spec").get("host")
+                if member_host != local_cluster_name:
+                    # Use next() to check if the member exists in user-specified clusters
+                    member_exists = next((True for user_member_host in clusters if user_member_host.get("cluster_hostname") == member_host), False)
+                    if not member_exists:
+                        remove_member_list.append(existing_member_hosts)
```
akinross, line 216: do we need the remove_member_list or can we do the remove request here and append to payload dict? This would prevent additional loops for in line 227

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,268 @@
...
+
+            # Add members specified by the users.
+            for user_member_host in clusters:
+                user_host = user_member_host.get("cluster_hostname")
+                # Use next() to check if the user-specified member exists in existing members
+                found = next((True for existing_member_hosts in federation_member_obj if existing_member_hosts.get("spec").get("host") == user_host), False)
+                if not found:
+                    add_member_list.append(user_member_host)
```
akinross, line 224: do we need the add_member_list or can we do the add request here and append to payload dict? This would prevent additional loops for in line 235

akinross (changes requested):

plugins/modules/nd_federation_member.py
anvitha-jain: done

plugins/modules/nd_federation_member.py
anvitha-jain: done

plugins/modules/nd_federation_member.py
anvitha-jain: done

plugins/modules/nd_federation_member.py
anvitha-jain: done

plugins/modules/nd_federation_member.py
anvitha-jain: Done

plugins/modules/nd_federation_member.py
anvitha-jain: modified

plugins/modules/nd_federation_member.py
anvitha-jain: modified

samiib (changes requested):

plugins/module_utils/nd.py
```diff
@@ -301,6 +301,10 @@ def request(self, path, method=None, data=None, file=None, qs=None, prefix="", f
                     self.fail_json(msg="ND Error {code}: {message}".format(**payload), data=data, info=info, payload=payload)
                 elif "messages" in payload and len(payload.get("messages")) > 0:
                     self.fail_json(msg="ND Error {code} ({severity}): {message}".format(**payload["messages"][0]), data=data, info=info, payload=payload)
+                elif "errors" in payload:
+                    if ignore_not_found_error:
+                        return {}
+                    self.fail_json(msg="ND Error: {0}".format(payload["errors"][0]), data=data, info=info, payload=payload)
```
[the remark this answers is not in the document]
anvitha-jain: There's usually only one error.

plugins/module_utils/nd.py
```diff
@@ -301,6 +301,10 @@ def request(self, path, method=None, data=None, file=None, qs=None, prefix="", f
                     self.fail_json(msg="ND Error {code}: {message}".format(**payload), data=data, info=info, payload=payload)
                 elif "messages" in payload and len(payload.get("messages")) > 0:
                     self.fail_json(msg="ND Error {code} ({severity}): {message}".format(**payload["messages"][0]), data=data, info=info, payload=payload)
+                elif "errors" in payload:
+                    if ignore_not_found_error:
+                        return {}
```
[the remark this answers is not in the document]
anvitha-jain: The reason, this was repeated here is because -> 
- in nd_federation member module
  - federation has to be created before creating the federation member 
  - the first call this module makes is a query to verify if the federation was added and it will return "errors" in payload, which means it will exit from the module but we want it to continue execution since we can add federation later 
  - if ignore_not_found_error: -> works as an additional check and lets us continue the execution of the code

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,271 @@
...
+author:
+  - Anvitha Jain (@anvjain)
+options:
+  clusters:
+    description:
+      - The IP address of the cluster.
+    type: list
+    elements: dict
+    suboptions:
+      cluster_hostname:
```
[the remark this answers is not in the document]
anvitha-jain: changed

tests/integration/targets/nd_federation_member/tasks/main.yml
```diff
@@ -0,0 +1,154 @@
+# Test code for the ND modules
+# Copyright: (c) 2024, Anvitha Jain (@anvitha-jain) <anvjain@cisco.com>
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+# - name: Test that we have an Nexus Dashboard Insights Group and Site Name defined
+#   fail:
+#     msg: "Please define the following variables: insights_group, site_name."
+#   when: insights_group is not defined or site_name is not defined
```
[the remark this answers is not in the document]
anvitha-jain: not required here, removed the code

plugins/module_utils/nd.py
```diff
@@ -301,6 +301,10 @@ def request(self, path, method=None, data=None, file=None, qs=None, prefix="", f
                     self.fail_json(msg="ND Error {code}: {message}".format(**payload), data=data, info=info, payload=payload)
                 elif "messages" in payload and len(payload.get("messages")) > 0:
                     self.fail_json(msg="ND Error {code} ({severity}): {message}".format(**payload["messages"][0]), data=data, info=info, payload=payload)
+                elif "errors" in payload:
```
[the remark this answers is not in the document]
anvitha-jain: yes added

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,271 @@
...
+      - The IP address of the cluster.
+    type: list
+    elements: dict
+    suboptions:
+      hostname:
+        description:
+          - The IP address of the federation member/cluster.
+        type: str
+        required: true
+        aliases: [ cluster_ip, hostname, ip_address, federation_member ]
```
samiib, line 38: Just got to remove the duplicated alias to fix sanity. :)
```suggestion
        aliases: [ cluster_ip, ip_address, federation_member ]
```

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,271 @@
...
+
+
+def main():
+    argument_spec = nd_argument_spec()
+    argument_spec.update(
+        clusters=dict(
+            type="list",
+            elements="dict",
+            options=dict(
+                hostname=dict(type="str", required=True, aliases=["cluster_ip", "hostname", "ip_address", "federation_member"]),
```
samiib, line 120:
```suggestion
                hostname=dict(type="str", required=True, aliases=["cluster_ip", "ip_address", "federation_member"]),
```

samiib (changes requested):

plugins/modules/nd_federation_member.py
anvitha-jain: changed

plugins/modules/nd_federation_member.py
anvitha-jain: chnaged

samiib (approved):
LGTM

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,271 @@
...
+                if member_host != local_cluster_name:
+                    # Use next() to check if the member exists in user-specified clusters
+                    member_exists = next((True for user_member_host in clusters if user_member_host.get("hostname") == member_host), False)
+                    if not member_exists:
+                        remove_member_list.append(existing_member_hosts)
+
+            # Add members specified by the users.
+            for user_member_host in clusters:
+                # Use next() to check if the user-specified member exists in existing members
+                found = found = next(
```
akinross, line 219: why the double found?

plugins/modules/nd_federation_member.py
akinross: Not sure you understood what I am trying to say, I do not see where this comment is affecting idempotency?

You would still do the request and add it to the payload dict just here ( and at the remove part ) thus the need of storing in list variables is not needed.

akinross: Wasn't the sleep for the creation on ND site? I also do not see this in your code anymore.

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,271 @@
...
+
+    clusters = nd.params.get("clusters")
+    state = nd.params.get("state")
+
+    # validate parameters
+    if clusters:
+        for cluster in clusters:
+            if state == "present":
+                if not (cluster.get("username") and cluster.get("password")):
+                    nd.fail_json(msg="'username' and 'password' are required when state is present.")
```
akinross, line 147: We could improve on the nesting and readability here I think:

```suggestion
    clusters = nd.params.get("clusters", [])
    state = nd.params.get("state")

    # validate parameters
    for cluster in clusters:
        if state == "present" and not (cluster.get("username") and cluster.get("password")):
            nd.fail_json(msg="'username' and 'password' are required when state is present.")
```

akinross (changes requested):

plugins/modules/nd_federation_member.py
anvitha-jain: removed.

plugins/modules/nd_federation_member.py
anvitha-jain: changed and then edited 139 -> clusters = nd.params.get("clusters") if nd.params.get("clusters") is not None else []

plugins/modules/nd_federation_member.py
anvitha-jain: Fixed this and mode changes

anvitha-jain: Fixed this and mode changes

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,254 @@
...
+        argument_spec=argument_spec,
+        supports_check_mode=True,
+        required_if=[
+            ["state", "present", ["clusters"]],
+        ],
+    )
+
+    nd = NDModule(module)
+
+    clusters = nd.params.get("clusters") if nd.params.get("clusters") is not None else []
```
samiib, line 139: Does using a default param work here instead?
Should return `[]` if `clusters` is None. 
```suggestion
    clusters = nd.params.get("clusters", [])
```

plugins/modules/nd_federation_member.py
anvitha-jain: fixed

plugins/modules/nd_federation_member.py
anvitha-jain: fixed

plugins/modules/nd_federation_member.py
anvitha-jain: fixed

samiib (changes requested):

plugins/modules/nd_federation_member.py
anvitha-jain: Nope, this doesn't work. Apparently the .get() is never returned empty it is always set to None

samiib: I see, this is to make sure its an empty list when `clusters` are not specified **and** if `clusters` are explicitly set to `null`. 👍

gmicol (approved):
LGTM

plugins/modules/nd_federation_member.py
```diff
@@ -0,0 +1,254 @@
...
+  - Connects multiple clusters together for a single pane of glass view and administration of the clusters and their sites, services, and configurations.
+  - M(cisco.nd.nd_federation_member) can only be used with python 3.7 and higher.
+  - Can only be used with Nexus Dashboard versions >= 2.3(2d).
+  - The ND version of the local cluster must be greater than or equal to the ND versions of the clusters being added.
+author:
+  - Anvitha Jain (@anvjain)
+options:
+  clusters:
+    description:
+      - The IP address of the cluster.
```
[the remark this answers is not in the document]
anvitha-jain: changed

lhercot (approved):
LGTM

PR #85, opened 2024-07-12
First PR to comply with requirements for inclusion in Ansible package

lhercot:
See https://github.com/ansible-collections/ansible-inclusion/discussions/77 for details

This PR addresses the following items:

**Standards and documentation:**
- [x] follows the [Ansible documentation standards](https://docs.ansible.com/ansible/devel/dev_guide/developing_modules_documenting.html) and the [style guide](https://docs.ansible.com/ansible/devel/dev_guide/style_guide/index.html#style-guide)
MUST FIX: check mode support is not reflected in docs
- [x] supports [all Python versions supported by all ansible-core versions its supports](https://docs.ansible.com/ansible/devel/reference_appendices/release_and_maintenance.html#ansible-core-support-matrix). If it does not, read the [full guidelines](https://docs.ansible.com/ansible/devel/community/collection_contributors/collection_requirements.html#python-compatibility) to see if you qualify for an exception and document the unsupported [Python versions](https://docs.ansible.com/ansible/devel/reference_appendices/release_and_maintenance.html#ansible-core-support-matrix) in the collection README.md and in every module and plugin (or in doc fragments)
COMMENT: does the collection support all Python versions required ^? Anyway, i would suggest including this info to README explicitly

**Collection management:**
- [x]  collection dependencies must have a lower bound on the version which is at least 1.0.0, and are all part of the ansible package
MUST FIX: Please set the lower bound for ansible.netcommon explicitly
- [x] meta/runtime.yml defines the minimal version of Ansible which the collection works with
SHOULD FIX: I see requires_ansible: '>=2.15.0' but in README it's 2.14+. Please update README if it's irrelevant (and in other places if needed).

### DIFF
--- a/plugins/doc_fragments/check_mode.py
+++ b/plugins/doc_fragments/check_mode.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2024, Lionel Hercot (@LHercot) <lhercot@cisco.com>
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+
+class ModuleDocFragment(object):
+    # Standard files documentation fragment
+    DOCUMENTATION = r"""
+attributes:
+  check_mode:
+    description: The check_mode / dry run support.
+    support: full
+"""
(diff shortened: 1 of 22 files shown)


anvitha-jain (approved):
LGTM

plugins/doc_fragments/check_mode.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
```
samiib, line 1: Are we adding this as a separate doc fragment because some modules don't need check_mode?

lhercot: I think they all need check_mode but currently they don't all support it and I didn't want to add that support in this PR.

samiib (approved):
LGTM

Issue #80, opened 2024-06-07
Enhancement: add deployment mode to nd_setup (DCNE-47)


akinross:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Prior to release 3.1(1), you had to download and install individual services after the initial cluster deployment was completed. As of release 3.1(1) you can choose to enable the services during the initial installation with deployment mode.

See step 22 in below guide:

https://www.cisco.com/c/en/us/td/docs/dcn/nd/3x/deployment/cisco-nexus-dashboard-and-services-deployment-guide-311/nd-deploy-esx-31x.html

### New or Affected Module(s):

* nd_setup

### ND version and ND Platform

* V 3.1(1)

### References

* #0000

akinross:
related to https://github.com/CiscoDevNet/ansible-nd/issues/78

PR #96, opened 2024-09-10
Add support and deployment mode to nd_setup for ND version 3.1.1 and later. (DCNE-46 DCNE-47)

gmicol:
Solves #78 and #80


plugins/modules/nd_setup.py
```diff
@@ -248,6 +255,31 @@
...
+    - This is used for enabling the available services between
+      Orchestrator, Fabric Controller and Insights during the intial installation.
+    - This option is only applicable for ND version 3.1.1 and later.
+    type: list
+    elements: str
+    choices: [ ndo, ndfc, ndi ]
+  external_services:
+    description:
+    - The persistent Service IPs/Pools to provide
+      if I(deployment_mode) includes C(ndfc) or C(ndi).
```
shrsr, line 269: Is "if I(deployment_mode) includes C(ndfc) or C(ndi)." required here?

plugins/modules/nd_setup.py
```diff
@@ -248,6 +255,31 @@
...
+    description:
+    - This is used for enabling the available services between
+      Orchestrator, Fabric Controller and Insights during the intial installation.
+    - This option is only applicable for ND version 3.1.1 and later.
+    type: list
+    elements: str
+    choices: [ ndo, ndfc, ndi ]
+  external_services:
+    description:
+    - The persistent Service IPs/Pools to provide
```
shrsr, line 268: - The persistent service IPs/Pools to be provided

plugins/modules/nd_setup.py
```diff
@@ -248,6 +255,31 @@
                 - The ASN of the BGP peer.
                 type: int
                 required: true
+  deployment_mode:
+    description:
+    - This is used for enabling the available services between
```
shrsr, line 260: - This is used for enabling the services between Orchestrator, Fabric Controller and Insights during the initial installation.

plugins/modules/nd_setup.py
```diff
@@ -407,7 +454,7 @@ def main():
                             "keyID": server.get("ntp_key_id"),
                             "prefer": server.get("preferred"),
                         }
-                        for server in ([] if ntp_config is None else ntp_config.get("servers"))
+                        for server in (ntp_config.get("servers") if ntp_config is not None and ntp_config.get("servers") is not None else [])
```
shrsr, line 457: Any reason for changing this?

shrsr (changes requested):

plugins/modules/nd_setup.py
gmicol: It is not required but could be phrased differently. instead of "If..." maybe something like "This is used when I(deployment_mode) includes C(ndfc) or C(ndi)."

plugins/modules/nd_setup.py
gmicol: Will fix it.

plugins/modules/nd_setup.py
gmicol: Better phrasing thanks, will fix it!

plugins/modules/nd_setup.py
gmicol: Actually no reason, I just forgot that "servers" was a required attribute when "ntp_config" is used. So the code did not need to change.

plugins/modules/nd_setup.py
shrsr: I think this sounds good:
"This can be used when I(deployment_mode) is C(ndfc) or C(ndi)"

plugins/modules/nd_setup.py
```diff
@@ -464,10 +512,34 @@ def main():
             ],
         }
 
+        path = "/bootstrap/cluster"
```
shrsr, line 515: I was also thinking whether the version check is required here? In other words, will the installation process be affected in anyway if the payload consisting of ["clusterConfig"]["deploymentMode"] is pushed to an older version of ND? If it's not affected we can avoid making an api call to check the version and probably always include ["clusterConfig"]["deploymentMode"] within the payload.

shrsr (changes requested):

plugins/modules/nd_setup.py
gmicol: the check version might required for the path not for deployment or the new feature introduced with 3.1.1. Seems like that they change the URL path to /v2/boostrap/cluster instead of /boostrap/cluster. I am gonna try with the older path to see if it's required. I agree with you, I would love to avoid this extra API call.

plugins/modules/nd_setup.py
```diff
@@ -248,6 +255,31 @@
...
+    - This is used for enabling the services between Orchestrator,
+      Fabric Controller and Insights during the initial installation.
+    - This option is only applicable for ND version 3.1.1 and later.
+    type: list
+    elements: str
+    choices: [ ndo, ndfc, ndi ]
+  external_services:
+    description:
+    - The persistent Service IPs/Pools to be provided.
+    - This can be used when I(deployment_mode) includes C(ndfc) or C(ndi) otherwise it will be ignored.
```
akinross, line 269:
```suggestion
    - This can be used when 0(deployment_mode=ndfc) or 0(deployment_mode=ndi) otherwise it will be ignored.
```

plugins/modules/nd_setup.py
```diff
@@ -248,6 +255,31 @@
                 - The ASN of the BGP peer.
                 type: int
                 required: true
+  deployment_mode:
```
akinross, line 258: perhaps add some additional information here from the documentation: 


Depending on the number of nodes in the cluster, some services or cohosting scenarios may not be supported. If you are unable to choose the desired number of services, click Back and ensure that you have provided enough secondary nodes in the previous step.The deployment mode cannot be changed after the cluster is deployed, so you must ensure that you have completed all service-specific prerequisites described in earlier chapters of this document:


reference: https://www.cisco.com/c/en/us/td/docs/dcn/nd/3x/deployment/cisco-nexus-dashboard-and-services-deployment-guide-321/nd-deploy-physical-32x.html

plugins/modules/nd_setup.py
akinross: https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#semantic-markup-within-module-documentation

akinross (changes requested):

plugins/modules/nd_setup.py
gmicol: Will do, thanks!

plugins/modules/nd_setup.py
gmicol:
```suggestion
    - This can be used when O(deployment_mode) includes V(ndfc) or V(ndi) otherwise it will be ignored.
```
Or since deployment_mode is a list:
```suggestion
    - This can be used when O(deployment_mode=[ndfc]) or O(deployment_mode=[ndi]) otherwise it will be ignored.
```
I found the latest more confusing because from a reader point of view, it seems like the only two options when to use external_services but you can also use it with a combination of ndi/ndo for example. what do you think

samiib (changes requested):

plugins/modules/nd_setup.py
```diff
@@ -248,6 +255,36 @@
...
```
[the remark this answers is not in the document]
gmicol: I am trying to stick to the UI with only `deployment_mode` but not a bad idea at all, I will add `mode`

plugins/modules/nd_setup.py
akinross: I like the O V combination better

plugins/modules/nd_setup.py
```diff
@@ -150,6 +152,11 @@
         - The serial number of the node.
         type: str
         required: true
+      role:
+        description:
+        - The role of the node.
+        type: str
+        choices: [ primary, secondary, standby ]
```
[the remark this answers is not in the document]
gmicol: there specific information regarding the default role for nodes in the API documentation. the problem is that this feature was introduced in ND v3.1.1 as well as deployment mode but the role is required. I tried several test without the role being defined and it seems like it defaults to `primary`. I am gonna propose something but let me know what you think.

akinross (approved):
LGTM

plugins/modules/nd_setup.py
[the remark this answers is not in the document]
gmicol: the API works so that for previous ND version, new features introduced in later ones are ignored if used in the payload. So this should not be a problem.

shrsr (approved):
LGTM

samiib (approved):
LGTM

sajagana (approved):
LGTM!

lhercot (approved):
LGTM

PR #123, opened 2025-05-01
[minor_change] Updated the `cisco.nd.nd` HTTPAPI plugin to be compatible with the `cisco.dcnm` collection. (DCNE-404)

samiib:


lhercot (approved):
LGTM

plugins/httpapi/nd.py
```diff
@@ -85,6 +85,24 @@ def set_backup_hosts(self):
         except Exception:
             return []
 
+    # Required for cisco.dcnm modules
+    # The cisco.nd HTTPAPI provider only supports NDFC 12+
+    # TODO Add support for more platforms.
+    # TODO Add support for dynamically returning platform versions.
+    def get_version(self, platform="ndfc"):
+        if platform == "ndfc":
```
akinross, line 93: Do we want to define this as a constant, type dictionary? Then we could use a `return PLATFORMS.get(platform)`. 

I am not familiar with the flow of DCNM modules but why return None instead of raising a undefined platform type error?

akinross (commented):
LGTM

plugins/httpapi/nd.py
samiib: I kept it simple for now to keep compatibility with the `ansible-dcnm` modules. We may need to enhance and use this function in the future for newer ND versions.
It will now throw a value error if an unsupported platform is used.

plugins/httpapi/nd.py
```diff
@@ -337,6 +367,13 @@ def _verify_response(self, response, method, path, data):
         #     self.info['msg'] = msg
         self.info["body"] = response_data
 
+        # Set info keys required for cisco.dcnm modules
+        self.info["RETURN_CODE"] = response_code
```
shrsr, line 371: I'm not sure about the behaviour but will this info be displayed for ND modules as well? If that's the case shall we segregate what's displayed for ND and NDFC by leveraging platform?

samiib: Based on my testing and review of our code, I don't believe we use the `send_request()` functions in a way that surfaces the entire `info` dictionary. We use the specific keys from the returned `info` instead. The new ones required by `ansible-dcnm` will be ignored.

samiib: The goal was to make the `ansible-nd` HTTPAPI plugin compatible without changes to `ansible-dcnm`. I am not sure if there is way to know which platform/module is being used without making more changes.

sajagana (approved):
LGTM!

akinross (approved):
LGTM

lhercot (approved):
LGTM

PR #134, opened 2025-06-26
Fix small changes in Documentation and comments in nd_setup module. (DCNE-459)

gmicol:


plugins/modules/nd_setup.py
```diff
@@ -588,7 +588,7 @@ def main():
             ],
         }
 
-        # Deployment mode options introduced between ND version 3.1.1 and 3.2.2
+        # Deployment mode options available from ND version 3.1.1 to 3.2.2
```
shrsr, line 591: Can we change this to? ->
```
# Deployment mode options available in ND from version 3.1.1 to 3.2.2
```

plugins/modules/nd_setup.py
```diff
@@ -276,17 +276,17 @@
...
     elements: str
     choices: [ ndo, ndfc, ndi-virtual, ndi-physical ]
     aliases: [ mode ]
   external_services:
     description:
     - The persistent Service IPs/Pools to be provided.
-    - For ND versions between 3.1.1 and 3.2.2,
-      This can only be used when O(deployment_mode) includes V(ndi) or V(ndo) otherwise it will be ignored.
-    - This option is only applicable for ND version 3.1.1 and later.
+    - For ND versions between 3.1.1 and 3.2.2, This can only be used when O(deployment_mode)
```
shrsr, line 287: Can we change this to ->
- For ND versions between 3.1.1 and 3.2.2, this can only be used when O(deployment_mode)

shrsr (changes requested):

anvitha-jain (approved):
LGTM

lhercot (approved):
LGTM

PR #146, opened 2025-07-17
Fixed attribute names and aliases to match ND 4.1 (DCNE-454)

shrsr:


plugins/modules/nd_delta_analysis.py
```diff
@@ -27,39 +27,41 @@
...
     - Epoch collection time, in ISO format, for the earlier epoch
     - Ignored if state is C(query) or C(absent)
     type: str
+    aliases: [ earlier_snapshot_time ]
   later_epoch_time:
     description:
     - Epoch collection time, in ISO format, for the later epoch
     - Ignored if state is C(query) or C(absent)
     type: str
+    aliases: [ later_snapshot_time ]
```
gmicol, line 64: Shouldn't we mention `snapshot` in all the `epoch` description?

gmicol, line 64: Shouldn't we also swap `snapshot` and epoch`, meaning that `snapshot` should be the name of the attribute and `epoch` the alias instead? in order to be consistent with the nd_snapshot module.

gmicol (changes requested):

anvitha-jain (approved):
LGTM

akinross (approved):
LGTM

lhercot (approved):
LGTM

PR #161, opened 2025-07-29
[minor_change] Update nd_pcv_delta_analysis for ND 4.1 with delay and timeout options (DCNE-493 DCNE-456)

samiib:
- Align nd_pcv_delta_analysis with ND 4.1 
- Introduced delay and timeout options for the validate state. 
- Refactor the waiting logic for pre-change analysis completion into a dedicated method. 
- Updated test cases to reflect changes in messaging and functionality.

Completes DCNE-456

### DIFF
--- a/plugins/modules/nd_pcv_delta_analysis.py
+++ b/plugins/modules/nd_pcv_delta_analysis.py
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 
 # Copyright: (c) 2022, Cindy Zhao (@cizhao) <cizhao@cisco.com>
+# Copyright: (c) 2025, Samita Bhattacharjee (@samiib) <samitab@cisco.com>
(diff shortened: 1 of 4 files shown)


plugins/modules/nd_pcv_delta_analysis.py
```diff
@@ -29,15 +31,15 @@
...
-  site_name:
+  fabric:
     description:
-    - Name of the Assurance Entity.
+    - Name of the Fabric.
     type: str
     required: true
-    aliases: [ site ]
+    aliases: [ site, site_name, fabric_name ]
   epoch_choice:
```
shrsr, line 43: Shall we update this to snapshot_choice to be consistent with the recent changes made to the other modules?

shrsr (changes requested):

anvitha-jain (approved):
LGTM

PR #168, opened 2025-08-19
Add a new module nd_local_user for local users on Nexus Dashboard

gmicol:
This module is only supported on ND v.4.1.0 or higher.


plugins/modules/nd_local_user.py
```diff
@@ -0,0 +1,274 @@
...
+- name: Query an existing local user
+  cisco.nd.nd_local_user:
+    login_id: local_user
+    state: query
+  register: query_result
+
+- name: Query all local users
+  cisco.nd.nd_local_user:
+    state: query
+  register: all_keys
```
sajagana, line 155:
```suggestion
  register: query_all
```

tests/integration/targets/nd_local_user/tasks/main.yml
```diff
@@ -0,0 +1,276 @@
...
+- name: Ensure local users do not exist before test starts
+  cisco.nd.nd_local_user:
+    <<: *nd_info
+    login_id: "{{ item }}"
+    state: absent
+  loop:
+    - ansible_local_user
+    - ansible_local_user_2
+
+#CREATE
```
sajagana, line 25:
```suggestion
# CREATE
```

Fix all other places.

tests/integration/targets/nd_local_user/tasks/main.yml
```diff
@@ -0,0 +1,276 @@
...
+    remote_user_authorization: true
+    state: present
+  check_mode: true
+  register: cm_create_local_user
+
+- name: Create local user (normal mode)
+  cisco.nd.nd_local_user:
+    <<: *create_local_user
+  register: nm_create_local_user
+
```
sajagana, line 51: Can you add normal mode create and update again to check the idempotent behaviour?

plugins/modules/nd_local_user.py
```diff
@@ -0,0 +1,274 @@
...
+                nd.existing = nd.request(path=updated_path, method="PUT", data=payload)
+            else:
+                nd.existing = nd.request(path=path, method="POST", data=payload)
+        else:
+            nd.existing = nd.proposed
+
+    elif state == "absent":
+        if nd.existing:
+            if not module.check_mode:
+                nd.request(path="{0}/{1}".format(path, login_id), method="DELETE")
```
sajagana, line 267: Can you use `updated_path` instead of formatting the path again?

plugins/modules/nd_local_user.py
```diff
@@ -0,0 +1,274 @@
...
+
+__metaclass__ = type
+
+ANSIBLE_METADATA = {"metadata_version": "1.1", "status": ["preview"], "supported_by": "community"}
+
+DOCUMENTATION = r"""
+---
+module: nd_local_user
+version_added: "1.4.0"
+short_description: Manage local users on Nexus Dashboard
```
sajagana, line 18:
```suggestion
short_description: Manage local users on Cisco Nexus Dashboard
```

sajagana (changes requested):

tests/integration/targets/nd_local_user/tasks/main.yml
gmicol: This module cannot be idempotent as it is used to set credentials.
Passwords can not be retrieved by the API which is normal but it makes it difficult to make comparison when the password is being updated.
Also a same password can be used several time in a row and the API will notice that as a change even though it's the same password (well depends on how you set your password policy to begin with but it will always be the case if `reuse_limitation` is set to `0`).

plugins/modules/nd_local_user.py
gmicol: you're right, forgot to fix that

plugins/modules/nd_local_user.py
```diff
@@ -0,0 +1,274 @@
...
+            payload["rbac"] = {
+                "domains": {
+                    security_domain.get("name"): {
+                        "roles": [USER_ROLES_MAPPING.get(role) for role in security_domain["roles"]] if isinstance(security_domain.get("roles"), list) else [],
+                    }
+                    for security_domain in security_domains
+                },
+            }
+        if reuse_limitation or time_interval_limitation:
+            payload["passwordPolicy"] = sanitize_dict(
```
shrsr, line 247: Is sanitize_dict() required here when we sanitize the entire payload in the next step?

plugins/modules/nd_local_user.py
```diff
@@ -0,0 +1,274 @@
...
+    type: int
+  security_domains:
+    description:
+    - The list of Security Domains and Roles for the local user.
+    type: list
+    elements: dict
+    suboptions:
+      name:
+        description:
+        - The name of the Security Domain to which give the local user access.
```
shrsr, line 63: Can we please change this to... if it's more meaningful?
- The name of the Security Domain to which the local user is given access.

shrsr (changes requested):

plugins/modules/nd_local_user.py
gmicol: It's required as `nd.sanitize()` is not recursive and just sanitize one level deep which is not enough in this case. Maybe, we should address this issue in the future so that the `nd.sanitize()` can be recursive as well

plugins/modules/nd_local_user.py
```diff
@@ -0,0 +1,280 @@
...
+            "password": user_password,
+            "remoteIDClaim": remote_id_claim,
+            "xLaunch": remote_user_authorization,
+        }
+
+        if security_domains:
+            payload["rbac"] = {
+                "domains": {
+                    security_domain.get("name"): {
+                        "roles": [USER_ROLES_MAPPING.get(role) for role in security_domain["roles"]] if isinstance(security_domain.get("roles"), list) else [],
```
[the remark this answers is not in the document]
gmicol: This is the only way I found to provide an empty list of `roles` assigned to a security domain. Assigning an None value to "roles" or sanitizing the key will result in an API error.
Also, as providing an empty list works, I wonder if this is a normal behavior or a bug as it seems weird to not assign any role to a local user for a security domain.

plugins/modules/nd_local_user.py
```diff
@@ -0,0 +1,280 @@
...
+      name: all
+      roles:
+        - observer
+        - support_engineer
+    remote_id_claim: remote_user
+    remote_user_authorization: true
+    state: present
+  register: result
+
+- name: Create local user with minimal configuration
```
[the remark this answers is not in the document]
gmicol: This is something that was introduced with the new nd_api_key module. So I thought this has become the new standard but if not, we should discuss if we should add this creation task with minimum config in the EXAMPLES section. 
I think this is interesting to have as it makes visually clearer what attributes are required when creating a new object.
@sajagana, @shrsr, @samiib, @lhercot, @anvitha-jain what do you think?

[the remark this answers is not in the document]
shrsr: I have no preference on this

sajagana (approved):
LGTM!

plugins/modules/nd_local_user.py
```diff
@@ -0,0 +1,280 @@
...
+    - Use C(absent) to delete an existing local user.
+    - Use C(query) for listing all the existing local users or a specific local user if O(login_id) is specified.
+    type: str
+    default: present
+    choices: [ present, absent, query ]
+extends_documentation_fragment:
+- cisco.nd.modules
+- cisco.nd.check_mode
+notes:
+- This module is only supported on Nexus Dashboard having version 4.1.0 or higher.
```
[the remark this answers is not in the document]
gmicol: I think you're right, it's immediately clearer for the user

samiib (approved):

anvitha-jain (approved):
LGTM

PR #169, opened 2025-08-20
Created nd_backup_schedule module using Ansible Network Resource Modules Approach (DCNE-496)

sajagana:


plugins/module_utils/utils.py
```diff
@@ -0,0 +1,67 @@
+# -*- coding: utf-8 -*-
```
akinross, line 1: do we want to start to write unit tests for all these newly introduced utils function?

samiib (changes requested):

allenrobel (commented):
General comments:

1. All method signatures should include modern type annotations, i.e.:

- Legacy -> Modern
- `Dict` -> `dict`
- `List` -> `list`
- `Set` -> `set`

For example:

```python
from typing import Any

def my_method(dict_key: str, dict_value: bool) -> dict[str, Any]:
    return { "key": dict_key, "value": dict_value}
```

2. Remove all lines matching each of the following lines across all files since these are no longer needed.

```python
from __future__ import absolute_import, division, print_function

__metaclass__ = type
```

mtarking:
Closing as not needed now.

PR #170, opened 2025-08-25
Addition of new network resource module for nd_multi_cluster_connectivity (DCNE-458)

shrsr:


plugins/modules/nd_multi_cluster_connectivity.py
```diff
@@ -0,0 +1,354 @@
...
+    )
+
+    nd = NDModule(module)
+    desired_clusters_from_module_input = nd.params.get("config")
+    state = nd.params.get("state")
+
+    desired_clusters_map = {}
+    for cluster_data_input in desired_clusters_from_module_input:
+        normalized_payload = module_input_to_api_payload(cluster_data_input)
+        desired_clusters_map[get_cluster_unique_key(normalized_payload)] = normalized_payload
```
akinross, line 321: so for my understanding, is this map reflecting the desired state of a unique id ( this case the onboardUrl from spec ) and then you get the existing configuration after and put it in another map with same format? Those two maps are then passed to the manage_state function which decides what to delete / add / update?

what is the purpose of the normalized_payload and why is this not done for existing? are you conforming the input to the existing configuration?

plugins/modules/nd_multi_cluster_connectivity.py
```diff
@@ -0,0 +1,354 @@
...
+        state=dict(type="str", default="merged", choices=["merged", "replaced", "overridden", "deleted"]),
+    )
+
+    module = AnsibleModule(
+        argument_spec=argument_spec,
+        supports_check_mode=True,
+    )
+
+    nd = NDModule(module)
+    desired_clusters_from_module_input = nd.params.get("config")
```
akinross, line 315: should we assign variables when they are only used once after in code?

plugins/modules/nd_multi_cluster_connectivity.py
akinross, line 321: Do you need this get_cluster_unique_key(normalized_payload) when you use basically use cluster_data_input.get("cluster_hostname")?

akinross, line 321: Should there be validations in place to avoid duplicate keys and thus overwrites of existing dict entries?

plugins/modules/nd_multi_cluster_connectivity.py
```diff
@@ -0,0 +1,354 @@
...
+    for cluster_data_input in desired_clusters_from_module_input:
+        normalized_payload = module_input_to_api_payload(cluster_data_input)
+        desired_clusters_map[get_cluster_unique_key(normalized_payload)] = normalized_payload
+
+    existing_clusters_raw = nd.query_objs("/api/v1/infra/clusters", key="clusters")
+
+    existing_clusters_map = {}
+    for cluster_data_raw in existing_clusters_raw:
+        # normalized_payload = convert_api_response_to_payload_format(cluster_data_raw)
+        existing_clusters_map[get_cluster_unique_key(cluster_data_raw)] = cluster_data_raw
```
akinross, line 328: when dict comprehension is possible should we try to use that technique?

tests/integration/targets/nd_multi_cluster_connectivity/tasks/main.yml
```diff
@@ -0,0 +1,177 @@
...
+      - cm_connect_cluster_merged.after["173.36.219.190"]["spec"]["clusterType"] == "APIC"
+      - cm_connect_cluster_merged.after["173.36.219.190"]["spec"]["aci"]["licenseTier"] == "advantage"
+      - cm_connect_cluster_merged.after["173.36.219.190"]["spec"]["aci"]["name"] == "ansible_test"
+      - cm_connect_cluster_merged.after["173.36.219.190"]["spec"]["aci"]['orchestration']["status"] == "disabled"
+      - cm_connect_cluster_merged.after["173.36.219.190"]["spec"]["aci"]["telemetry"]["status"] == "disabled"
+      - cm_connect_cluster_merged.after["173.36.219.190"]["spec"]["onboardUrl"] == "173.36.219.190"
+      - nm_connect_cluster_merged is changed
+      - nm_connect_cluster_merged.before["173.36.219.189"] == {}
+      - nm_connect_cluster_merged.before["173.36.219.190"] == {}
+      - nm_connect_cluster_merged.after["173.36.219.189"]["spec"]["clusterType"] == "APIC"
```
akinross, line 67: what is the benefit of this spec container in the output?

anvitha-jain (changes requested):

mtarking:
Closing as not needed now.

PR #184, opened 2026-03-02
Add Log class and logging config

allenrobel:
## Summary

Adds the `Log` class for configuring Python logging across the ND collection. This PR is one of two companion PRs targeting `nd42_integration`:

- **This PR `nd42_logging`** — `Log` class, logging config, and `test_log.py`
- **#185 `nd42_rest_send`** — RestSend framework, `enums.py`, and all shared test infrastructure

> **Note:** `test_log.py` imports from `common_utils` and `enums` which reside in `nd42_rest_send` (#185). Tests will not run until both branches are merged into `nd42_integration`.

---

## Files added

| File | Description |
|------|-------------|
| `plugins/module_utils/log.py` | `Log` class — configures Python logging from an external JSON `dictConfig` file; validates that only file-based handlers are used (no stdout/stderr interference with Ansible) |
| `plugins/module_utils/logging_config.json` | Ready-to-use `logging.config.dictConfig`-compatible config writing to `/tmp/nd.log` via `RotatingFileHandler` |
| `tests/unit/test_log.py` | Unit tests for the `Log` class |

## Log class behaviour

- Reads `ND_LOGGING_CONFIG` env var to locate the logging config JSON file
- Validates that the config contains only `file`-based handlers to avoid interfering with Ansible's JSON output
- Calls `logging.config.dictConfig()` to configure the root `nd` logger
- Disables logging entirely (adds `NullHandler`) when no config is provided
- `develop` property controls whether the logging module raises internal exceptions (default: `False`)

## Merge order

1. Merge `nd42_rest_send` (#185) → `nd42_integration` *(provides enums.py and shared test infrastructure)*
2. Merge this PR → `nd42_integration`

## Test plan

- [ ] After merging both PRs into `nd42_integration`: `python -m pytest tests/unit/test_log.py` passes cleanly
- [ ] `tox -e linters` passes (black, isort, pylint, mypy)
- [ ] `ansible-test sanity --docker` passes

### DIFF
--- a/plugins/module_utils/common/log.py
+++ b/plugins/module_utils/common/log.py
@@ -121,7 +121,26 @@ class Log:
 
     ## Example module code using the Log() object
 
-    In the main() function of a module.
+    The `setup_logging()` helper is the recommended way to configure logging in module `main()` functions.
+    It handles exceptions internally by calling `module.fail_json()`.
+
+    ```python
+    from ansible_collections.cisco.nd.plugins.module_utils.common.log import setup_logging
+
+    def main():
+        module = AnsibleModule(...)
+        log = setup_logging(module)
+
+        task = AnsibleTask()
+    ```
+
+    To enable logging exceptions during development, pass `develop=True`:
+
+    ```python
+    log = setup_logging(module, develop=True)
+    ```
+
+    Alternatively, `Log()` can be used directly when finer control is needed:
 
     ```python
     from ansible_collections.cisco.nd.plugins.module_utils.common.log import Log
@@ -404,7 +423,7 @@ class Log:
         logging.raiseExceptions = value
 
 
-def setup_logging(module: "AnsibleModule") -> Log:
+def setup_logging(module: "AnsibleModule", develop: bool = False) -> Log:
     """
     # Summary
 
@@ -431,9 +450,15 @@ def setup_logging(module: "AnsibleModule") -> Log:
         module = AnsibleModule(...)
         log = setup_logging(module)
     ```
+
+    To enable logging exceptions during development, pass `develop=True`:
+
+    ```python
+    log = setup_logging(module, develop=True)
+    ```
     """
     try:
-        log = Log()
+        log = Log(develop=develop)
         log.commit()
     except ValueError as error:
         module.fail_json(msg=str(error))
--- a/tests/unit/test_log.py
+++ b/tests/unit/test_log.py
@@ -844,6 +844,37 @@ def test_setup_logging_00010(tmp_path, monkeypatch) -> None:
     mock_module.fail_json.assert_not_called()
 
 
+@pytest.mark.parametrize("develop", [(True), (False)])
+def test_setup_logging_00040(monkeypatch, develop) -> None:
+    """
+    # Summary
+
+    Verify `setup_logging()` passes `develop` through to the `Log` instance.
+
+    ## Test
+
+    - `ND_LOGGING_CONFIG` is not set.
+    - `setup_logging()` is called with `develop` set to `True` or `False`.
+    - `result.develop` matches the value passed.
+    - `logging.raiseExceptions` matches the value passed.
+
+    ## Classes and Methods
+
+    - `setup_logging()`
+    """
+    monkeypatch.delenv("ND_LOGGING_CONFIG", raising=False)
+
+    mock_module = MagicMock()
+
+    with does_not_raise():
+        result = setup_logging(mock_module, develop=develop)
+
+    assert isinstance(result, Log)
+    assert result.develop is develop
+    assert logging.raiseExceptions is develop
+    mock_module.fail_json.assert_not_called()
+
+
 def test_setup_logging_00020(monkeypatch) -> None:
     """
     # Summary


plugins/module_utils/logging_config.json
```diff
@@ -0,0 +1,36 @@
+{
```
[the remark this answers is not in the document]
samiib: I agree with @akinross. I feel like this is an important request for customer usability. I am concerned that it may get lost in the backlog of issues once this PR is merged in. Would it be okay to prioritise this issue?

tests/unit/test_log.py
```diff
@@ -0,0 +1,569 @@
+# -*- coding: utf-8 -*-
```
samiib: Since this may be one of the first set of unit tests created for this repo, the Unit test CI step can now be enabled. If it is working in this branch already it can be turned on in this PR or later in the `nd42_integrated` branch. 
Edit:
As the note mentions it will only work in `nd42_integrated`, please enable the Unit test step when all is merged.

tests/unit/test_log.py
```diff
@@ -0,0 +1,569 @@
...
+
+    with does_not_raise():
+        instance = Log()
+        instance.commit()
+
+    info_msg = "foo"
+    debug_msg = "bing"
+    warning_msg = "bar"
+    critical_msg = "baz"
+    log = logging.getLogger("dcnm.test_logger")
```
lhercot:
```suggestion
    log = logging.getLogger("nd.test_logger")
```

tests/unit/test_log.py
```diff
@@ -0,0 +1,569 @@
...
```
lhercot:
```suggestion
    log = logging.getLogger("nd.test_logger")
```

tests/unit/test_log.py
```diff
@@ -0,0 +1,569 @@
...
+def test_log_00231(tmp_path) -> None:
+    """
+    ### Methods
+    -   Log().commit()
+
+    ### Test
+    -   No ``ValueError`` is raised when a handler uses a non-standard name
+        but a valid handler class (e.g. ``logging.handlers.RotatingFileHandler``).
+    -   Previously, validation checked the handler key name rather than the
+        class, so ``"my_file_handler"`` would have been incorrectly rejected.
```
lhercot: Just a comment but it seems that you limit your line length to ~80 char while ansible allows for up to 160.
I think it makes it easier to read with line length closer to 160.

plugins/module_utils/common/log.py
```diff
@@ -0,0 +1,437 @@
...
+            "level": "INFO",
+            "handlers": [
+                "file"
+            ]
+        }
+    }
+    ```
+    """
+
+    def __init__(self):
```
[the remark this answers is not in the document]
lhercot: I am also in favor of allowing setting of these option in constructor.

lhercot (changes requested):

tests/unit/test_log.py
mikewiebe: Thanks @samiib

plugins/module_utils/common/log.py
```diff
@@ -0,0 +1,437 @@
...
+            with open(self.config, "r", encoding="utf-8") as file:
+                try:
+                    logging_config = json.load(file)
+                except json.JSONDecodeError as error:
+                    msg = f"error parsing logging config from {self.config}. "
+                    msg += f"Error detail: {error}"
+                    raise ValueError(msg) from error
+        except IOError as error:
+            msg = f"error reading logging config from {self.config}. "
+            msg += f"Error detail: {error}"
```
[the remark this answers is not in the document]
allenrobel: Agree that we should be using 159 characters and my local linters are set for that length.

In the case above, it was more personal preference (just prefer to be able to scan the code quickly to see what is being logged.  To my eyes, the above is more clear than:

```python
msg = f"logging.config.dictConfig: error reading logging config from {self.config}.  Error detail: {error}"
```

But if there's a team preference, I'm more than happy to abide by it.

In the case of the docstring, black doesn't reformat these AFAIK, that's just me being conservative.  I can reformat the docstrings to leverage the longer length.

plugins/module_utils/common/log.py
```diff
@@ -0,0 +1,437 @@
...
+    """
+
+    def __init__(self):
+        self.class_name = self.__class__.__name__
+        # Disable exceptions raised by the logging module.
+        # Set this to True during development to catch logging errors.
+        logging.raiseExceptions = False
+
+        self._config: Optional[str] = environ.get("ND_LOGGING_CONFIG", None)
+        self._develop: bool = False
```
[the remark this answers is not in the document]
allenrobel: Done: https://github.com/CiscoDevNet/ansible-nd/pull/184/changes/045f498e4c52c40f710d67dd266cc5d21582bbdc

This commit also modifies the function signature for `setup_logging` to add the ability to pass `develop` as a parameter (defaults to False).

tests/unit/test_log.py
samiib: 👍 I can enable the step when the PRs are merged together. #196

shrsr (dismissed):

PR #198, opened 2026-03-11
Ansible ND 4.X | ND Manage Switches Module + Pydantic Models + Smart Endpoints

AKDRG:
This PR introduces the initial implementation of ND Manage switch lifecycle support in the cisco.nd collection.
It adds a new switch management resource layer, endpoint wrappers, Pydantic model hierarchy, utility helpers.

What’s Included
1. New switch lifecycle resource implementation

Added a full resource engine in plugins/module_utils/manage_switches/nd_switch_resources.py
Implements state handling for:
- gathered
- merged
- replaced
- overridden
- deleted
- POAP flows (bootstrap, pre-provision, serial swap)
- RMA flow

Adds structured operation sequencing:
- discovery
- add/remove
- role updates
- credentials save
- wait-for-manageability
- save/deploy finalize

2. ND Manage endpoint layer

Added endpoint models under plugins/module_utils/endpoints/v1/:
- manage_credentials_switches.py
- manage_fabrics_actions.py
- manage_fabrics_bootstrap.py
- manage_fabrics_inventory.py
- manage_fabrics_switchactions.py
- manage_fabrics_switches.py 
- manage_fabrics.py 

4. Pydantic model framework and nd_manage_switches models

- Added nd_manage_switches model package and exports:
- bootstrap, config, discovery, preprovision, rma, switch_actions, switch_data, enums, validators
- package-level init re-exports for consistent imports

5. Utilities for lifecycle orchestration
Added helpers in plugins/module_utils/manage_switches:
- utils.py

6. Unit Tests for the Endpoints and Integration Tests for the States supported by the Module. 

These utilities centralize grouping, payload shaping, wait/retry logic, and fabric operations.


plugins/module_utils/endpoints/v1/manage/nd_manage_switches/credentials.py
```diff
@@ -0,0 +1,153 @@
...
+    params = CredentialsSwitchesEndpointParams(ticket_id="CHG12345")
+    query_string = params.to_query_string()
+    # Returns: "ticketId=CHG12345"
+    ```
+    """
+
+    ticket_id: Optional[str] = Field(default=None, min_length=1, description="Change control ticket ID")
+
+
+class _V1ManageCredentialsSwitchesBase(BaseModel):
```
akinross: should this be NDEndpointBaseModel?

```suggestion
class _V1ManageCredentialsSwitchesBase(NDEndpointBaseModel):
```

plugins/module_utils/endpoints/v1/manage/nd_manage_switches/credentials.py
```diff
@@ -0,0 +1,153 @@
...
+    ## Usage
+
+    ```python
+    params = CredentialsSwitchesEndpointParams(ticket_id="CHG12345")
+    query_string = params.to_query_string()
+    # Returns: "ticketId=CHG12345"
+    ```
+    """
+
+    ticket_id: Optional[str] = Field(default=None, min_length=1, description="Change control ticket ID")
```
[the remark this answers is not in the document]
akinross: same comment applies to other endpoints

plugins/module_utils/endpoints/v1/manage/nd_manage_switches/credentials.py
akinross: same comment applies to other endpoints

plugins/module_utils/endpoints/v1/manage/nd_manage_switches/fabric_bootstrap.py
```diff
@@ -0,0 +1,164 @@
...
+    BasePath,
+)
+from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import (
+    BaseModel,
+    ConfigDict,
+    Field,
+)
+
+# Common config for basic validation
+COMMON_CONFIG = ConfigDict(validate_assignment=True)
```
akinross: why is this constant defined here and only used one time in _V1ManageFabricBootstrapBase?

plugins/module_utils/endpoints/v1/manage/nd_manage_switches/fabric_bootstrap.py
```diff
@@ -0,0 +1,164 @@
...
+    """
+    # Summary
+
+    Endpoint-specific query parameters for fabric bootstrap endpoint.
+
+    ## Parameters
+
+    - max: Maximum number of results to return (optional)
+    - offset: Pagination offset (optional)
+    - filter: Lucene filter expression (optional)
```
akinross: Why do some of the endpoint params include filter? What is this used for?

akinross (changes requested):

plugins/module_utils/endpoints/v1/manage/nd_manage_switches/fabric_config.py
```diff
@@ -0,0 +1,285 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2026, Akshayanat Chengam Saravanan (@achengam) <achengam@cisco.com>
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+"""
+ND Manage Fabric Config endpoint models.
```
mikewiebe: There is some overlap here with the fabric module endpoints so we should discuss consolidation and moving things like config-save and config-deploy to a more common location instead of under nd_manage_switches

plugins/module_utils/models/nd_manage_switches/bootstrap_models.py
```diff
@@ -0,0 +1,398 @@
...
+__metaclass__ = type
+
+from pydantic import Field, computed_field, field_validator, model_validator
+from typing import Any, Dict, List, Optional, ClassVar, Literal
+from typing_extensions import Self
+
+from ansible_collections.cisco.nd.plugins.module_utils.models.base import NDBaseModel
+from ansible_collections.cisco.nd.plugins.module_utils.models.nested import NDNestedModel
+
+from .enums import (
```
mikewiebe: Use full path for imports, (here and elsewhere)

`from ansible_collections.cisco.nd.plugins.module_utils.models.nd_manage_switches.enums`

mikewiebe (changes requested):

plugins/modules/nd_manage_switches.py
```diff
@@ -0,0 +1,622 @@
...
+)
+from ansible_collections.cisco.nd.plugins.module_utils.rest.results import Results
+
+
+def main():
+    """Main entry point for the nd_manage_switches module."""
+    
+    # Build argument spec
+    argument_spec = nd_argument_spec()
+    argument_spec.update(
```
mikewiebe: We decided that it would be better to include this as part of our pydantic models and update argument_spec from the model.

Here is an example:

https://github.com/CiscoDevNet/ansible-nd/pull/207/changes#diff-e8e57d176067341e720b4349c4569795b5155338208dd655a3d4ad739e462be4R1368

https://github.com/CiscoDevNet/ansible-nd/pull/207/changes#diff-e8e57d176067341e720b4349c4569795b5155338208dd655a3d4ad739e462be4R1283

Probably makes sense to update your plugins/module_utils/models/nd_manage_switches/config_models.py

plugins/modules/nd_manage_switches.py
```diff
@@ -0,0 +1,622 @@
...
+        required_if=[
+            ("state", "merged", ["config"]),
+            ("state", "overridden", ["config"]),
+        ],
+    )
+
+    # Initialize logging
+    try:
+        log_config = Log()
+        log_config.config = "/Users/achengam/Documents/Ansible_Dev/NDBranch/ansible_collections/cisco/nd/ansible_cisco_log_r.json"
```
mikewiebe: We should remove the hardcoded path

plugins/module_utils/models/manage_switches/bootstrap_models.py
```diff
@@ -0,0 +1,398 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2026, Akshayanat C S (@achengam) <achengam@cisco.com>
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
```
mikewiebe, line 6: **UPDATE:**

We settled on the following rule for organizing our models

**Rule:** Directory name that matches shortest part of module name with model files under it but drop `nd_` prefix

In this case your directory should be named `plugins/module_utils/models/manage_switches/`.

plugins/module_utils/endpoints/v1/manage/manage_credentials_switches.py
```diff
@@ -0,0 +1,146 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2026, Akshayanat Chengam Saravanan (@achengam) <achengam@cisco.com>
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+"""
```
mikewiebe, line 6: **Update:** We settled on a rule for how we should organize endpoints

**Rule:** Name files consistent with the API (one file per endpoint path)

For example:

* Endpoint Path: 
    * `/api/v1/manage/credentials/switches`
* FileName: `manage_credentials_switches.py`
* All GET/POST/PUT/DELETE classes associated with this endpoint would go in this file

------

* Endpoint Path: 
    * `/api/v1/manage/fabrics/{fabricName}/bootstrap`
* FileName: `manage_fabrics_bootstrap.py`
* All GET/POST/PUT/DELETE classes associated with this endpoint would go in this file

This will result in more files for each endpoint path but it will make it easier to find endpoints based on the filename that can be reused across modules if needed.

plugins/module_utils/utils/manage_switches/__init__.py
```diff
@@ -0,0 +1,52 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2025, Akshayant Chengam Saravanan (@achengam) <achengam@cisco.com>
+
```
mikewiebe: I saw a conflict when I was testing this code with `plugins/module_utils/utils/utils.py` that already exists in the repo.  How did you reconcile that in your workspace?

plugins/modules/nd_manage_switches.py
```diff
@@ -0,0 +1,622 @@
...
+
+# Copyright: (c) 2026, Akshayanat Chengam Saravanan (@achengam) <achengam@cisco.com>
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+__copyright__ = "Copyright (c) 2026 Cisco and/or its affiliates."
+__author__ = "Akshayanat Chengam Saravanan"
+
```
mikewiebe, line 11: Couple of general comments when testing the module:

* I was able to add switches using discovery mode (awesome!)
* I was not able to add both discovery and pre-provision switches in the same playbook task.  The module code blocked me but we should be able to add in the same task.  Let's discuss
* I was not able to add pre-provision switches even after placing it in a separate task.  I got a 400 error indicating I had to enable dhcp in my fabric settings but I don't think we should need that for just pre-provision (WITHOUT POAP)

plugins/modules/nd_manage_switches.py
```diff
@@ -0,0 +1,622 @@
...
+    # Get parameters
+    state = module.params.get("state")
+    fabric = module.params.get("fabric")
+    output_level = module.params.get("output_level")
+
+    # Initialize Results - this collects all operation results
+    results = Results()
+    results.state = state
+    results.check_mode = module.check_mode
```
mikewiebe, line 442: Does not look like you handle check_mode at the moment (unless I am misreading something) but we should be able to display what switches would be added but not actually add them.  We can discuss more in our meeting tomorrow.

plugins/module_utils/nd_switch_resources.py
```diff
@@ -0,0 +1,2972 @@
...
+import logging
+from copy import deepcopy
+from dataclasses import dataclass
+from typing import Any, Dict, List, Optional, Tuple, Union
+
+from pydantic import ValidationError
+
+from ansible_collections.cisco.nd.plugins.module_utils.nd_v2 import NDModule
+from ansible_collections.cisco.nd.plugins.module_utils.enums import OperationType
+from ansible_collections.cisco.nd.plugins.module_utils.nd_output import NDOutput
```
mikewiebe, line 31: Notes about NDOutput behavior from my discussion today

* Before and after state can contain read-only properties (things like config-sync state)
* After by default should include a query of the fabric at the end to reflect updated state
* Explore control knobs to improve performance and not have to query at the end for after state
    * Example knobs: https://github.com/CiscoDevNet/ansible-aci/blob/master/docs/optimizing.md

Don't worry about the control knobs for now.  We can do that in a subsequent update.

plugins/modules/nd_manage_switches.py
```diff
@@ -0,0 +1,622 @@
...
+        - Name of the target fabric for switch operations.
+        type: str
+        required: yes
+    state:
+        description:
+        - The state of ND and switch(es) after module completion.
+        - C(merged) and C(query) are the only states supported for POAP.
+        - C(merged) is the only state supported for RMA.
+        type: str
+        default: merged
```
mikewiebe, line 40: We discussed that it would be good to support `state: gathered`

This should follow the Ansible argspec and if used in a playbook with state overridden would result in no changes to the controller

This IMO can be a subsequent update to the module

plugins/module_utils/endpoints/v1/manage/nd_manage_switches/credentials.py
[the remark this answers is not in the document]
AKDRG: Ticket_id, max, offset, filter have been made as separate classes in the common mixin file and included wherever it's being used. 
However, for some unique params such as incl_all_msd_switches, hostname might not be used in other endpoints. So, for now have kept them in their respective base endpoint.

plugins/module_utils/endpoints/v1/manage/nd_manage_switches/credentials.py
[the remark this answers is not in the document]
AKDRG: All endpoint names changed in the latest.

AKDRG: Have made the change to all endpoints in the latest.

plugins/module_utils/endpoints/v1/manage/nd_manage_switches/credentials.py
```diff
@@ -0,0 +1,153 @@
...
+    request.endpoint_params.ticket_id = "CHG12345"
+    path = request.path
+    verb = request.verb
+    # Path will be: /api/v1/manage/credentials/switches?ticketId=CHG12345
+    ```
+    """
+
+    # Version metadata
+    api_version: Literal["v1"] = Field(default="v1", description="ND API version for this endpoint")
+    min_controller_version: str = Field(default="3.0.0", description="Minimum ND version supporting this endpoint")
```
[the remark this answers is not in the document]
AKDRG: Done.

plugins/module_utils/endpoints/v1/manage/nd_manage_switches/fabric_bootstrap.py
AKDRG: It was defined in all the endpoints files and used locally within the file in the base class. Now that NDEndpointBaseModel contains    model_config = ConfigDict(validate_assignment=True), we don't require these anymore. Have removed it in the latest.

plugins/module_utils/endpoints/v1/manage/nd_manage_switches/fabric_bootstrap.py
AKDRG: Some endpoints support the use of filter to narrow down on the query request.

For example,
/api/v1/manage/fabrics/fabricname/switches?filter=switchRole:spine AND model:N9K-C9500v

plugins/module_utils/endpoints/v1/manage/nd_manage_switches/credentials.py
```diff
@@ -0,0 +1,153 @@
...
+
+    class_name: Literal["V1ManageCredentialsSwitchesPost"] = Field(
+        default="V1ManageCredentialsSwitchesPost", description="Class name for backward compatibility"
+    )
```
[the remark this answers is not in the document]
AKDRG: Done across endpoint classes.

plugins/modules/nd_manage_switches.py
```diff
@@ -0,0 +1,512 @@
...
+                type: str
+                required: true
+            auth_proto:
+                description:
+                - SNMP authentication protocol to use.
+                - For POAP and RMA, should be C(MD5).
+                type: str
+                default: MD5
+                choices: ['MD5', 'SHA', 'MD5_DES', 'MD5_AES', 'SHA_DES', 'SHA_AES']
+            user_name:
```
mtarking: Can we just make this username?

plugins/modules/nd_manage_switches.py
```diff
@@ -0,0 +1,512 @@
...
+                        type: str
+                        no_log: true
+                    serial_number:
+                        description:
+                        - Serial number of the physical switch to Bootstrap.
+                        - When used together with C(preprovision_serial), performs a swap operation
+                          that changes the serial number of a pre-provisioned switch and then
+                          imports it via bootstrap.
+                        type: str
+                    preprovision_serial:
```
mtarking: Can we give prevision it's own entry point, separate from poap, as I can preprovision without poap, e.g.

```
- name: Preprovision a switch
  cisco.nd.nd_manage_switches:
    fabric: my-fabric
    config:
      - seed_ip: 192.168.10.1
        user_name: admin
        password: "{{ switch_password }}"
        preprovision:
          serial: SAL1234ABCD
          model: N9K-C93180YC-EX
          version: "10.3(1)"
          hostname: leaf-preprov
          gateway_ip: 192.168.10.1/24
        poap:
          serial: SAL1234ABCD
    state: merged
```

plugins/modules/nd_manage_switches.py
```diff
@@ -0,0 +1,512 @@
...
+            preserve_config:
+                description:
+                - Set to C(false) for greenfield deployment, C(true) for brownfield.
+                type: bool
+                default: false
+            poap:
+                description:
+                - POAP (PowerOn Auto Provisioning) configurations for bootstrap/preprovision.
+                - POAP and DHCP must be enabled in fabric before using.
+                type: list
```
mtarking: Curious, why is this a ```list```?

plugins/modules/nd_manage_switches.py
```diff
@@ -0,0 +1,512 @@
...
+                                required: true
+                        - Serial number of new replacement switch.
+                        type: str
+                        required: true
+
+extends_documentation_fragment:
+- cisco.nd.modules
+- cisco.nd.check_mode
+notes:
+- This module requires ND 12.x or higher.
```
mtarking: `- This module requires ND 4.2 or higher.`

plugins/modules/nd_manage_switches.py
```diff
@@ -0,0 +1,512 @@
...
+                                elements: str
+                            gateway:
+                                description:
+                                - Gateway IP with subnet mask (e.g., 192.168.0.1/24).
+                                type: str
+            rma:
+                description:
+                - RMA an existing switch with a new one.
+                - Please note that the existing switch should be configured and deployed in maintenance mode.
+                - Please note that the existing switch being replaced should be shutdown state or out of network.
```
mtarking: These contradict each other a bit; I'll draft something to help reword.

plugins/modules/nd_manage_switches.py
```diff
@@ -0,0 +1,512 @@
...
+                suboptions:
+                    discovery_username:
+                        description:
+                        - Username for device discovery during POAP and RMA discovery.
+                        type: str
+                    discovery_password:
+                        description:
+                        - Password for device discovery during POAP and RMA discovery.
+                        type: str
+                    serial_number:
```
mtarking: Can we place do `new_serial_number`?

plugins/modules/nd_manage_switches.py
```diff
@@ -0,0 +1,512 @@
...
+                        description:
+                        - Serial number of switch to Bootstrap for RMA.
+                        type: str
+                        required: true
+                    old_serial:
```
mtarking: And can we do `old_serial_number`?

plugins/modules/nd_manage_switches.py
```diff
@@ -0,0 +1,512 @@
...
+                            gateway:
+                                description:
+                                - Gateway IP with subnet mask (e.g., 192.168.0.1/24).
+                                type: str
+            rma:
+                description:
+                - RMA an existing switch with a new one.
+                - Please note that the existing switch should be configured and deployed in maintenance mode.
+                - Please note that the existing switch being replaced should be shutdown state or out of network.
+                type: list
```
mtarking, line 230: Same question as above (poap), is there a reason this is a `list`?

plugins/modules/nd_manage_switches.py
```diff
@@ -0,0 +1,512 @@
...
+                                - List of module models in the switch.
+                                type: list
+                                elements: str
+                                required: true
+                            gateway:
+                                description:
+                                - Gateway IP with subnet mask (e.g., 192.168.0.1/24).
+                                type: str
+                                required: true
+                        - Serial number of new replacement switch.
```
mtarking: This sees out of place or am I missing something?

mtarking (changes requested):

plugins/module_utils/utils/manage_switches/__init__.py
mikewiebe: Drop leading `utils` directory to avoid conflict with utils.py

plugins/modules/nd_manage_switches.py
mikewiebe: Supported Use Cases:

* Preprovision without POAP
* Preprovision followed by second run with POAP (triggered by S/N change)
* POAP only using S/N under POAP key

* If both preprovision and poap S/N provided day 0 (currently fails if preprovision device is not configured)

plugins/modules/nd_manage_switches.py
mikewiebe: New structure will address this

plugins/modules/nd_manage_switches.py
mikewiebe: New structure will address this

plugins/modules/nd_manage_switches.py
AKDRG: Check mode has been added

plugins/modules/nd_manage_switches.py
AKDRG: Addressed.

plugins/modules/nd_manage_switches.py
AKDRG: Changes made.

plugins/modules/nd_manage_switches.py
AKDRG: Changed.

plugins/modules/nd_manage_switches.py
AKDRG: Changed.

plugins/modules/nd_manage_switches.py
AKDRG: Have reworded this, please check and let me know if we need to change.

plugins/modules/nd_manage_switches.py
AKDRG: Done.

plugins/modules/nd_manage_switches.py
AKDRG: Have split up POAP and Preprovision.

plugins/modules/nd_manage_switches.py
AKDRG: Done.

plugins/modules/nd_manage_switches.py
AKDRG: Removed it.

plugins/modules/nd_manage_switches.py
AKDRG: Gathered is now supported in the module.

plugins/module_utils/nd_switch_resources.py
AKDRG: Most of the changes have been integrated. Query for "after " and performance control knob is being explored.

plugins/module_utils/utils/manage_switches/__init__.py
AKDRG: Handled the changes

mikewiebe:
@AKDRG Couple of testing notes:

* We should support `state: overridden` with an empty config list which would result in the removal of all switches from the fabric.

```yaml
    - name: Manage Fabric Switches
      cisco.nd.nd_manage_switches:
        fabric: mike_test
        state: overridden
        deploy: false
        config: []
```

Currently I see the following:

```bash
"msg": "'config' is required for 'overridden' state."
```

* If we add switches using `deploy: false` and then in a subsequent run change the flag to `deploy: true` this should result in the switches getting deployed if they are not already deployed.

* After I add switches to the fabric and then re-run the playbook it runs very fast since the switches are added but the change flag still reflects `true` even though nothing has changed

* I added switches using `state: overridden` and then tried to add a preprovision switch but got the following error messaged indicating I have to use `state: merged` for pre-provision.  Is this by design?  This also seems to impact `state: deleted` so that we cannot delete pre-provisioned devices so looks like this needs to be fixed.

```bash
"msg": "Configuration validation failed for config index 13: [{'type': 'value_error', 'loc': (), 'msg': \"Value error, POAP/Pre-provision operations require 'merged' state, got 'overridden' (switch: 192.168.10.1)\", 'input': {'seed_ip': '192.168.10.1', 'username': 'admin', 'password': '********', 'preprovision': {'serial_number': 'SAL1234ABCD', 'model': 'N9K-C93180YC-EX', 'version': '10.3(1)', 'hostname': 'leaf-preprov', 'config_data': {'models': ['N9K-C93180YC-EX'], 'gateway': '192.168.10.1/24'}, 'discovery_username': None, 'discovery_password': None, 'image_policy': None}, 'auth_proto': 'MD5', 'role': 'leaf', 'preserve_config': False, 'poap': None, 'rma': None}, 'ctx': {'error': ValueError(\"POAP/Pre-provision operations require 'merged' state, got 'overridden' (switch: 192.168.10.1)\")}, 'url': 'https://errors.pydantic.dev/2.12/v/value_error'}]"
}
```

* Item to discuss: We should consider a query IF the user provides pre-provisioned and or poap devices in the playbook and fail fast if the boostrap settings don't support these workflows

plugins/modules/nd_manage_switches.py
```diff
@@ -0,0 +1,550 @@
...
+    state: merged
+
+- name: Add multiple switches
+  cisco.nd.nd_manage_switches:
+    fabric: my-fabric
+    config:
+      - seed_ip: 192.168.10.201
+        username: admin
+        password: "{{ switch_password }}"
+        role: leaf
```
mikewiebe, line 292: Lets add `preserve_config: false` to this `Add multiple switches` example

plugins/modules/nd_manage_switches.py
```diff
@@ -0,0 +1,550 @@
...
+        role: spine
+    state: merged
+
+- name: Preprovision a switch via POAP
+  cisco.nd.nd_manage_switches:
+    fabric: my-fabric
+    config:
+      - seed_ip: 192.168.10.1
+        username: admin
+        password: "{{ switch_password }}"
```
mikewiebe, line 307: Let's add `role` to these examples to be more explicit

mikewiebe:
I was not able to add more then one switch to the fabric until I added the following line

```python
diff --git a/plugins/module_utils/nd_config_collection.py b/plugins/module_utils/nd_config_collection.py
index 832cc13..7610d05 100644
--- a/plugins/module_utils/nd_config_collection.py
+++ b/plugins/module_utils/nd_config_collection.py
@@ -56,7 +56,7 @@ class NDConfigCollection:
 
         key = self._extract_key(item)
 
-        if key in self._index:
+        if key is not None and key in self._index:
             raise ValueError(f"Item with identifier {key} already exists. Use replace() to update")
 
         position = len(self._items)
```

This is just a workaround but I am wondering if `merge` function from NDConfigCollection should be used instead?

plugins/modules/nd_manage_switches.py
AKDRG: Fixed.

plugins/modules/nd_manage_switches.py
AKDRG: Fixed.

tests/integration/targets/nd_manage_switches/tests/deleted.yaml
```diff
@@ -0,0 +1,142 @@
+---
```
mikewiebe, line 1: Why are these tests under a directory called `nd`?  Can we just move them directly under `tests`?

AKDRG: Fixed.

plugins/modules/nd_manage_switches.py
```diff
@@ -0,0 +1,504 @@
...
+        - replaced
+        - overridden
+        - deleted
+        - gathered
+    save:
+        description:
+        - Save/Recalculate the configuration of the fabric after inventory is updated.
+        type: bool
+        default: true
+    deploy:
```
mikewiebe: We need to support separate save and deploy options

Use the following property structures

```yaml
config_actions:
  save: true
  deploy: true
  type: switch [Enum Choices: switch and global]
```

If `config_actions.save` is `false` and `config_actions.deploy` is `true` we should fail

plugins/action/nd_switches_validate.py
```diff
@@ -0,0 +1,255 @@
...
+switch in the ND API response (fabricManagementIp == seed_ip,
+switchRole == role).
+
+Supports an optional ``mode`` argument:
+  - ``"both"`` (default): match by seed_ip AND role.
+  - ``"ip"``:   match by seed_ip only  (role is ignored).
+  - ``"role"``: match by role only     (seed_ip is ignored).
+"""
+
+from __future__ import absolute_import, division, print_function
```
allenrobel: Remove.  No longer needed.

Comment applies for all files.

plugins/action/nd_switches_validate.py
```diff
@@ -0,0 +1,255 @@
...
+
+__metaclass__ = type  # pylint: disable=invalid-name
```
allenrobel: Remove. No longer needed.

Comment applies for all files.

plugins/action/nd_switches_validate.py
```diff
@@ -0,0 +1,255 @@
...
+
+import json
+from typing import Any, Dict, List, Optional, Union
```
allenrobel: Remove imports for `Dict`, `List`, `Optional`, `Union` and use modern type annotations.

Easiest way to do this if you're using Claude Code is to ask Claude to update his memory, or Claude.md file, to always use modern type annotations for all new code.  You can also ask him to migrate all existing legacy type annotations to modern annotations.

Comment applies for all files in this PR.

plugins/action/nd_switches_validate.py
```diff
@@ -0,0 +1,255 @@
...
+
+                seed_ip_match = (seed_ip is not None and ip_address is not None and ip_address == seed_ip) or bool(ignore_fields["seed_ip"])
+                role_match = (role_expected is not None and switch_role is not None and switch_role == role_expected) or bool(ignore_fields["role"])
+
+                if seed_ip_match and role_match:
+                    matched_indices.add(i)
+                    found_match = True
+                    if ignore_fields["seed_ip"]:
+                        break
+                elif (seed_ip_match and role_expected is not None and switch_role is not None and switch_role != role_expected) or ignore_fields["role"]:
```
allenrobel: Doesn't the `or ignore_fields["role"]` at the end of this conditional mean that in role-only mode, this branch fires for every unmatched `nd_item` in the inner loop, so would record spurious role mismatches against arbitrary pairings?  Shouldn't role-only mode not be producing role mismatch diagnostics at all?  Or, if it should, maybe the logic needs rethinking?

plugins/action/nd_switches_validate.py
```diff
@@ -0,0 +1,255 @@
...
+        if not config_data and not nd_data_list:
+            self.response = True
+            return self
+        if not config_data or not nd_data_list:
+            self.response = False
+            return self
+
+        missing_ips = []
+        role_mismatches = {}
+        nd_data_copy = nd_data_list.copy()
```
allenrobel: `nd_data_copy` is never mutated, right?  So the copy() is not needed and `nd_data_list` could be iterated directly.

plugins/action/nd_switches_validate.py
```diff
@@ -0,0 +1,255 @@
...
+
+display = Display()
+
+
+# ---------------------------------------------------------------------------
+# Validation orchestration model
+# ---------------------------------------------------------------------------
+
+
+class SwitchesValidate(BaseModel):
```
allenrobel: Can you separate this out into a separate file that has no Ansible dependencies?  Below is a pure python implementation that uses modern type annotations and fixes some of the issues raised by my comments.

Diagnostics (previously surfaced with Display) are now collected in self.missing_ips and self.role_mismatches. The caller decides how to surface them.

This is also now trivially testable since we don't have to mock ActionBash and capture Display output to test the same logic e.g.:

```python
def test_missing_seed_ip():
    v = SwitchesValidate(
        config_data=[{"seed_ip": "10.0.0.1", "role": "leaf"}],
        nd_data=[{"fabricManagementIp": "10.0.0.2", "switchRole": "leaf"}],
    )
    assert v.response is False
    assert v.missing_ips == ["10.0.0.1"]
    assert v.role_mismatches == {}
```

New SwitchesValidate:

```python
# -*- coding: utf-8 -*-

# Copyright: (c) 2026, Cisco Systems

"""Pure-Python validator for ND switch inventory data.

Validates that every entry in a user-supplied config list has a matching
switch in the ND API response. No Ansible dependencies — reusable from
any Python context.

Matching modes are expressed via ``ignore_fields``:
  - ``{"seed_ip": 0, "role": 0}`` (default): match by seed_ip AND role.
  - ``{"seed_ip": 0, "role": 1}``: match by seed_ip only.
  - ``{"seed_ip": 1, "role": 0}``: match by role only.
"""

from typing import Any

from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import (
    BaseModel,
    Field,
    ValidationError,
    field_validator,
    model_validator,
)
from ansible_collections.cisco.nd.plugins.module_utils.models.manage_switches.config_models import (
    SwitchConfigModel,
)
from ansible_collections.cisco.nd.plugins.module_utils.models.manage_switches.switch_data_models import (
    SwitchDataModel,
)


class SwitchesValidate(BaseModel):
    """Matches playbook config entries against live ND inventory.

    After construction, inspect:
      - ``response``: True if all config entries matched, False otherwise.
      - ``missing_ips``: seed_ips from config with no matching ND switch.
      - ``role_mismatches``: {seed_ip: {"expected_role": ..., "response_role": ...}}
    """

    config_data: list[SwitchConfigModel] | None = None
    nd_data: list[SwitchDataModel] | None = None
    ignore_fields: dict[str, int] = Field(default_factory=lambda: {"seed_ip": 0, "role": 0})
    response: bool | None = None
    missing_ips: list[str] = Field(default_factory=list)
    role_mismatches: dict[str, dict[str, Any]] = Field(default_factory=dict)

    @field_validator("config_data", mode="before")
    @classmethod
    def parse_config_data(cls, value: Any) -> list[SwitchConfigModel] | None:
        """Coerce raw dicts into SwitchConfigModel instances."""
        if value is None:
            return None
        if isinstance(value, dict):
            return [SwitchConfigModel.model_validate(value)]
        if isinstance(value, list):
            try:
                return [
                    SwitchConfigModel.model_validate(item) if isinstance(item, dict) else item
                    for item in value
                ]
            except (ValidationError, ValueError) as e:
                raise ValueError(f"Invalid format in Config Data: {e}")
        raise ValueError("Config Data must be a dict, list of dicts, or None.")

    @field_validator("nd_data", mode="before")
    @classmethod
    def parse_nd_data(cls, value: Any) -> list[SwitchDataModel] | None:
        """Coerce raw ND API switch dicts into SwitchDataModel instances."""
        if value is None:
            return None
        if isinstance(value, list):
            try:
                return [
                    SwitchDataModel.from_response(item) if isinstance(item, dict) else item
                    for item in value
                ]
            except (ValidationError, ValueError) as e:
                raise ValueError(f"Invalid format in ND Response: {e}")
        raise ValueError("ND Response must be a list of dictionaries.")

    @field_validator("ignore_fields", mode="before")
    @classmethod
    def default_ignore_fields(cls, value: dict[str, int] | None) -> dict[str, int]:
        """Ensure both keys are always present, defaulting to 0 (strict match)."""
        if value is None:
            return {"seed_ip": 0, "role": 0}
        return {"seed_ip": value.get("seed_ip", 0), "role": value.get("role", 0)}

    @model_validator(mode="after")
    def validate_lists_equality(self) -> SwitchesValidate:
        """Match every config entry against the live ND switch inventory."""
        config_data = self.config_data
        nd_data_list = self.nd_data
        ignore_seed_ip = bool(self.ignore_fields["seed_ip"])
        ignore_role = bool(self.ignore_fields["role"])

        # Both empty → nothing to validate, treat as success.
        # Exactly one empty → mismatch, treat as failure.
        if not config_data and not nd_data_list:
            self.response = True
            return self
        if not config_data or not nd_data_list:
            self.response = False
            return self

        matched_indices: set[int] = set()

        for config_item in config_data:
            seed_ip = config_item.seed_ip
            role_expected = config_item.role  # SwitchRole enum or None
            found_match = False
            role_mismatch_for_this_entry: dict[str, Any] | None = None

            for i, nd_item in enumerate(nd_data_list):
                if i in matched_indices:
                    continue

                ip_address = nd_item.fabric_management_ip
                switch_role = nd_item.switch_role  # SwitchRole enum or None

                seed_ip_match = ignore_seed_ip or (
                    seed_ip is not None
                    and ip_address is not None
                    and ip_address == seed_ip
                )
                role_match = ignore_role or (
                    role_expected is not None
                    and switch_role is not None
                    and switch_role == role_expected
                )

                if seed_ip_match and role_match:
                    matched_indices.add(i)
                    found_match = True
                    break

                # Full match failed. If seed_ip matched but role didn't,
                # remember it as a candidate role mismatch. We only commit
                # this diagnostic if no better match turns up later.
                if (
                    not ignore_role
                    and seed_ip_match
                    and not ignore_seed_ip
                    and role_expected is not None
                    and switch_role is not None
                    and switch_role != role_expected
                ):
                    role_mismatch_for_this_entry = {
                        "nd_index": i,
                        "ip_address": ip_address,
                        "expected_role": role_expected.value,
                        "response_role": switch_role.value,
                    }

            if not found_match:
                if role_mismatch_for_this_entry is not None:
                    # Seed IP was found but role disagreed — record and
                    # consume that nd_item so it's not re-matched.
                    rm = role_mismatch_for_this_entry
                    matched_indices.add(rm["nd_index"])
                    self.role_mismatches[seed_ip or rm["ip_address"]] = {
                        "expected_role": rm["expected_role"],
                        "response_role": rm["response_role"],
                    }
                elif seed_ip is not None:
                    self.missing_ips.append(seed_ip)

        self.response = not self.missing_ips and not self.role_mismatches
        return self
```

plugins/action/tests/integration/_nd_switches_validate.py
```diff
@@ -0,0 +1,255 @@
...
+
+        return self
+
+
+# ---------------------------------------------------------------------------
+# Action plugin
+# ---------------------------------------------------------------------------
+
+
+class ActionModule(ActionBase):
```
allenrobel, line 106: Below is a rewritten ActionModule using modern type annotations, that should go into a separate file that contains only code that requires the Ansible dependencies.

```python
# -*- coding: utf-8 -*-

# Copyright: (c) 2026, Cisco Systems

"""ND Switches Validation Action Plugin (thin Ansible adapter)."""

import json
from typing import Any

from ansible.plugins.action import ActionBase
from ansible.utils.display import Display

from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import (
    HAS_PYDANTIC,
)

try:
    from ansible_collections.cisco.nd.plugins.module_utils.validators.switches_validator import (
        SwitchesValidate,
    )

    HAS_VALIDATOR = True
except ImportError:
    HAS_VALIDATOR = False

display = Display()


class ActionModule(ActionBase):
    """Ansible action plugin for validating ND switch inventory data.

    Task arguments:
        nd_data   (dict): The registered result of a cisco.nd.nd_rest GET call.
        test_data (list|dict): Expected switch entries, each with ``seed_ip``
                               and optionally ``role``.
        changed   (bool, optional): If provided and False, the task fails
                                    immediately (used to assert an upstream
                                    operation produced a change).
        mode      (str, optional): ``"both"`` (default), ``"ip"``, or ``"role"``.
    """

    def run(
        self,
        tmp: Any = None,
        task_vars: dict[str, Any] | None = None,
    ) -> dict[str, Any]:
        results = super().run(tmp, task_vars)
        results["failed"] = False

        if not HAS_PYDANTIC or not HAS_VALIDATOR:
            results["failed"] = True
            results["msg"] = (
                "pydantic and the ND collection validators are required "
                "for nd_switches_validate"
            )
            return results

        args = self._task.args
        nd_data = args["nd_data"]
        test_data = args["test_data"]

        # Fail fast if the caller signals that no change occurred when one was expected.
        if "changed" in args and not args["changed"]:
            results["failed"] = True
            results["msg"] = 'Changed is "false"'
            return results

        # Fail fast if the upstream nd_rest task itself failed.
        if nd_data.get("failed"):
            results["failed"] = True
            results["msg"] = nd_data.get("msg", "ND module returned a failure")
            return results

        switches = nd_data.get("current", {}).get("switches", [])

        if isinstance(test_data, dict):
            test_data = [test_data]

        if not switches and not test_data:
            results["msg"] = "Validation Successful!"
            return results

        if not switches:
            results["failed"] = True
            results["msg"] = "No switches found in ND response"
            return results

        # Resolve matching mode.
        ignore_fields: dict[str, int] = {"seed_ip": 0, "role": 0}
        mode = args.get("mode", "both").lower()
        if mode == "ip":
            ignore_fields["role"] = 1
        elif mode == "role":
            ignore_fields["seed_ip"] = 1

        validation = SwitchesValidate(
            config_data=test_data,
            nd_data=switches,
            ignore_fields=ignore_fields,
        )

        if validation.response:
            results["msg"] = "Validation Successful!"
            return results

        # Surface diagnostics via Ansible's Display so they show up with -v.
        display.display("Invalid Data:")
        if validation.missing_ips:
            display.display(f"  Missing IPs: {validation.missing_ips}")
        if validation.role_mismatches:
            display.display(
                f"  Role mismatches: {json.dumps(validation.role_mismatches, indent=2)}"
            )

        results["failed"] = True
        results["msg"] = "Validation Failed! Please check output above."
        results["missing_ips"] = validation.missing_ips
        results["role_mismatches"] = validation.role_mismatches
        return results
```

allenrobel (changes requested):
Added several comments.  Mainly around removing all legacy type annotations and removing unneeded `__metaclass = type`.

Added comment regarding divorcing as much code as possible from Ansible dependencies.  So, split the action plugin file into two files; one containing Ansible dependencies and the other free of Ansible dependencies.

Also, possible bug with conditional in SwitchesValidate `elif` conditional.

plugins/action/nd_switches_validate.py
AKDRG: Fixed.

plugins/action/nd_switches_validate.py
AKDRG: Fixed.

plugins/action/nd_switches_validate.py
AKDRG: Fixed.

plugins/action/nd_switches_validate.py
AKDRG: Thank you Allen, have used your code suggestions in the split - out validate files.

plugins/action/tests/integration/_nd_switches_validate.py
AKDRG: Done

plugins/action/nd_switches_validate.py
AKDRG: Fixed. Role-only mode no longer records spurious role_mismatches.  The mismatch path now requires not ignore_seed_ip, so it only fires when IP matching is active.

plugins/action/nd_switches_validate.py
AKDRG: Fixed.

plugins/module_utils/endpoints/mixins.py
```diff
@@ -68,16 +74,28 @@ class LoginIdMixin(BaseModel):
...
+    """Mixin for endpoints that require a pagination offset parameter."""
 
-    node_name: Optional[str] = Field(default=None, min_length=1, description="Node name")
+    offset: Optional[int] = Field(default=None, ge=0, description="Pagination offset")
+
+
+class OffsetMixin(BaseModel):
+    """Mixin for endpoints that require a pagination offset parameter."""
+
+    offset: Optional[int] = Field(default=None, ge=0, description="Pagination offset")
```
allenrobel: OffsetMixin defined twice (also line 89); the second shadows the first. Remove this duplicate.

plugins/module_utils/manage_switches/nd_switch_resources.py
```diff
@@ -0,0 +1,3593 @@
...
+# Copyright: (c) 2026, Akshayanat C S (@achengam) <achengam@cisco.com>
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+"""Manage ND fabric switch lifecycle workflows.
+
+This module validates desired switch state, performs discovery and fabric
+operations, and coordinates POAP and RMA workflows.
+"""
+
```
allenrobel, line 12: No unit tests for the 3,600-line engine, models, or fabric_inventory.py — only integration tests.

Please add unit coverage.

ansible_cisco_log_r.json
```diff
@@ -0,0 +1,36 @@
...
+                "class": "logging.Formatter",
+                "format": "%(asctime)s - %(levelname)s - [%(name)s.%(funcName)s.%(lineno)d] %(message)s"
+            }
+        },
+        "handlers": {
+            "file": {
+                "class": "logging.handlers.RotatingFileHandler",
+                "formatter": "standard",
+                "level": "DEBUG",
+                "filename": "/Users/achengam/Documents/Ansible_Dev/NDBranch/ansible_collections/cisco/nd/Manage_Switches_R1.txt",
```
allenrobel: Please remove this file.

plugins/module_utils/endpoints/v1/manage/manage_fabrics_switchactions.py
```diff
@@ -0,0 +1,601 @@
...
+    params = SwitchActionsRemoveEndpointParams(force=True, ticket_id="CHG12345")
+    query_string = params.to_query_string()
+    # Returns: "force=true&ticketId=CHG12345"
+    ```
+    """
+
+    force: bool | None = Field(default=None, description="Force removal of switches")
+
+
+class SwitchActionsTicketEndpointParams(TicketIdMixin, EndpointQueryParams):
```
allenrobel: This is duplicated by manage_fabrics_switches.py:268

Can we lift this into a file and import into both places to reduce code duplication?

plugins/module_utils/endpoints/v1/manage/manage_fabrics_switches.py
```diff
@@ -0,0 +1,461 @@
...
+        """Return the HTTP verb for this endpoint."""
+        return HttpVerbEnum.POST
+
+
+# ============================================================================
+# Per-Switch Action Endpoints
+# ============================================================================
+
+
+class SwitchActionsTicketEndpointParams(TicketIdMixin, EndpointQueryParams):
```
allenrobel: This is duplicated by manage_fabrics_switchactions.py:73

Can we lift this into a file and import into both places to reduce code duplication?

plugins/module_utils/models/manage_switches/switch_data_models.py
```diff
@@ -0,0 +1,338 @@
...
+
+__all__ = [
+    "TelemetryIpCollection",
+    "VpcData",
+    "SwitchMetadata",
+    "AdditionalSwitchData",
+    "AdditionalAciSwitchData",
+    "Metadata",
+    "SwitchDataModel",
+]
```
allenrobel: AFAIK, `__all__` is not required except in `__init__.py` files (which Ansible sanity doesn't allow anyway).

Please remove __all__ unless you have a reason for keeping it (please let me know, since my understanding is that this serves no purpose, but I'm open to learning something if I'm mistaken).

plugins/module_utils/models/manage_switches/switch_data_models.py
```diff
@@ -0,0 +1,338 @@
...
+    )
+
+
+class VpcData(NDNestedModel):
+    """
+    vPC pair configuration and operational status for a switch.
+    """
+
+    identifiers: ClassVar[list[str]] = []
+    vpc_domain: int = Field(alias="vpcDomain", ge=1, le=1000, description="vPC domain ID")
```
allenrobel, line 66: Just checking, but do we need mode="before" str->int coercion here?  What does vpcDomain look like in the wire response?  If it's an int on the wire, please ignore.

allenrobel (changes requested):
One overall comment is that nd_v2.py is going to be deprecated, so it would be good not to leverage this file and move exclusively to Gaspard's framework.  But this can wait until after it's merged if OK with Matt/Mike.

I left a few more comments.  Can you look into them?

Thank you Akshay!

plugins/module_utils/endpoints/mixins.py
AKDRG: Fixed. Great Catch, Thanks!

plugins/module_utils/manage_switches/nd_switch_resources.py
AKDRG: Added now.

ansible_cisco_log_r.json
AKDRG: Removed

plugins/module_utils/endpoints/v1/manage/manage_fabrics_switchactions.py
AKDRG: Done.

plugins/module_utils/endpoints/v1/manage/manage_fabrics_switches.py
AKDRG: Done.

plugins/module_utils/models/manage_switches/switch_data_models.py
AKDRG: Fixed.

AKDRG: All is not being used anywhere. It was kept there. As you rightly pointed out, it's not needed, so removed.

plugins/module_utils/models/manage_switches/switch_data_models.py
AKDRG: {
  "hostname": "nac-mcfg-fabric1-leaf1",
  "fabricManagementIp": "192.168.1.1",
  "switchId": "1A2BC34567D",
  "vpcConfigured": true,
  "vpcData": {
    "vpcDomain": 1,
    "vpcRole": "primary"
  }
}

The value is an int from the ND Data, and defined as an int from the DataModel too. So we don't need the str-int with before here.

mikewiebe (approved):
LGTM

allenrobel (commented):
Full review pass on this PR. Overall the state engine held up well under scrutiny — several suspicious paths I chased (credentials-save skip, `config_data` deref, post-`bulk_add` serial tracking) turned out to be properly guarded by the `SwitchConfigModel` validators. Nice work there.

Four items below as inline comments. I'd address these in this PR:

1. Empty `publicKey`/`fingerPrint` sent to ND on RMA/bootstrap-import flows (correctness)
2. Empty switch list treated as an API failure in the wait helpers (correctness)
3. Shared `module_utils/utils.py` now pulls in the manage endpoints package for every module in the collection (coupling / public-API one-way door)
4. `nd_switches_validate` ships as a public, undocumented action plugin with sanity ignores (public-API one-way door — the rename is cheap now, breaking after release)

Lower-severity findings (exit-path inventory re-fetch, state-handler/builder duplication, `issubset()` relocation changelog note, the `filter` encoding special case in `to_query_string`) are being deferred to separate issues so they don't hold this PR.

plugins/module_utils/manage_switches/nd_switch_resources.py
```diff
@@ -0,0 +1,3594 @@
...
+        )
+
+        bs_data = bootstrap_data.get("data") or {}
+
+        gateway_ip_mask = bootstrap_data.get("gatewayIpMask") or bs_data.get("gatewayIpMask", "")
+        data_models = bs_data.get("models", [])
+        model = bootstrap_data.get("model", "")
+        software_version = bootstrap_data.get("softwareVersion", "")
+        public_key = bootstrap_data.get("publicKey", "")
+        finger_print = bootstrap_data.get("fingerPrint") or bootstrap_data.get("fingerprint", "")
```
allenrobel, line 2411: **Correctness:** if the bootstrap entry doesn't yet contain `publicKey`/`fingerPrint` (e.g. the switch hasn't fully called home when we query), these default to `""` and the RMA payload is sent anyway. ND then rejects the RMA — or worse, accepts a switch without fingerprint verification — and nothing in the module output points at the missing field. The same pattern is in `_build_bootstrap_import_model` at lines 1778-1779.

Suggest validating both values are non-empty before building the model and failing with a message that names the serial number and the missing field, so the user knows to wait for the switch to finish calling home and retry.

plugins/module_utils/manage_switches/utils.py
```diff
@@ -0,0 +1,876 @@
...
+        """
+        try:
+            response = self.nd.request(
+                self.ep_switches_get.path,
+                verb=self.ep_switches_get.verb,
+            )
+            switch_data = response.get("switches", [])
+            if not switch_data:
+                self.log.error("No switch data returned for fabric")
+                return None
```
allenrobel: **Correctness:** `{"switches": []}` is a legitimate response (fabric briefly empty, or pending switches removed between polling cycles), but this treats it the same as a request failure: callers receive `None` and either abort the wait as an "API failure" (`_poll_system_mode`, `_wait_for_discovery_state`) or spin to full timeout (`_wait_for_switches_in_fabric`).

Suggest returning `[]` here and reserving `None` for the actual request-error path in the `except` block below, letting each caller decide what an empty inventory means for its wait condition.

plugins/module_utils/utils.py
```diff
@@ -2,10 +2,21 @@
...
 
-from __future__ import absolute_import, division, print_function
-
+import logging
+import time
 from copy import deepcopy
-from typing import Any, Dict, List, Union
+from typing import Any, Dict, List, Optional, Union
+
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.v1.manage.manage_fabrics import (
```
allenrobel: **Coupling / public API:** these endpoint imports execute at import time for *every* consumer of `utils.py` — and since `nd.py` now imports `issubset` from here, every legacy `nd_*`/`ndi_*` module in the collection transitively loads the manage endpoints package on every run. Any future import-time problem in `endpoints/v1/manage` would break the whole collection, not just `nd_manage_switches`.

It's also a one-way door: `module_utils` is public API once released, so relocating `SwitchOperationError`/`ApiDataChecker`/`FabricUtils` after a release is a breaking change, while moving them now is a mechanical file move.

Suggest moving the three classes (and these endpoint imports) into `plugins/module_utils/manage_switches/` (e.g. alongside or into the existing `utils.py` there), keeping this shared module free of endpoint dependencies.

plugins/action/tests/integration/_nd_switches_validate.py
```diff
@@ -0,0 +1,117 @@
+# -*- coding: utf-8 -*-
```
allenrobel, line 1: **Public API / docs:** on merge, `cisco.nd.nd_switches_validate` becomes a public plugin name, but it's only invoked from the integration-test YAML, has no `DOCUMENTATION`/`EXAMPLES`/`RETURN`, and the gap is suppressed via `action-plugin-docs` entries added to all four `tests/sanity/ignore-2.1x.txt` files.

Since it's a test-support helper, suggest renaming it `_nd_switches_validate` (underscore prefix marks it private and exempts it from the docs check, so the four ignore lines can be dropped). Alternatively, add a full `DOCUMENTATION` block and drop the ignores. The rename is a one-line-per-callsite change now, but a breaking change once 2.x ships with the public name.

Relatedly, `module_utils/validators/switches_validator.py` backing this plugin is an inventory matcher rather than a field validator — worth relocating/renaming in a follow-up so the `validators/` namespace stays coherent.

plugins/module_utils/utils.py
AKDRG: Agreed. Fixed!

plugins/module_utils/manage_switches/nd_switch_resources.py
AKDRG: Fixed it with the help of Pydantic Validators.

plugins/module_utils/manage_switches/utils.py
AKDRG: Done.

plugins/action/tests/integration/_nd_switches_validate.py
AKDRG: _nd_switches_validate.py is now under: plugins/action/tests/integration/_nd_switches_validate.py

Integration test YAML now calls it as:
cisco.nd.tests.integration._nd_switches_validate:

The old collection-level action plugin path is gone:
plugins/action/_nd_switches_validate.py

Since action-plugin-docs still requires a matching module even for the private underscored action plugin, the sanity ignore is maintained in:
tests/sanity/ignore-2.16.txt
tests/sanity/ignore-2.17.txt
tests/sanity/ignore-2.18.txt
tests/sanity/ignore-2.19.txt

The switch inventory matcher was moved from: plugins/module_utils/validators/switches_validator.py 
to:
plugins/module_utils/manage_switches/switches_inventory_matcher.py

AKDRG: The folder structure (plugins/action/tests/integration/) is in line with the VPC intg action plugin which is already present in the develop branch

mikewiebe: @AKDRG and @allenrobel The only real way to remove this from  the ignore.txt files is to add a module with the same name that has the documentation header.  IMO we should just stick with the ignores.

allenrobel: Sounds good guys.

allenrobel (approved):
LGTM

mtarking (approved):

PR #202, opened 2026-03-12
Add a new network resource module nd_local_user for local users on Nexus Dashboard v4.1.0 or higher (DCNE-543)

gmicol:


plugins/module_utils/nd_state_machine.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+class NDStateMachine:
+    """
+    Generic Network Resource Module for Nexus Dashboard.
+    """
+
+    def __init__(self, module: AnsibleModule, model_orchestrator: Type[NDBaseOrchestrator]):
+        """
+        Initialize the Network Resource Module.
```
lhercot:
```suggestion
        Initialize the State Machine
```

plugins/module_utils/nd_config_collection.py
```diff
@@ -0,0 +1,219 @@
...
+        items = [model_class.from_config(item_data, **kwargs) for item_data in data]
+        return cls(model_class=model_class, items=items)
+
+    @classmethod
+    def from_api_response(cls, response_data: List[Dict[str, Any]], model_class: type[NDBaseModel], **kwargs) -> "NDConfigCollection":
+        """
+        Create collection from API response.
+        """
+        items = [model_class.from_response(item_data, **kwargs) for item_data in response_data]
+        return cls(model_class=model_class, items=items)
```
lhercot: Shouldn't these be static methods?

lhercot (changes requested):

plugins/modules/nd_local_user.py
```diff
@@ -0,0 +1,211 @@
...
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+ANSIBLE_METADATA = {"metadata_version": "1.1", "status": ["preview"], "supported_by": "community"}
+
+DOCUMENTATION = r"""
+---
+module: nd_local_user
+version_added: "1.4.0"
```
akinross: 1.4.0 is already released, so I think this should be a different version. Or is release handling this automatically?

tests/integration/targets/nd_local_user/tasks/main.yml
```diff
@@ -0,0 +1,651 @@
...
+      - nm_override_local_users.proposed.1.security_domains.0.name == "all"
+      - nm_override_local_users.proposed.1.security_domains.0.roles | length == 1
+      - nm_override_local_users.proposed.1.security_domains.0.roles.0 == "super_admin"
+      - nm_override_local_users.proposed.1.time_interval_limitation == 0
+      - nm_override_local_users.proposed.2.login_id == "ansible_local_user_3"
+      - nm_override_local_users.proposed.2.security_domains.0.name == "all"
+
+
+# DELETE
+- name: Delete local user by name (check mode)
```
akinross: Just a generic question regarding delete functionality, is it possible to delete all existing local users by providing an empty list or are specified items only deleted?

akinross (changes requested):

mikewiebe (dismissed):

plugins/modules/nd_local_user.py
samiib: This should be 1.6.0 now.

plugins/modules/nd_local_user.py
```diff
@@ -0,0 +1,211 @@
...
+short_description: Manage local users on Cisco Nexus Dashboard
+description:
+- Manage local users on Cisco Nexus Dashboard (ND).
+- It supports creating, updating, querying, and deleting local users.
```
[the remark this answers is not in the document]
mikewiebe: Agree  with @akinross that this should be deleted since the module only supports the following states:

`choices: [ merged, replaced, overridden, deleted ]`

As an additional data point, I have removed the query function from the `nd_manage_fabric_*` modules.  Back in July we discussed supporting the following instead of a query state in the module

The query state should be moved to a _info module as specified by "MUST NOT query information using special state option values like get, list, query, or info

  * create new _info or _facts modules instead
  * (for more information, refer to the Developing modules guidelines)." in https://docs.ansible.com/ansible/devel/community/collection_contributors/collection_requirements.html and specifically holding our inclusion here: https://github.com/ansible-collections/ansible-inclusion/discussions/77

We need to discuss _info vs. _facts modules and what we want to support here moving forward.

tests/integration/targets/nd_local_user/tasks/main.yml
```diff
@@ -0,0 +1,651 @@
...
+- name: Ensure local users do not exist before test starts
+  cisco.nd.nd_local_user:
+    <<: *nd_info
+    config:
+      - login_id: ansible_local_user
+      - login_id: ansible_local_user_2
+      - login_id: ansible_local_user_3
+    state: deleted
+
+# CREATE
```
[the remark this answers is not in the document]
mikewiebe: If the resource(s) do not exist in ND then there should be no functional difference between `merged, replaced, overridden`.  The difference only comes if the resources already exists for `merged/replaced` and then in the case of `overridden` any resources (users in this case) that exist in ND but don't in the playbook should be purged.

plugins/module_utils/nd_state_machine.py
```diff
@@ -0,0 +1,176 @@
...
+                    # Replace or create
+                    if diff_status == "changed":
+                        self.existing.replace(proposed_item)
+                    else:
+                        self.existing.add(proposed_item)
+                    final_item = proposed_item
+
+                # Execute API operation
+                if diff_status == "changed":
+                    if not self.module.check_mode:
```
mikewiebe, line 100: Seems like `self.sent` here is only populated inside the check_mode guard `i not self.module.check_mode`.  Should this be moved outside the guard?

```python
if not self.module.check_mode:
    self.model_orchestrator.update(final_item)
self.sent.add(final_item)  # ← always track what would be sent
```

mikewiebe: Check for all states to ensure check_mode is supported properly.

plugins/module_utils/nd_state_machine.py
```diff
@@ -0,0 +1,176 @@
...
+            self.proposed = NDConfigCollection(model_class=self.model_class)
+            for config in self.module.params.get("config", []):
+                try:
+                    # Parse config into model
+                    item = self.model_class.from_config(config)
+                    self.proposed.add(item)
+                except ValidationError as e:
+                    raise NDStateMachineError(f"Invalid configuration. for config {config}: {str(e)}")
+            self.output.assign(after=self.existing, before=self.before, proposed=self.proposed)
+        except Exception as e:
```
mikewiebe, line 50: Here you seem to be wrapping erros with "Initialization failed:" losing the original traceback context.

Should you do the following instead to preserve the chain?

```python
# Suggested - preserve chain
except Exception as e:
    raise NDStateMachineError(f"Initialization failed: {str(e)}") from e
```

plugins/modules/nd_local_user.py
```diff
@@ -0,0 +1,211 @@
...
+            module=module,
+            model_orchestrator=LocalUserOrchestrator,
+        )
+
+        # Manage state
+        nd_state_machine.manage_state()
+
+        module.exit_json(**nd_state_machine.output.format())
+
+    except Exception as e:
```
mikewiebe, line 204: NDStateMachineError is raised internally but here you only catch `Exception`.  The specific error type is lost and all errors will look the same to the user.

Suggest the following instead:

```python
except NDStateMachineError as e:
    module.fail_json(msg=str(e))
except Exception as e:
    module.fail_json(msg=f"Unexpected error: {str(e)}")
```

tests/integration/targets/nd_local_user/tasks/main.yml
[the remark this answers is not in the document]
akinross: @mikewiebe, in case that the resource does not exist, do you think we should test there is indeed no functional difference between the merged, replaced, overridden? All the exist flows I think @gmicol is already testing.

plugins/module_utils/nd_state_machine.py
akinross: @mikewiebe I think this is related to https://github.com/CiscoDevNet/ansible-nd/pull/202#discussion_r2929718437, is this a topic we should discuss next meeting?

plugins/modules/nd_local_user.py
akinross: @mikewiebe I think this is related to https://github.com/CiscoDevNet/ansible-nd/pull/202#discussion_r2929718437, is this a topic we should discuss next meeting?

plugins/modules/nd_local_user.py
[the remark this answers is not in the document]
mikewiebe: **UPDATE:** Instead of creating a new _info or _facts module it seems to make more sense to use `state: gathered` in our modules so proposing we adopt this state in all of our modules.

plugins/module_utils/common/exceptions.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2026, Gaspard Micol (@gmicol) <gmicol@cisco.com>
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
```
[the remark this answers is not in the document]
gmicol: I removed the ``` __metaclass__ = type``` as this was needed in the past for Python 2.7 compatibility. I have not yet touched ```from __future__ import absolute_import, division, print_function``` as the `absolute_import` is needed but not sure about the rest. This is something we need to discuss

plugins/module_utils/endpoints/v1/infra/aaa_local_users.py
```diff
@@ -0,0 +1,201 @@
...
+
+    Provides common functionality for all HTTP methods on the /api/v1/infra/aaa/localUsers endpoint.
+    """
+
+    @property
+    def path(self) -> str:
+        """
+        # Summary
+
+        Build the endpoint path.
```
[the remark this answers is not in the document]
gmicol: I agree but I also think we should revisit all docstrings in the future

plugins/module_utils/orchestrators/base.py
```diff
@@ -0,0 +1,77 @@
...
+            return self.sender.request(path=api_endpoint.path, method=api_endpoint.verb, data=model_instance.to_payload())
+        except Exception as e:
+            raise Exception(f"Create failed for {model_instance.get_identifier_value()}: {e}") from e
+
+    def update(self, model_instance: NDBaseModel, **kwargs) -> ResponseType:
+        try:
+            api_endpoint = self.update_endpoint()
+            api_endpoint.set_identifiers(model_instance.get_identifier_value())
+            return self.sender.request(path=api_endpoint.path, method=api_endpoint.verb, data=model_instance.to_payload())
+        except Exception as e:
```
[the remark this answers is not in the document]
gmicol: We should handle this in another issue because as you mentioned it is likely a wider discussion.

The problem is that multiple exceptions can occur and there is no way to know for now which one it is as we have not properly set proper specific Exceptions but I agree with your point but this is a wider discussion

plugins/module_utils/models/local_user/local_user.py
```diff
@@ -0,0 +1,232 @@
...
+    field_validator,
+    model_validator,
+    FieldSerializationInfo,
+    SerializationInfo,
+)
+from ansible_collections.cisco.nd.plugins.module_utils.models.base import NDBaseModel
+from ansible_collections.cisco.nd.plugins.module_utils.models.nested import NDNestedModel
+from ansible_collections.cisco.nd.plugins.module_utils.constants import NDConstantMapping
+
+USER_ROLES_MAPPING = NDConstantMapping(
```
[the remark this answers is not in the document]
gmicol: We will keep it for now as it is just relevant for local_user and no other model but should be revisited in the future as I think we need to discuss and set solid rules for all constants and where they should be stored (dict, enums, types, etc...)

tests/integration/targets/nd_local_user/tasks/main.yml
```diff
@@ -0,0 +1,651 @@
...
+      - cm_create_local_users.before.0.login_id == "admin"
+      - cm_create_local_users.before.0.first_name == "admin"
+      - cm_create_local_users.before.0.remote_user_authorization == false
+      - cm_create_local_users.before.0.reuse_limitation == 0
+      - cm_create_local_users.before.0.security_domains | length == 1
+      - cm_create_local_users.before.0.security_domains.0.name == "all"
+      - cm_create_local_users.before.0.security_domains.0.roles | length == 1
+      - cm_create_local_users.before.0.security_domains.0.roles.0 == "super_admin"
+      - cm_create_local_users.before.0.time_interval_limitation == 0
+      - cm_create_local_users.diff == []
```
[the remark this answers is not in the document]
gmicol: You are right, but this is actually pointing to a broader discussion as we need to agree on what exactly would be this diff output

tests/integration/targets/nd_local_user/tasks/main.yml
```diff
@@ -0,0 +1,651 @@
...
+      - nm_create_local_users.proposed.0.security_domains.0.name == "all"
+      - nm_create_local_users.proposed.0.security_domains.0.roles | length == 2
+      - nm_create_local_users.proposed.0.security_domains.0.roles.0 == "observer"
+      - nm_create_local_users.proposed.0.security_domains.0.roles.1 == "support_engineer"
+      - nm_create_local_users.proposed.0.time_interval_limitation == 10
+      - nm_create_local_users.proposed.1.login_id == "ansible_local_user_2"
+      - nm_create_local_users.proposed.1.security_domains | length == 1
+      - nm_create_local_users.proposed.1.security_domains.0.name == "all"
+
+# UPDATE
```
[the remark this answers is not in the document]
gmicol: I think you are right, I would go even further with this idea. Should we do something like?
```
- MERGED
  --> CREATE
  --> UPDATE
- REPLACED
  --> CREATE
  --> UPDATE
- OVERRIDDEN
  --> CREATE
  --> UPDATE
- DELETED
```

tests/integration/targets/nd_local_user/tasks/main.yml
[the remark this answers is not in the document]
gmicol: @akinross, in case that the resource does not exist, how would you test there is no functional difference between the merged, replaced and overridden? especially with the `local_user` module which is not idempotent for creation (the attribute `password` prevents it)

tests/integration/targets/nd_local_user/tasks/main.yml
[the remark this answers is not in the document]
gmicol: this is incorrect, will be fixed

gmicol: I don't think `deleted` should be behaving like that (it could be a bit counter intuitive) , I think `overridden` should be behaving like this if providing an empty list

tests/integration/targets/nd_local_user/tasks/main.yml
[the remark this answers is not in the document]
akinross: @gmicol I think your proposal https://github.com/CiscoDevNet/ansible-nd/pull/202#discussion_r2948070442 already to some extend does the create for merged, replaced, overridden. In that case can assert similar behaviour that we expect? so not password but at least the expected results should be similar for those attributes that are similar on create for each state?

tests/unit/module_utils/endpoints/test_endpoints_api_v1_infra_aaa_local_users.py
```diff
@@ -0,0 +1,437 @@
+# Copyright: (c) 2026, Allen Robel (@arobel) <arobel@cisco.com>
```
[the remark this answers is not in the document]
gmicol: just by Allen

tests/integration/targets/nd_local_user/tasks/main.yml
akinross: Are we testing that scenario? Are we documenting that scenario?

gmicol: It's not documented and not tested, we should do both! But just wanted to make sure we all agree on the states behavior when it comes to deletion.

plugins/module_utils/endpoints/v1/infra/aaa_local_users.py
```diff
@@ -0,0 +1,201 @@
...
+
+    Provides common functionality for all HTTP methods on the /api/v1/infra/aaa/localUsers endpoint.
+    """
+
+    @property
+    def path(self) -> str:
+        """
+        # Summary
+
+        Build the /api/v1/infra/aaa/localUsers endpoint path.
```
mikewiebe, line 35: Are you planning to rename this file based on the new naming rule we agreed to?  In this case this file would be renamed to:

`infra_aaa_local_users.py`

plugins/module_utils/models/local_user/local_user.py
```diff
@@ -0,0 +1,228 @@
+# Copyright: (c) 2026, Gaspard Micol (@gmicol) <gmicol@cisco.com>
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
```
mikewiebe, line 4: Based on the new rules we discussed:

1. Create new directory called `local_user`
2. Move `local_user.py` under `local_user` directory

mikewiebe (dismissed):
LGTM

lhercot (dismissed):
LGTM

plugins/modules/nd_local_user.py
```diff
@@ -0,0 +1,206 @@
...
+
+from ansible.module_utils.basic import AnsibleModule
+from ansible_collections.cisco.nd.plugins.module_utils.nd import nd_argument_spec
+from ansible_collections.cisco.nd.plugins.module_utils.nd_state_machine import NDStateMachine
+from ansible_collections.cisco.nd.plugins.module_utils.models.local_user.local_user import LocalUserModel
+from ansible_collections.cisco.nd.plugins.module_utils.orchestrators.local_user import LocalUserOrchestrator
+
+
+def main():
+    argument_spec = nd_argument_spec()
```
samiib, line 183: Should we be using the `require_pydantic` util call here as suggested in the function doc?

plugins/module_utils/constants.py
```diff
@@ -1,13 +1,29 @@
...
 
 from __future__ import absolute_import, division, print_function
 
-__metaclass__ = type
+from typing import Dict
+from types import MappingProxyType
+from copy import deepcopy
+
+
+class NDConstantMapping(Dict):
```
samiib: Is the Dict inheritance needed in this class? There doesn't seem to be any need for it as there isn't any calls to the base dictionary. If anyone was to use Dict methods eg. len() it would always return an empty dictionary.

```suggestion
class NDConstantMapping:
```

plugins/modules/nd_local_user.py
```diff
@@ -0,0 +1,206 @@
...
+        remote_user_authorization: true
+    state: merged
+  register: result
+
+- name: Create local user with minimal configuration
+  cisco.nd.nd_local_user:
+    config:
+      - login_id: local_user_min
+        user_password: localUserMinuser_password
+        security_domain: all
```
samiib: Should this be security_domains?

plugins/module_utils/models/local_user/local_user.py
```diff
@@ -0,0 +1,228 @@
...
+                                choices=USER_ROLES_MAPPING.get_original_data(),
+                            ),
+                        ),
+                        aliases=["domains"],
+                    ),
+                    remote_id_claim=dict(type="str"),
+                    remote_user_authorization=dict(type="bool"),
+                ),
+            ),
+            state=dict(
```
samiib, line 223: Would this state dictionary likely be repeated in future modules? If so could it be a constant spec defined elsewhere?

plugins/modules/nd_local_user.py
```diff
@@ -0,0 +1,206 @@
...
+        last_name: Updated user last name
+        user_password: updatedLocalUserPassword1%
+        reuse_limitation: 25
+        time_interval_limitation: 15
+        security_domains:
+          - name: all
+            roles: super_admin
+          - name: ansible_domain
+            roles: observer
+        roles: super_admin
```
samiib:
```suggestion
```
This looks a little out of place?

plugins/modules/nd_local_user.py
```diff
@@ -0,0 +1,206 @@
...
+    config:
+      - login_id: local_user_min
+        user_password: localUserMinuser_password
+        security_domain: all
+    state: merged
+
+- name: Update local user
+  cisco.nd.nd_local_user:
+    config:
+      - email: udpateduser@example.com
```
samiib:
```suggestion
      - email: updateduser@example.com
```

samiib (changes requested):

plugins/module_utils/orchestrators/base.py
```diff
@@ -0,0 +1,73 @@
...
+
+from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import BaseModel, ConfigDict
+from typing import ClassVar, Type, Optional
+from ansible_collections.cisco.nd.plugins.module_utils.models.base import NDBaseModel
+from ansible_collections.cisco.nd.plugins.module_utils.nd import NDModule
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.base import NDEndpointBaseModel
+from ansible_collections.cisco.nd.plugins.module_utils.orchestrators.types import ResponseType
+
+
+class NDBaseOrchestrator(BaseModel):
```
allenrobel: Sorry, I wasn't asked to review this but hit a typing issue when trying to use this in the nd_interface_loopback module.  The fix was to make this generic so that subclasses can narrow parameter types. e.g.:

```python
ModelT = TypeVar("ModelT", bound=NDBaseModel)

class NDBaseOrchestrator(BaseModel, Generic[ModelT]):

    model_config = ConfigDict()  # ommited contents for brevity

    model_class: ClassVar[Type[NDBaseModel]] = NDBaseModel

    # ...

    def create(self, model_instance: ModelT, **kwargs) -> ResponseType:

   # etc...
```

Then, when defining orchestrator for interface_loopback...

```python
class LoopbackInterfaceOrchestrator(NDBaseOrchestrator[LoopbackInterfaceModel]):
    model_class: ClassVar[Type[NDBaseModel]] = LoopbackInterfaceModel
    # etc...

    def create(self, model_instance: LoopbackInterfaceModel, **kwargs) -> ResponseType:
        # etc...

```

So, by making `NDBaseOrchestrator` a generic, we can specify `NDBaseOrchestrator[LoopbackInterfaceModel]`, which narrows the type parameter `ModelT` to `LoopbackInterfaceModel` and my `create()` method in `LoopbackInterfaceOrchestrator` isn't flagged by type checkers (e.g. mypy) for an invalid signature.

allenrobel (commented):
Leaving one comment based on my experience using this for the nd_interface_loopback module.

plugins/modules/nd_local_user.py
gmicol: You are right, I think we should use it. I did not use it because I was using an older version of the `pydantic_compat.py` where this function did not exist yet. Will add it!

plugins/module_utils/constants.py
gmicol: You are right, forgot to remove it

plugins/module_utils/models/local_user/local_user.py
gmicol: You are right but this should related to a different PR once we start generating the argument_spec from the model.

plugins/module_utils/orchestrators/base.py
gmicol: I think that would be a good improvement, I just made the change.

lhercot (approved):
LGTM

akinross (approved):
LGTM

mikewiebe (approved):
Rebased fabric module on your latest updates so  LGTM

mtarking (approved):

allenrobel (approved):
Approving with a general comment.

We should discuss importing annotations from __future__ so that we can use the new type annotations vs e.g. Dict, List, etc.

samiib (approved):
LGTM

PR #209, opened 2026-03-27
Ansible ND 4.X Fabric Modules for iBGP, eBGP and External Fabric Types

mikewiebe:
This PR introduces 3 new modules to configure ibgp, ebgp and external fabric types.

Additionally this has a fix for state: merged when using the orchestrartor and ND state machine

Fix Details:

### Root Cause

When `state: merged` is used and a field like `banner` is not specified in the task config, Pydantic still populates it with its default value (`""`). The `merge()` method in `NDBaseModel` was applying **all** non-None field values from the proposed config — including these unintended defaults — overwriting existing values.


plugins/module_utils/models/base.py
```diff
@@ -196,16 +196,64 @@ def to_diff_dict(self, **kwargs) -> Dict[str, Any]:
...
+            only_set_fields: When True, only compare fields explicitly set in
+                ``other`` (via model_fields_set). This prevents default values
+                from triggering false diffs during merge operations.
+        """
         self_data = self.to_diff_dict()
-        other_data = other.to_diff_dict()
+        if only_set_fields:
+            other_data = other._to_set_fields_diff_dict()
+        else:
+            other_data = other.to_diff_dict()
```
gmicol: Instead of using `_to_set_fields_diff_dict` and `_filter_set_fields` methods combined to filter unset fields, we could leverage the boolean argument `exclude_unset` from Pydantic built-in dump method (`model_dump`) like this:

```PYTHON
other_data = other.to_diff_dict(exclude_unset=only_set_fields)
```

This solution would avoid us to add extra functions while leveraging Pydantic built-in capabilities, and more specifically in the case of `_filter_set_fields` method, it would prevent looping over the model again to not simply exclude unset but also fields from `exclude_from_diff` and fields with  `None` values (which the `to_diff_dict` method is already doing)

gmicol: And maybe rename the boolean argument `only_set_fields` to `exclude_unset` to be more on parse with Pydantic?

gmicol (changes requested):

plugins/module_utils/models/base.py
mikewiebe: Good suggestion and a far simpler approach.  Changes made

plugins/module_utils/endpoints/v1/manage/manage_fabrics.py
```diff
@@ -0,0 +1,500 @@
...
+"""
+
+from __future__ import absolute_import, annotations, division, print_function
+
+# from plugins.module_utils.endpoints.base import NDBaseEndpoint
+
+# pylint: disable=invalid-name
+__metaclass__ = type
+# pylint: enable=invalid-name
+
```
allenrobel, line 31: 32-34 - We can remove all of these.

plugins/module_utils/endpoints/v1/manage/manage_fabrics.py
```diff
@@ -0,0 +1,500 @@
...
+  (POST /api/v1/manage/fabrics)
+- `EpApiV1ManageFabricsPut` - Update a specific fabric
+  (PUT /api/v1/manage/fabrics/{fabric_name})
+- `EpApiV1ManageFabricsDelete` - Delete a specific fabric
+  (DELETE /api/v1/manage/fabrics/{fabric_name})
+- `EpApiV1ManageFabricsSummaryGet` - Get summary for a specific fabric
+  (GET /api/v1/manage/fabrics/{fabric_name}/summary)
+"""
+
+from __future__ import absolute_import, annotations, division, print_function
```
allenrobel: 28 - AFAIK, the only thing we need to import for __future__ is annotations (for modern annotations/type hints e.g. `dict` vs older `Dict`).

plugins/module_utils/models/base.py
```diff
@@ -216,6 +226,10 @@ def merge(self, other: "NDBaseModel") -> "NDBaseModel":
             if value is None:
                 continue
 
+            # Only merge fields that were explicitly provided, not defaults
+            if field_name not in other.model_fields_set:
+                continue
+
```
allenrobel, line 232: Since this is a recursive method and `other.module_fields_set` contains (AFAIK) only top-level fields, how are fields not explicitly provided in the user's config handled in the nested sections?  Wouldn't we need to pass a `model_fields_set` for each nested level or pass a nested `other` object with identical nesting?

allenrobel (approved):
Approving with comments.

plugins/module_utils/endpoints/v1/manage/manage_fabrics.py
mikewiebe: Done

plugins/module_utils/utils.py
```diff
@@ -36,7 +36,18 @@ def issubset(subset: Any, superset: Any) -> bool:
 
     if not isinstance(subset, dict):
         if isinstance(subset, list):
-            return all(item in superset for item in subset)
+            if len(subset) != len(superset):
```
mikewiebe, line 39: This change to utils.py is needed so that properties like the following will support removal of items from the list for properties like `bootstrap_subnet_collection`

```yaml
    - name: Create an iBGP VXLAN fabric using state merged
      cisco.nd.nd_manage_fabric_ibgp:
        state: replaced
        config:
          - fabric_name: mike-ibgp-test
            category: fabric
            location:
              latitude: 35.78
              longitude: -78.64
            license_tier: premier
            alert_suspend: disabled
            security_domain: all
            telemetry_collection: false
            management:
              type: vxlanIbgp
              bgp_asn: "65001"
              site_id: "65001"
              dns_collection:
                - "192.88.88.44"
                - "192.88.88.46"
                - "192.88.88.49"
                - "192.88.88.50"
              bootstrap_subnet_collection:
                # - start_ip: "192.168.55.2"
                #   end_ip: "192.168.55.244"
                #   default_gateway: "192.168.55.1"
                #   subnet_prefix: 24
                - start_ip: "192.168.88.2"
                  end_ip: "192.168.88.244"
                  default_gateway: "192.168.88.1"
                  subnet_prefix: 24
                - start_ip: "192.168.89.2"
                  end_ip: "192.168.89.244"
                  default_gateway: "192.168.89.1"
                  subnet_prefix: 24
              target_subnet_mask: 30
```

Without this change, setting the property to `[]` or removing items from the list would result in no changes making it impossible to remove configuration

plugins/module_utils/models/base.py
mikewiebe: I would like to address this in a separate PR.  I don't see any direct impact of this for the fabric modules but we probably have other use cases there this could be required.  I would prefer to open an issue and not address it in this PR since it needs more investigation

mtarking (approved):

PR #210, opened 2026-03-30
[bugfix] Fix sanitize function for nd_rest module (DCNE-684)

### LINKED ISSUE #181: nd_rest module crashes with TypeError when API returns empty response body (DCNE-684)
Collection: cisco.nd
Version: 1.4.0
Module: nd_rest

Environment:
Python: 3.10.4
Ansible: 2.17.6
cisco.nd: 1.4.0
OS: macOS 15.3 (Darwin 25.3.0 arm64)
NDFC Version: 4.1.1 (Nexus Dashboard Fabric Controller)

Problem Summary:
The nd_rest module crashes with a TypeError when the NDFC API returns an empty/null response body on a successful POST request. This happens with certain NDFC endpoints that return HTTP 200 with no body, such as the VPC pair creation endpoint.

Steps to repro:
- name: Create VPC pair
  cisco.nd.nd_rest:
    host: "{{ ansible_host }}"
    username: "{{ ansible_user }}"
    password: "{{ ansible_password }}"
    use_ssl: true
    validate_certs: false
    login_domain: local
    path: /appcenter/cisco/ndfc/api/v1/lan-fabric/rest/vpcpair
    method: POST
    content: |
      {
        "useVirtualPeerLink": false,
        "peerOneId": "SERIAL1",
        "peerTwoId": "SERIAL2",
        "templateName": "vpc_pair",
        "nvPairs": { ... }
      }

Expected Behavior
Module completes successfully and reports changed: true when VPC pair is created.

Actual Behavior:
TypeError: object to sanitize can only be of type list or dict. Got <class 'NoneType'>

Traceback:
File ".../cisco/nd/plugins/modules/nd_rest.py", line 262, in main
File ".../cisco/nd/plugins/module_utils/nd.py", line 66, in sanitize
TypeError: object to sanitize can only be of type list or dict. Got <class 'NoneType'>

Note: The API call itself succeeds — the VPC pair is created in NDFC. The error occurs when the module tries to parse the response.

The sanitize() function in plugins/module_utils/nd.py (line 55-66) only accepts list or dict:
def sanitize(obj_to_sanitize, keys=None, values=None, recursive=True, remove_none_values=True):
    if isinstance(obj_to_sanitize, dict):
        return sanitize_dict(...)
    elif isinstance(obj_to_sanitize, list):
        return sanitize_list(...)
    else:
        raise TypeError("object to sanitize can only be of type list or dict. Got {}".format(type(obj_to_sanitize)))

Possible fix to handle None gracefully:

def sanitize(obj_to_sanitize, keys=None, values=None, recursive=True, remove_none_values=True):
    """Clean up a Python object of type list or dict from specific keys, values and None values if specified"""
    if obj_to_sanitize is None:
        return {}  # or return None, depending on desired behavior
    if isinstance(obj_to_sanitize, dict):
        return sanitize_dict(obj_to_sanitize, keys, values, recursive, remove_none_values)
    elif isinstance(obj_to_sanitize, list):
        return sanitize_list(obj_to_sanitize, keys, values, recursive, recursive, remove_none_values)
    else:
        raise TypeError("object to sanitize can only be of type list or dict. Got {}".format(type(obj_to_sanitize)))

gmicol:
fix #181 

The `sanitize` function will now return an empty dict when the `obj_to_sanitize` is None: Deals with cases where the API returns None.

### DIFF
--- a/plugins/module_utils/nd.py
+++ b/plugins/module_utils/nd.py
@@ -59,7 +59,7 @@ def sanitize_list(list_to_sanitize, keys=None, values=None, list_recursive=True,
 def sanitize(obj_to_sanitize, keys=None, values=None, recursive=True, remove_none_values=True):
     """Clean up a Python object of type list or dict from specific keys, values and None values if specified"""
     if obj_to_sanitize is None:
-        return {}
+        return None
     elif isinstance(obj_to_sanitize, dict):
         return sanitize_dict(obj_to_sanitize, keys, values, recursive, remove_none_values)
     elif isinstance(obj_to_sanitize, list):
--- a/plugins/module_utils/nd.py
+++ b/plugins/module_utils/nd.py
@@ -58,7 +58,9 @@ def sanitize_list(list_to_sanitize, keys=None, values=None, list_recursive=True,
 
 def sanitize(obj_to_sanitize, keys=None, values=None, recursive=True, remove_none_values=True):
     """Clean up a Python object of type list or dict from specific keys, values and None values if specified"""
-    if isinstance(obj_to_sanitize, dict):
+    if obj_to_sanitize is None:
+        return {}
+    elif isinstance(obj_to_sanitize, dict):
         return sanitize_dict(obj_to_sanitize, keys, values, recursive, remove_none_values)
     elif isinstance(obj_to_sanitize, list):
         return sanitize_list(obj_to_sanitize, keys, values, recursive, recursive, remove_none_values)


shrsr (approved):
LGTM

plugins/module_utils/nd.py
```diff
@@ -58,7 +58,9 @@ def sanitize_list(list_to_sanitize, keys=None, values=None, list_recursive=True,
 
 def sanitize(obj_to_sanitize, keys=None, values=None, recursive=True, remove_none_values=True):
     """Clean up a Python object of type list or dict from specific keys, values and None values if specified"""
-    if isinstance(obj_to_sanitize, dict):
+    if obj_to_sanitize is None:
+        return {}
```
akinross: Why determine to return a dict and not a list because the function cleans list or dict? Why not return None in this case?

gmicol: `None` makes definitely more sense. Changed it

shrsr: @gmicol I checked the modules to see where sanitize() is being used and looks like nd_rest is the only module using it so far. However, In nd_rest could you please verify whether we could replace:

```
if sanitize(nd.result["jsondata"], ND_REST_KEYS_TO_SANITIZE) != nd.previous and method != "GET":
    nd.result["changed"] = True
```
with
```
if nd.existing != nd.previous and method != "GET":
    nd.result["changed"] = True
```

akinross (approved):
LGTM

mikewiebe (approved):
LGTM

allenrobel (approved):
LGTM

plugins/module_utils/nd.py
```diff
@@ -58,7 +58,9 @@ def sanitize_list(list_to_sanitize, keys=None, values=None, list_recursive=True,
 
 def sanitize(obj_to_sanitize, keys=None, values=None, recursive=True, remove_none_values=True):
     """Clean up a Python object of type list or dict from specific keys, values and None values if specified"""
-    if isinstance(obj_to_sanitize, dict):
+    if obj_to_sanitize is None:
+        return None
```
samiib, line 62: Although None makes sense in isolation here, I tend to agree with the initial change of returning `{}`. This is based on how sanitize is used in nd_rest.
Since existing & previous all default to `dict()` there might be issues related to comparison of `{} != None`. 
It could surface issues like always setting changed=True on line 279 in nd_rest.py for certain cases. Although it might only be an issue for edge cases.

gmicol: I tend to disagree with that statement. I think `nd_rest` should return exactly what the API is returning: If a POST or PUT request has been made and the API returns no content, we should return `None` as well.

If we go back to the initial change of returning to an empty dict, we face the opposite issue of the one you are mentioning: `changed` will always be equal to `False` on line 279 if the API returns `None` even though potential changes have been pushed.

it's true that might be issues related to comparison of `{} != None`. However, I think there would just a few of edge cases compared to the opposite scenario. I would rather indicate to our users that a change has potentially been made rather than the opposite.

But it raises an issue on how we want the `nd_rest` module to behave when indicating a change, knowing that the API is very inconsistent.

samiib: There are places we adjust what the API returns for usability. Example in other modules that have comma separated string properties which get converted to lists. Or error messages that get additional validation. Since nd_rest is simply a passthrough for the API we just need to make sure the module clearly shows what the API is returning. I can see `changed` being false is also a potential usability issue aswell. 
Does a `None` in the existing/previous results returned as explicit `null` or get omitted?

plugins/module_utils/nd.py
```diff
@@ -58,7 +58,9 @@ def sanitize_list(list_to_sanitize, keys=None, values=None, list_recursive=True,
 
 def sanitize(obj_to_sanitize, keys=None, values=None, recursive=True, remove_none_values=True):
     """Clean up a Python object of type list or dict from specific keys, values and None values if specified"""
-    if isinstance(obj_to_sanitize, dict):
+    if obj_to_sanitize is None:
```
samiib, line 61: I can see an opportunity to add tests for this to increase test coverage of `nd_rest`.

samiib (changes requested):

mtarking:
Closing for now as nd_rest is to be reworked as part of 2.0.0 refactor and alignment with new OpenAPIs.

PR #216, opened 2026-04-01
Ansible ND 4.X | WIP | ND Manage Policies Module + Pydantic Models + Smart Endpoints 

nikhilsrikrishna:
This PR adds the nd_policy module for switch policy management in the cisco.nd collection, including the core resource handler, endpoint wrappers, and Pydantic models.

## What's Included

- **Policy resource handler** — `nd_policy_resources.py` supports `gathered`, `merged`, and `deleted` states. Uses bulk create, bulk delete with 207 response handling, the mark-delete → push-config → remove pipeline, and deploy through switch actions.

- **Endpoint definitions** — Three endpoint classes: `manage_fabrics_policies.py`, `manage_fabrics_policy_actions.py`, `manage_fabrics_switch_actions.py`.

- **Pydantic model layer** — Model classes: `config_models.py`, `gathered_models.py`, `policy_actions.py`, `policy_base.py`, `policy_crud.py`. These cover input validation, API response parsing, and gathered output formatting.

- **Input aliasing** — `switch_id` is aliased as `switch_ip` in the argument spec, allowing users to specify either name interchangeably. Pydantic validators normalize both to the field expected by the API.

- **Template inputs validation** — User-provided `template_inputs` are validated at runtime against the template's parameter schema fetched from the controller, checking for unknown keys, missing required parameters, and basic type correctness. System-injected keys are stripped from gathered output so only user-defined variables are returned.

- **Unit tests** — Endpoint-level tests covering all three endpoint classes.

## Notes

- **Delete workflow** — Delete follows a 3-step flow: markDelete → pushConfig → remove. PYTHON content-type templates (e.g., switch_freeform, Ext_VRF_Lite_SVI) fail on markDelete with "Content type is PYTHON, cannot mark for deletion". Instead of maintaining a hardcoded list of these template names, the module inspects the 207 response — any policy that fails with this specific message is automatically retried via direct DELETE /policies/{policyId}, then deployed via switchActions/deploy to push the config removal to the switch.

- **Gathered state and policy_id** — The gathered output includes a policy_id field (e.g., POLICY-28440) alongside the template name. When this output is fed back into state=merged, the policy_id can be directly used to identify the exact policy.

## Work In Progress

- ND output format structure
- Integration tests


mikewiebe:
I see sanity test and formatting failures.  Make sure to run the following commands locally and fix any issues

```
ansible-test sanity --docker -v --color --truncate 0

black --color -l 159 .
```

PR #220, opened 2026-04-03
nd_interface_loopback

### LINKED ISSUE #243: [ND 4.2+] Add Loopback Interfaces Module

```shell
A new module to support ND 4.2+ API Loopback Interfaces is required.
```

manage

```shell

```

```yaml
TBD
```

### Ansible Version

```shell

```

```shell

```

```shell
4.2.1
```

### Cisco NX-OS Version

```shell

```

### Cisco ACI Version

```shell

```

allenrobel:
## Summary

- Adds the `nd_interface_loopback` Ansible module for managing loopback interfaces on Nexus Dashboard 4.2
- Fixes `state:overridden` incorrectly attempting to delete system-provisioned loopback interfaces (Loopback0/Loopback1 with `policyType: "underlayLoopback"`)
- Adds shared `AsciiDescription` Annotated type to reject non-ASCII interface descriptions client-side (applied to `LoopbackPolicyModel.description`)
- Adds integration test suite covering all four states (merged, replaced, overridden, deleted)
- Adds dynamic inventory generator (`tests/integration/inventory.py`) for running integration tests against live ND testbeds
- Adds unit tests for manage_interfaces endpoint classes
- File-based logging integrated via `setup_logging()` in `main()`. Module-level `nd.nd_interface_loopback` logger emits debug records at `manage_state` begin/end, and `log.exception` is called from both the `NDStateMachineError` handler and the broad `Exception` handler so tracebacks reach the log file regardless of `output_level`. Logging activates only when `ND_LOGGING_CONFIG` points at a `logging.config.dictConfig` JSON file; otherwise the calls are no-ops. The test target's `tasks/main.yaml` wraps its `include_tasks` in a block that forwards `nd_logging_config` (set in `inventory.networking [nd:vars]`) into the module subprocess as `ND_LOGGING_CONFIG`, since `ansible-test` strips the controller's shell environment.

## Merge & rebase note

This PR is part of a stack (`develop` → `nd_interface_loopback` → `nd_interface_ethernet_access` → `nd_interface_ethernet_trunk`). Because the team uses squash merges, **after this PR is merged, @allenrobel will rebase the dependent branches onto `develop`** before the next PR (#222) can be merged. Please ping him once this is squashed so he can rebase.

### System loopback filtering fix

`query_all()` in `LoopbackInterfaceOrchestrator` previously filtered only by `interfaceType == "loopback"`, which included system-provisioned loopbacks that ND creates during switch role configuration. These interfaces have `policyType: "underlayLoopback"` and cannot be deleted.

The fix adds a second filter that checks `policyType` against the set of policy types this module manages (`loopback`, `ipfmLoopback`, `userDefined`). System loopbacks are now excluded from the before collection, ensuring accurate `changed` flags and no unnecessary API calls.

### Non-ASCII description rejection (`AsciiDescription` type)

Adds `plugins/module_utils/models/types.py` with a shared `AsciiDescription` Annotated type (`str | None` + `AfterValidator(ascii_only)`). The Cisco backend pipes interface descriptions through CLI generators that return a generic 500 ("unexpected error during policy execution") when the description contains non-ASCII characters (em-dash, smart quotes, emoji, latin-1, etc.). `AsciiDescription` catches this client-side with a clear validation error before the request is sent.

`LoopbackPolicyModel.description` now uses `AsciiDescription` while keeping its `min_length=1, max_length=254` constraints intact via `Field(...)`. Parametrized tests cover em-dash, smart quotes, emoji, and latin-1 rejection alongside ASCII passthrough. The same type is reused by the ethernet_access, ethernet_trunk_host, and svi policy models in the stacked PRs.

### Integration tests

Test coverage organized into per-state task files:

- **Merged**: create single/multiple, idempotency, update, `deploy: false`
- **Replaced**: full replace, idempotency, multi-interface replace
- **Overridden**: reduce to single, idempotency, system loopback filtering validation, swap interfaces
- **Deleted**: single/multi delete, idempotency, non-existent interface

### Dynamic inventory generator

`tests/integration/inventory.py` reads ND connection params from environment variables and generates a static `inventory.networking` for `ansible-test`, which sanitizes the subprocess environment. Also works directly as a dynamic inventory with `ansible-playbook -i`.

## Test plan

- [x] All 365 unit tests pass (`python -m pytest tests/unit/`)
- [x] All integration tests pass against live ND 4.2 testbed via `ansible-test network-integration nd_interface_loopback`
- [x] System loopback filtering verified: Loopback0/Loopback1 excluded from before collection in overridden state
- [ ] Peer review


tests/integration/inventory.py
```diff
@@ -0,0 +1,228 @@
...
+    ## Raises
+
+    None
+    """
+
+    host: str = _required("ND_IP4", "ND controller management IP")
+    username: str = _required("ND_USERNAME", "ND login username")
+    password: str = _required("ND_PASSWORD", "ND login password")
+    domain: str = _required("ND_DOMAIN", "ND login domain e.g. 'local', 'radius'")
+    use_ssl: bool = True
```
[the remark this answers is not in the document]
allenrobel: Addressed via commit 37ad7b9f2cfb3e365b018e27a85ab2eea0d92f42.

tests/integration/inventory.py
```diff
@@ -0,0 +1,228 @@
...
+                "ansible_user": conn.username,
+                "ansible_password": conn.password,
+                "ansible_python_interpreter": "python",
+                # Testbed topology — targets reference these via Jinja defaults
+                # e.g. {{ nd_test_fabric_name | default('test_fabric') }}
+                "nd_test_fabric_name": topo.fabric_name,
+                "nd_test_switch_ip": topo.switch_ip,
+            },
+        },
+    }
```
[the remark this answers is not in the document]
allenrobel: Addressed via commit e17aafaab6e0213b84edb2edb783a6d5b0ee22c0

plugins/module_utils/models/interfaces/loopback_interface.py
```diff
@@ -0,0 +1,302 @@
...
+
+    ## Raises
+
+    None
+    """
+
+    # --- Identifier Configuration ---
+
+    identifiers: ClassVar[Optional[List[str]]] = ["switch_ip", "interface_name"]
+    identifier_strategy: ClassVar[Optional[Literal["single", "composite", "hierarchical", "singleton"]]] = "composite"
```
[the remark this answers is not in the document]
allenrobel: Addressed via commit 161c72c603967c6885d50965d97f0f44acc727f2

tests/unit/module_utils/models/test_loopback_interface.py
```diff
@@ -0,0 +1,1847 @@
...
+
+    ## Classes and Methods
+
+    - LoopbackPolicyModel.serialize_policy_type()
+    """
+    instance = LoopbackPolicyModel()
+    payload_result = instance.model_dump(by_alias=True, context={"mode": "payload"})
+    config_result = instance.model_dump(by_alias=False, context={"mode": "config"})
+    assert payload_result["policyType"] is None
+    assert config_result["policy_type"] is None
```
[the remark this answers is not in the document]
allenrobel: Addressed via commit 12edebcac34e034d47d22dc644942f8f5d14c275

plugins/module_utils/orchestrators/base_interface.py
```diff
@@ -0,0 +1,239 @@
...
+    - Via `_resolve_switch_id` if no switch matches the given IP in the fabric.
+    - Via `deploy_pending` if the bulk deploy API request fails.
+    - Via `remove_pending` if the bulk remove API request fails.
+    """
+
+    deploy: bool = True
+
+    _fabric_context: Optional[FabricContext] = None
+    _pending_deploys: list[tuple[str, str]] = []
+    _pending_removes: list[tuple[str, str]] = []
```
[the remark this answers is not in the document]
allenrobel: Addressed with the following commits:

- d0fb785b50239159a2263a1eed93744312484e12
- b6ff4717d6143f23ac4c9224015ffa8a1c9e9a36

plugins/modules/nd_interface_loopback.py
```diff
@@ -0,0 +1,300 @@
...
+            nd_state_machine.model_orchestrator.deploy_pending()
+
+        module.exit_json(**nd_state_machine.output.format())
+
+    except NDStateMachineError as e:
+        output = nd_state_machine.output.format() if nd_state_machine else {}
+        error_msg = f"Module execution failed: {str(e)}"
+        if module.params.get("output_level") == "debug":
+            error_msg += f"\nTraceback:\n{traceback.format_exc()}"
+        module.fail_json(msg=error_msg, **output)
```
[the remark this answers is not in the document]
allenrobel: Addressed via commit abc59bcef2b7c622f4e41f25a8eaff77d7a9f9f4

plugins/modules/nd_interface_loopback.py
```diff
@@ -0,0 +1,322 @@
...
+        - The management IP address of the switch on which to manage this loopback interface.
+        - This is resolved to the switch serial number (switchId) internally.
+        type: str
+        required: true
+      interface_name:
+        description:
+        - The name of the loopback interface (e.g., C(loopback0), C(Loopback10)).
+        type: str
+        required: true
+      interface_type:
```
akinross: Is interface_type different from policy_type? they both expose loopback as value? Also I do not see interface_type in the examples, what is this used for and is this required to be exposed to users? 

should we restrict options if there are any for this type attribute? or keep it freefrom input? If not restricting should we document the allowed values? The same would apply to other inputs in config data like mode, network_os_type. 

I do see you mention it for policy_type.

plugins/modules/nd_interface_loopback.py
```diff
@@ -0,0 +1,322 @@
...
+            description:
+            - The interface management mode.
+            type: str
+            default: managed
+          network_os:
+            description:
+            - Network OS specific configuration.
+            type: dict
+            suboptions:
+              network_os_type:
```
akinross: do we currently support other os-types?

plugins/modules/nd_interface_loopback.py
```diff
@@ -0,0 +1,322 @@
...
+    choices: [ merged, replaced, overridden, deleted ]
+extends_documentation_fragment:
+- cisco.nd.modules
+- cisco.nd.check_mode
+notes:
+- This module is only supported on Nexus Dashboard.
+- This module currently supports NX-OS loopback interfaces only.
+"""
+
+EXAMPLES = r"""
```
akinross, line 125: I am missing an example with extra config, which might be good to have since that is likely a more complicated scenario. 

Is there a clear rule as to what to include into examples? because some attributes are not shown in the examples.

plugins/modules/nd_interface_loopback.py
```diff
@@ -0,0 +1,322 @@
...
+                    type: str
+                  description:
+                    description:
+                    - The description of the loopback interface.
+                    - Maximum 254 characters.
+                    type: str
+                  extra_config:
+                    description:
+                    - Additional CLI configuration commands to apply to the interface.
+                    type: str
```
akinross, line 93: did you consider making this a list of commands?

plugins/modules/nd_interface_loopback.py
akinross, line 125: I am missing an overridden example

plugins/modules/nd_interface_loopback.py
```diff
@@ -0,0 +1,322 @@
...
+    - Use O(state=merged) to create new resources and update existing ones as defined in your configuration.
+      Resources on ND that are not specified in the configuration will be left unchanged.
+    - Use O(state=replaced) to replace the resources specified in the configuration.
+    - Use O(state=overridden) to enforce the configuration as the single source of truth.
+      The resources on ND will be modified to exactly match the configuration.
+      Any resource existing on ND but not present in the configuration will be deleted. Use with extra caution.
+    - Use O(state=deleted) to remove the resources specified in the configuration from the Cisco Nexus Dashboard.
+    type: str
+    default: merged
+    choices: [ merged, replaced, overridden, deleted ]
```
akinross, line 115: in some other PRs I see gathered state? is this intentially skipped?

plugins/modules/nd_interface_loopback.py
allenrobel: Thanks Akini.  Yes, these should not be exposed to users.  Addressed with commit: 56a45a8

I was on the fence about `policy_type` for the loopback module since it _could_ be either "loopback", "ipfmLoopback", or "userDefined" but I think it would be better to have three separate loopback modules so am hardcoding `policy_type` here like I did in the ethernet and port-channel modules.  The other params (mode, network_os_type, interface_type) are also now hard-coded in the model and all integration tests and unit tests modified accordingly.

Much better UX now so thanks again.

plugins/modules/nd_interface_loopback.py
allenrobel: Addressed with same commit as previous comment: https://github.com/CiscoDevNet/ansible-nd/commit/56a45a85852f7029fe4a35344831f9e01d03b545

plugins/modules/nd_interface_loopback.py
allenrobel: Yes, I considered it, but a list breaks down for sub-mode config (HSRP/VRRP/OSPF blocks) where indentation is semantic, would diverge from how other nd_interface_*  modules will need to handle the same field, and works against the common 'paste a snippet from running-config' workflow. The | block scalar already gives the per-line readability a list would offer e.g.:

```yaml
- name: Create a loopback interface with extra CLI configuration
    fabric_name: my_fabric
    config:
      - switch_ip: 192.168.1.1
        interface_name: loopback0
        config_data:
          network_os:
            policy:
              ip: 10.1.1.1
              ipv6: 2001:db8::1/128
              description: Loopback with PIM and OSPF tuning
              vrf: default
              extra_config: |
                ip pim sparse-mode
                ip ospf network point-to-point
                no ip redirects
    state: merged
```

plugins/modules/nd_interface_loopback.py
allenrobel: Added example via commit: 6257ba9

allenrobel: Added via commit: 96a62b3

plugins/modules/nd_interface_loopback.py
allenrobel: Yes, intentionally skipped since I'm waiting on Gaspard's orchestrator to support it rather than rolling my own.

plugins/modules/nd_interface_loopback.py
akinross: Ok makes sense, do we document the indentation semantics somewhere? Or provide examples with proper indentation?

plugins/modules/nd_interface_loopback.py
akinross: ok makes sense, got confused with other PRs. but assume the change when Gaspard implemented it should be minimal in the code also.

plugins/modules/nd_interface_loopback.py
allenrobel: Let me add an example to all interface modules where the possibility of an indented block exists.

Checking with CIRCUIT, loopback, L2 access, L2 trunk interfaces have no indented blocks.  Port-channels do, for example (evpn esi multihoming must be enabled at global config level first):

VPC PO

```yaml
extra_config: |
  ethernet-segment vpc
    esi 0012.0000.0000.1200.0102 tag 1012
```

non-VPC PO

```yaml
extra_config: |
  ethernet-segment 2011
    system-mac 0000.0000.2011  
```

I could add a generic example for loopback (in case there's a case CIRCUIT missed, and I missed via direct NX-OS console).  For example:

```yaml
extra_config: |
  config line 1
  config line 2
    indented config line 3
```

What do you think?

akinross: yeah that is fine by me, if not applicable for loopback would not add generic

akinross (dismissed):

tests/integration/targets/nd_interface_loopback/tasks/setup.yaml
```diff
@@ -0,0 +1,23 @@
...
+  cisco.nd.nd_interface_loopback:
+    output_level: "{{ nd_info.output_level }}"
+    fabric_name: "{{ test_fabric_name }}"
+    config:
+      - switch_ip: "{{ test_switch_ip }}"
+        interface_name: loopback100
+      - switch_ip: "{{ test_switch_ip }}"
+        interface_name: loopback101
+      - switch_ip: "{{ test_switch_ip }}"
+        interface_name: loopback102
```
samiib, line 21: Should we also clear loopback103 just in case?

plugins/module_utils/endpoints/v1/manage/manage_switches.py
```diff
@@ -0,0 +1,196 @@
...
+
+        ## Raises
+
+        ### ValueError
+
+        - If `fabric_name` is not set before accessing `path`.
+        """
+        if self.fabric_name is None:
+            raise ValueError(f"{type(self).__name__}.path: fabric_name must be set before accessing path.")
+        segments = ["fabrics", self.fabric_name, "switches"]
```
samiib: Should fabric_name also be quote encoded?

plugins/module_utils/endpoints/v1/manage/manage_switches.py
```diff
@@ -0,0 +1,196 @@
...
+
+        ## Raises
+
+        ### ValueError
+
+        - If `fabric_name` is not set before accessing `path`.
+        """
+        if self.fabric_name is None:
+            raise ValueError(f"{type(self).__name__}.path: fabric_name must be set before accessing path.")
+        return BasePath.path("fabrics", self.fabric_name, "switchActions", "deploy")
```
samiib: Should fabric_name also be quote encoded here?

plugins/modules/nd_interface_loopback.py
```diff
@@ -0,0 +1,339 @@
...
+ANSIBLE_METADATA = {"metadata_version": "1.1", "status": ["preview"], "supported_by": "community"}
+
+DOCUMENTATION = r"""
+---
+module: nd_interface_loopback
+version_added: "1.4.0"
+short_description: Manage loopback interfaces on Cisco Nexus Dashboard
+description:
+- Manage loopback interfaces on Cisco Nexus Dashboard.
+- It supports creating, updating, querying, and deleting loopback interfaces on switches within a fabric.
```
samiib: Is querying an option?

plugins/module_utils/fabric_context.py
```diff
@@ -0,0 +1,313 @@
...
+        None
+        """
+        return self._fabric_name
+
+    @property
+    def fabric_summary(self) -> dict | None:
+        """
+        # Summary
+
+        Return the cached fabric detail dict, fetching it from the `/fabrics/{fabric_name}` endpoint on first access.
```
samiib:
```suggestion
        Return the cached fabric detail dict, fetching it from the `/api/v1/manage/fabrics/{fabric_name}` endpoint on first access.
```

tests/integration/targets/nd_interface_loopback/tasks/setup.yaml
allenrobel: Good catch!  Addressed with commit bff5d88.

plugins/module_utils/endpoints/v1/manage/manage_switches.py
allenrobel: Yep, it should.  Thanks Samita!  Addressed with commit bff5d88.

plugins/module_utils/endpoints/v1/manage/manage_switches.py
allenrobel: Yep, thanks!  Addressed with commit bff5d88.

plugins/modules/nd_interface_loopback.py
allenrobel: Doh!  No, it's not.  Removed via commit bff5d88.

plugins/module_utils/fabric_context.py
allenrobel: Thanks Samita.  Addressed with commit bff5d88 (using the path from the endpoint definition which includes trailing .../summary).

samiib (approved):
LGTM

plugins/modules/nd_interface_loopback.py
```diff
@@ -0,0 +1,339 @@
...
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+"""Ansible module for managing loopback interfaces on Cisco Nexus Dashboard."""
+
+ANSIBLE_METADATA = {"metadata_version": "1.1", "status": ["preview"], "supported_by": "community"}
+
+DOCUMENTATION = r"""
+---
+module: nd_interface_loopback
+version_added: "1.4.0"
```
akinross, line 14: small thing i noticed is to adjust this version added, since we are already on 1.5.0 currently. I assume this is a widespread thing that should be checked and adjusted pre-release. created issue to track this: https://github.com/CiscoDevNet/ansible-nd/issues/290

akinross (approved):

PR #221, opened 2026-04-06
Addition of module for ND 4.2 Links comprising manage and one_manage (NDA-21)

### LINKED ISSUE #234: [ND 4.2+] Add Links Module  (NDA-21)

```shell
A new module to support ND 4.2+ API Links is required.
Review opportunity of module usage between manage and onemanage.
```

manage

```shell

```

```yaml
TBD
```

### Ansible Version

```shell

```

```shell

```

```shell
4.2.1
```

### Cisco NX-OS Version

```shell

```

### Cisco ACI Version

```shell

```

shrsr:
## Related Issue(s)

Related to #234


plugins/action/nd_action_base.py
```diff
@@ -0,0 +1,156 @@
...
+
+    class ActionModule(NDActionBase):
+        MODULE_NAME = "nd_links"
+
+Cross worker "one login per play" is achieved via the on disk session cache
+in ``NDClient.ensure_session()``; the class level ``_client_cache`` only
+covers the current worker's in process reuse.
+"""
+
+from __future__ import absolute_import, division, print_function
```
allenrobel, line 16: No longer needed.  Please remove.

Instead replace with the following so that modern annotations are accepted:

```python
```

Same comment applies for all files.

We should be using modern annotations throughout i.e.:

- `Dict` -> `dict`
- `List` -> `list`
- `Set` -> `set`

Please replace all legacy annotions (if any) with the above across all files commited in this PR.

plugins/action/nd_action_base.py
```diff
@@ -0,0 +1,156 @@
...
+
+__metaclass__ = type
```
allenrobel, line 18: `__metaclass__ = type` is no longer needed.  Please remove.

Same comment applies for all files committed in this PR.

plugins/action/nd_action_base.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+class NDActionBase(ActionBase):
+    """Shared base for every cisco.nd action plugin. Subclasses set ``MODULE_NAME``."""
+
+    MODULE_NAME = None
+
+    _client_cache = {}
+
+    def run(self, tmp=None, task_vars=None):
```
allenrobel, line 48: Add modern method annotations for all methods in files commited in this PR.

The above should be e.g.:

```python
    def run(self, tmp: <type> = None, task_vars: dict[str, Any] = None) -> <type>:
```

Currently, the reader (and the reader's IDE) has no idea what type `tmp` or `task_vars` are, or what type is returned by this method unless they read through the code.  And, even then, it's still unclear what type `tmp` should be.

allenrobel (changes requested):
Leaving initial set of comments:

1. Fix/add modern type annotations throughout
2. Replace `from __future__ import absolute_import, division, print_function` with `from __future__ import annotations` throughout
3. Remove `__metaclass__ = type` throughout

Once these are addressed will review the rest.

Thanks!

plugins/callback/nd_session_cleanup.py
```diff
@@ -0,0 +1,52 @@
...
+
+_XDG_CACHE = os.environ.get("XDG_CACHE_HOME") or os.path.expanduser("~/.cache")
+_CACHE_GLOB = os.path.join(_XDG_CACHE, "ansible-cisco-nd", "nd_session_*.json")
+
+
+class CallbackModule(CallbackBase):
+    CALLBACK_VERSION = 2.0
+    CALLBACK_TYPE = "aggregate"
+    CALLBACK_NAME = "cisco.nd.nd_session_cleanup"
+    CALLBACK_NEEDS_ENABLED = True
```
allenrobel, line 44: If these are intended to be constants, please either annotate with typing.Final.

```python
from typing import Final

class CallbackModule(CallbackBase):
    CALLBACK_VERSION: Final = 2.0
```

Or use a frozen dataclass.

```python
from dataclasses import dataclass

@dataclass(frozen=True)
class CallbackConfig:
    CALLBACK_VERSION: float = 2.0
    CALLBACK_TYPE: str = "aggregate"
    CALLBACK_NAME: str = "cisco.nd.nd_session_cleanup"
    CALLBACK_NEEDS_ENABLED: bool = True

callback_config = CallbackConfig

# Usage

if callback_config.CALLBACK_VERSION < 2.0:
    # handle error
```

If any of these are NOT constants, they should be lowercase.

allenrobel (changes requested):

plugins/action/nd_action_base.py
```diff
@@ -0,0 +1,156 @@
...
+        """Run the module in process with a shared NDClient instead of forking."""
+        result = super(NDActionBase, self).run(tmp, task_vars)
+        del tmp
+        task_vars = task_vars or {}
```
akinross, line 52: why do the {} assignment, instead of failing? When empty the host/user/password will all be None and will error.

plugins/action/nd_action_base.py
```diff
@@ -0,0 +1,156 @@
...
+    """Shared base for every cisco.nd action plugin. Subclasses set ``MODULE_NAME``."""
+
+    MODULE_NAME = None
+
+    _client_cache = {}
+
+    def run(self, tmp=None, task_vars=None):
+        """Run the module in process with a shared NDClient instead of forking."""
+        result = super(NDActionBase, self).run(tmp, task_vars)
+        del tmp
```
akinross, line 51: what is the purpose of this del? perhaps comment?

plugins/action/nd_action_base.py
```diff
@@ -0,0 +1,156 @@
...
+        user = task_vars.get("ansible_user")
+        password = (
+            task_vars.get("ansible_password")
+            or task_vars.get("ansible_httpapi_password")
+            or task_vars.get("ansible_httpapi_pass")
+        )
+        port = task_vars.get("ansible_httpapi_port", 443)
+        verify_certs = task_vars.get("ansible_httpapi_validate_certs", True)
+        login_domain = task_vars.get("ansible_httpapi_login_domain", "DefaultAuth")
+        timeout = task_vars.get("ansible_command_timeout", 30)
```
akinross, line 71: why assign and not define inline in the _get_or_create_client() function?

plugins/action/nd_action_base.py
```diff
@@ -0,0 +1,156 @@
...
+            check_mode=self._task.check_mode,
+            diff=task_vars.get("ansible_diff_mode", False),
+        )
+
+        try:
+            module_pkg.main()
+            result["failed"] = True
+            result["msg"] = (
+                "Module {0} returned without calling exit_json or fail_json".format(self.MODULE_NAME)
+            )
```
akinross, line 123: Can you comment/explain the try/except block? It is kind of counter-intuitive that the NDInlineModuleExit is handled in an except flow ( success ) and the regular try should never continue after and thus failing in regular python success flow.

plugins/callback/nd_session_cleanup.py
```diff
@@ -0,0 +1,52 @@
...
+DOCUMENTATION = """
+  name: nd_session_cleanup
+  type: aggregate
+  short_description: Delete cisco.nd on-disk session caches at end of playbook.
+  description:
+    - Deletes the per-user token cache files used by the cisco.nd fast path
+      once a playbook completes.
+    - Enable when you want zero on-disk persistence of ND tokens between plays.
+  requirements:
+    - Enable via callbacks_enabled in ansible.cfg under [defaults].
```
akinross, line 27: can it also be enabled through env vars? should we document that also?

plugins/module_utils/client/nd_client_connection.py
```diff
@@ -0,0 +1,82 @@
...
+
+        self._params = {}
+        self._messages = []
+        self._connected = True
+
+    def send_request(self, method, path, data=None):
+        """Forward a request to the underlying ``NDClient``."""
+        return self._client.request(method, path, data=data)
+
+    def send_file_request(self, method, path, file=None, data=None,
```
akinross, line 38: why is this not supported? should we comment this?

plugins/action/nd_action_base.py
akinross, line 71: why define defaults here and also in the NDClient init object? should a none pass through not enforce the init defaults? This keeps the default definition in a single location where they belong

plugins/action/nd_action_base.py
```diff
@@ -0,0 +1,156 @@
+# Copyright: (c) 2026, Shreyas Srish (@shrsr) <ssrish@cisco.com>
```
akinross, line 1: how do we handle logging for the action plugin and client?

plugins/module_utils/client/nd_client.py
```diff
@@ -0,0 +1,342 @@
...
+
+        self.host = host
+        self.username = username
+        self.password = password
+        self.port = int(port)
+        self.login_domain = login_domain
+        self.verify_certs = bool(verify_certs)
+        self.timeout = int(timeout)
+
+        self.base_url = "https://{0}:{1}".format(host, self.port)
```
akinross, line 66: i notice a mix of concatenation, format() function. is there a reason to avoid f string usage? since min version i think is  Python 3.10 so newer f strings should be safe to use consistently

plugins/module_utils/client/nd_client.py
```diff
@@ -0,0 +1,342 @@
...
+            try:
+                self._session.close()
+            except Exception:
+                pass
+            self._session = None
+            self._token = None
+
+    def invalidate_session_cache(self):
+        """Delete the shared on disk session cache (explicit logout only)."""
+        self._invalidate_session_cache()
```
akinross, line 154: why have the private function call here when public method does literally nothing the private one doesn't?

plugins/module_utils/client/nd_client.py
```diff
@@ -0,0 +1,342 @@
...
+                os.chmod(path, 0o600)
+            except OSError:
+                pass
+        except OSError:
+            pass
+
+    def _invalidate_session_cache(self):
+        """Remove the on disk cache file (best effort)."""
+        try:
+            os.unlink(self._cache_file_path())
```
akinross, line 239: should this function also clear _token / _session on the client?

plugins/module_utils/client/nd_client.py
```diff
@@ -0,0 +1,342 @@
...
+            except requests.RequestException as e:
+                err = {
+                    "error": {
+                        "code": -1,
+                        "message": "Network error contacting ND: {0}".format(e),
+                    }
+                }
+                raise ConnectionError(json_lib.dumps(err))
+
+            if resp.status_code == 401 and not auth_retried:
```
akinross, line 294: I it intended that the attempt does not increase for a 401?

plugins/module_utils/client/nd_client.py
```diff
@@ -0,0 +1,342 @@
...
+                self._connect()
+                auth_retried = True
+                continue
+            break
+
+        return self._build_info(resp, method, url)
```
akinross, line 300: is it possible tho have resp remain None? if so it might lead to crash in the _build_info function. If not the resp initialisations seems redundant.

plugins/module_utils/client/nd_client.py
```diff
@@ -0,0 +1,342 @@
...
+                if attempt >= self.MAX_NETWORK_ATTEMPTS:
+                    err = {
+                        "error": {
+                            "code": -1,
+                            "message": "Network error contacting ND after {0} attempts: {1}".format(
+                                attempt, e
+                            ),
+                        }
+                    }
+                    raise ConnectionError(json_lib.dumps(err))
```
akinross, line 282: should these ConnectionError be NDClientError or its own implementation of a connection error?

plugins/module_utils/client/nd_client.py
```diff
@@ -0,0 +1,342 @@
...
+        if platform == "ndfc":
+            return 12
+        info = self.request("GET", "/version.json")
+        body = info.get("body")
+        if isinstance(body, dict):
+            try:
+                return "{0}.{1}.{2}".format(body["major"], body["minor"], body["maintenance"])
+            except KeyError:
+                pass
+        return "unknown"
```
akinross, line 313: why return "unknown" instead of raising an error?

plugins/module_utils/client/nd_client.py
```diff
@@ -0,0 +1,342 @@
...
+        if self._session is not None:
+            return
+        if self._try_load_session():
+            return
+        self._connect()
+
+    def _connect(self):
+        """Perform a /login, store the Bearer token, and persist to the on disk cache."""
+        if not self.verify_certs:
+            urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
```
akinross, line 89: why do this in connect and not in the init? I see it called twice, here and in _try_load_session?

plugins/module_utils/client/nd_client.py
```diff
@@ -0,0 +1,342 @@
...
+
+        if self._session is not None:
+            self._session.close()
+
+        session = requests.Session()
+        session.verify = self.verify_certs
+        adapter = HTTPAdapter(pool_connections=1, pool_maxsize=10)
```
akinross, line 96: why are these values 1 and 10? is 10 useful when only 1 connection is allowed? should these be configurable values or defined as constants in the file somewhere?

plugins/module_utils/client/nd_client.py
```diff
@@ -0,0 +1,342 @@
...
+
+    def _connect(self):
+        """Perform a /login, store the Bearer token, and persist to the on disk cache."""
+        if not self.verify_certs:
+            urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+
+        if self._session is not None:
+            self._session.close()
+
+        session = requests.Session()
```
akinross, line 94: there is quite some overlap in the function _connect and _try_load_session, should we create functions for the duplicated code?

plugins/module_utils/client/nd_client.py
```diff
@@ -0,0 +1,342 @@
...
+            return
+        cache_dir = self._session_cache_dir()
+        try:
+            os.makedirs(cache_dir, mode=0o700, exist_ok=True)
+            try:
+                os.chmod(cache_dir, 0o700)
+            except OSError:
+                pass
+        except OSError:
+            return
```
akinross, line 213: I think we should log a warning of some kind to indicate something can not be written

plugins/module_utils/client/nd_client.py
```diff
@@ -0,0 +1,342 @@
...
+        self._connect()
+
+    def _connect(self):
+        """Perform a /login, store the Bearer token, and persist to the on disk cache."""
+        if not self.verify_certs:
+            urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+
+        if self._session is not None:
+            self._session.close()
```
akinross, line 92: is there a refresh option that could be called in the api for when a session is not None?

plugins/module_utils/client/nd_client.py
```diff
@@ -0,0 +1,342 @@
...
+            {
+                "Authorization": "Bearer {0}".format(token),
+                "Cookie": "AuthCookie={0}".format(token),
+            }
+        )
+
+        self._session = session
+        self._save_session()
+
+    def close(self):
```
akinross, line 138: what happens when the exceptions are hit? should we not handle or log those exceptions?

plugins/module_utils/client/nd_client.py
```diff
@@ -0,0 +1,342 @@
...
+    HAS_REQUESTS = True
+except ImportError:
+    HAS_REQUESTS = False
+
+
+class NDClientError(Exception):
+    """Raised for any ND client level error (connection, auth, parsing)."""
+
+
+class NDClient(object):
```
akinross, line 35: object not needed in python 3
```suggestion
class NDClient:
```

plugins/module_utils/models/links/links.py
```diff
@@ -0,0 +1,253 @@
...
+
+    def merge(self, other: "LinkConfigDataModel") -> "LinkConfigDataModel":
+        """Reject inline policy_type changes; ND requires delete and recreate for those."""
+        if (
+            isinstance(other, LinkConfigDataModel)
+            and self.policy_type
+            and other.policy_type
+            and self.policy_type != other.policy_type
+        ):
+            raise Exception(
```
akinross, line 29: should this be a more specific exception?

plugins/module_utils/models/links/links.py
```diff
@@ -0,0 +1,253 @@
...
+        tmpl_cls = type(self.config_data.template_inputs)
+        for field_name, field_info in tmpl_cls.model_fields.items():
+            if field_info.exclude:
+                continue
+            alias = field_info.alias or field_name
+            if template_inputs.get(alias) is None:
+                template_inputs[alias] = self._empty_for_annotation(field_info.annotation)
+
+    @staticmethod
+    def _empty_for_annotation(annotation: Any) -> Any:
```
akinross, line 179: why is the functionality needed, is this because of ND api? 

why have the same static function defined in template/base and links?

should this be a helper function or perhaps even some base level enhancement which would set empty values when not provided?

sivakasi-cisco:
https://github.com/CiscoDevNet/ansible-nd/issues/234

PR #222, opened 2026-04-08
nd_interface_ethernet_access

allenrobel:
## Summary

- Add `nd_interface_ethernet_access` Ansible module for managing ethernet accessHost interfaces on Nexus Dashboard
- Fan-out playbook shape: `interface_names` list with shared `config_data`, expanded into flat items before the state machine
- Shared `EthernetBaseOrchestrator` base class for all future ethernet interface modules (trunk, routed, breakout, freeform)
- Port-channel membership enforcement with whitelisted modifiable fields (`description`, `admin_state`, `extra_config`)
- Bulk normalize-based delete via `interfaceActions/normalize` with the `int_trunk_host` template defaults
- `InterfaceDefaultConfig` model providing the full `int_trunk_host` normalize payload
- `EpManageInterfacesDelete` and `EpManageInterfacesNormalize` endpoint classes
- File-based logging integrated via `setup_logging()` in `main()`. Module-level `nd.nd_interface_ethernet_access` logger emits debug records at `expand_config`, `manage_state` begin/end, and on `NDStateMachineError` (via `log.exception` so tracebacks reach the log file regardless of `output_level`). Logging activates only when `ND_LOGGING_CONFIG` points at a `logging.config.dictConfig` JSON file; otherwise the calls are no-ops.
- Integration tests for all four states: merged, replaced, overridden, deleted. The test target's `tasks/main.yaml` wraps its `include_tasks` in a block that forwards `nd_logging_config` (set in `inventory.networking [nd:vars]`) into the module subprocess as `ND_LOGGING_CONFIG`, since `ansible-test` strips the controller's shell environment.

### Review fixes (commit 999e19d5)

Three correctness fixes from a high-effort code review:

- Interface-name abbreviations (`eth1/41`, `e1/41`) are now expanded to ND's canonical `Ethernet1/41` so abbreviated input is idempotent.
- Port-channel-member change detection parses the wire policy through the same model before comparing, eliminating float/int-vs-string false positives that spuriously blocked idempotent re-runs.
- `_existing_port_channel_id` treats only a positive integer as membership; all call sites use `is None`/`is not None` consistently.

**Known limitation (tracked in #307):** under `state: overridden`, a port-channel member that `delete_bulk` deliberately skips is still removed from the reported `after` state by `NDStateMachine._delete_items`. The fix belongs in `nd_state_machine.py` (already on develop), so it is out of scope for this PR.

### Delete mechanism for physical ethernet interfaces

`/api/v1/manage/fabrics/{fabricName}/interfaceActions/normalize` is used with the full `int_trunk_host` config template defaults (`mode: "trunk"`, `policyType: "trunkHost"`, etc). This resets the interface to fabric defaults and changes its `policyType` to `trunkHost`, which naturally removes it from the accessHost `query_all()` filter, enabling idempotent `state: overridden`.

## Merge & rebase note

This PR is part of a stack with a follow-on branch (`nd_interface_ethernet_trunk`). Because the team uses squash merges, **after this PR is merged, @allenrobel will rebase the remaining stack branch onto `develop`** before the next PR can be opened/merged. Please ping him once this is squashed.

## Unit tests

Unit tests now cover the branch-new source files with 137 tests. Measured line coverage from the unit suite: model `ethernet_access_interface.py` 97%, orchestrator subclass `ethernet_access_interface.py` 92%, shared `ethernet_base.py` 78% (the remainder of `ethernet_base.py` is exercised by the integration suite):

- `tests/unit/module_utils/models/test_ethernet_access_interface.py` exercises every nested Pydantic model (`Policy`, `NetworkOS`, `ConfigData`, `Interface`) with field validation, range and enum constraints, `normalize_policy_type`, `serialize_policy_type`, round-trip `to_payload` / `to_config` / `from_response` / `from_config`, composite identifier, diff, merge, and argument_spec shape.
- `tests/unit/module_utils/orchestrators/test_ethernet_access_interface.py` wires a real `RestSend` with the file-based `Sender` from `tests/unit/module_utils/sender_file.py` and `ResponseHandler` to verify `model_class`, `supports_bulk_*` inheritance, and `_managed_policy_types` returning `{"accessHost"}`. A `query_all` smoke test exercises the inherited base-class logic end-to-end across two switches, confirming non-accessHost interfaces are filtered out and `switchIp` is injected on the kept interfaces.

Parametrize collapses all value-table tests (range checks, enum choice checks, serializer modes, normalizer inputs) into single functions with descriptive `ids=` labels so pytest output still names each case.

## Test plan

- [x] Integration tests pass for `state: merged` (create, fan-out, idempotency, update, deploy:false)
- [x] Integration tests pass for `state: replaced` (full replace, fan-out, idempotency)
- [x] Integration tests pass for `state: overridden` (reduce, idempotency, swap, fan-out, non-accessHost filtering)
- [x] Integration tests pass for `state: deleted` (single, fan-out, idempotency, non-existent)
- [x] Unit tests pass: `python -m pytest tests/unit/module_utils/models/test_ethernet_access_interface.py tests/unit/module_utils/orchestrators/test_ethernet_access_interface.py` (137 tests)


plugins/modules/nd_interface_ethernet_access.py
```diff
@@ -0,0 +1,392 @@
...
+              admin_state: true
+              access_vlan: 100
+              bpdu_guard: enable
+              cdp: true
+              description: Access Host Interface
+              speed: auto
+    state: merged
+  register: result
+
+- name: Create accessHost interfaces across multiple switches
```
akinross, line 316: I assume we can also provide different configs on the same switch, should we add this to the example also?

Furthermore there appears validation preventing the switch/interface configuration to be specified twice, via configcollection add() error. Which would result in ValueError raise: https://github.com/CiscoDevNet/ansible-nd/blob/develop/plugins/module_utils/nd_config_collection.py#L60. This error to me seems kind of confusing for users that are providing configuration: "Module execution failed: Initialisation failed: Item with identifier ('192.168.1.1', 'Ethernet1/1') already exists. Use replace() to update" I think we should handle this differently to which it is more clear to the user. Perhaps the some custom validation could be add to the expand_config() function?

plugins/modules/nd_interface_ethernet_access.py
```diff
@@ -0,0 +1,392 @@
...
+from ansible_collections.cisco.nd.plugins.module_utils.common.log import setup_logging
+from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import require_pydantic
+from ansible_collections.cisco.nd.plugins.module_utils.models.interfaces.ethernet_access_interface import EthernetAccessInterfaceModel
+from ansible_collections.cisco.nd.plugins.module_utils.nd import nd_argument_spec
+from ansible_collections.cisco.nd.plugins.module_utils.nd_state_machine import NDStateMachine
+from ansible_collections.cisco.nd.plugins.module_utils.orchestrators.base_interface import NDBaseInterfaceOrchestrator
+from ansible_collections.cisco.nd.plugins.module_utils.orchestrators.ethernet_access_interface import EthernetAccessInterfaceOrchestrator
+
+
+def expand_config(config_list):
```
akinross: Is this function shared between modules, see similar function https://github.com/CiscoDevNet/ansible-nd/pull/265/changes#diff-f7581ec3bfcae3e122df059e557bd92dafdb2f4b31e8339187987a8d14c337baR448. Should we define it one time?

plugins/modules/nd_interface_ethernet_access.py
```diff
@@ -0,0 +1,392 @@
...
+    description:
+    - The list of ethernet accessHost interface groups to configure.
+    - Each item specifies the target switch, a list of interface names, and a shared configuration.
+    - Multiple switches can be configured in a single task.
+    - The structure mirrors the ND Manage Interfaces API payload.
+    type: list
+    elements: dict
+    required: true
+    suboptions:
+      switch_ip:
```
akinross, line 39: did you consider making this a list to be able to provide same access config across multiple switches? Related to https://github.com/CiscoDevNet/ansible-nd/pull/222/changes#r3199683197

allenrobel: Good question — I did consider it, but kept switch_ip as a single string. My reasoning mirrored 1 and 4 below, but Claude added 2,3,5 which I hadn't considered.

1. It would break the "playbook config mirrors API payload" design principle. The interface models are deliberately structured so to_payload() and from_response() work via standard Pydantic serialization with no custom wrapping or flattening (see the LoopbackInterfaceModel docstring). The ND Manage Interfaces API has one switch per record. If switch_ip is a list on input, the playbook shape no longer mirrors the API shape, and the upcoming gathered/query state work (which renders ND state back as playbook config) would have to either emit one item per switch — defeating the ergonomic gain — or re-aggregate switches with identical configs, which introduces new logic and edge cases around what counts as "identical." Today neither problem exists.
2. The Ansible argument spec can't cleanly express it. argument_spec has no union type, so switch_ip becomes either type="list", elements="str" (a hard break for every existing playbook using switch_ip: 192.168.1.1, and divergent from every other interface module) or accept-both with manual post-parse coercion in expand_config, which scatters validation between AnsibleModule parsing and the expand step. Either way, validation error messages degrade — the parser no longer cleanly says "switch_ip must be an IP string."
3. Error and diff reporting degrade with M(switch)×N(interface) expansion. Today one user-facing config item maps to N records (one per interface) and the index in expand_config matches the user's mental model. With switch_ip: list, one item produces M×N records; when a duplicate (switch_ip, interface_name) collision happens, the error has to explain "item 0 expanded into (s1, Eth1/1) and item 2 also produced (s1, Eth1/1)" — the user's view (one item) and the system's view (many records) diverge. replaced/overridden modes get worse too: if two items have overlapping switch lists with conflicting configs, precedence has to be defined and documented. Today the rule is just "two explicit items, duplicate detection catches it."
4. YAML anchors already cover the same-config-across-switches case without a schema change:

```yaml
 config:
   - switch_ip: 192.168.1.1
     interface_names: [Ethernet1/1]
     config_data: &shared
       network_os:
         policy:
           access_vlan: 100
   - switch_ip: 192.168.1.2
     interface_names: [Ethernet1/1]
     config_data: *shared
```

5. It wouldn't address the related concern in #3199683197 — a user could still repeat a switch in the list and hit the same opaque duplicate-identifier error. That UX issue is better fixed inside expand_config, which I'll address in a separate reply. 

Happy to revisit if there's a use case where anchors don't suffice — let me know what you think.

plugins/modules/nd_interface_ethernet_access.py
allenrobel: Good points Akini.  Addressed with commit: fe17d78 

The confusing error cited is Gaspard's code, which he should probably address separately.  For now, commit fe17d78 handles config expansion locally in the module so that we can generate user-friendly errors via fail_json.

The commit message for commit fe17d78 explains the changes made in detail.

plugins/modules/nd_interface_ethernet_access.py
allenrobel: Addressed as part of commit: fe17d78 where we added TODO comments and are deferring until all interface modules are merged into develop.

plugins/modules/nd_interface_ethernet_access.py
```diff
@@ -0,0 +1,470 @@
...
+
+        module.exit_json(**nd_state_machine.output.format())
+
+    except NDStateMachineError as e:
+        module_log.exception("NDStateMachineError during module execution")
+        output = nd_state_machine.output.format() if nd_state_machine else {}
+        error_msg = f"Module execution failed: {str(e)}"
+        if module.params.get("output_level") == "debug":
+            error_msg += f"\nTraceback:\n{traceback.format_exc()}"
+        module.fail_json(msg=error_msg, **output)
```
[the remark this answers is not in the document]
allenrobel: Addressed in 0d8dae0. Added a broad `except Exception` handler to `main()` matching the `nd_interface_loopback` reference pattern, so the docstring contract ("catches all exceptions and calls `module.fail_json`") now holds and unexpected exceptions produce structured Ansible output instead of a raw traceback.

plugins/module_utils/orchestrators/ethernet_base.py
```diff
@@ -0,0 +1,437 @@
...
+    delete_endpoint: type[NDEndpointBaseModel] = NDEndpointBaseModel  # unused; delete() uses bulk normalize
+    query_one_endpoint: type[NDEndpointBaseModel] = EpManageInterfacesGet
+    query_all_endpoint: type[NDEndpointBaseModel] = EpManageInterfacesListGet
+    create_bulk_endpoint: type[NDEndpointBaseModel] | None = EpManageInterfacesPost
+    delete_bulk_endpoint: type[NDEndpointBaseModel] | None = EpManageInterfacesNormalize
+
+    PORT_CHANNEL_MODIFIABLE_FIELDS: ClassVar[set[str]] = {"description", "admin_state", "extra_config"}
+
+    _pending_normalizes: list[tuple[str, str]] = []
+
```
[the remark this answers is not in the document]
allenrobel: The leak described here doesn't occur at runtime. `EthernetBaseOrchestrator` is a Pydantic model, and `_pending_normalizes: list[...] = []` was registered as a Pydantic private attribute. Pydantic v2 `smart_deepcopy`s the `[]` default per instance in `__init_private_attributes__`, so each orchestrator instance receives a distinct list -- verified empirically on Pydantic 2.12.5 (two instances, `a is b` -> `False`).

The class-body form *would* be a shared attribute only under the `pydantic_compat` fallback shim (no Pydantic -- `ansible-test sanity`), but orchestrators are never instantiated there, so it's inert.

Applied the suggested change anyway for consistency: `_pending_normalizes` is the same kind of deferred-action queue as `_pending_deploys` / `_pending_removes`, which `NDBaseInterfaceOrchestrator` initializes in `model_post_init`. Initializing one of the three sibling queues via a class-body default was the real inconsistency. Removed the class-body declaration; added a `model_post_init` override that calls `super()` and sets `self._pending_normalizes = []`.

Addressed in 00de022.

plugins/module_utils/orchestrators/ethernet_base.py
```diff
@@ -0,0 +1,437 @@
...
+        ### RuntimeError
+
+        - If the interface is a port-channel member and non-whitelisted fields are being modified.
+        """
+        if existing_data is None:
+            return
+
+        port_channel_id = existing_data.get("configData", {}).get("networkOS", {}).get("policy", {}).get("portChannelId")
+        if not port_channel_id:
+            return
```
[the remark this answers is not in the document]
allenrobel: Fixed in f9403fa.

`_check_port_channel_restrictions` was effectively dead code: `existing_data` was always `None` because `NDStateMachine._execute_operation` never passes that kwarg. The orchestrator now fetches the interface's current wire state itself (review option (a)) via a new `_existing_interface` helper, backed by a per-switch `_switch_interfaces` cache -- one `interfaceList` GET per switch per run, shared with `query_all`, so no N+1. `create` / `update` / `create_bulk` now all validate against real data.

Took option (a) rather than changing the `NDStateMachine` -> orchestrator contract because option (b) would have modified develop-level `nd_state_machine.py`, outside this branch.

Test coverage added in 882c922: five orchestrator tests for the check. The three raise-tests fail with "DID NOT RAISE" against the pre-fix code, confirming they exercise the enforcement rather than passing vacuously.

plugins/modules/nd_interface_ethernet_access.py
```diff
@@ -0,0 +1,470 @@
...
+
+    - If an interface name appears more than once within a single config item's `interface_names` list
+    - If the same `(switch_ip, interface_name)` pair appears in more than one config item
+    """
+    validate_within_item_duplicates(config_list)
+    validate_across_item_duplicates(config_list)
+
+    expanded = []
+    for group in config_list:
+        interface_names = group.get("interface_names", [])
```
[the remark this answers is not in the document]
allenrobel: Fixed in 731a114.

`expand_config` now uses `group.get("interface_names") or []`, so an explicit null `interface_names:` in YAML is treated as empty instead of raising `TypeError: 'NoneType' object is not iterable`. This matches `validate_within_item_duplicates` and `validate_across_item_duplicates`, which already used that idiom.

plugins/modules/nd_interface_ethernet_access.py
```diff
@@ -0,0 +1,597 @@
...
+ANSIBLE_METADATA = {"metadata_version": "1.1", "status": ["preview"], "supported_by": "community"}
+
+DOCUMENTATION = r"""
+---
+module: nd_interface_ethernet_access
+version_added: "2.0.0"
+short_description: Manage ethernet accessHost interfaces on Cisco Nexus Dashboard
+description:
+- Manage ethernet accessHost interfaces on Cisco Nexus Dashboard.
+- It supports creating, updating, querying, and deleting accessHost interface configurations on switches within a fabric.
```
akinross: querying should be removed as we only support merged, replaced, overridden, deleted states

plugins/modules/nd_interface_ethernet_access.py
```diff
@@ -0,0 +1,597 @@
...
+from ansible_collections.cisco.nd.plugins.module_utils.nd_state_machine import NDStateMachine
+from ansible_collections.cisco.nd.plugins.module_utils.orchestrators.base_interface import NDBaseInterfaceOrchestrator
+from ansible_collections.cisco.nd.plugins.module_utils.orchestrators.ethernet_access_interface import EthernetAccessInterfaceOrchestrator
+
+
+# TODO: When all interface modules using `interface_names: list` are merged, lift
+# `validate_interface_names`, `validate_within_item_duplicates`,
+# `validate_across_item_duplicates`, and `expand_config` into a shared helper module
+# (e.g. `plugins/module_utils/interfaces/config_expansion.py`) and import from there.
+def validate_interface_names(config_list):
```
akinross: i notice there are several functions that do not have typehinting, please add.

plugins/modules/nd_interface_ethernet_access.py
```diff
@@ -0,0 +1,597 @@
...
+
+    ### ValueError
+
+    - If any element of `interface_names` is `None` or an empty string.
+    """
+    for item_index, group in enumerate(config_list):
+        switch_ip = group.get("switch_ip")
+        interface_names = group.get("interface_names") or []
+        for entry_index, name in enumerate(interface_names):
+            if name is None or (isinstance(name, str) and not name):
```
akinross: unlikely to hit since it is ansible restricted input of string elements but the conditional leaves room when called with other types to pass
```suggestion
            if not isinstance(name, str) or not name:
```

plugins/module_utils/models/interfaces/interface_default_config.py
```diff
@@ -0,0 +1,198 @@
...
+
+    ## Raises
+
+    None
+    """
+
+    interface_type: str = Field(default="ethernet", alias="interfaceType")
+    config_data: InterfaceDefaultConfigDataModel = Field(default_factory=InterfaceDefaultConfigDataModel, alias="configData")
+
+    PAYLOAD_FIELDS: ClassVar[list[str]] = []
```
akinross: what is this used for, appears stale but I do not know the context in combination with other modules under development?

plugins/module_utils/models/interfaces/ethernet_access_interface.py
```diff
@@ -0,0 +1,362 @@
...
+        the field value before serialization runs — returning None from a field_serializer is too late to
+        drop the key.
+
+        ## Raises
+
+        None
+        """
+        result = handler(self)
+        mode = (info.context or {}).get("mode", "payload")
+        if mode == "config" and isinstance(result, dict):
```
akinross: can result ever be something else then dict? if it is not should we raise an error instead of ignoring?

plugins/module_utils/orchestrators/ethernet_base.py
```diff
@@ -0,0 +1,684 @@
...
+        ### Exception
+
+        - If any PUT request fails (propagated to caller).
+        """
+        results: list[ResponseType] = []
+        for interface_name, switch_id in self._pending_resets:
+            api_endpoint = self._configure_endpoint(self.update_endpoint(), switch_sn=switch_id)
+            api_endpoint.set_identifiers(interface_name)
+            payload = InterfaceDefaultConfig.to_reset_payload(interface_name, switch_id)
+            results.append(self._request(path=api_endpoint.path, verb=api_endpoint.verb, data=payload))
```
akinross: what happens when self._request(path=api_endpoint.path, verb=api_endpoint.verb, data=payload) fails in the after a few interfaces were successfully reset? Should it fail and not attempt the remainder, how should this error be displayed to the user, because now it will display error for all? Can also you add a test for this scenario when decided?

akinross (changes requested):

plugins/modules/nd_interface_ethernet_access.py
allenrobel: Agreed — the module never wired up a query state; the argspec choices are already `[merged, replaced, overridden, deleted]`, so the description line was just stale prose. Removed "querying" in 99732694. Read-back will be handled by a `gathered` state once it lands in the NDStateMachine/NDOutput framework.

plugins/modules/nd_interface_ethernet_access.py
allenrobel: Added in 15398b94 — annotated all five module-level functions (the three validators, `expand_config`, `main`). Config items are dicts, so the validators take `list[dict]`, `expand_config` returns `list[dict]`, and the rest return `None`. Also annotated the local `seen` accumulators that mypy then required once the enclosing functions were typed.

plugins/modules/nd_interface_ethernet_access.py
allenrobel: Adopted your broader check in fa864fae. You're right that the old guard let a non-string (e.g. an int from a templated value) slip through to `name.lower()` and raise a cryptic AttributeError. Now it reports an accurate reason ("not a string (got int)") alongside the existing null/empty cases, and I added non-string parametrize cases proving the rejection path.

plugins/module_utils/models/interfaces/interface_default_config.py
allenrobel: Good catch — that `PAYLOAD_FIELDS` ClassVar was vestigial, defined but never read anywhere. Removed in 697e83ea.

plugins/module_utils/models/interfaces/ethernet_access_interface.py
allenrobel: You're right — `result` is always a dict for a model-level wrap serializer, so the `isinstance` guard silently skipped the strip in a can't-happen case (which would leak `policy_type` into before/after/gathered output). Made it an explicit `AssertionError` in 3577c3b8 so it fails loudly instead, and simplified the condition to `if mode == "config":`. The same `_strip_policy_type_in_config` pattern is in the svi/trunk_host/subinterface branches, so I'll propagate this there too.

plugins/module_utils/orchestrators/ethernet_base.py
allenrobel: Good catch — confirmed and fixed in 3c2e1b92. The loop is fail-fast (stops on the first PUT failure, doesn't attempt the remainder), but you're right that the reporting was wrong: `remove_pending` interpolated the full, still-uncleared `_pending_resets` list, so it reported *every* queued interface as failed even though the earlier ones had already been reset on the controller — and ND has no rollback for a per-interface PUT, so those genuinely stay at fabric default.

`_reset_interfaces` now catches the failing PUT and raises a RuntimeError that names the failed interface, the ones successfully reset before it (now at fabric default, not rolled back), and the ones not attempted — so the user can reconcile the partial state. `remove_pending` no longer re-wraps that detail; the normalize path gets its own accurate message too. Behavior stays fail-fast; only the reporting changed.

Added `test_ethernet_access_orchestrator_00630` for the scenario (first PUT 200, second PUT 500) asserting the partial-state message. Since this is the shared ethernet base, the same fix carries into the trunk_host/svi/subinterface branches.

plugins/modules/nd_interface_ethernet_access.py
```diff
@@ -0,0 +1,604 @@
...
+                suboptions:
+                  admin_state:
+                    description:
+                    - The administrative state of the interface.
+                    - It defaults to C(true) when unset during creation.
+                    type: bool
+                  access_vlan:
+                    description:
+                    - The access VLAN for the interface.
+                    - Valid range is 1-4094.
```
samiib, line 75: Are there any other defaults we should document among the options that are not required?

plugins/modules/nd_interface_ethernet_access.py
```diff
@@ -0,0 +1,604 @@
...
+        interface_names = group.get("interface_names") or []
+        seen: dict[str, bool] = {}
+        for name in interface_names:
+            key = name.lower()
+            if key in seen:
+                raise ValueError(
+                    f"Duplicate interface '{name}' in interface_names for switch '{switch_ip}' "
+                    f"(config item {item_index}). Each interface may appear only once per config item."
+                )
+            seen[key] = True
```
samiib:
```suggestion
        seen: set[str] = set()
        for name in interface_names:
            key = name.lower()
            if key in seen:
                raise ValueError(
                    f"Duplicate interface '{name}' in interface_names for switch '{switch_ip}' "
                    f"(config item {item_index}). Each interface may appear only once per config item."
                )
            seen.add(key)
```
Would it be better to use a set here instead? The stored value of True doesn't seem to be used.

allenrobel: Adopted in e0e27b4. You're right — the stored `True` was never read; membership is the only operation, so `set[str]` is the right structure. Note `validate_across_item_duplicates` keeps its dict deliberately: its stored value (the config item index) is interpolated into the duplicate-pair error message.

plugins/modules/nd_interface_ethernet_access.py
allenrobel: Good catch — there were 19 more. Verified against a live ND 4.2.1 controller rather than trusting the spec: created an accessHost interface with only `policyType` set and read back the stored config. All 19 spec-declared defaults (`intAccessHostTemplate`) matched observed behavior, so 80bd38c documents each with the same "defaults to C(x) when unset during creation" phrasing as `admin_state`. Options the API stores no creation default for (`access_vlan`, `bandwidth`, `description`, storm-control levels, etc.) deliberately make no claim. Doc-only — the module still omits unset fields from payloads.

plugins/modules/nd_interface_ethernet_access.py
```diff
@@ -134,24 +141,29 @@
...
                     choices: [ auto, pointToPoint, shared ]
                   monitor:
                     description:
                     - Whether switchport monitor for SPAN / ERSPAN is enabled.
+                    - It defaults to C(false) when unset during creation.
                     type: bool
                   mtu:
                     description:
                     - The MTU setting for the interface.
+                    - It defaults to C(jumbo) when unset during creation.
```
samiib, line 158: Interesting that it doesn't default to `default`? 😄

allenrobel: It's sorta confusing to me @samiib and I'm actually conflicted on this. 🤷‍♂️ 🫤 

# On a switch console

```bash
S1_LE1# sh run inter e1/41 | beg ^in
interface Ethernet1/41
  mtu 9216

S1_LE1# conf
Enter configuration commands, one per line. End with CNTL/Z.
S1_LE1(config)# default interface eth1/41
S1_LE1(config)# sh run inter e1/41 all | grep mtu
  mtu 1500
S1_LE1(config)#
```

# ND

But, setting the interface in ND to its default trunk/host shows "jumbo" on that screen for MTU (options being "default" and "jumbo" with "jumbo" selected).  So, the template defaults to "jumbo", but NX-OS 'default interface' defaults to 1500.

querying the interface after applying the ND default trunk/host config and then deploying gives the below:

`configData.policy.mtu: "jumbo"`
`operData.mtu: 9216`

`GET /api/v1/manage/fabrics/SITE1/switches/97AJB2MJXO7/interfaces/:interfaceName`

```json
{
  "configData": {
    "mode": "trunk",
    "networkOS": {
      "networkOSType": "nx-os",
      "policy": {
        "adminState": true,
        "allowedVlans": "none",
        "bpduFilter": "default",
        "bpduGuard": "default",
        "cdp": true,
        "debounceTimer": 100,
        "duplexMode": "auto",
        "errorDetectionAcl": true,
        "fec": "auto",
        "linkType": "auto",
        "monitor": false,
        "mtu": "jumbo",
        "negotiateAuto": true,
        "netflow": false,
        "orphanPort": false,
        "pfc": false,
        "policyType": "trunkHost",
        "portTypeEdgeTrunk": true,
        "ptp": false,
        "qos": false,
        "speed": "auto",
        "stormControl": false,
        "vlanMapping": false
      }
    }
  },
  "interfaceName": "Ethernet1/41",
  "interfaceType": "ethernet",
  "operData": {
    "adminStatus": "up",
    "mode": "access",
    "mtu": 9216,
    "operationalDescription": "Link not connected",
    "operationalStatus": "down",
    "portChannelId": -1,
    "speed": "10Gb",
    "switchName": "S1_LE1",
    "vlanRange": "-1"
  },
  "switchId": "97AJB2MJXO7"
}
```

I actually wonder if the interface default profile changes if the fabric type is changed?  I'll try a few experiments...

plugins/modules/nd_interface_ethernet_access.py
allenrobel: Am going to resolve this after 4 days.

plugins/modules/nd_interface_ethernet_access.py
allenrobel: Am going to resolve this after 4 days.

plugins/modules/nd_interface_ethernet_access.py
allenrobel: Am going to resolve this after 4 days.

plugins/module_utils/models/interfaces/ethernet_access_interface.py
allenrobel: Am going to resolve this after 4 days.

plugins/module_utils/orchestrators/ethernet_base.py
allenrobel: Am going to resolve this after 4 days.

gmicol (approved):
LGTM

allenrobel:
### Code review

Found 2 issues:

1. Missing `TODO(4.2.1)` workaround marker on the normalize-as-delete workaround (CLAUDE.md says "Any code that works around an ND API discrepancy ... MUST be flagged with a `TODO(X.Y.Z)` marker so the workaround can be audited and removed when the ND release cadence permits"). `delete_bulk_endpoint = EpManageInterfacesNormalize` encodes the lab-verified ND 4.2.1 behavior that physical interfaces cannot be deleted via `interfaceActions/remove` (silent no-op) or per-interface DELETE (HTTP 500), so delete is implemented as normalize-to-`int_trunk_host`-defaults. Without the marker, `/nd-workaround-audit` cannot find this site.

https://github.com/CiscoDevNet/ansible-nd/blob/cbfb25b0fe140c976833f56e39d5660951a310e6/plugins/module_utils/orchestrators/ethernet_base.py#L87-L89

2. Dead `/deploymentFreeze` fixture and unexercised freeze check. The fixture key `test_query_all_happy_path_00400_freeze` (a `GET /deploymentFreeze` response) is never yielded by any test in `tests/unit/module_utils/orchestrators/test_ethernet_access_interface.py`. `FabricContext.fabric_is_deployment_frozen()` actually reads `fabricStatus` from the `/summary` response, and none of the summary fixtures in this file include a `fabricStatus` field, so the freeze prerequisite check passes vacuously in every orchestrator test (neither the frozen nor the not-frozen branch is exercised). The identical issue was flagged on PR #269 and fixed there in 363dcb8 (dead `*_freeze` entries removed, TEST_NOTES corrected); the loopback fixtures on develop include `fabricStatus` plus a dedicated frozen-path test (00730) that this file could mirror.

https://github.com/CiscoDevNet/ansible-nd/blob/cbfb25b0fe140c976833f56e39d5660951a310e6/tests/unit/module_utils/fixtures/fixture_data/test_ethernet_access_interface.json#L22-L31

plugins/modules/nd_interface_ethernet_access.py
allenrobel: Confirmed against the lab and documented the semantics in 4e591b11.

`jumbo` is the accessHost template's own creation default on ND 4.2.1 (the GUI and API both apply it when `mtu` is unset), so the module follows the platform rather than inventing a different default. The two choices render as:

- `default` -> no `mtu` line, i.e. the NX-OS system default of 1500
- `jumbo` -> `mtu 9216`

I also checked whether `jumbo` tracks fabric settings (the worry being that the effective MTU might float with `fabricMtu` / `l2HostInterfaceMtu`). It doesn't: on SITE1 I set `l2HostInterfaceMtu` 9216 -> 9000, then both an existing and a newly created `jumbo` accessHost interface still rendered `mtu 9216`, and a full `switchActions/deploy?forceShowRun=true` reported "No Commands to execute" (zero drift). So `jumbo` is a fixed 9216, independent of fabric settings. That's now captured in the option description.

akinross (approved):

samiib (approved):

gmicol (approved):

allenrobel:
Merged after 3 approvals.  Thank you @akinross @samiib and @gmicol !

sivakasi-cisco:
https://github.com/CiscoDevNet/ansible-nd/issues/317

PR #223, opened 2026-04-08
Add Bulk create and delete operation support

gmicol:
<html><head></head><body><h2>Summary of Changes (backwards compatible with existing modules)</h2>

<p>Introduced <strong>optional bulk operation support</strong> across three layers:</p>
<h3>1. Orchestrator Layer (<code>NDBaseOrchestrator</code>)</h3>
<ul>
<li>Added <code>supports_bulk_create</code> / <code>supports_bulk_delete</code> class variable flags (default <code>False</code>)</li>
<li>Added <code>create_bulk</code> / <code>delete_bulk</code> methods guarded by a <code>requires_bulk_support</code> decorator — calling them on an unsupported orchestrator raises <code>AttributeError</code></li>
<li>Added <code>create_bulk_endpoint</code> / <code>delete_bulk_endpoint</code> as conditionally required fields — validated at instantiation via <code>@model_validator</code>, only required when their respective flag is <code>True</code></li>
<li>Subclasses opt-in by setting the flag and defining the endpoint:</li>
</ul>
<pre><code class="language-python">class MyOrchestrator(NDBaseOrchestrator[MyModel]):
    supports_bulk_create: ClassVar[bool] = True
    create_bulk_endpoint = EpMyModelBulkPost
</code></pre>
<h3>2. Collection Layer (<code>NDConfigCollection</code>)</h3>
<ul>
<li>Added <code>add_many()</code> / <code>delete_many()</code> to support batch mutations</li>
<li><code>delete_many()</code> performs a <strong>single index rebuild</strong> instead of one per deletion, reducing complexity from O(D×N) to O(N)</li>
</ul>
<h3>3. State Machine Layer (<code>NDStateMachine</code>)</h3>
<ul>
<li>Adopted a <strong>categorize-then-execute</strong> pattern: items are first classified (create/update/delete), then executed in batch</li>
<li>When bulk is supported → single API call for all creates/deletes</li>
<li>When bulk is not supported → falls back to individual API calls transparently</li>
<li><code>self.sent</code> tracking deferred until <strong>after</strong> successful API execution to ensure correctness</li>
</ul>

<p>Bulk support is <strong>fully opt-in</strong> — existing orchestrators require zero changes.</p>
</body></html>

### DIFF
--- a/plugins/module_utils/nd_state_machine.py
+++ b/plugins/module_utils/nd_state_machine.py
@@ -109,6 +109,7 @@ class NDStateMachine:
         items_to_update: List[NDBaseModel] = []
 
         for proposed_item in self.proposed:
+            identifier = None
             try:
                 # Extract identifier
                 identifier = proposed_item.get_identifier_value()
@@ -142,7 +143,10 @@ class NDStateMachine:
                     items_to_create.append(final_item)
 
             except Exception as e:
-                error_msg = f"Failed to process {identifier}: {e}"
+                if identifier:
+                    error_msg = f"Failed to process {identifier}: {e}"
+                else:
+                    error_msg = f"Failed to process: {e}"
                 if not self.ignore_errors:
                     raise NDStateMachineError(error_msg) from e
 
--- a/plugins/module_utils/nd_state_machine.py
+++ b/plugins/module_utils/nd_state_machine.py
@@ -5,7 +5,7 @@
 
 from __future__ import absolute_import, division, print_function
 
-from typing import Type, Union, List
+from typing import Type
 from ansible.module_utils.basic import AnsibleModule
 from ansible_collections.cisco.nd.plugins.module_utils.nd import NDModule
 from ansible_collections.cisco.nd.plugins.module_utils.nd_output import NDOutput
@@ -38,6 +38,12 @@ class NDStateMachine:
 
         self.model_class = self.model_orchestrator.model_class
         self.state = self.module.params["state"]
+
+        # Cached flags
+        self.check_mode = self.module.check_mode
+        self.ignore_errors = self.module.params.get("ignore_errors", False)
+        self.supports_bulk_create = getattr(self.model_orchestrator, "supports_bulk_create", False)
+        self.supports_bulk_delete = getattr(self.model_orchestrator, "supports_bulk_delete", False)
 
         # Initialize collections
         try:
@@ -78,7 +84,7 @@ class NDStateMachine:
         """
         Handle merged/replaced/overridden states.
         """
-        items_to_create_bulk: List = []
+        items_to_create_bulk = []
 
         for proposed_item in self.proposed:
             # Extract identifier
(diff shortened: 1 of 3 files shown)


shrsr (dismissed):
LGTM

mikewiebe:
@gmicol Are there any changes required to modules that are using this infrastructure currently without bulk APIs?  Is it backwards compatible?

gmicol:
@mikewiebe there is no change needed to modules that are using this infrastructure currently without bulk APIs. I made sure it was backwards compatible.

plugins/module_utils/orchestrators/base.py
```diff
@@ -73,3 +95,19 @@ def query_all(self, model_instance: Optional[ModelType] = None, **kwargs) -> Res
...
+    def validate_bulk_endpoints(self):
+        if self.supports_bulk_create and self.create_bulk_endpoint is None:
+            raise ValueError(f"'{self.__class__.__name__}' has 'supports_bulk_create=True' but 'create_bulk_endpoint' is not defined.")
+        if self.supports_bulk_delete and self.delete_bulk_endpoint is None:
+            raise ValueError(f"'{self.__class__.__name__}' has 'supports_bulk_delete=True' but 'delete_bulk_endpoint' is not defined.")
+        return self
+
+    @requires_bulk_support("supports_bulk_create")
+    def create_bulk(self, model_instances: List[ModelType], **kwargs) -> ResponseType:
+        pass
```
allenrobel: Since subclasses are expected to implement these, should we consider raising NotImplementedError rather than pass for create_bulk and delete_bulk?  This way, if they are called, but not implemented, we'll catch it right away.

plugins/module_utils/nd_state_machine.py
```diff
@@ -97,69 +111,88 @@ def _manage_create_update_state(self) -> None:
...
                         self.model_orchestrator.update(final_item)
                 elif diff_status == "new":
-                    if not self.module.check_mode:
-                        self.model_orchestrator.create(final_item)
+                    if not self.check_mode:
+                        if self.supports_bulk_create:
+                            items_to_create_bulk.append(final_item)
+                        else:
+                            self.model_orchestrator.create(final_item)
                 self.sent.add(final_item)
```
allenrobel: What's the expected behavior if create_bulk() fails and ignore_errors is True?  Are the items in self.sent considered sent and would still appear in the output?  Wondering if self.sent.add() should be moved to after the bulk call succeeds?

plugins/module_utils/nd_state_machine.py
```diff
@@ -97,69 +111,88 @@ def _manage_create_update_state(self) -> None:
...
+                identifier = item.get_identifier_value()
+
+                if not self.check_mode:
+                    if self.supports_bulk_delete:
+                        items_to_delete_bulk.append(item)
+                    else:
+                        self.model_orchestrator.delete(item)
 
                 # Remove from collection
                 self.existing.delete(identifier)
```
allenrobel: Same comment as with self.sent.add() being called before create_bulk.  What is the expected behavior when ignore_errors is True and the delete_bulk call fails?

plugins/module_utils/nd_state_machine.py
```diff
@@ -97,69 +111,88 @@ def _manage_create_update_state(self) -> None:
...
+        self._delete_items(items_to_delete)
 
     def _manage_delete_state(self) -> None:
         """Handle deleted state."""
+        items_to_delete = []
+
         for proposed_item in self.proposed:
-            try:
-                identifier = proposed_item.get_identifier_value()
+            identifier = proposed_item.get_identifier_value()
```
allenrobel: What's the expected behavior if get_identifier_value raises and ignore_errors is True?  Previously, this was wrapped in a try/except with the except branch containing a conditional for ignore_errors.

plugins/module_utils/nd_state_machine.py
```diff
@@ -29,10 +30,21 @@ def __init__(self, module: AnsibleModule, model_orchestrator: Type[NDBaseOrchest
         self.output = NDOutput(output_level=module.params.get("output_level", "normal"))
 
         # Configuration
-        self.model_orchestrator = model_orchestrator(sender=self.nd_module)
+        # Accept either an orchestrator instance or a class.
+        if isinstance(model_orchestrator, type):
```
allenrobel: Nit, but anything that's a class will pass this check, not just orchestrator classes.  Fine as long as folks are careful.

plugins/module_utils/nd_state_machine.py
```diff
@@ -97,69 +111,88 @@ def _manage_create_update_state(self) -> None:
...
+        if items_to_delete_bulk:
+            try:
+                self.model_orchestrator.delete_bulk(items_to_delete_bulk)
+            except Exception as e:
+                error_msg = f"Failed to delete in bulk: {e}"
+                if not self.ignore_errors:
                     raise NDStateMachineError(error_msg) from e
+
+        # Log deletion
+        self.output.assign(after=self.existing)
```
allenrobel, line 205: Is this needed given the same on line 139?  Just asking since I'm not sure.

allenrobel (commented):
Similar to @mikewiebe, would be good for reviewers if there were a more detailed description of the changes and design decisions in the PR.

plugins/module_utils/orchestrators/base.py
gmicol: That's a great idea! I will update these methods

plugins/module_utils/nd_state_machine.py
gmicol: You are right, in the case of a bulk operation when `ignore_errors` is `True`, the items would still appear in `self.sent` in the output even though the request failed. `self.sent.add()` should be moved after the bulk call succeeds.

plugins/module_utils/nd_state_machine.py
gmicol: same answer as before https://github.com/CiscoDevNet/ansible-nd/pull/223#discussion_r3055226537

plugins/module_utils/nd_state_machine.py
gmicol: The error would not be catch correctly, will add it back to the try/except

plugins/module_utils/nd_state_machine.py
gmicol: You definitely right. Therefore, I will make the condition more specific to avoid any mistake by devs on the orchestrator class type

plugins/module_utils/nd_state_machine.py
gmicol: It is needed as this is for delete operation output (`_manage_override_deletions` and `_manage_delete_state` methods), whereas line 139 is for saving `_manage_create_update_state()` output.

But you could argue that this repetitive code that could be set directly in the  broader `mnage_state` function.

gmicol:
> Similar to @mikewiebe, would be good for reviewers if there were a more detailed description of the changes and design decisions in the PR.

I agree, I added a concise description of the main changes.

shrsr (approved):
LGTM

allenrobel (approved):
LGTM

mikewiebe (approved):
LGTM

PR #246, opened 2026-04-09
Add VXLAN (MSD) Fabric Group Module

### LINKED ISSUE #245: [ND 4.2+] Add VXLAN Fabric Group (MSD) Type Module

```shell
A new module to support ND 4.2+ API VXLAN fabric group (MSD) is required.
```

manage

```shell

```

```yaml
TBD
```

### Ansible Version

```shell

```

```shell

```

```shell
4.2.1
```

### Cisco NX-OS Version

```shell

```

### Cisco ACI Version

```shell

```

mtarking:
## Related Issue(s)

Resolves #245

## Proposed Changes

Add new `nd_manage_fabric_group_vxlan` module for managing VXLAN Fabric Group (MSD) fabrics on Nexus Dashboard, aligned with the shared fabric framework used by the other `nd_manage_fabric_*` modules.

New files:

- `plugins/module_utils/models/manage_fabric_group/enums.py` — Fabric-group-specific enums (`FabricGroupTypeEnum`, `MultisiteOverlayInterConnectTypeEnum`, `CloudSecAlgorithmEnum`, `CloudSecEnforcementEnum`, `SecurityGroupTagEnum`)
- `plugins/module_utils/models/manage_fabric_group/manage_fabric_group_vxlan.py` — Pydantic models (`FabricGroupVxlanModel`, `VxlanFabricGroupManagementModel`, `RouteServerModel`) with `Literal[FabricGroupTypeEnum.VXLAN]` type discriminator. `fabric_name` uses the shared `NdFabricName` type and is propagated to `management.name`. The module argspec is auto-generated from the model via the shared `_build_options_from_model` helper (single-value `Literal` fields such as `category` and `management.type` are intentionally excluded from the argspec).
- `plugins/module_utils/orchestrators/manage_fabric_group_vxlan.py` — Orchestrator built on `ConfigActionsMixin` + `NDBaseOrchestrator`, reusing the `EpManageFabrics*` endpoints with a custom `query_all` that uses the `?category=fabricGroup` query parameter for server-side filtering plus client-side filtering on `management.type == "vxlan"`
- `plugins/modules/nd_manage_fabric_group_vxlan.py` — Module built on `NDStateMachine` supporting `merged`, `replaced`, `overridden`, and `deleted` states, plus `config_actions` (save/deploy)

New tests:

- `tests/unit/module_utils/endpoints/test_endpoints_api_v1_manage_fabric_group_vxlan.py` — 14 endpoint unit tests (orchestrator constructed via `RestSend`; `query_all` verified against a monkeypatched `_request`, asserting the `?category=fabricGroup` path)
- `tests/integration/targets/nd_manage_fabric_group_vxlan/tasks/main.yaml` — Integration tests covering create, update (merged), replace, override, delete, check mode, and idempotency
- `tests/integration/targets/nd_manage_fabric_group_vxlan/vars/main.yaml` — Test variables

## Test Notes

- `ansible-test units` — 14 endpoint unit tests passing, 0 failures
- `ansible-test sanity` — 24/24 passing (ansible-core 2.18), including `validate-modules`, `import`, `ansible-doc`, and `yamllint`
- Integration tests cover: single fabric group creation, merged updates, replaced state, overridden state, deletion, check mode for all states, and idempotency verification

4.2.1

- [ ] analyze
- [ ] infra
- [x] manage
- [ ] onemanage
- [ ] other

## Checklist

- [x] Assigned the proper reviewers


plugins/module_utils/models/manage_fabric_group/manage_fabric_group_vxlan.py
```diff
@@ -0,0 +1,474 @@
...
+        alias="multisiteInterConnectBgpAuthentication",
+        description="Enables or disables the BGP authentication for inter-site links",
+        default=False,
+    )
+    multisite_inter_connect_bgp_auth_key_type: BgpAuthenticationKeyTypeEnum = Field(
+        alias="multisiteInterConnectBgpAuthKeyType",
+        description="BGP key encryption type: 3 - 3DES, 6 - Cisco type 6, 7 - Cisco type 7",
+        default=BgpAuthenticationKeyTypeEnum.THREE_DES,
+    )
+    multisite_inter_connect_bgp_key: Optional[str] = Field(
```
mikewiebe, line 256: **@mtarking Need to decide if we want to handle this now or just handle this as part of the larger [#359](https://github.com/CiscoDevNet/ansible-nd/issues/359)**

#### High: BGP and CloudSec keys are exposed through arguments and module output

**Issue**

Both secret-bearing fields are ordinary strings. The generated nested argspec does not mark them `no_log`, and the model does not distinguish between serialization for user-visible module output and serialization for an ND request. As a result, the real values remain visible instead of being unmasked only for the narrowly controlled API-payload path and, if exact stable comparison is proven, a strictly private comparison path.

**Evidence**

- [`multisite_inter_connect_bgp_key` lines 256-262](https://github.com/CiscoDevNet/ansible-nd/pull/246/files#diff-f8dd1b48f7f49aaa13938a58a299992966cac7ff56c351bf8da7197a931f149cR256) is a plain `Optional[str]`.
- [`cloud_sec_key` lines 356-362](https://github.com/CiscoDevNet/ansible-nd/pull/246/files#diff-f8dd1b48f7f49aaa13938a58a299992966cac7ff56c351bf8da7197a931f149cR356) is also a plain `Optional[str]`.
- [`main()` lines 590-597](https://github.com/CiscoDevNet/ansible-nd/pull/246/files#diff-0ab850f6805fd6865ddf740fb131bc0ae3d8b245906629c501384ccc6aad2a6dR590) installs the generated argspec directly, while [`proposed` and `api_payload` lines 532-578](https://github.com/CiscoDevNet/ansible-nd/pull/246/files#diff-0ab850f6805fd6865ddf740fb131bc0ae3d8b245906629c501384ccc6aad2a6dR532) expose serialized configuration at higher output levels.
- The exact-head argspec for each field was only `{'type': 'str'}`; Ansible's `_list_no_log_values()` returned `set()`. `to_config()` and `to_payload()` retained the supplied values verbatim.
- The established [`FabricBgpDetailsModel.auth_key` implementation](https://github.com/CiscoDevNet/ansible-nd/blob/d723f8efaa2759702650a88ed4ebb83c8ca26da0/plugins/module_utils/models/l3out/l3out.py#L57-L69) demonstrates `SecretStr` and context-aware serialization. Its use of a private `diff` context does not, by itself, prove that these two fabric-group keys are returned by ND byte-for-byte and can be compared safely.

**Existing PR overlap**

Tracked by open issue [#359](https://github.com/CiscoDevNet/ansible-nd/issues/359), which explicitly covers authentication/MACsec-style keys, nested argspec `no_log`, aliases, model/output masking, and success/failure tests. This PR adds two new affected fields.

**Impact**

Reversible or reusable BGP and CloudSec key material can appear in task arguments, registered results, `before`/`after`/`proposed`, verbose API payloads, and failure diagnostics.

For example, a user can supply both keys in an otherwise normal fabric-group task:

```yaml
- name: Configure authentication for the production fabric group
  cisco.nd.nd_manage_fabric_group_vxlan:
    state: merged
    config:
      - fabric_name: MSD-PROD
        management:
          multisite_inter_connect_bgp_authentication: true
          multisite_inter_connect_bgp_key: "070C285F4D06"
          auto_configure_cloud_sec: true
          cloud_sec_key: "060506324F41"
  register: result
```

In the current implementation, `result.proposed` can contain the snake-case values and verbose `api_payload` output can contain their API aliases:

```json
{
  "proposed": {
    "management": {
      "multisite_inter_connect_bgp_key": "070C285F4D06",
      "cloud_sec_key": "060506324F41"
    }
  },
  "api_payload": {
    "management": {
      "multisiteInterConnectBgpKey": "070C285F4D06",
      "cloudSecKey": "060506324F41"
    }
  }
}
```

Those Cisco Type 7-style example values are reversible and may be reusable on the switches. After the fix, user-visible output should contain `VALUE_SPECIFIED_IN_NO_LOG_PARAMETER`, while the actual POST/PUT body sent to ND must still use the original values. A private comparison representation may use the original values only if API or lab evidence proves that ND returns each key byte-for-byte; otherwise these fields must follow write-only/transformed-secret behavior.

**Suggested fix**

Enable the existing secret pattern for these two properties as follows:

1. Make secret metadata the authoritative declaration and retain `SecretStr` as defense in depth. Import `SecretStr`, `FieldSerializationInfo`, and `field_serializer` from `common.pydantic_compat`, and change both annotations in `VxlanFabricGroupManagementModel`:

   ```python
   multisite_inter_connect_bgp_key: Optional[SecretStr] = Field(
       ...,
       json_schema_extra={"secret": True},
   )
   cloud_sec_key: Optional[SecretStr] = Field(
       ...,
       json_schema_extra={"secret": True},
   )
   ```

2. Add one field serializer covering both properties. It should return the real value for API payload generation (and an explicitly private `internal` handoff, if this module needs one), preserve empty values, and return the standard mask for every user-visible, missing, or unknown context:

   ```python
   @field_serializer(
       "multisite_inter_connect_bgp_key",
       "cloud_sec_key",
   )
   def serialize_secret_key(
       self,
       value: Optional[SecretStr],
       info: FieldSerializationInfo,
   ) -> Optional[str]:
       if value is None:
           return None
       raw_value = value.get_secret_value()
       if raw_value == "":
           return ""
       mode = (info.context or {}).get("mode")
       if mode in ("payload", "internal"):
           return raw_value
       return "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER"
   ```

3. Classify each key before defining its comparison behavior. Do **not** add `diff` to the plaintext serializer allowlist merely because L3Out currently does so. First obtain API or lab evidence that ND returns the exact original value:

   - If ND returns the value byte-for-byte, it may be compared only through a strictly private representation that cannot enter public diff, verbose results, errors, or logs.
   - If ND blanks, transforms, or re-encrypts the value, omit it from diff and gathered/replayable output. Explicit user input then means update/rotation intent, while omission must remain idempotent; the `nd_manage_links` write-only-secret handling is the safer reference.

   `NDBaseModel.to_payload()` already supplies `context={"mode": "payload"}`, so POST/PUT requests can continue to carry the usable values.

4. Mark both generated nested options as `no_log=True`. The local targeted change can set the flags in `get_argument_spec()` immediately after `_build_options_from_model(cls)` returns:

   ```python
   config_options = _build_options_from_model(cls)
   management_options = config_options["management"]["options"]
   management_options["multisite_inter_connect_bgp_key"]["no_log"] = True
   management_options["cloud_sec_key"]["no_log"] = True
   ```

   The shared implementation should teach `_build_options_from_model()` to emit `{"type": "str", "no_log": True}` when field metadata contains `{"secret": True}`. Metadata must be the authoritative signal, with `SecretStr` detection retained only as a compatibility backstop. `SecretStr` protects model serialization, while `no_log=True` gives Ansible the original values it must scrub from task arguments, returned payloads, and exception text; neither control replaces the other.

5. Both properties are nested beneath `management`, while the current generic metadata handling is root-only. Issue [#359](https://github.com/CiscoDevNet/ansible-nd/issues/359) may provide recursive discovery, masking, and diff removal (more investigation needed) but until that is a reality, add the equivalent explicit nested/root handling in this PR. Also ensure raw `api_payload`, `api_response`, `api_result`, controller errors, and verbose logs are sanitized by key/alias as well as by exact input value, since transformed controller values can evade Ansible's value-based scrubber.

Add tests proving that both nested argspec entries have `no_log=True`; `to_config()`, `before`, `after`, `proposed`, public diff, verbose output, and failure output never contain the real values; and `to_payload()` retains the real values. Exercise input by both snake-case field names and API aliases. For comparison behavior, test either (a) private stable-value comparison backed by exact-echo evidence or (b) explicit-secret update/rotation intent plus omitted-secret idempotency for blanked/transformed values.

plugins/module_utils/models/manage_fabric_group/manage_fabric_group_vxlan.py
```diff
@@ -0,0 +1,474 @@
...
+        alias="l3VniRange",
+        description="Overlay VRF identifier range (minimum: 1, maximum: 16777214)",
+        default="50000-59000",
+    )
+    downstream_vni: bool = Field(
+        alias="downstreamVni",
+        description="Enable unique per-fabric virtual network identifier (VNI)",
+        default=False,
+    )
+    downstream_l2_vni_range: str = Field(
```
mikewiebe, line 130: ### Medium: Enabling downstream VNI sends blank ranges instead of the documented defaults

**Issue**

The module documentation and ND schema advertise downstream L2/L3 defaults, but the model defaults both fields to empty strings and serializes those values into the request.

**Evidence**

- [`downstream_l2_vni_range` and `downstream_l3_vni_range` lines 130-139](https://github.com/CiscoDevNet/ansible-nd/pull/246/files#diff-f8dd1b48f7f49aaa13938a58a299992966cac7ff56c351bf8da7197a931f149cR130) default to `""`.
- [`DOCUMENTATION` lines 61-72](https://github.com/CiscoDevNet/ansible-nd/pull/246/files#diff-0ab850f6805fd6865ddf740fb131bc0ae3d8b245906629c501384ccc6aad2a6dR61) advertises `60000-69000` and `80000-89000`; [the ND schema](https://github.com/allenrobel/nd-openapi-mcp/blob/bdf5d57bcda663c8aa03cb7c972fa72f01480929/schemas/4.2.1/manage.json#L94037) defines the same defaults.
- Exact-head serialization of `management: {downstream_vni: true}` produced:

  ```text
  {'downstreamVni': True,
   'downstreamL2VniRange': '',
   'downstreamL3VniRange': ''}
  ```

- The integration test always supplies both ranges when enabling downstream VNI, so it cannot catch the default path.

**Existing PR overlap**

No matching open issue found.

**Impact**

A user enables downstream VNI allocation and relies on the documented defaults:

```yaml
- name: Enable downstream VNI allocation
  cisco.nd.nd_manage_fabric_group_vxlan:
    state: merged
    config:
      - fabric_name: MSD-Production
        management:
          downstream_vni: true
```

Based on the module documentation, they expect the request to use:

```json
{
  "downstreamVni": true,
  "downstreamL2VniRange": "60000-69000",
  "downstreamL3VniRange": "80000-89000"
}
```

Instead, the model currently generates:

```json
{
  "downstreamVni": true,
  "downstreamL2VniRange": "",
  "downstreamL3VniRange": ""
}
```

ND could reject the request or enable downstream VNI without usable allocation ranges.

The `replaced` case is more concerning. Suppose the fabric group already has valid ranges:

```text
L2: 60000-69000
L3: 80000-89000
```

A user submits `downstream_vni: true` while omitting the ranges, expecting omitted properties to revert to their documented defaults. The module instead sends empty strings, potentially clearing those working ranges.

**Suggested fix**

When downstream VNI is enabled, either default to the documented ranges or require both ranges explicitly. Preserve blank values only for the disabled controller-response shape. Test enabled-without-ranges, disabled-without-ranges, create, and `replaced` serialization.

---

plugins/module_utils/models/manage_fabric_group/manage_fabric_group_vxlan.py
```diff
@@ -0,0 +1,474 @@
...
+    Route server configuration for multi-site overlay interconnect.
+
+    ## Raises
+
+    - `ValueError` - If IP address or ASN format is invalid
+    """
+
+    model_config = ConfigDict(str_strip_whitespace=True, validate_assignment=True, populate_by_name=True, extra="allow")
+
+    route_server_ip: str = Field(alias="routeServerIp", description="Route Server IP Address")
```
mikewiebe, line 83: ### Medium: The model accepts values that violate explicit OpenAPI formats

**Issue**

Several fields are plain strings even though the ND schema supplies IP/CIDR formats or regex constraints. Invalid data therefore reaches the controller instead of failing consistently during model validation.

**Evidence**

- [`RouteServerModel` line 83](https://github.com/CiscoDevNet/ansible-nd/pull/246/files#diff-f8dd1b48f7f49aaa13938a58a299992966cac7ff56c351bf8da7197a931f149cR83) leaves the route-server IP unvalidated; [the multi-site IP ranges lines 279-305](https://github.com/CiscoDevNet/ansible-nd/pull/246/files#diff-f8dd1b48f7f49aaa13938a58a299992966cac7ff56c351bf8da7197a931f149cR279) are also plain strings.
- [`security_group_tag_prefix` lines 327-332](https://github.com/CiscoDevNet/ansible-nd/pull/246/files#diff-f8dd1b48f7f49aaa13938a58a299992966cac7ff56c351bf8da7197a931f149cR327), [`cloud_sec_key` lines 356-362](https://github.com/CiscoDevNet/ansible-nd/pull/246/files#diff-f8dd1b48f7f49aaa13938a58a299992966cac7ff56c351bf8da7197a931f149cR356), and [`scheduled_backup_time` lines 387-390](https://github.com/CiscoDevNet/ansible-nd/pull/246/files#diff-f8dd1b48f7f49aaa13938a58a299992966cac7ff56c351bf8da7197a931f149cR387) enforce no format beyond length where present.
- ND 4.2.1 defines an IP regex for `routeServerIp`, `cidrv4`/`cidrv6` formats for the multi-site ranges, an allowlisted-character regex for `securityGroupTagPrefix`, a hex regex for `cloudSecKey`, and a 24-hour regex for [`scheduledBackupTime`](https://github.com/allenrobel/nd-openapi-mcp/blob/bdf5d57bcda663c8aa03cb7c972fa72f01480929/schemas/4.2.1/manage.json#L88423).
- An exact-head model probe accepted all of `not-ip`, `not-cidr`, `not-v6`, `bad!`, `not-hex`, and `99:99`.

**Existing PR overlap**

No matching open issue found.

**Impact**

Users get late, controller-specific failures instead of deterministic Ansible validation. In a multi-group task, an earlier group can already be mutated before a later invalid value is rejected by ND.

**Suggested fix**

**Reuse the repository's existing validation framework; this PR does not need to invent a parallel validation mechanism.** Current `develop` already provides [`validate_ip_address()` and `validate_cidr()`](https://github.com/CiscoDevNet/ansible-nd/blob/67197ec9bab9822912e2768b72863370bad967fa/plugins/module_utils/common/validators.py#L81-L123), along with the shared [`IPv4CIDR` and `IPv6CIDR` annotated types](https://github.com/CiscoDevNet/ansible-nd/blob/67197ec9bab9822912e2768b72863370bad967fa/plugins/module_utils/models/types.py#L74-L124). [`SwitchDataModel`](https://github.com/CiscoDevNet/ansible-nd/blob/67197ec9bab9822912e2768b72863370bad967fa/plugins/module_utils/models/manage_switches/switch_data_models.py#L260-L263) is a concrete example of calling a shared network validator from a Pydantic `field_validator`.

For ordinary API-specific string formats, the repository convention is to use declarative Pydantic `Field(pattern=...)` constraints, as demonstrated by [`RouteMapModel.name` and `tenant_name`](https://github.com/CiscoDevNet/ansible-nd/blob/67197ec9bab9822912e2768b72863370bad967fa/plugins/module_utils/models/manage_route_map/manage_route_map.py#L512-L535). Cross-field semantics and values requiring special handling belong in `field_validator` or `model_validator` methods.

Apply those existing patterns here as follows:

- Validate `route_server_ip` with the shared IP-address validator, while enforcing the IPv4 family required by this API field.
- Type `multisite_loopback_ip_range` and `multisite_underlay_subnet_range` with the shared `IPv4CIDR` type, and use `IPv6CIDR` for the corresponding IPv6 ranges.
- Put the exact OpenAPI patterns directly on the ordinary string fields `security_group_tag_prefix` and `scheduled_backup_time` using `Field(pattern=...)`. If `HH:MM` will be reused across the fabric family, a shared annotated type followed by migration of the older models would be preferable.
- Apply the separate secret-handling contract from [#359](https://github.com/CiscoDevNet/ansible-nd/issues/359) to `cloud_sec_key`. Because that field should become `SecretStr`, do not apply a string `Field(pattern=...)` constraint directly to it—Pydantic 2 cannot apply that constraint after conversion to `SecretStr`. Validate its unwrapped value in a secret-aware `field_validator`, return the `SecretStr`, use a generic error message, and enable `hide_input_in_errors=True` so a rejected key is not reproduced in the validation error.

There is no shared arbitrary-hexadecimal or `HH:MM` type yet. Older fabric-family models also leave [`scheduled_backup_time`](https://github.com/CiscoDevNet/ansible-nd/blob/67197ec9bab9822912e2768b72863370bad967fa/plugins/module_utils/models/manage_fabric/manage_fabric_ibgp_vxlan.py#L329-L333) and [`security_group_tag_prefix`](https://github.com/CiscoDevNet/ansible-nd/blob/67197ec9bab9822912e2768b72863370bad967fa/plugins/module_utils/models/manage_fabric/manage_fabric_ibgp_vxlan.py#L637-L640) as unconstrained strings. That part is an inherited family inconsistency rather than a regression introduced only by PR #246, but this new module should follow the repository's newer Pydantic-first convention instead of extending the gap.

Add parameterized valid-boundary and invalid-value tests for every field, including IPv4/IPv6 family enforcement and the exact `00:00` through `23:59` time boundary.

---

plugins/module_utils/models/manage_fabric_group/manage_fabric_group_vxlan.py
mtarking: Thanks for the thorough review, @mikewiebe. I decided to **handle the secret exposure in this PR** rather than defer it to #359, and I also went one step further and retrofitted the same protection onto the existing eBGP/iBGP (and inherited AI eBGP/iBGP) fabric models so the family is consistent. #359 can still track the broader recursive/automated audit, but the two new fields this PR introduces — plus the pre-existing sibling keys — are fully protected here.

I got lab evidence to resolve the open comparison question: on a live ND 4.2.1 cluster, **ND echoes both `cloudSecKey` and `multisiteInterConnectBgpKey` back byte-for-byte on GET**. Even with that, I chose the safer **write-only / masked** behavior (per the `nd_manage_links` reference you cited) rather than a private stable-compare path — the byte-for-byte echo just guarantees there's no perpetual diff, since both the proposed and existing values mask to the same sentinel.

Fixed in this PR. Implemented exactly along the lines you suggested:

1. **`SecretStr` + authoritative metadata** — both `multisite_inter_connect_bgp_key` and `cloud_sec_key` are now `Optional[SecretStr]` with `json_schema_extra={"secret": True}`.
2. **Context-aware serializer** — a single `serialize_secret_key` field serializer (delegating to a shared `serialize_secret_value` helper) returns the real value **only** for `mode == "payload"`, preserves empty values, and returns `VALUE_SPECIFIED_IN_NO_LOG_PARAMETER` for every other/user-visible/missing context. `to_diff_dict()` sets no mode, so the diff path is masked too.
3. **`no_log` is now driven by metadata, not a local hack** — I taught the shared `_build_options_from_model()` to emit `{"type": "str", "no_log": True}` whenever a field's metadata carries `{"secret": True}`, with `SecretStr` detection retained as a compatibility backstop. This works for the nested `management` options, so no root-only special-casing is needed.
4. **Comparison classification** — on live ND 4.2.1, both keys are returned byte-for-byte on GET. I still went with **write-only masking** (the safer `nd_manage_links` model): `to_payload()` carries the real values via `context={"mode": "payload"}`, while `before`/`after`/`proposed`/public diff/verbose/errors never contain them. Idempotency holds because both sides mask to the same sentinel; explicit input means rotation intent.

**Tests added** proving: nested argspec entries have `no_log=True`; `to_config()` / `to_diff_dict()` mask both fields; `to_payload()` retains the real values; input accepted by both snake_case and API alias.

I also applied the same pattern to the existing eBGP (5 keys) and iBGP (9 keys) fabric models — AI eBGP/iBGP inherit it automatically — with dedicated masking tests, so the whole fabric family is consistent.

plugins/module_utils/models/manage_fabric_group/manage_fabric_group_vxlan.py
mtarking: Fixed. `downstream_l2_vni_range` and `downstream_l3_vni_range` now carry concrete defaults instead of `""`:

- `downstream_l2_vni_range` → `10030000-10049000`
- `downstream_l3_vni_range` → `10050000-10059000`

These are the actual ND 4.2.1 schema defaults for the downstream ranges, and I reconciled the DOCUMENTATION to match the model (the previously-advertised `60000-69000`/`80000-89000` were the non-downstream `l3_vni_range`-style values). Enabling `downstream_vni: true` without supplying ranges now serializes the documented defaults rather than empty strings, so there's no risk of ND rejecting the request or of the `replaced` path clearing existing working ranges.

Integration/unit tests now cover the enabled-without-ranges path so the default serialization is actually exercised.

plugins/module_utils/models/manage_fabric_group/manage_fabric_group_vxlan.py
mtarking: Fixed using the existing repository validation framework — no parallel mechanism introduced:

- **`route_server_ip`** — validated with the shared IP-address validator and enforced to the IPv4 family required by this API field.
- **Multi-site ranges** — `multisite_loopback_ip_range` and `multisite_underlay_subnet_range` now use the shared `IPv4CIDR` annotated type; the IPv6 ranges use `IPv6CIDR`.
- **`security_group_tag_prefix`** — declarative `Field(pattern=r"^[a-zA-Z0-9-_]*$")`.
- **`scheduled_backup_time`** — declarative `Field(pattern=r"^([01]\d|2[0-3]):([0-5]\d)$")` (00:00–23:59).
- **`cloud_sec_key`** — since it's now `SecretStr`, I did **not** put a string `Field(pattern=...)` on it. Instead it's validated in a secret-aware `field_validator` on the unwrapped value with a generic error message, and `hide_input_in_errors=True` is set so a rejected key is never reproduced in the validation error. I also added a cross-field `model_validator` enforcing ND's exact length rule (66 hex for `AES_128_CMAC`, 130 for `AES_256_CMAC`), which I confirmed against live ND — the message never echoes the secret.

Parameterized valid-boundary and invalid-value tests were added for each field, including IPv4/IPv6 family enforcement and the `00:00`/`23:59` time boundaries.

Re: the inherited family inconsistency you noted (older models leaving `scheduled_backup_time`/`security_group_tag_prefix` unconstrained) — this new module follows the newer Pydantic-first convention rather than extending the gap.

plugins/module_utils/models/manage_fabric_group/manage_fabric_group_vxlan.py
mtarking: Thanks for the thorough review, @mikewiebe. I handled the secret exposure in this PR rather than deferring to #359, and retrofitted the same protection onto the existing eBGP/iBGP (and inherited AI eBGP/iBGP) fabric models so the family is consistent. #359 can still track the broader recursive audit.

On the open comparison question: on a live ND 4.2.1 cluster, ND echoes both `cloudSecKey` and `multisiteInterConnectBgpKey` back byte-for-byte on GET. Even so, I chose the safer write-only/masked behaviour (per the `nd_manage_links` reference you cited) rather than a private stable-compare path. The byte-for-byte echo guarantees there's no perpetual diff, since proposed and existing both mask to the same sentinel.

What's implemented:

1. **`SecretStr` + authoritative metadata** — both fields are `Optional[SecretStr]` with `json_schema_extra={"secret": True}`.
2. **Context-aware serializer** — a single `serialize_secret_key` serializer (delegating to the shared `serialize_secret_value` helper) returns the real value only for `mode == "payload"`, and `VALUE_SPECIFIED_IN_NO_LOG_PARAMETER` for every other context. `to_diff_dict()` sets no mode, so the diff path is masked too.
3. **`no_log` driven by metadata** — the shared `_build_options_from_model()` emits `{"type": "str", "no_log": True}` whenever a field's metadata carries `{"secret": True}`, with `SecretStr` detection retained as a backstop. This works for the nested `management` options, so no root-only special-casing.
4. **`hide_input_in_errors=True`** so a rejected key is never reproduced in a validation error.

plugins/module_utils/models/manage_fabric_group/manage_fabric_group_vxlan.py
mtarking: Fixed. `downstream_l2_vni_range` and `downstream_l3_vni_range` now carry concrete defaults instead of `""`:

- `downstream_l2_vni_range` → `10030000-10049000`
- `downstream_l3_vni_range` → `10050000-10059000`

These are the actual ND 4.2.1 schema defaults for the downstream ranges. I also reconciled DOCUMENTATION to match — the previously advertised `60000-69000` / `80000-89000` were the non-downstream `l3_vni_range`-style values.

plugins/module_utils/models/manage_fabric_group/manage_fabric_group_vxlan.py
mtarking: Fixed using the existing repository validation framework — no parallel mechanism introduced:

- **`route_server_ip`** — shared `validate_ip_address` validator. The ND schema declares `format: ip`, which permits both families, so the validator accepts IPv4 and IPv6 and rejects anything that isn't an IP.
- **Multi-site ranges** — `multisite_loopback_ip_range` and `multisite_underlay_subnet_range` use the shared `IPv4CIDR` annotated type; the IPv6 ranges use `IPv6CIDR`.
- **`security_group_tag_prefix`** — declarative `Field(pattern=r"^[a-zA-Z0-9-_]*$", min_length=1, max_length=10)`.
- **`scheduled_backup_time`** — declarative `Field(pattern=r"^([01]\d|2[0-3]):([0-5]\d)$")` (00:00–23:59).
- **`cloud_sec_key`** — since it's now `SecretStr`, I did not put a string `Field(pattern=...)` on it. It's validated in a secret-aware `field_validator` on the unwrapped value with a generic error message, plus a cross-field `model_validator` enforcing ND's exact length rule (66 hex for `AES_128_CMAC`, 130 for `AES_256_CMAC`), confirmed against live ND. With `hide_input_in_errors=True`, the message never echoes the secret.

plugins/module_utils/models/manage_fabric_group/manage_fabric_group_vxlan.py
mikewiebe: Thanks @mtarking, there were a few more things identified but the remaining work is being deferred to [issue #359](https://github.com/CiscoDevNet/ansible-nd/issues/359) for implementation in a subsequent PR.

Issue #359 has now been updated to explicitly cover the PR #246 case where ND returns `cloudSecKey` or `multisiteInterConnectBgpKey` from GET while the current task either omits the key or supplies a different replacement value. Its acceptance criteria require `-vvv` tests proving that the controller-returned value is absent from the complete result—including `api_response`, `api_result`, merged `api_payload`, errors, and logs—without altering the private request body intentionally sent to ND.

Issue #359 also remains the source of truth for explicit secret rotation versus omitted-secret idempotency. The local `SecretStr`, metadata, and `no_log` work in this PR can remain as implemented; resolving this thread as tracked by #359.

plugins/module_utils/models/base.py
```diff
@@ -215,6 +215,48 @@ def to_gathered_config(self, **kwargs) -> Dict[str, Any]:
 
     # --- Core Deserialization ---
 
+    @model_validator(mode="before")
+    @classmethod
+    def _treat_empty_string_as_unset(cls, data: Any, info: Any) -> Any:
```
allenrobel, line 226: This before-validator sits on `NDBaseModel`, and `from_config()` (below, unchanged) sets `mode="config"` for every state, so the `""` -> unset coercion now applies to every model in the collection and to `merged` state, not just the fabric models this PR converted for ND 4.3.1.

Concrete regression: every interface model declares `description: AsciiDescription = Field(default=None, ...)` (e.g. `interfaces/ethernet_access_interface.py:111`, `loopback_interface.py:120`). A merged-state task with `description: ""` to clear a description is now dropped before validation, never lands in `model_fields_set`, and is never sent. The same happens for an explicit `""` on `Optional[SecretStr]` fields meant to clear a key.

Commit bb985d74 (#422) deliberately confined empty-equals-absent to the replaced/overridden reverse diff and states "Merged-state semantics are unchanged." The new `test_fabric_empty_string_payload.py` covers only fabric models.

Suggest scoping this: an opt-in `ClassVar` on the fabric model families, or gating on the state carried in the context, rather than unconditional on the base class.

plugins/module_utils/orchestrators/config_actions/mixin.py
```diff
@@ -0,0 +1,369 @@
...
+
+        - Via `_get_fabric_switches` when the switches query fails with a non-404 status.
+        """
+        allowlist = set(only_switch_ids) if only_switch_ids is not None else None
+        eligible_fabrics: list[str] = []
+        switch_ids_by_fabric: dict[str, tuple[str, ...]] = {}
+        for fabric_name in fabric_names:
+            switches = self._get_fabric_switches(fabric_name)
+            if not switches:
+                self.rest_send.warn(f"Skipping config save/deploy for '{fabric_name}': fabric has no switches.")
```
allenrobel, line 187: Please verify this path against a live controller for a fabric group before merge. `ManageFabricGroupVxlanOrchestrator` overrides only `deploy_global_endpoint`; `switches_endpoint` is inherited, so this guard calls `GET /fabrics/{group}/switches` on the group's own name. If ND returns an empty list for a fabric group (the group is a container; its switches belong to the member fabrics, and your own docstring notes the deploy API does not reach member fabrics by default), the group is dropped from `eligible_fabrics` here and the controller returns `skipped`/`no_fabrics`, so `save`, `deploy type: global`, and `deploy type: switch` all silently no-op with a warning. That would contradict the module's "Save and deploy fabric group configuration after changes" example and make the new `inclAllFabricGroupsSwitches` override unreachable.

The 4.2.1 spec says `configSave` applies to "the specified fabric or fabric group" but says nothing about whether the switches listing aggregates member switches, and no unit or integration test in this PR exercises `config_actions` on a fabric group. A single integration task with `config_actions: {save: true, deploy: true, type: global}` against a group with members would settle it either way.

plugins/modules/nd_manage_fabric_group_vxlan.py
```diff
@@ -0,0 +1,660 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2026, Matt Tarkington (@mtarking)
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
```
allenrobel: New file. CLAUDE.md: "Boilerplate `from future ...` imports are no longer needed and can be removed (with the exception of annotations)" and "`__metaclass__ = type` is no longer needed and can be removed." Same pattern in the new `models/manage_fabric_group/manage_fabric_group_vxlan.py:7-9`, `orchestrators/manage_fabric_group_vxlan.py:7-9`, `models/manage_fabric_group/enums.py:24` (`__metaclass__` only), and the four new test files. The new `orchestrators/config_actions/` package already uses only `from __future__ import annotations`, which is the target.

plugins/module_utils/models/manage_fabric_group/manage_fabric_group_vxlan.py
```diff
@@ -0,0 +1,513 @@
...
+# Copyright: (c) 2026, Matt Tarkington (@mtarking)
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+import re
+from typing import List, Dict, Optional, ClassVar, Literal
```
allenrobel: CLAUDE.md: "Modern type annotations (PEP 585 / PEP 604) for all new code, and when touching existing annotations: `X | None` instead of `Optional[X]` ... `list[X]`, `dict[K, V]`, `set[X]` ... instead of `List`, `Dict`, `Set` ... Drop the now-unused `Optional`, `Union`, `List`, `Dict`, `Set`, `Tuple` imports from `typing`." Applies to this new file (`Optional[List[RouteServerModel]]`, `get_argument_spec(cls) -> Dict`), to the new `_nullable_default_keys(cls) -> Set[str]` in `models/base.py:244-246`, and to the fields this PR converted to `Optional[SecretStr]` / `Optional[str] = None` in `manage_fabric_ebgp_vxlan.py`, `manage_fabric_ibgp_vxlan.py`, and `manage_fabric_external.py`.

allenrobel (commented):
### Code review

One collection-wide behavior change in `models/base.py`, one config_actions path on fabric groups that needs a live-controller check before merge, and two CLAUDE.md convention items on the new files. Details inline.

plugins/module_utils/models/manage_fabric_group/manage_fabric_group_vxlan.py
mtarking: Resolved.

plugins/modules/nd_manage_fabric_group_vxlan.py
mtarking: Resolved.

plugins/module_utils/orchestrators/config_actions/mixin.py
mtarking: This is intended at this time as we cannot save and deploy without switches.

plugins/module_utils/models/base.py
mtarking: Resolved.

mikewiebe (approved):
LGTM

PR #248, opened 2026-04-09
Add Fabric Group Members Module

### LINKED ISSUE #247: [ND 4.2+] Add VXLAN Fabric Group Members Module

```shell
A new module to support ND 4.2+ API fabric group members is required.
```

manage

```shell

```

```yaml
TBD
```

### Ansible Version

```shell

```

```shell

```

```shell
4.2.1
```

### Cisco NX-OS Version

```shell

```

### Cisco ACI Version

```shell

```

mtarking:
## Related Issue(s)

Resolves #247

## Proposed Changes

Add a new `nd_manage_fabric_group_members` module for managing fabric group membership on Nexus Dashboard, aligned to the current orchestrator / `NDStateMachine` framework and extended to multi-cluster fabric groups (OneManage) with optional config save/deploy.

### Module & framework alignment
- `plugins/modules/nd_manage_fabric_group_members.py` — supports `merged`, `deleted`, and `gathered` states. `merged`/`deleted` route through `NDStateMachine`; `gathered` runs via a RestSend-backed orchestrator. Full `RETURN` documentation.
- `plugins/module_utils/orchestrators/manage_fabric_group_members.py` — orchestrator using the aligned `rest_send` / `_request` bulk pattern (replaces the earlier non-framework `sender.*` usage).
- `plugins/module_utils/models/manage_fabric_group/manage_fabric_group_members.py` — Pydantic model. Read-only `fabric_type` is excluded from the diff for `merged` idempotency; `fabric_name` is the top-level parent-group parameter.

### Multi-cluster fabric group (OneManage)
- Auto-detects whether `fabric_name` is a plain fabric group (Manage) or a multi-cluster fabric group (OneManage) via a cached OneManage fabric probe (`category == multiClusterFabricGroup`) and routes member add/remove/get to the correct surface.
- Adds per-member `cluster_name` with a composite `(cluster_name, member_name)` identity for MCFG members.
- `plugins/module_utils/endpoints/v1/onemanage/onemanage_fabrics.py` — adds OneManage member add/remove endpoints.

### Config actions (save/deploy)
- Adds a `config_actions` option (`save` / `deploy` / `type`) to save and optionally deploy the fabric group after membership changes (skipped for `deleted`/`gathered` and when nothing changed).
- `plugins/module_utils/orchestrators/config_actions_mixin.py` — endpoint selection made overridable via hooks (Manage remains the default for all existing consumers).
- OneManage `configSave` / `deploy` / `switches` / `switchActions/deploy` endpoints added; the members orchestrator overrides the hooks so save/deploy target the correct surface based on `is_multicluster`.

## Test Notes

- New unit tests: model, orchestrator (Manage + MCFG member operations and Manage/MCFG save & deploy paths), and OneManage endpoint coverage; existing Manage member-endpoint tests retained.
- Full `tests/unit/module_utils/orchestrators` and `tests/unit/module_utils/endpoints` suites pass (1373 tests) — confirming the `ConfigActionsMixin` hook refactor introduces no regressions for the fabric/ToR/vPC consumers.
- Integration tests cover add (merged), idempotency on re-add, remove (deleted), idempotency on re-delete, gathered state, and check mode.

4.2.1

* [ ] analyze
* [ ] infa
* [x] manage
* [x] onemanage
* [ ] other

## Checklist


plugins/module_utils/orchestrators/manage_fabric_group_members.py
```diff
@@ -0,0 +1,212 @@
...
+        try:
+            api_endpoint = self._query_endpoint()
+            result = self._request(path=api_endpoint.path, verb=api_endpoint.verb, not_found_ok=True)
+            if isinstance(result, dict):
+                return result.get("fabrics", []) or []
+            return result or []
+        except Exception as e:
+            raise Exception(f"Query all members failed: {e}") from e
+
+    def create_bulk(self, model_instances: List[FabricGroupMemberModel], **kwargs) -> ResponseType:
```
mikewiebe, line 184: ### High: Multi-cluster membership writes use the single-cluster bulk request body

**Issue**

The orchestrator sends `{"members": [...]}` for both API surfaces. That is the Manage bulk contract, but the OneManage add/remove operations accept one `multiClusterFabricGroupMemberUpdate` object per request, with `clusterName` and `name` at the root.

**Evidence**

- [`create_bulk()` constructs the same wrapper for every endpoint](https://github.com/CiscoDevNet/ansible-nd/pull/248/files#diff-61156c401096b3eb4f77b8e12fd1577f0cbd146e69da42f53cb6a8978c7d5e7bR184-R195), and [`delete_bulk()` repeats it](https://github.com/CiscoDevNet/ansible-nd/pull/248/files#diff-61156c401096b3eb4f77b8e12fd1577f0cbd146e69da42f53cb6a8978c7d5e7bR197-R209).
- [The OneManage add-members operation](https://github.com/allenrobel/nd-openapi-mcp/blob/bdf5d57bcda663c8aa03cb7c972fa72f01480929/schemas/4.2.1/oneManage.json#L2127-L2149) consumes the singular [`multiClusterFabricGroupMemberUpdate` schema](https://github.com/allenrobel/nd-openapi-mcp/blob/bdf5d57bcda663c8aa03cb7c972fa72f01480929/schemas/4.2.1/oneManage.json#L25723-L25743). The remove operation has the same singular shape.
- [`test_add_members()` and `test_remove_members()`](https://github.com/CiscoDevNet/ansible-nd/pull/248/files#diff-1145b9cefe8d411a4fcdc3253b257137fec326955e892779594b2f4124a58665R254-R289) assert the incorrect wrapper, so the tests preserve rather than catch the mismatch.

**Existing PR overlap**

No matching open issue found. [#247](https://github.com/CiscoDevNet/ansible-nd/issues/247) is the implementation umbrella, not a defect report for this request shape.

**Impact**

For an MCFG named `MCFG-PROD`, this desired configuration:

```yaml
config:
  - member_name: SITE-A
    fabric_type: mcfg
    cluster_name: CLUSTER-1
  - member_name: SITE-B
    fabric_type: mcfg
    cluster_name: CLUSTER-2
```

currently produces one request like:

```json
{
  "members": [
    {"name": "SITE-A", "clusterName": "CLUSTER-1"},
    {"name": "SITE-B", "clusterName": "CLUSTER-2"}
  ]
}
```

The OneManage contract instead requires two requests whose bodies are respectively `{"name":"SITE-A","clusterName":"CLUSTER-1"}` and `{"name":"SITE-B","clusterName":"CLUSTER-2"}`. A valid multi-cluster task can therefore fail before adding or removing any member.

**Suggested fix**

Branch on the selected API surface: retain the bulk `members` envelope for Manage, but fan out OneManage mutations as one direct member object per call. Preserve accepted items if a later call fails, and add add/remove tests for multiple clusters and a partial failure.

---

plugins/module_utils/orchestrators/manage_fabric_group_members.py
```diff
@@ -0,0 +1,212 @@
...
+
+        Detection mirrors the VRF/Network resolver: a plain fabric group is unknown to OneManage
+        (the probe GET 404s), whereas a multi-cluster fabric group is reported by OneManage with
+        category 'multiClusterFabricGroup'. Probed once and cached for the orchestrator's lifetime.
+        """
+        if self._multicluster is None:
+            self._multicluster = self._detect_multicluster()
+        return self._multicluster
+
+    def _detect_multicluster(self) -> bool:
```
mikewiebe, line 84: ### High: API-surface discovery converts every probe failure into a Manage result

**Issue**

The multi-cluster probe catches every exception and returns `False`. Authentication failures, authorization failures, timeouts, server errors, and malformed responses are therefore classified as proof that the fabric is a normal Manage fabric group.

**Evidence**

- [`_detect_multicluster()` suppresses every exception](https://github.com/CiscoDevNet/ansible-nd/pull/248/files#diff-61156c401096b3eb4f77b8e12fd1577f0cbd146e69da42f53cb6a8978c7d5e7bR84-R92).
- That result immediately selects [the Manage rather than OneManage endpoint family](https://github.com/CiscoDevNet/ansible-nd/pull/248/files#diff-61156c401096b3eb4f77b8e12fd1577f0cbd146e69da42f53cb6a8978c7d5e7bR94-R113).
- Current NaC explicitly recognizes an expected first-run OneManage `400` response while preparing MCFG membership; that is evidence that surface discovery needs response-specific handling rather than a blanket exception fallback.

**Existing PR overlap**

No matching open issue found.

**Impact**

For example, a transient `503` from the OneManage lookup for `MCFG-PROD` is interpreted as “not multi-cluster.” The same run can then issue a write to `/api/v1/manage/fabrics/MCFG-PROD/...` with a single-cluster payload. The diagnostic that mattered—the `503`—is lost, and a read failure is allowed to redirect later mutations.

**Suggested fix**

Only map an explicitly documented “not found/not this category” response to the Manage surface. Propagate authentication, authorization, transport, 5xx, malformed-response, and otherwise ambiguous failures with no mutation. Add one test for the accepted negative response and separate zero-write tests for 401, 403, 5xx, timeout, and malformed data.

---

plugins/module_utils/orchestrators/manage_fabric_group_members.py
```diff
@@ -0,0 +1,212 @@
...
+
+    # --- ConfigActionsMixin hook overrides ---
+    # Route save/deploy to the OneManage surface for a multi-cluster fabric group; otherwise the
+    # mixin's Manage defaults apply. The OneManage endpoints mirror Manage 1:1 (same bodies), so
+    # the mixin's save/switch-filter/deploy logic is reused unchanged.
+    def _config_save_endpoint(self, fabric_name: str) -> NDEndpointBaseModel:
+        endpoint_cls = EpOneManageFabricsConfigSavePost if self.is_multicluster else EpFabricConfigSavePost
+        return endpoint_cls(fabric_name=fabric_name)
+
+    def _deploy_global_endpoint(self, fabric_name: str) -> NDEndpointBaseModel:
```
mikewiebe, line 123: ### High: Manage global deployment omits the flag required to include member fabrics

**Issue**

The module documents global deployment as deployment to all switches in the fabric group, but its Manage global hook uses a bare deploy endpoint. The Manage API explicitly defaults `inclAllFabricGroupsSwitches` to `false`, which means changes are not deployed to member fabrics. The OneManage hook also omits its analogous parameter, but that schema does not declare the omission default, so the OneManage runtime effect requires live confirmation.

**Evidence**

- [`get_config_deploy_endpoint()` returns the ordinary bare global endpoint](https://github.com/CiscoDevNet/ansible-nd/pull/248/files#diff-61156c401096b3eb4f77b8e12fd1577f0cbd146e69da42f53cb6a8978c7d5e7bR115-R129).
- The module promises [`type: global` applies to all switches](https://github.com/CiscoDevNet/ansible-nd/pull/248/files#diff-91a09630837dd294bc29ffb4df119e5daf6b55785ee1d7800e6c13bddf293a47R56-R84).
- [The ND 4.2.1 Manage deploy contract](https://developer.cisco.com/docs/nexus-dashboard/latest/execute-fabric-config-deploy/) defines `inclAllFabricGroupsSwitches`, explicitly defaults it to `false`, and says `false` does not deploy changes to member fabrics.
- The OneManage operation exposes `includeAllFabricGroupSwitches`, but its schema provides `example: false` without a declared default. It is therefore valid to say the PR does not explicitly request all children, but not yet to claim that omission definitively excludes them.
- `ConfigActionsMixin` queries the parent's switches before deployment, but for `type: global` that collection is used only as a non-empty guard. The mutually exclusive `type: switch` branch is the only path that sends `switchIds`; global mode ignores the prefetched switch list and sends one request without a body.

**Existing PR overlap**

Related open issue [#368](https://github.com/CiscoDevNet/ansible-nd/issues/368) tracks the collection-wide `config_actions` contract. It does not spell out these two fabric-group parameters, but this finding belongs in that workstream.

**Impact**

Assume `MSD-PROD` contains three child fabrics:

| Member fabric | Switches |
| --- | --- |
| `SITE-A` | `leaf-a1`, `leaf-a2` |
| `SITE-B` | `leaf-b1`, `leaf-b2` |
| `SITE-C` | `leaf-c1`, `leaf-c2` |

The user adds `SITE-C` and requests a global save and deploy:

```yaml
- name: Add SITE-C and deploy the fabric group
  cisco.nd.nd_manage_fabric_group_members:
    fabric_name: MSD-PROD
    config:
      - member_name: SITE-C
    state: merged
    config_actions:
      save: true
      deploy: true
      type: global
```

The user expects this sequence:

1. Add `SITE-C` to `MSD-PROD`.
2. Save the pending configuration.
3. Deploy to all six switches belonging to the group's member fabrics.
4. Finish with the switches in a deployed/in-sync state.

The module currently sends a request equivalent to:

```http
POST /api/v1/manage/fabrics/MSD-PROD/actions/deploy
```

The API treats this as:

```text
inclAllFabricGroupsSwitches=false
```

Because the fabric group is a container and its switches belong to its child fabrics, the operation does not expand into `SITE-A`, `SITE-B`, and `SITE-C`. ND may return success even though those switches remain pending.

The request needs to be:

```http
POST /api/v1/manage/fabrics/MSD-PROD/actions/deploy?inclAllFabricGroupsSwitches=true
```

For an MCFG through OneManage, the API exposes the analogous parameter:

```text
includeAllFabricGroupSwitches=true
```

However, OneManage does not declare the default applied when that optional parameter is omitted, and its description is preview-oriented. A live OneManage test is therefore needed before asserting that the bare request necessarily leaves all MCFG children pending.

For the ordinary Manage/MSD example above, the visible symptom is conclusive from the contract: **the Ansible task succeeds, but the ND UI still shows pending configuration on switches in the member fabrics.**

**Suggested fix**

Set `inclAllFabricGroupsSwitches=true` on the Manage global deploy endpoint and assert the exact URL. Model `includeAllFabricGroupSwitches` on the OneManage endpoint as well, then validate its omitted/true behavior live and set it explicitly to `true` when `type: global` is intended to mean every child switch. Add integration coverage that verifies the final deployment state on member-fabric switches for both surfaces.

---

plugins/modules/nd_manage_fabric_group_members.py
```diff
@@ -0,0 +1,276 @@
...
+            nd_state_machine = NDStateMachine(
+                module=module,
+                model_orchestrator=ManageFabricGroupMembersOrchestrator,
+            )
+
+            nd_state_machine.manage_state()
+
+            # Save/deploy the parent fabric group only when membership actually changed.
+            if module.params["state"] != "deleted" and len(nd_state_machine.sent) > 0:
+                nd_state_machine.model_orchestrator.execute_config_actions(
```
mikewiebe, line 262: ### Medium: Switch deployment filters a snapshot taken before config-save

**Issue**

The shared config-actions workflow queries switch synchronization state before saving the fabric configuration. It then reuses that cached response after config-save to decide which switches require deployment.

Config-save can create new pending configuration and change a switch from `inSync` to `outOfSync`. The cached response cannot reflect that change.

**Evidence**

- The module invokes actions [after membership reconciliation](https://github.com/CiscoDevNet/ansible-nd/pull/248/files#diff-91a09630837dd294bc29ffb4df119e5daf6b55785ee1d7800e6c13bddf293a47R260-R267).
- [`execute_config_actions()` queries switches before config-save](https://github.com/CiscoDevNet/ansible-nd/blob/0ad04328b086f9ad775678fc0695e004d4de93a9/plugins/module_utils/orchestrators/config_actions_mixin.py#L262-L272).
- [`config_deploy()` receives and reuses that prefetched list](https://github.com/CiscoDevNet/ansible-nd/blob/0ad04328b086f9ad775678fc0695e004d4de93a9/plugins/module_utils/orchestrators/config_actions_mixin.py#L125-L147).
- The existing “deploy after saving” unit test encodes GET → save → deploy without simulating a post-save synchronization-state change.

**Existing PR overlap**

Related open issue [#368](https://github.com/CiscoDevNet/ansible-nd/issues/368) tracks shared `config_actions` behavior, including switch-target selection and save/deploy ordering.

**Impact**

Suppose `leaf-1` initially has no pending configuration:

```json
{
  "serialNumber": "FDO123",
  "additionalData": {
    "configSyncStatus": "inSync"
  }
}
```

The user adds a member and requests switch-scoped deployment:

```yaml
- name: Add SITE-C and deploy pending switch configuration
  cisco.nd.nd_manage_fabric_group_members:
    fabric_name: MSD-PROD
    config:
      - member_name: SITE-C
    state: merged
    config_actions:
      save: true
      deploy: true
      type: switch
```

The module performs this sequence:

1. Add `SITE-C`.
2. Query the switches; `leaf-1` reports `inSync`.
3. Run config-save.
4. Config-save recalculates intent, making `leaf-1` `outOfSync`.
5. Filter the cached result from step 2.
6. Find no out-of-sync switches and skip deployment.

The current run can therefore report success while ND shows `leaf-1` with pending configuration.

**Suggested fix**

Change the ordering to:

1. Run config-save.
2. Query the current, fully paginated switch state.
3. Select switches that now require deployment.
4. Call `switchActions/deploy` with those identifiers.

Add a regression test where the pre-save response is `inSync`, the post-save response is `outOfSync`, and deployment includes the switch.

---

plugins/modules/nd_manage_fabric_group_members.py
```diff
@@ -0,0 +1,276 @@
...
+        else:
+            nd_state_machine = NDStateMachine(
+                module=module,
+                model_orchestrator=ManageFabricGroupMembersOrchestrator,
+            )
+
+            nd_state_machine.manage_state()
+
+            # Save/deploy the parent fabric group only when membership actually changed.
+            if module.params["state"] != "deleted" and len(nd_state_machine.sent) > 0:
```
mikewiebe, line 261: ### Medium: A failed save or deploy cannot be completed by an ordinary retry

**Issue**

Config actions run only when the membership state machine sent a CRUD request during the current invocation.

If the membership change succeeds but config-save or deployment fails, the next run sees the membership as already converged. Because no new membership request is sent, the module skips the requested save and deployment actions.

**Evidence**

- Actions are gated on [`len(state_machine.sent) > 0`](https://github.com/CiscoDevNet/ansible-nd/pull/248/files#diff-91a09630837dd294bc29ffb4df119e5daf6b55785ee1d7800e6c13bddf293a47R260-R267).
- The exception path returns only [`fail_json(msg=...)`](https://github.com/CiscoDevNet/ansible-nd/pull/248/files#diff-91a09630837dd294bc29ffb4df119e5daf6b55785ee1d7800e6c13bddf293a47R269-R272), so accepted CRUD progress is not preserved in normal result fields.
- Issue #368's shared contract specifically identifies action eligibility on an idempotent resource run as a decision that must be handled explicitly.

**Existing PR overlap**

Related open issue [#368](https://github.com/CiscoDevNet/ansible-nd/issues/368) covers the shared config-actions contract and no-diff/pending-state eligibility.

Related issue [#546](https://github.com/CiscoDevNet/ansible-nd/issues/546) covers controller-accepted mutations becoming stranded after later failures, but it is currently scoped to interface modules rather than this resource.

**Impact**

The user runs:

```yaml
- name: Add SITE-C and deploy the change
  cisco.nd.nd_manage_fabric_group_members:
    fabric_name: MSD-PROD
    config:
      - member_name: SITE-C
    state: merged
    config_actions:
      save: true
      deploy: true
      type: switch
```

On the first run:

1. `addMembers` succeeds and `SITE-C` becomes a member.
2. Config-save succeeds.
3. Deployment returns `500`.
4. The module reports failure.
5. ND retains pending switch configuration.

The user fixes the transient controller problem and runs the same playbook again:

1. The module sees that `SITE-C` is already a member.
2. The membership state machine sends no CRUD request.
3. `nd_state_machine.sent` is empty.
4. The module skips config-save and deployment entirely.
5. The task reports no change and succeeds.
6. The pending configuration remains undeployed.

The same retry hole occurs when membership succeeds but config-save fails.

**Suggested fix**

Do not base action eligibility solely on whether membership CRUD occurred during the current invocation.

When the user explicitly requests config actions, allow the module to complete them on a converged membership run when the controller reports pending or out-of-sync state. Also preserve successfully accepted membership changes and completed action steps in structured failure output.

Add a two-run regression test:

1. Membership succeeds and save/deploy fails.
2. The identical retry detects pending state and completes the requested action without trying to add the member again.

---

plugins/module_utils/endpoints/v1/manage/manage_fabric_group_members.py
```diff
@@ -0,0 +1,227 @@
...
+    Provides common functionality for all HTTP methods on the
+    /api/v1/manage/fabrics/{fabric_name}/members and
+    /api/v1/manage/fabrics/{fabric_name}/actions/addMembers|removeMembers endpoints.
+
+    Subclasses override ``_path_suffix`` to build the correct path.
+    """
+
+    _path_suffix: ClassVar[str] = "members"
+
+    endpoint_params: EndpointQueryParams = Field(default_factory=EndpointQueryParams, description="Endpoint-specific query parameters")
```
mikewiebe, line 48: ### Medium: The PR reintroduces unescaped fabric names through parallel endpoint classes

**Issue**

The new endpoint classes interpolate `fabric_name` directly into URL paths without encoding it.

The branch already contains canonical Manage membership endpoints that encode path components. PR #248 adds a parallel endpoint family that bypasses that behavior and reintroduces the same defect previously fixed elsewhere.

**Evidence**

- The new Manage membership endpoint builds its path from the [raw `fabric_name`](https://github.com/CiscoDevNet/ansible-nd/pull/248/files#diff-9eb04ffa57d41a219bed7e628f1c2da9fc6bdfec364554e26d5aec201ab7de9fR48-R62).
- `BasePath.path()` simply joins its arguments with `/`; it does not URL-encode them.
- The new OneManage endpoint family likewise interpolates its [raw group name](https://github.com/CiscoDevNet/ansible-nd/pull/248/files#diff-8ad15b240efe64ceb76193c07237fe6faceae1096c9d22afa049273ae12997cdR25-R35).
- Current `develop` contains [canonical Manage membership endpoints that encode the fabric name](https://github.com/CiscoDevNet/ansible-nd/blob/67197ec9bab9822912e2768b72863370bad967fa/plugins/module_utils/endpoints/v1/manage/manage_fabrics.py#L658-L730).

**Existing PR overlap**

No matching open issue found. Closed issue [#292](https://github.com/CiscoDevNet/ansible-nd/issues/292) tracked this same raw path-parameter class of defect and was fixed by PR #491.

**Impact**

Suppose a valid fabric group is named:

```text
MSD West/Blue
```

The user runs:

```yaml
- name: Add SITE-A to the fabric group
  cisco.nd.nd_manage_fabric_group_members:
    fabric_name: "MSD West/Blue"
    config:
      - member_name: SITE-A
    state: merged
```

The module constructs a path equivalent to:

```text
/api/v1/manage/fabrics/MSD West/Blue/actions/addMembers
```

The `/` inside the fabric name is interpreted as a path separator. Instead of identifying one fabric named `MSD West/Blue`, the router sees separate `MSD West` and `Blue` path components.

The correctly encoded request is:

```text
/api/v1/manage/fabrics/MSD%20West%2FBlue/actions/addMembers
```

Other valid or accepted characters can cause similar failures:

| Character | Possible effect when unescaped |
| --- | --- |
| Space | Invalid or inconsistently normalized URL |
| `/` | Creates an unintended path segment |
| `?` | Starts a query string |
| `#` | Starts a fragment that is not sent to the server |
| `%` | Is interpreted as the beginning of an encoded sequence |

As a result, membership queries, additions, removals, save operations, and deployment can fail or target the wrong route for affected names.

**Suggested fix**

Semantically rebase onto current `develop` and reuse its canonical encoded Manage membership endpoint classes instead of maintaining a second family.

For OneManage, encode the fabric-name path component using:

```python
quote(self.fabric_name, safe="")
```

Add exact endpoint tests for names containing spaces, `/`, `#`, `%`, and `?`.

---

plugins/module_utils/orchestrators/manage_fabric_group_members.py
mtarking: Fixed in `419f0d1a`. Live testing showed this finding was broader than reported.

You flagged the OneManage path. Testing against ND 4.2.1 and 4.3.1 showed **Manage rejects multi-member bodies too**:

```
POST .../actions/addMembers     {"members":[{...},{...}]}  -> 400 "Only one member fabric can be added at a time"
POST .../actions/removeMembers  {"members":[{...},{...}]}  -> 400 "Only one member fabric can be deleted at a time"
POST .../actions/addMembers     {"members":[{...}]}        -> passes, reaches field validation
```

The Manage schema types `members` as an array, which is what led to the batching code. The schema is misleading about what the implementation actually accepts.

`_send_members()` now fans out to one request per member on **both** surfaces. Only the envelope differs: Manage takes a one-element `{"members": [{...}]}`, OneManage takes a flat `{"clusterName": ..., "name": ...}`.

The previous unit test asserted a two-element body — it encoded the bug — and has been rewritten to assert fan-out.

Caveat: verified on hardware for the Manage surface only. The OneManage branch is unit-tested but no MCFG controller was available.

plugins/module_utils/orchestrators/manage_fabric_group_members.py
mtarking: Fixed in `419f0d1a`. Worth recording that my first attempt at this was wrong.

I initially replaced the blanket `except` with `not_found_ok=True`, on the reasoning that a non-MCFG fabric 404s. That broke **every single-cluster run** at initialization. A single-cluster controller rejects the entire OneManage surface before it resolves the fabric name:

```
GET /api/v1/oneManage/manage/fabrics/fg1                       -> 400
GET /api/v1/oneManage/manage/fabrics/ansible_test_fabric_group -> 400
GET /api/v1/oneManage/manage/fabrics/definitely_no_such_fabric -> 400

{"code":400,"message":"Multi-cluster environment must be configured before using this feature"}
```

The 404 never occurs there. So the original blanket `except` was masking real errors — your point stands — but it was also the only thing making single-cluster work.

`ONEMANAGE_UNAVAILABLE_CODES = {400, 404}` now means "not MCFG". Authentication, transport and 5xx failures still propagate, so the diagnostic you wanted preserved is preserved.

Three tests cover this. I verified the new 400 test fails without the fix, reproducing the exact production error, and that the 5xx-propagation test still passes.

plugins/module_utils/orchestrators/manage_fabric_group_members.py
mtarking: Fixed in `419f0d1a`. `deploy_global_endpoint()` now passes `FabricDeployQueryParams(incl_all_fabric_groups_switches=True)` on the Manage path; OneManage continues to use its own deploy endpoint.

Covered by `test_manage_fabric_group_members_deploy_global_manage_includes_member_switches`.

plugins/modules/nd_manage_fabric_group_members.py
mtarking: Fixed in `419f0d1a`, via the shared mixin rather than locally.

This module now uses `plugins/module_utils/orchestrators/config_actions/mixin.py`. Its `resolve_switch_deploy_targets()` re-queries switch state after config-save, for exactly the reason you gave: config-save can push a previously `inSync` switch out of sync, and that switch still needs deploying.

plugins/modules/nd_manage_fabric_group_members.py
mtarking: **Not resolved — only partially addressed.** Leaving this open rather than closing it.

Addressed: failures now preserve `before` / `after` / `diff` through `fail_json(**output)`, so a failed run is diagnosable rather than opaque.

Not addressed: the gate you identified is unchanged.

```python
if state != "deleted" and len(nd_state_machine.sent) > 0:
```

If the membership change succeeds and config-save or deploy fails, the next run sees membership already converged, sends no CRUD request, and skips the requested actions.

This is the same design question as #368 and applies to the other config-actions modules, not just this one, so I'd rather not change the contract here unilaterally. 

I'd like to defer this for now.

plugins/module_utils/endpoints/v1/manage/manage_fabric_group_members.py
mtarking: Fixed in `419f0d1a` by deletion rather than by adding encoding.

The parallel endpoint family and its test file are removed (−500 lines). The orchestrator now uses the canonical `EpManageFabricsMembersGet` / `EpManageFabricsMembersAddPost` / `EpManageFabricsMembersRemovePost`, which already encode path components.

No `manage_fabric_group_members` endpoint file remains under `plugins/module_utils/endpoints/v1/manage/`.

mtarking:
### Verification summary

Ran the integration target against live ND 4.2.1: `ok=27, changed=12, failed=0` and live ND 4.3.1: `ok=27, changed=10, failed=0, ignored=0`, controller restored exactly.

, with the controller's fabric inventory restored to its exact pre-run state.

The target previously assumed four pre-existing fabrics and never created them, so on an unprepared controller it passed vacuously — empty member lists satisfied the `gathered` and idempotency assertions. It now builds and tears down its own fabric group and member fabrics.

Two findings in this review were only caught by running against hardware, and in both cases the unit fixtures had confidently encoded an incorrect assumption about the API (see threads 1 and 2).

Gates: 4984 unit tests passed · `black --check` clean · `ansible-test sanity` exit 0.

mikewiebe (approved):

PR #253, opened 2026-04-13
Exposing Timeout and Check_Mode Properties for Rest Send

AKDRG:
Issue:
Before the first request is made, the rest send object is not initialised. As a result, we are not able to set the timeout and check mode properties. Setting of timeout is crucial for restricting retries when errors occur and setting of check mode allows us to make a genuine GET request which is otherwise blocked.

Fix:
Introduce setters in the NDModule which sets the properties when the lazy initialisation happens, at the time of invoking the request method.

### DIFF
--- a/plugins/module_utils/nd_v2.py
+++ b/plugins/module_utils/nd_v2.py
@@ -146,9 +146,31 @@ class NDModule:
         self._sender: Optional[SenderProtocol] = None
         self._response_handler: Optional[ResponseHandlerProtocol] = None
 
+        self._rest_send_timeout: Optional[int] = None
+        self._rest_send_check_mode: Optional[bool] = None
         if self.module._debug:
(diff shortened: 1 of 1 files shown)


allenrobel (approved):
Hi Akshay, this looks fine to me.  However `nd_v2.py` will be deprecated once #250 is merged.

Since modules are using `nd_v2.py` we'll, of course, keep it until everyone has transitioned to use Gaspard's framework (which #250 merges RestSend directly into).  But we'll need to look into providing a means (perhaps in the base orchestrator class) to pass `timeout` and `check_mode` to RestSend via the orchestrator.

AKDRG:
PR https://github.com/CiscoDevNet/ansible-nd/pull/250 has been merged. Should address the check mode problem. Timeout might need to be handled separately.

PR #256, opened 2026-04-13
Add AI iBGP / eBGP VXLAN Fabric Modules & Add Config Actions Mixin

### LINKED ISSUE #254: [ND 4.2+] Add AI eBGP VXLAN Fabric Type Module

```shell
A new module supporting the AI eBGP VXLAN fabric type is required.
```

manage

```shell

```

```yaml
TBD
```

### Ansible Version

```shell

```

```shell

```

```shell
4.2.1
```

### Cisco NX-OS Version

```shell

```

### Cisco ACI Version

```shell

```

### LINKED ISSUE #255: [ND 4.2+] Add AI iBGP VXLAN Fabric Type Module

```shell
A new module supporting the AI iBGP VXLAN fabric type is required.
```

manage

```shell

```

```yaml
TBD
```

### Ansible Version

```shell

```

```shell

```

```shell
4.2.1
```

### Cisco NX-OS Version

```shell

```

### Cisco ACI Version

```shell

```

mtarking:
## Related Issue(s)

Resolves #254  
Resolves #255

## Proposed Changes

This PR adds and refines support for AI VXLAN fabric types in the ND collection.

- Added AI VXLAN fabric modules:
  - nd_manage_fabric_ai_ibgp_vxlan
  - nd_manage_fabric_ai_ebgp_vxlan
- Added AI model and orchestrator wiring aligned with existing fabric module architecture.
- Added integration target and task wiring for AI iBGP and AI eBGP fabric scenarios.
- Fixed module wrapper syntax and indentation issue found during integration execution.
- Updated AI module defaults and behavior to enforce AI QoS policy:
  - aiml_qos is now always enabled for AI iBGP and AI eBGP flows.
  - aiml_qos is no longer exposed as a user-configurable module option.
- Updated module documentation to reflect enforced AI QoS behavior.
- Added and updated unit tests to validate:
  - AI models enforce aiml_qos = true.
  - AI argument specs do not expose aiml_qos.

- Added fabric config save/deploy into a composable `ConfigActionsMixin`:
  - Created `plugins/module_utils/orchestrators/config_actions_mixin.py` — reusable
    mixin leveraging `NDBaseOrchestrator._request()` for automatic check-mode,
    error handling, and Results API call tracking.
  - Supports two deploy modes: `global` (fabric-wide via `/actions/deploy`) and
    `switch` (targeted via `/switchActions/deploy` using `configSyncStatus` filtering).
  - Config save uses `/actions/configSave` (no request body, always fabric-wide).
  - Applied mixin to all 5 fabric orchestrators (iBGP, eBGP, AI iBGP, AI eBGP, External).
  - Updated all 5 fabric module `main()` functions to use orchestrator mixin methods.

## Test Notes

- Unit tests for updated AI model and argspec behavior pass.
- Integration tasks were run and now reach runtime execution for AI flows.
- Remaining integration failures are environment-dependent (inventory, ND endpoint, auth, connectivity), not module syntax or parsing regressions.
- Added unit tests for `ConfigActionsMixin` covering save, global deploy,
  switch-level deploy, skip-when-in-sync, invalid type, and batch execution.

4.2.1

manage

## Notes for Reviewers

- Local workspace-only artifacts under .ansible/collections are not part of the intended PR scope.
- Intended review scope is the cisco.nd source and tests for AI fabric module behavior and coverage updates.

## Checklist

- [x] Latest branch rebased or merged with develop
- [x] Documentation updated to reflect behavior changes
- [x] Unit coverage updated for new enforced defaults
- [x] Integration scenarios updated for AI module paths


plugins/modules/nd_manage_fabric_ai_ibgp_vxlan.py
```diff
@@ -0,0 +1,1712 @@
...
+            description:
+            - Enable MVPN VRI ID generation for Tenant Routed Multicast with IPv4 underlay.
+            type: bool
+            default: true
+          mvpn_vrf_route_import_id_range:
+            description:
+            - MVPN VRI ID range (minimum 1, maximum 65535) for vPC.
+            - Applicable when TRM is enabled with IPv6 underlay, or mvpn_vrf_route_import_id is enabled with IPv4 underlay.
+            type: str
+            default: ""
```
mikewiebe, line 269: This property specifies the default as `default: ""`. Couple of issues here

1. The pydantic model that this inherits from the default is 
    * ibgp: [`None`](https://github.com/CiscoDevNet/ansible-nd/blob/develop/plugins/module_utils/models/manage_fabric/manage_fabric_ibgp_vxlan.py#L364)
    * ebgp [`None`](https://github.com/CiscoDevNet/ansible-nd/blob/develop/plugins/module_utils/models/manage_fabric/manage_fabric_ebgp_vxlan.py#L251)
2. The ebgp version of this AI fabric does not have the default documented

Same is true for other properties

* flowlet_aging
* priority_flow_control_watch_interval
* real_time_backup
* scheduled_backup

Pydantic model for these values is also `None`

tests/integration/targets/nd_manage_fabric/vars/main.yaml
```diff
@@ -326,3 +334,278 @@ fabric_config_ebgp:
...
+    tcam_allocation: true
+    inband_management: false
+
+    # Security
+    security_group_tag: false
+    private_vlan: false
+
+    # QoS / Queuing
+    default_queuing_policy: false
+    aiml_qos: false
```
mikewiebe: The PR says AI QoS is always enabled and not user-configurable, and the AI model hides aiml_qos, which is good. But the PR’s integration vars still include aiml_qos: false. Should this be removed or changed, otherwise the tests/examples don't match the documentation?

plugins/modules/nd_manage_fabric_ai_ebgp_vxlan.py
```diff
@@ -0,0 +1,1697 @@
...
+def main():
+    argument_spec = nd_argument_spec()
+    argument_spec.update(FabricAiEbgpVxlanModel.get_argument_spec())
+
+    module = AnsibleModule(
+        argument_spec=argument_spec,
+        supports_check_mode=True,
+    )
+
+    try:
```
mikewiebe, line 1767: ### Medium: Module wrappers can mask constructor failures with an unbound local

**Issue**

Both new module wrappers assign `nd_state_machine` for the first time inside the `try` block, then reference it from the exception handlers. If `NDStateMachine(...)` fails during construction, the handler itself can raise `UnboundLocalError`, masking the original failure. The wrappers also do not call `require_pydantic(module)`, which is the newer model-backed wrapper pattern for a clear dependency error.

**Evidence**

- `plugins/modules/nd_manage_fabric_ai_ebgp_vxlan.py:1673-1693` assigns `nd_state_machine` inside `try` and later references it in both exception handlers.
- `plugins/modules/nd_manage_fabric_ai_ibgp_vxlan.py:1688-1708` has the same pattern.
- Base fabric wrappers initialize the guard first, for example `plugins/modules/nd_manage_fabric_ibgp_vxlan.py` in `develop` sets `nd_state_machine = None` before `try`.
- Newer model-backed wrappers such as `plugins/modules/nd_interface_loopback.py` call `require_pydantic(module)` before state-machine initialization and also initialize `nd_state_machine = None`.

**Existing PR overlap**

**Impact**

If model validation, orchestrator initialization, or the initial query fails inside `NDStateMachine.__init__`, the user can lose the real root cause and see a secondary local-variable failure instead. Missing `require_pydantic(module)` also makes dependency failures less consistent with newer modules.

**Suggested fix**

After `AnsibleModule(...)`, call `require_pydantic(module)`, then set `nd_state_machine = None` before entering `try`. Add a small mocked wrapper test that forces `NDStateMachine(...)` to raise and verifies the original error is returned through `fail_json`.

plugins/modules/nd_manage_fabric_ai_ebgp_vxlan.py
```diff
@@ -0,0 +1,1697 @@
...
+          syslog:
+            description:
+            - Syslog streaming configuration.
+            type: dict
+          webhooks:
+            description:
+            - Webhook configuration.
+            type: list
+            elements: dict
+  state:
```
mikewiebe, line 1356: ### Medium: Fabric save/deploy behavior is not surfaced through `config_actions`

@mtarking Let's discuss then when you are back.  For consistency need to decide if we should support this across all of the fabric modules and align with other modules that support config_actions

**Issue**

The new AI fabric modules expose create, update, replace, override, and delete, but they do not expose `config_actions.save`, `config_actions.deploy`, or `config_actions.type`. This is inherited from the current fabric-family module shape, but it leaves the new modules short of the newer module contract and short of the OpenAPI fabric lifecycle capabilities.

**Evidence**

- Dynamic argspec verification shows both AI modules expose only:

```text
state choices: merged, replaced, deleted, overridden
config: {required: false, type: list, elements: dict}
```

- `plugins/module_utils/models/manage_fabric/manage_fabric_base.py:159-167` is the inherited shallow fabric argspec and has no `config_actions`.
- `plugins/module_utils/endpoints/v1/manage/manage_fabrics.py:540-586` defines a fabric `configDeploy` endpoint class.
- OpenAPI inspection confirmed `POST /fabrics/{fabricName}/actions/configSave` and `POST /fabrics/{fabricName}/actions/configDeploy`; both accept optional `ticketId`.
- `plugins/modules/nd_manage_switches.py` in `develop` documents the current `config_actions.save`, `config_actions.deploy`, and `config_actions.type` reference pattern.

**Existing PR overlap**

**Impact**

Users can create or mutate fabric intent but have no module-level control over save/recalculate, deploy, deploy scope, or ticket-aware action behavior. That forces a manual/API follow-up for a complete workflow and makes these new modules inconsistent with newer `config_actions`-based modules.

**Suggested fix**

Add the standard `config_actions` shape and wire it to fabric save/deploy behavior in the fabric orchestrator layer. Include examples for no action, save only, save plus deploy, and ticket-aware save/deploy if change-control tickets are supported. If fabric lifecycle modules intentionally stop before save/deploy, document that explicitly in the module notes and tests.

plugins/modules/nd_manage_fabric_ai_ebgp_vxlan.py
```diff
@@ -0,0 +1,1697 @@
...
+  cisco.nd.nd_manage_fabric_ai_ebgp_vxlan:
+    state: deleted
+    config:
+      - fabric_name: fabric_east
+      - fabric_name: fabric_west
+      - fabric_name: fabric_old
+  register: result
+"""
+
+RETURN = r"""
```
mikewiebe, line 1658: ### Medium: Empty `RETURN` docs for both new modules

**Issue**

Both new module wrappers have empty `RETURN` blocks even though they return the shared state-machine output through `exit_json`.

**Evidence**

- `plugins/modules/nd_manage_fabric_ai_ebgp_vxlan.py:1653` is an empty `RETURN = r""" """` block.
- `plugins/modules/nd_manage_fabric_ai_ibgp_vxlan.py:1668-1669` is also empty.
- `plugins/modules/nd_manage_fabric_ai_ebgp_vxlan.py:1683-1684` and `plugins/modules/nd_manage_fabric_ai_ibgp_vxlan.py:1698-1699` call `module.exit_json(**nd_state_machine.output.format_with_verbosity(...))`.

**Impact**

Generated module docs do not tell users what result keys are returned, especially for `before`, `after`, `proposed`, `sent`, `changed`, and verbosity-dependent API detail. That makes it harder to write playbooks that consume or assert module output.

**Suggested fix**

Document the shared state-machine return shape used by these modules, including the normal output keys and any verbosity-dependent details. If the exact return shape is intentionally shared across state-machine modules, consider a reusable docs fragment or generated helper to avoid copy/paste drift.

**Follow-up tracking**

For the broader cleanup across fabric modules already merged into `develop`, see https://github.com/CiscoDevNet/ansible-nd/issues/358.

plugins/modules/nd_manage_fabric_ai_ebgp_vxlan.py
```diff
@@ -0,0 +1,1697 @@
...
+          bgp_asn: "65004"
+          bgp_asn_auto_allocation: false
+          site_id: "65004"
+          bgp_as_mode: multiAS
+          target_subnet_mask: 30
+          anycast_gateway_mac: "2020.0000.00dd"
+          performance_monitoring: true
+          replication_mode: multicast
+          multicast_group_subnet: "239.1.3.0/25"
+          rendezvous_point_count: 3
```
mikewiebe: ### Medium: AI eBGP example uses an invalid `rendezvous_point_count`

**Issue**

The AI eBGP `replaced` example sets `rendezvous_point_count: 3`, but both the docs and model only allow `2` or `4`.

**Evidence**

- `plugins/modules/nd_manage_fabric_ai_ebgp_vxlan.py:1533` uses `rendezvous_point_count: 3` in the `state: replaced` example.
- `plugins/modules/nd_manage_fabric_ai_ebgp_vxlan.py:326-331` documents choices `[2, 4]`.
- `plugins/module_utils/models/manage_fabric/enums.py:346-356` defines `RendezvousPointCountEnum` as only `TWO = 2` and `FOUR = 4`.
- `inspect_module_contract.py examples-check` reported one choice mismatch for `config[].management.rendezvous_point_count`.

**Impact**

The example will fail validation or teach users an unsupported value. This is especially risky because examples are commonly copied directly into playbooks.

**Suggested fix**

Change the example to `2` or `4` and keep the examples-check helper clean for both new modules.

plugins/modules/nd_manage_fabric_ai_ibgp_vxlan.py
```diff
@@ -0,0 +1,1712 @@
...
+                    description:
+                    - Primary exporter name.
+                    type: str
+                    required: true
+                  exporter2_name:
+                    description:
+                    - Secondary exporter name.
+                    type: str
+                    default: ""
+  state:
```
mikewiebe, line 1555: ### Medium: Advertised states are not fully covered by examples or tests

**Issue**

Both AI fabric modules advertise `overridden`, but the AI integration task files do not test it. AI iBGP docs also provide only `merged` and `deleted` examples, even though the module advertises `replaced` and `overridden`. The modules also omit `gathered`; that matches the current fabric family, but it remains a visible state-surface gap versus newer modules.

**Impact**

`overridden` is the most destructive state, but it lacks coverage in the new AI integration scenarios. Missing examples for advertised states also makes the module contract harder for users to understand and harder for reviewers to validate.

**Suggested fix**

Add examples and tests for every advertised state, especially `overridden`. If `gathered` is planned as part of the broader gathered-support work, document the current limitation or add follow-up tracking in the module notes.

**Follow-up tracking**

For the broader `state: gathered` support work across NDStateMachine/orchestrator-backed modules, see https://github.com/CiscoDevNet/ansible-nd/issues/340.

plugins/modules/nd_manage_fabric_ai_ibgp_vxlan.py
```diff
@@ -0,0 +1,1712 @@
...
+            - Enable BGP authentication.
+            type: bool
+            default: false
+          bgp_authentication_key_type:
+            description:
+            - "BGP key encryption type: 3 - 3DES, 6 - Cisco type 6, 7 - Cisco type 7."
+            type: str
+            default: 3des
+            choices: [ 3des, type6, type7 ]
+          bgp_authentication_key:
```
mikewiebe, line 437: ### Medium: Secret-bearing nested fields are not protected by Ansible `no_log`

**Issue**

The AI fabric docs expose BGP, BFD, OSPF, PIM, IS-IS, MACsec, and VRF Lite MACsec key fields under the shallow `config` dict. Because the fabric argspec is only `config: list[dict]`, Ansible cannot mark those nested secret fields with field-level `no_log=True`.

**Impact**

Secret values can leak through task args, diffs, debug output, or failure payloads because Ansible does not know these nested keys are sensitive. That is risky for authentication and MACsec material.

**Suggested fix**

Either deepen the fabric argspec enough to mark secret fields with `no_log=True`, or add shared model/output masking for sensitive fabric aliases before returning, logging, or diffing values. Add tests proving these fields are masked in success and failure paths.

**Follow-up tracking**

For the broader collection-wide `no_log` audit, see https://github.com/CiscoDevNet/ansible-nd/issues/359.

plugins/modules/nd_manage_fabric_ai_ibgp_vxlan.py
mtarking: Resolved.

tests/integration/targets/nd_manage_fabric/vars/main.yaml
mtarking: Updated.

plugins/modules/nd_manage_fabric_ai_ebgp_vxlan.py
mtarking: Added.

plugins/modules/nd_manage_fabric_ai_ebgp_vxlan.py
mtarking: Resolved.

plugins/modules/nd_manage_fabric_ai_ibgp_vxlan.py
mtarking: Skipped overridden as original fabric types skipped. Merged, replaced, and deleted are covered.

plugins/modules/nd_manage_fabric_ai_ibgp_vxlan.py
mtarking: argspec dynamically expanded. Will wait for open issue to resolve no_log issue.

plugins/module_utils/models/manage_fabric/manage_fabric_base.py
```diff
@@ -45,6 +50,157 @@
...
+
+        annotation = field_info.annotation
+        if annotation is None:
+            continue
+
+        # Unwrap Optional
+        inner_type, is_optional = _unwrap_optional(annotation)
+
+        # Skip single-value Literal fields (not user-configurable)
+        if _is_single_literal(inner_type):
```
mikewiebe, line 142: ### High: Examples include `category`, but the generated argspec rejects it

**Issue**

The module docs and examples tell users to provide `category: fabric`, but `category` is a single-value `Literal["fabric"]` in the base model and the generated argspec intentionally skips single-value literals. The generated nested `config.options` therefore does not include `category`, while every fabric example in this PR still includes it.

**Evidence**

- [`plugins/module_utils/models/manage_fabric/manage_fabric_base.py` line 142](https://github.com/CiscoDevNet/ansible-nd/pull/256/files#diff-3c1962b86bac13c47c3ca773114ef9c91338329018ad9360fe49ee94194e29f0R142) skips single-value literal fields while building argspec options.
- `plugins/module_utils/models/manage_fabric/manage_fabric_base.py:239` defines `category: Literal["fabric"] = Field(..., default="fabric")`.
- Runtime argspec probe against head `a456feca` returned `category_in_argspec False`.
- `plugins/modules/nd_manage_fabric_ai_ibgp_vxlan.py:1590` and `plugins/modules/nd_manage_fabric_ai_ebgp_vxlan.py:1393` include `category: fabric` in examples.
- The same mismatch appears in modified non-AI fabric modules, for example `nd_manage_fabric_ibgp_vxlan.py:1760`, `nd_manage_fabric_ebgp_vxlan.py:1462`, and `nd_manage_fabric_external.py:712`.

**Existing PR overlap**

**Impact**

Copy/paste examples can fail before reaching Pydantic or the controller because `category` is not accepted by the generated module argspec. This affects the new AI modules and the modified fabric modules in the same PR.

**Suggested fix**

Either expose `category` in the generated argspec with fixed choice/default semantics, or remove it from DOCUMENTATION and EXAMPLES as an internal discriminator. Add a lightweight example replay or argspec assertion to prevent future docs/argspec drift.

---

plugins/modules/nd_manage_fabric_ai_ebgp_vxlan.py
```diff
@@ -0,0 +1,1763 @@
...
+    default: merged
+    choices: [ merged, replaced, overridden, deleted ]
+extends_documentation_fragment:
+- cisco.nd.modules
+- cisco.nd.check_mode
+notes:
+- This module is only supported on Nexus Dashboard having version 4.2.0 or higher.
+- Only AI/ML eBGP VXLAN fabric type (C(aimlVxlanEbgp)) is supported by this module.
+- When using O(state=replaced) with only required fields, all optional management settings revert to their defaults.
+- The O(config.management.bgp_asn) field is optional when O(config.management.bgp_asn_auto_allocation) is C(true).
```
mikewiebe: ### High: AI eBGP auto-ASN examples fail local model validation

**Issue**

The AI eBGP documentation says `config.management.bgp_asn` is optional when `bgp_asn_auto_allocation` is true, and the first AI eBGP create example omits `bgp_asn`. The model still inherits `bgp_asn: str | None = Field(...)` without a default, so Pydantic treats it as required even when auto allocation and `bgp_asn_range` are supplied.

**Evidence**

- [`plugins/modules/nd_manage_fabric_ai_ebgp_vxlan.py` line 1381](https://github.com/CiscoDevNet/ansible-nd/pull/256/files#diff-2ff0b5539ac44ff331feebc2a91259f74152e253b7ac6c628807b0955b53df2cR1381) documents `bgp_asn` as optional when auto allocation is true.
- `plugins/modules/nd_manage_fabric_ai_ebgp_vxlan.py:1402` creates an AI eBGP fabric with `bgp_asn_auto_allocation: true` and `bgp_asn_range`, but no `bgp_asn`.
- `plugins/module_utils/models/manage_fabric/manage_fabric_ai_ebgp_vxlan.py:37` inherits `VxlanEbgpManagementModel`.
- `plugins/module_utils/models/manage_fabric/manage_fabric_ebgp_vxlan.py:101` defines `bgp_asn: str | None = Field(...)` without `default=None`.
- Runtime probe with the documented auto-ASN shape returned `ValidationError` before any API call.

**Existing PR overlap**

**Impact**

The primary AI eBGP create example and partial-update patterns can be locally rejected, even though the docs say they are valid. This is a user-facing contract break for the eBGP AI module.

**Suggested fix**

Make `bgp_asn` optional in the model when `bgp_asn_auto_allocation` is true, and enforce the inverse condition when auto allocation is false. Add unit tests for both auto-ASN and static-ASN eBGP create/update paths.

---

tests/integration/targets/nd_manage_fabric/tasks/fabric_ai_ibgp.yaml
```diff
@@ -0,0 +1,632 @@
...
+
+- name: Test that we have a Nexus Dashboard host, username and password
+  ansible.builtin.fail:
+    msg: 'Please define the following variables: ansible_host, ansible_user and ansible_password.'
+  when: ansible_host is not defined or ansible_user is not defined or ansible_password is not defined
+
+- name: Set vars
+  ansible.builtin.set_fact:
+    nd_info: &nd_info
+      output_level: '{{ api_key_output_level | default("debug") }}'
```
mikewiebe, line 15: ### High: Integration tasks expose ND credentials and JWT tokens

**Issue**

The new AI integration task files manually authenticate to ND using `ansible.builtin.uri`, register the response containing a JWT, and reuse the token in an `Authorization` header. Those tasks are not marked `no_log: true`, and the task defaults set `output_level` to debug.

**Evidence**

- [`tests/integration/targets/nd_manage_fabric/tasks/fabric_ai_ibgp.yaml` line 15](https://github.com/CiscoDevNet/ansible-nd/pull/256/files#diff-511f68c25b8d9850440b0e7977be080a94f6b415a6f6a8815ee5587cd4a56d8dR15) and `fabric_ai_ebgp.yaml:15` default `output_level` to `debug`.
- `tests/integration/targets/nd_manage_fabric/tasks/fabric_ai_ibgp.yaml:169` and `fabric_ai_ebgp.yaml:174` send `userPasswd: "{{ ansible_password }}"`.
- `tests/integration/targets/nd_manage_fabric/tasks/fabric_ai_ibgp.yaml:183` and `fabric_ai_ebgp.yaml:188` send `Authorization: "Bearer {{ nd_auth_response.json.jwttoken }}"`.
- `rg` found no `no_log` around those manual login/token tasks.

**Existing PR overlap**

No matching existing PR comment found. The existing `no_log` PR thread is about module fields, not the new integration auth flow.

**Impact**

Live test logs can expose the ND password or bearer token when a task fails, when output is captured in CI, or when debug output is retained. This is a concrete secret-handling issue in the PR test assets.

**Suggested fix**

Add `no_log: true` to the manual authentication tasks and the token-bearing GET tasks, or reuse an existing ND test helper that handles authentication without exposing secrets.

---

plugins/modules/nd_manage_fabric_ai_ibgp_vxlan.py
```diff
@@ -0,0 +1,1779 @@
...
+            type: str
+            default: p2p
+            choices: [ p2p, unNumbered ]
+          link_state_routing_protocol:
+            description:
+            - The underlay link-state routing protocol.
+            type: str
+            default: ospf
+            choices: [ ospf, isis ]
+          target_subnet_mask:
```
mikewiebe, line 131: ### Low: AI models allow `target_subnet_mask` values the AI OpenAPI schema rejects

**Issue**

The inherited iBGP/eBGP models allow `target_subnet_mask` values from 24 to 31, while the AI OpenAPI schema constrains `targetSubnetMask` to 30 through 31.

**Evidence**

- [`plugins/modules/nd_manage_fabric_ai_ibgp_vxlan.py` line 136](https://github.com/CiscoDevNet/ansible-nd/pull/256/files#diff-b06fa22292127a4c8c4e41f726032522a743493abcafebfc65cf8010cbb65f20R136) documents `target_subnet_mask` as valid for 24-31.
- `plugins/module_utils/models/manage_fabric/manage_fabric_ibgp_vxlan.py:173` uses `ge=24, le=31`.
- `plugins/module_utils/models/manage_fabric/manage_fabric_ebgp_vxlan.py:259` uses `ge=24, le=31`.
- Bundled OpenAPI schema at `nd-openapi-mcp/schemas/manage.json:88438` constrains `targetSubnetMask` to minimum 30 and maximum 31.

**Existing PR overlap**

**Impact**

The module can accept values that fail only at controller/API time. This is lower severity than the example-breaking issues above, but it is still a model/OpenAPI mismatch.

**Suggested fix**

Override the AI model constraint to `ge=30, le=31`, or confirm the OpenAPI schema is wrong and update the generated reference.

---

mtarking: Day1 issue and updated to match spec.

tests/integration/targets/nd_manage_fabric/tasks/fabric_ai_ibgp.yaml
mtarking: Deferring for #293

plugins/modules/nd_manage_fabric_ai_ebgp_vxlan.py
mtarking: Removed incorrect comments/notes for bgp asn as optional.

plugins/module_utils/models/manage_fabric/manage_fabric_base.py
mtarking: Cleaned up examples to remove category.

plugins/modules/nd_manage_fabric_ai_ebgp_vxlan.py
mtarking: Added a config actions mixin for orchestrator that reuses _request from base orchestrator. Support added for global config save, global config deploy, and switch level deploy (using pre-existing endpoints).

mikewiebe: This comment is related to the newly added config_actions

### High: Config-action validation runs after mutation and is skipped when idempotent

**Issue**

`execute_config_actions()` explicitly defines `deploy=true, save=false` as invalid. The wrappers nevertheless call `NDStateMachine.manage_state()` before that validation, and call `execute_config_actions()` only when the state machine sent a change. As a result, the first invalid run mutates ND and then fails, while an identical no-drift second run skips validation and succeeds.

**Evidence**

- [`nd_manage_fabric_ai_ebgp_vxlan.py` line 1760](https://github.com/CiscoDevNet/ansible-nd/pull/256/files#diff-2ff0b5539ac44ff331feebc2a91259f74152e253b7ac6c628807b0955b53df2cR1760) calls `manage_state()` before reading or executing `config_actions`; all five changed fabric wrappers repeat this ordering.
- [`nd_manage_fabric_ai_ebgp_vxlan.py` line 1769](https://github.com/CiscoDevNet/ansible-nd/pull/256/files#diff-2ff0b5539ac44ff331feebc2a91259f74152e253b7ac6c628807b0955b53df2cR1769) gates action execution on a non-empty `sent` collection.
- [`config_actions_mixin.py` line 187](https://github.com/CiscoDevNet/ansible-nd/pull/256/files#diff-4ba7d6ca64d751b8a222dd6df9adc4f079e24e40f51280aaa6a49c6ff153f420R187) contains the validation, after the wrapper has already run the state machine.
- Two live reproductions on ND `10.122.84.113` confirmed the behavior. Starting from MTU `9216`, the first request for MTU `9000` with `save=false, deploy=true` returned `failed=true, changed=true` and `config_actions: deploy=True requires save=True`; ND had already stored MTU `9000`. The identical second request returned `failed=false, changed=false`, proving the no-drift path bypassed validation.

**NOTE:** I used the following playbook to reproduce the issue and added notes inline

```yaml
---
- name: Reproduce config-actions validation ordering
  hosts: nd113
  gather_facts: false
  vars:
    test_fabric: pr256-config-actions-min

  tasks:
    - block:
        - name: Create baseline fabric with MTU 9216
          cisco.nd.nd_manage_fabric_ai_ibgp_vxlan:
            state: merged
            config:
              - fabric_name: "{{ test_fabric }}"
                management:
                  bgp_asn: "65001.55"
                  fabric_mtu: 9216

        # MWIEBE NOTE: This next task does fail with a message indicating that save and deploy cannot
        # both be true but the MTU still gets updated in the fabric settings which it should not be
        
        - name: First invalid run
          cisco.nd.nd_manage_fabric_ai_ibgp_vxlan:
            state: merged
            config:
              - fabric_name: "{{ test_fabric }}"
                management:
                  bgp_asn: "65001.55"
                  fabric_mtu: 9000
            config_actions: &invalid_actions
              save: false
              deploy: true
              type: switch
          register: first_run
          ignore_errors: true

        # MWIEBE Note: This task should also fail but it does not because the fabric is already in the
        # desired state.
   
        - name: Second identical invalid run
          cisco.nd.nd_manage_fabric_ai_ibgp_vxlan:
            state: merged
            config:
              - fabric_name: "{{ test_fabric }}"
                management:
                  bgp_asn: "65001.55"
                  fabric_mtu: 9000
            config_actions: *invalid_actions
          register: second_run
          ignore_errors: true

        - name: Verify both defects
          ansible.builtin.assert:
            that:
              - first_run is failed
              - first_run is changed
              - "'deploy=True requires save=True' in first_run.msg"
              - first_run.after[0].management.fabric_mtu | int == 9000
              - second_run is not failed
              - second_run is not changed

      always:
        - name: Delete temporary fabric
          cisco.nd.nd_manage_fabric_ai_ibgp_vxlan:
            state: deleted
            config:
              - fabric_name: "{{ test_fabric }}"
```

**Existing PR overlap**

This follows up on the config-actions implementation added in response to the existing thread. The option is now present, but its validation occurs after mutation and is skipped on no-drift runs.

**Impact**

The first run reports failure even though ND intent changed and switches were not saved or deployed. The identical retry then succeeds because there is no drift. This creates controller-versus-switch divergence and makes validation depend on prior failed execution.

**Suggested fix**

Validate `config_actions` unconditionally before creating or running `NDStateMachine`. Keep action execution conditional on actual changes if desired, but do not condition input validation on `nd_state_machine.sent`. Add wrapper-level changed and no-drift tests proving both invalid calls fail without making mutating requests.

plugins/module_utils/orchestrators/config_actions_mixin.py
```diff
@@ -0,0 +1,197 @@
...
+        ``configSyncStatus`` that is not ``inSync``, and deploys
+        to them via the switchActions/deploy endpoint.
+
+        Returns None if all switches are already in sync.
+        """
+        switch_ids = self._get_switches_needing_deploy(fabric_name)
+        if not switch_ids:
+            return None
+
+        ep = EpManageFabricsSwitchActionsDeployPost(fabric_name=fabric_name)
```
mikewiebe, line 134: ### High: Switch deploy silently accepts per-switch failures in HTTP 207 responses

**Issue**

HTTP 207 means the bulk request was accepted and contains a separate result for each switch; it does not mean every switch succeeded. The switch deploy path returns that response without checking the individual results. Because the shared response strategy treats 207 as transport success, failed switch items do not fail the Ansible task.

**Evidence**

- [`config_actions_mixin.py` lines 136-142](https://github.com/CiscoDevNet/ansible-nd/pull/256/files#diff-4ba7d6ca64d751b8a222dd6df9adc4f079e24e40f51280aaa6a49c6ff153f420R136) returns `_request()` directly after switch deploy.
- The bundled OpenAPI operation `POST /fabrics/{fabricName}/switchActions/deploy` returns HTTP 207 with `switchIds[*].switchId`, `status`, and `message`.
- [`nd_v1_strategy.py` line 79](https://github.com/CiscoDevNet/ansible-nd/blob/e806bc01af914ecf62f00b7d9d85f2e9cf0b34c4/plugins/module_utils/rest/response_strategies/nd_v1_strategy.py#L79) includes 207 in its success codes.
- A current-head probe returned normally for `{"switchIds":[{"switchId":"FAIL1","status":"failed","message":"deploy rejected"}]}`.
- [`test_config_actions_mixin.py` line 336](https://github.com/CiscoDevNet/ansible-nd/pull/256/files#diff-6ef247c42588339b0e25df90b91ba460ba602720b26b8d5650ebec8b18e65acdR336) covers only successful switch deployment; there is no mixed or failed 207 case.

For example, ND can accept one bulk deploy request and return:

```json
{
  "switchIds": [
    {
      "switchId": "LEAF1",
      "status": "success",
      "message": "Configuration deployed"
    },
    {
      "switchId": "LEAF2",
      "status": "failed",
      "message": "Switch is unreachable"
    }
  ]
}
```

The HTTP request itself completed, and `LEAF1` deployed successfully, but the operation is incomplete because `LEAF2` failed. The current code accepts the 207 response and returns normally without interpreting `LEAF2`'s status. ND reported the failure; the module silently discards its significance.

**Existing PR overlap**

**Impact**

Ansible can report a successful deployment while one or more switches remain undeployed or out of sync. Automation therefore cannot distinguish complete deployment from partial deployment and may proceed as though the entire fabric change is active.

**Suggested fix**

Do not reject every 207 response, because a 207 can contain all successful items. Inspect every switch item in the body, classify failure/error statuses case-insensitively, and raise one aggregated error containing failed switch IDs and messages when any item fails. Preserve successful-item details and add all-success, mixed-success, and all-failed unit tests. The existing ACL orchestrator's per-item 207 validation is a useful local pattern.

---

mikewiebe: This is a collection-wide HTTP 207 response-handling concern and is already tracked by [issue #295](https://github.com/CiscoDevNet/ansible-nd/issues/295). That issue covers shared `RestSend`/response-handler behavior and the other modules that inspect endpoint-specific 207 bodies. The fabric switch-deploy `switchIds[]` response shape and its all-success, mixed-success, and all-failed cases should be included in that implementation and test scope.

Since the work will be implemented and tracked through #295 rather than specifically in this PR, I recommend resolving this review thread and using #295 as the source of truth for completion.

mtarking: Resolving as conscensus is to handle in #295

plugins/modules/nd_manage_fabric_ai_ebgp_vxlan.py
mtarking: Good catch — fixed in both respects.

**1. Validation now runs before any mutation.** I extracted the check into a side-effect-free `@staticmethod ConfigActionsMixin.validate_config_actions(save, deploy, deploy_type)` and call it in each module's `main()` *before* the `NDStateMachine` is constructed. Since it's a staticmethod on the orchestrator class, no instance/state machine exists yet when it runs, so an invalid combo (`deploy=true` + `save=false`) fails fast via `module.fail_json` before ND is touched. This mirrors the existing precedent in `nd_manage_vpc_pair.py`.

**2. No longer gated on `sent`.** The validation is now independent of `len(nd_state_machine.sent) > 0`, so it fails deterministically on every run — including idempotent no-drift re-runs. The action-execution block remains gated on `sent`, which is correct since there's nothing to save/deploy when nothing changed.

While here, I also corrected the `config_actions.type` default from `global` to `switch`.

Applied across all 5 fabric modules (`ibgp`, `ebgp`, `ai_ibgp`, `ai_ebgp`, `external`) plus the mixin.

**Tests:**
- 6 new unit tests in `test_config_actions_mixin.py` (deploy-without-save, invalid type, valid combos, class-level callable without an instance). Full orchestrator suite: 411 passed.
- New "TEST 4: CONFIG ACTIONS VALIDATION" integration block in each fabric target that asserts the invalid config fails on both the first and idempotent second run, and that the target field is never mutated.

plugins/module_utils/orchestrators/config_actions_mixin.py
```diff
@@ -0,0 +1,222 @@
...
+            not_found_ok=True,
+            operation_type=OperationType.QUERY,
+        )
+
+        switches = result.get("switches", []) if result else []
+        switch_ids = []
+        for switch in switches:
+            additional_data = switch.get("additionalData", {})
+            config_status = additional_data.get("configSyncStatus", "")
+            if config_status and config_status != "inSync":
```
allenrobel: Bug: this guard silently drops switches whose `configSyncStatus` is empty or missing. `additional_data.get("configSyncStatus", "")` defaults to `""`, which is falsy, so `if config_status and config_status != "inSync":` short-circuits and the switch is excluded from `switch_ids` — even though an empty/unknown status is "not inSync". This contradicts the method's own docstring: *"A switch needs deployment if its `configSyncStatus` is not `inSync`."* A newly-registered switch whose status is not yet populated would be silently skipped by `switchActions/deploy`, and the task would still report success with no signal that a switch was missed. Suggest dropping the truthiness pre-check — `if config_status != "inSync":` — matching the conservative polarity used elsewhere (`nd_switch_resources.py` treats missing/unknown as needing deploy). The new unit tests only exercise `inSync` / `outOfSync` / `pending`, so the empty/missing case isn't covered.

allenrobel (commented):
### Code review

Reviewed the new AI iBGP/eBGP VXLAN fabric modules and `ConfigActionsMixin`. Findings inline: one switch-deploy correctness bug, a wrong-default concern on `flowlet_aging` / `priority_flow_control_watch_interval` (follow-up to the resolved defaults thread), and two PEP 585 typing-convention items. Existing unresolved threads were checked to avoid duplication.

plugins/module_utils/orchestrators/config_actions_mixin.py
```diff
@@ -0,0 +1,222 @@
...
+
+    orchestrator.execute_config_actions(
+        fabric_names=["fab1", "fab2"],
+        save=True, deploy=True, deploy_type="switch",
+    )
+"""
+
+from __future__ import annotations
+
+from typing import List
```
allenrobel: CLAUDE.md: *"Modern type annotations (PEP 585 / PEP 604) for all new code ... `list[X]` ... instead of `List` ... Drop the now-unused ... `List` ... imports from typing."* This new file already has `from __future__ import annotations`, so `list[str]` works directly — replace `List[str]` (the `_get_switches_needing_deploy` return type and the `execute_config_actions` `fabric_names` param) and drop this import.

plugins/module_utils/models/manage_fabric/manage_fabric_base.py
```diff
@@ -22,11 +22,16 @@
 
 from __future__ import annotations
 
-from typing import ClassVar, Literal
+import enum
+import types
+import typing
+from typing import Any, ClassVar, Dict, Literal, Set, get_args, get_origin
```
allenrobel: CLAUDE.md: *"`dict[K, V]` ... `set[X]` ... instead of ... `Dict` ... `Set`."* The new `_build_options_from_model` helper mixes modern and legacy syntax in one signature — `exclude_fields: Set[str] | None = None) -> Dict[str, Any]`. Prefer `set[str] | None` / `dict[str, Any]` and drop `Dict` / `Set` from this import.

plugins/module_utils/models/manage_fabric/manage_fabric_ebgp_vxlan.py
```diff
@@ -651,7 +653,7 @@ class VxlanEbgpManagementModel(NDNestedModel):
             "Flowlet aging timer in microseconds. Valid range depends on platform: "
             "Cloud Scale (CS)=1-2000000 (default 500), Silicon One (S1)=1-1024 (default 256)"
         ),
-        default=None,
+        default=1,
```
allenrobel: Follow-up to the earlier (now-resolved) defaults discussion: changing `flowlet_aging` from `default=None` to `default=1` sends a value that contradicts this field's own documented ND defaults — *"Cloud Scale (CS)=1-2000000 (default 500), Silicon One (S1)=1-1024 (default 256)"*. Because `NDBaseModel.to_payload()` uses `model_dump(exclude_none=True)` (not `exclude_unset`), a concrete default is no longer omitted, so every create where the user leaves this unset now sends `flowletAging: 1` to ND, overriding ND's platform default (500/256). Same change in the iBGP model (`manage_fabric_ibgp_vxlan.py` line 717). Consider keeping `None` (let ND apply its default) or documenting why `1` is intended.

plugins/module_utils/models/manage_fabric/manage_fabric_ebgp_vxlan.py
```diff
@@ -677,7 +679,7 @@ class VxlanEbgpManagementModel(NDNestedModel):
     priority_flow_control_watch_interval: int | None = Field(
         alias="priorityFlowControlWatchInterval",
         description=("Acceptable values from 101 to 1000 (milliseconds). Leave blank for system default (100ms)."),
-        default=None,
+        default=101,
```
allenrobel: Same concern as `flowlet_aging`: `priority_flow_control_watch_interval` changed from `default=None` to `default=101`, but the docstring says *"Acceptable values from 101 to 1000 (milliseconds). Leave blank for system default (100ms)."* With `to_payload()`'s `exclude_none=True`, this now sends `priorityFlowControlWatchInterval: 101` on every create where the user omits it, overriding ND's 100ms system default. Mirror site: `manage_fabric_ibgp_vxlan.py` line 741. Keeping `None` preserves the documented "leave blank for system default" behavior.

allenrobel (commented):
### Code review (follow-up)

One additional convention item on the AI fabric management models.

plugins/module_utils/models/manage_fabric/manage_fabric_ai_ebgp_vxlan.py
```diff
@@ -0,0 +1,90 @@
...
+    ## Raises
+
+    - `ValueError` - If BGP ASN, VLAN ranges, or IP ranges are invalid
+    - `TypeError` - If required string fields are not provided
+    """
+
+    _argspec_exclude_fields: ClassVar[set[str]] = {"name", "aiml_qos"}
+
+    type: Literal["aimlVxlanEbgp"] = Field(description="Type of the fabric", default="aimlVxlanEbgp")
+    aiml_qos: bool = Field(
```
allenrobel: `aiml_qos` is intended to be fixed/non-configurable ("Always enabled for AI eBGP VXLAN fabrics"), but that intent is enforced only by excluding it from the argspec (line 52). The field itself is a plain `bool = Field(default=True)`, so the "always enabled" guarantee holds only for the module-input path — `from_response()` (parsing a wire payload where ND could echo `aimlQos: false`) and `merge()` can still put a non-`True` value into the model. The `type` field one line up (line 54) already uses the lock-down idiom for this: `Literal["aimlVxlanEbgp"]`. Suggest the same here — `aiml_qos: Literal[True] = Field(alias="aimlQos", ..., default=True, frozen=True)` — which rejects any non-`True` value on all construction paths. This mirrors a prior request in #260 on the `evpn` field (`frozen=True` + `Literal[True]`). Same applies to the iBGP sibling (`manage_fabric_ai_ibgp_vxlan.py` line 55). Not urgent — deferring to @mtarking on whether to address.

plugins/module_utils/models/manage_fabric/manage_fabric_ai_ibgp_vxlan.py
```diff
@@ -0,0 +1,81 @@
...
+    ## Raises
+
+    - `ValueError` - If BGP ASN, VLAN ranges, or IP ranges are invalid
+    - `TypeError` - If required string fields are not provided
+    """
+
+    _argspec_exclude_fields: ClassVar[set[str]] = {"name", "aiml_qos"}
+
+    type: Literal["aimlVxlanIbgp"] = Field(description="Type of the fabric", default="aimlVxlanIbgp")
+    aiml_qos: bool = Field(
```
allenrobel: Same as the eBGP sibling: `aiml_qos` is documented as always-on but declared as a plain `bool = Field(default=True)`, so only the argspec path (exclusion on line 52) enforces it — `from_response()` / `merge()` can still set a non-`True` value. The adjacent `type` field (line 54) already uses `Literal["aimlVxlanIbgp"]`; suggest `aiml_qos: Literal[True] = Field(alias="aimlQos", ..., default=True, frozen=True)` for the same guarantee on all construction paths. Echoes the #260 `evpn` request. Deferring to @mtarking on whether to address.

plugins/module_utils/orchestrators/config_actions_mixin.py
mtarking: Good catch. Changed the guard from if config_status and config_status != "inSync": to if config_status != "inSync":. An empty or missing configSyncStatus was silently skipping the switch, which could leave it undeployed. Now any non-inSync value (including empty/missing) is treated as needing deployment. Added test_deploy_switch_includes_empty_or_missing_status to cover the empty-string and missing-key cases.

plugins/module_utils/orchestrators/config_actions_mixin.py
mtarking: Done. Dropped the List import and switched the annotations to the built-in list[str] (the module already has [from __future__ import annotations]).

plugins/module_utils/models/manage_fabric/manage_fabric_base.py
mtarking: Done. Removed the Dict/Set imports and switched to set[str] / dict[str, Any] for consistency with PEP 585.

plugins/module_utils/models/manage_fabric/manage_fabric_ebgp_vxlan.py
mtarking: You're right — because to_payload() uses model_dump(exclude_none=True), a concrete default was always sent and overrode the ND platform defaults (500/256 and 100ms). Reverted both to default=None in the eBGP and iBGP models so ND applies its own defaults when the user doesn't set them. Confirmed the module docs don't advertise a default: for these fields, so no doc change was needed.

plugins/module_utils/models/manage_fabric/manage_fabric_ebgp_vxlan.py
mtarking: You're right — because to_payload() uses model_dump(exclude_none=True), a concrete default was always sent and overrode the ND platform defaults (500/256 and 100ms). Reverted both to default=None in the eBGP and iBGP models so ND applies its own defaults when the user doesn't set them. Confirmed the module docs don't advertise a default: for these fields, so no doc change was needed.

plugins/module_utils/models/manage_fabric/manage_fabric_ai_ebgp_vxlan.py
mtarking: Agreed — enforcing it only via argspec exclusion left from_response/merge paths able to set False. Changed it to [Literal[True] = Field(..., default=True, frozen=True)](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) in both AI models, matching the immutability idiom we already use for the [type](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) field. Now a non-True value raises [ValidationError](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) on every construction path. Added [test_manage_fabric_ai_vxlan_00040](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) covering direct construction and wire payload (aimlQos: false).

plugins/module_utils/models/manage_fabric/manage_fabric_ai_ibgp_vxlan.py
mtarking: Agreed — enforcing it only via argspec exclusion left from_response/merge paths able to set False. Changed it to [Literal[True] = Field(..., default=True, frozen=True)](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) in both AI models, matching the immutability idiom we already use for the [type](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) field. Now a non-True value raises [ValidationError](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) on every construction path. Added [test_manage_fabric_ai_vxlan_00040](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) covering direct construction and wire payload (aimlQos: false).

mikewiebe (approved):
LGTM

allenrobel (approved):
LGTM after comments were resolved.

PR #260, opened 2026-04-14
Update iBGP, eBGP, and External Fabric Type Modules (NDA-24 NDA-23)

### LINKED ISSUE #259: Update iBGP, eBGP, and External Fabric Modules Defaults (NDA-24)
### Ansible Version

```shell
2.19
```

```shell
develop
```

### Ansible Collection Module

```shell

```

```shell
4.2.1
```

### Cisco NX-OS Version

```shell

```

### Cisco ACI Version

```shell

```

### Expected Behavior

```shell
iBGP, eBGP, and External fabrics should create fabrics with default settings per the controller.
```

### Actual Behavior

```shell
iBGP, eBGP, and External fabrics are creating telemetry settings differently between themselves as well as the defaults from the controller.
```

### Ansible Playbook

```yaml
- name: "Merge Fabric iBGP configuration"
      cisco.nd.nd_manage_fabric_ibgp:
        state: merged
        config:
          - fabric_name: "{{ inventory_hostname }}"
            management:
              bgp_asn: "65000"
```

### Steps to Reproduce

```shell
N/A
```

### Relevant Debug Output

```shell

```

### LINKED ISSUE #263: Update iBGP, eBGP, and External Fabric Modules to Use FabricBaseModel (NDA-23)

```shell
Create an abstract `class FabricBaseModel(NDBaseModel)` that includes fabric name, location, license, telemetry, etc.

Update existing modules, e.g.

`class FabricIbgpModel(FabricBaseModel):`
```

manage

```shell

```

```yaml
N/A
```

### Ansible Version

```shell

```

```shell

```

```shell
4.2.1
```

### Cisco NX-OS Version

```shell

```

### Cisco ACI Version

```shell

```

mtarking:
## Related Issue(s)

Resolves #259
Resolves #263

### Shared Base Model Extraction

- Created `FabricBaseModel` in `manage_fabric_base.py` — abstract base class for all fabric models
- Extracted 13 shared fields (`category`, `fabric_name`, `location`, `license_tier`, `alert_suspend`, `telemetry_collection`, `telemetry_collection_type`, `telemetry_streaming_protocol`, `telemetry_source_interface`, `telemetry_source_vrf`, `security_domain`, `telemetry_settings`, `external_streaming_settings`) into the base class
- Refactored `FabricEbgpModel`, `FabricIbgpModel`, and `FabricExternalConnectivityModel` to inherit from `FabricBaseModel`
- Base class includes `_fabric_type` ClassVar, `_post_validate_consistency()` hook, `validate_fabric_consistency()`, and `get_argument_spec()`

### Module and File Renames

- Renamed eBGP and iBGP model, orchestrator, and module files to include `_vxlan` suffix for clarity:
  - `manage_fabric_ebgp.py` → `manage_fabric_ebgp_vxlan.py`
  - `manage_fabric_ibgp.py` → `manage_fabric_ibgp_vxlan.py`
  - `nd_manage_fabric_ebgp.py` → `nd_manage_fabric_ebgp_vxlan.py`
  - `nd_manage_fabric_ibgp.py` → `nd_manage_fabric_ibgp_vxlan.py`
- Updated integration tests (`fabric_ebgp.yaml`, `fabric_ibgp.yaml`) to reference the new module names

### Default and Type Alignment

- Aligned defaults across all three fabric models to use consistent values from `FabricBaseModel`:
  - `license_tier` → `LicenseTierEnum.ESSENTIALS`
  - `telemetry_collection` → `True`
  - `telemetry_collection_type` → `TelemetryCollectionTypeEnum.IN_BAND`
  - `telemetry_source_interface` → `"loopback0"`
  - `telemetry_source_vrf` → `"default"`
- Changed `telemetry_collection_type` and `telemetry_streaming_protocol` from `str` to their proper enum types
- Removed duplicate field block in `manage_fabric_external.py` where `license_tier` through `security_domain` were defined twice with conflicting values

### Module Documentation Fixes

- Updated DOCUMENTATION defaults in all three module files to match the pydantic model defaults:
  - `license_tier`: `premier` → `essentials`
  - `telemetry_collection`: `false` → `true`
  - `telemetry_collection_type`: `outOfBand` → `inBand`
  - `telemetry_source_interface`: `""` → `loopback0`
  - `telemetry_source_vrf`: `""` → `default`
  - `nxapi` (eBGP + iBGP): `false` → `true`
- Removed `evpn` from eBGP DOCUMENTATION (internal field, not user-configurable)

### Bug Fixes

• Added missing `enable_dpu_pinning` field to iBGP model and module documentation (already present in eBGP model)

### Modern Type Annotations

- Replaced `from __future__ import absolute_import, division, print_function` with `from __future__ import annotations` across all fabric model files
- Removed `__metaclass__ = type` (Python 2 artifact) from all fabric model files
- Modernized type hints throughout:
  - `Optional[X]` → `X | None`
  - `List[X]` → `list[X]`
  - `Dict[K, V]` → `dict[K, V]`
  - `ClassVar[Optional[List[str]]]` → `ClassVar[list[str] | None]`
- Removed unused `typing` imports (`Dict`, `List`, `Optional`)

### Shared Types File

- Created `plugins/module_utils/models/types.py` with reusable `NdFabricName` annotated type
- `NdFabricName` uses `Annotated` with `BeforeValidator` for fabric name validation (letters, digits, underscores, hyphens; 1-64 chars)
- Replaced inline `validate_fabric_name` field validator in `FabricBaseModel` with the shared `NdFabricName` type

### eBGP Model Cleanup

- Removed `to_diff_dict()` override that excluded `nxapiHttp` — ND now correctly returns the configured value, so the workaround is no longer needed
- Removed unused `Dict` and `Any` imports from eBGP model

### Code Organization

- Alphabetized enum and `manage_fabric_common` imports in all three fabric model files
- Fixed module docstrings to accurately list only the classes defined in each file
- Removed `__all__` exports from all non-`__init__.py` model files (per PEP8, wildcard imports are discouraged, making `__all__` unnecessary overhead)
- Alphabetized class definitions in `manage_fabric_common.py`

## Test Notes

- All 368 unit tests pass
- 29 fabric endpoint tests pass
- Import verification confirms all 3 models correctly inherit `FabricBaseModel` with proper `_fabric_type` values (`vxlanEbgp`, `vxlanIbgp`, `externalConnectivity`)
- Verified base fields are consistent across all models
- No behavioral changes to API payloads

4.2.1

* [ ] analyze
* [ ] infa
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist


plugins/module_utils/models/manage_fabric/manage_fabric_base.py
```diff
@@ -0,0 +1,177 @@
...
+Subclasses only need to define:
+
+- `_fabric_type` — the `FabricTypeEnum` value for this fabric kind
+- `management` field — typed to the specific management model
+
+Optionally override `_post_validate_consistency()` for type-specific logic
+(e.g. site_id propagation from BGP ASN).
+"""
+
+from __future__ import absolute_import, division, print_function
```
allenrobel: Comment applies throughout...

`from __future__` can be removed.

Though, it would be good to import annotations to leverage modern type annotations since the legacy annotations will eventually be deprecated.  So:

```python
```

Then remove:

```python
from typing import Dict, List, Optional
```

Then replace (everywhere):

- `Dict` -> `dict`
- `List` -> `list`
- `Set` -> `set`
- `Optional` -> `str | None`  (for example)
- `Union` -> `str | int`  (for example)

plugins/module_utils/models/manage_fabric/manage_fabric_base.py
```diff
@@ -0,0 +1,177 @@
...
+
+__metaclass__ = type
```
allenrobel: `_metaclass__ = type` can be removed.

Comment applies throughout.

plugins/module_utils/models/manage_fabric/manage_fabric_base.py
```diff
@@ -0,0 +1,177 @@
...
+    ## Raises
+
+    - `ValueError` - If required fields are missing or invalid
+    - `TypeError` - If field types don't match expected types
+    """
+
+    model_config = ConfigDict(str_strip_whitespace=True, validate_assignment=True, populate_by_name=True, extra="allow")
+
+    # ── ClassVars (shared across all fabric models) ──
+    identifiers: ClassVar[Optional[List[str]]] = ["fabric_name"]
```
allenrobel: Example of modern type annotation here would be:

```python
    identifiers: ClassVar[list[str] | None] = ["fabric_name"]
```

plugins/module_utils/models/manage_fabric/manage_fabric_base.py
```diff
@@ -0,0 +1,177 @@
...
+    identifier_strategy: ClassVar[Optional[Literal["single", "composite", "hierarchical", "singleton"]]] = "single"
+
+    # Subclass must set this to the appropriate FabricTypeEnum member
+    _fabric_type: ClassVar[FabricTypeEnum]
+
+    # ── Basic Fabric Properties ──
+    category: Literal["fabric"] = Field(description="Resource category", default="fabric")
+    fabric_name: str = Field(alias="name", description="Fabric name", min_length=1, max_length=64)
```
allenrobel: Might be good to consolidate `fabric_name` and other common fields into a shared file.

Something along the lines of:

1. Create shared model types file (rename to taste e.g. shared_fields.py, etc)

```python
# plugins/module_utils/models/types.py

import re
from typing import Annotated

from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import BeforeValidator, Field

_FABRIC_NAME_RE = re.compile(r"^[a-zA-Z0-9_-]+$")

def _validate_fabric_name(value: str) -> str:
    """
    # Summary

    Validate that a fabric name contains only letters, digits, underscores, and hyphens.

    ## Raises

    ### ValueError

    - If `value` contains characters outside `[a-zA-Z0-9_-]`.
    """
    if not _FABRIC_NAME_RE.match(value):
        raise ValueError(
            f"Fabric name can only contain letters, numbers, underscores, and hyphens, got: {value}"
        )
    return value

# Using NdFabricName here, but probably need a naming convention for shared fields?
NdFabricName = Annotated[
    str,
    Field(min_length=1, max_length=64, description="Fabric name"),
    BeforeValidator(_validate_fabric_name),
]
```

Then, in the model:

```python
from ansible_collections.cisco.nd.plugins.module_utils.models.types import NdFabricName

class FabricBaseModel(NDBaseModel):
# etc...

fabric_name: NdFabricName
# etc...
```

plugins/module_utils/models/manage_fabric/manage_fabric_common.py
```diff
@@ -298,42 +319,21 @@ class TelemetrySettingsModel(NDNestedModel):
...
-
-    email: List[Dict[str, Any]] = Field(description="Email streaming configuration", default_factory=list)
-    message_bus: List[Dict[str, Any]] = Field(alias="messageBus", description="Message bus configuration", default_factory=list)
-    syslog: Dict[str, Any] = Field(
-        description="Syslog streaming configuration", default_factory=lambda: {"collectionSettings": {"anomalies": []}, "facility": "", "servers": []}
-    )
-    webhooks: List[Dict[str, Any]] = Field(description="Webhook configuration", default_factory=list)
-
-
 # Export all models for external use
```
allenrobel: AFAIK, `__all__` is not needed except in `__init__.py` files (and Ansible sanity tests flag non-empty `__init__.py` files as an error, though Python allows them for package definitions).

`__all__` is used mainly to limit what would be imported with:

```python
from X import *
```

But wildcard imports are considered bad practice per PEP8 anyway, so removing this would not hurt anything and would be lower maintenance.

I guess some IDEs use it, but perhaps not worth the trouble of maintaining?

allenrobel (commented):
Leaving some comments.  Will defer to you though, so feel free to ignore and resolve them as appropriate.

plugins/module_utils/models/manage_fabric/manage_fabric_base.py
mtarking: Implemented and updated PR description.

plugins/module_utils/models/manage_fabric/manage_fabric_base.py
mtarking: I think I implemented per your recommendation. Please verify. Updated PR description.

plugins/module_utils/models/manage_fabric/manage_fabric_common.py
mtarking: Removed.

plugins/module_utils/models/manage_fabric/manage_fabric_base.py
mtarking: Removed.

plugins/module_utils/models/manage_fabric/manage_fabric_base.py
mtarking: Implemented.

plugins/module_utils/models/types.py
```diff
@@ -113,3 +115,19 @@ def validate_ipv6_cidr(value: str | None) -> str | None:
 
 IPv6CIDR = Annotated[Optional[str], BeforeValidator(validate_ipv6_cidr)]
 """IPv6 interface address in CIDR notation (`str | None`). Layer with `Field(...)` for alias/description as usual."""
+
+_FABRIC_NAME_RE = re.compile(r"^[a-zA-Z0-9_-]+$")
```
allenrobel: The OpenAPI spec is silent on fabric-name constraints (both `FabricName`/`fabricName` schemas and the `name` field in `baseFabric` are bare unconstrained strings — verified by dumping the resolved `POST /api/v1/manage/fabrics` schema and grepping for `pattern`/`maxLength`/`minLength`: zero matches). So the regex is a client-side convention. Live GUI probes on ND 4.2.x show the real rule is stricter than `^[a-zA-Z0-9_-]+$`:

| Input | Client regex | ND GUI |
|---|---|---|
| `_fabric` | accepts | rejects |
| `fabric_` | accepts | rejects |
| `-fabric` | accepts | rejects |
| `fabric-` | accepts | rejects |
| `12345` | accepts | rejects |
| `9foo` | accepts | accepts |
| `foo9` | accepts | accepts |

GUI error text: *"Invalid fabric name. Only alphanumeric(a-z, A-Z, 0-9), underscore(_) and hyphen(-) characters are allowed. Fabric name must start and end with an alphanumeric character, numbers alone are not allowed and max length should be 64."*

Client-side too-lax means Ansible accepts names that ND will 400 on, with a backend error harder to attribute to the input than a clean local ValueError.

**Suggested fix (both bugs in one edit):**

```python
# Fabric name rules (verified against ND 4.2.x GUI):
#   - Allowed chars: a-z, A-Z, 0-9, _, -
#   - Must start AND end with an alphanumeric (no leading/trailing _ or -)
#   - Numbers alone are not allowed (must contain at least one letter)
#   - Length 1-64 (enforced by the regex itself)
_FABRIC_NAME_RE = re.compile(
    r"^(?=.*[A-Za-z])"            # at least one letter (rejects all-digit)
    r"[A-Za-z0-9]"                 # start: alphanumeric
    r"(?:[A-Za-z0-9_-]{0,62}"      # middle: up to 62 inner chars
    r"[A-Za-z0-9])?$"              # end: alphanumeric (skipped if length is 1)
)


def _validate_fabric_name(value: str) -> str:
    if not _FABRIC_NAME_RE.match(value):
        raise ValueError(
            f"Invalid fabric name {value!r}. Allowed chars: a-z, A-Z, 0-9, _, -. "
            "Must start and end with an alphanumeric character, must contain at "
            "least one letter, and must be 1-64 characters."
        )
    return value


NdFabricName = Annotated[
    str,
    Field(alias="name", min_length=1, max_length=64, description="Fabric name"),
    AfterValidator(_validate_fabric_name),   # was BeforeValidator
]
```

The `min_length=1, max_length=64` on the Field become redundant once the regex enforces length itself — keep it for the cleaner Pydantic error on out-of-bounds input, or drop it; doesn't matter.  Personally, I'd keep it.

plugins/module_utils/models/manage_fabric/manage_fabric_ebgp_vxlan.py
```diff
@@ -135,19 +124,23 @@ class VxlanEbgpManagementModel(NDNestedModel):
     assign_ipv4_to_loopback0: bool = Field(
         alias="assignIpv4ToLoopback0",
         description=(
-            "In an IPv6 routed fabric or VXLAN EVPN fabric with IPv6 underlay, assign IPv4 address " "used for BGP Router ID to the routing loopback interface"
+            "In an IPv6 routed fabric or VXLAN EVPN fabric with IPv6 underlay, assign IPv4 address used for BGP Router ID to the routing loopback interface"
         ),
         default=True,
     )
     evpn: bool = Field(description=("Enable BGP EVPN as the control plane and VXLAN as the data plane for this fabric"), default=True)
```
allenrobel: evpn was removed as a user-configurable option in the main module's `DOCUMENTATION` header.  But a user could still set it to False since the model does not forbid it.

If the intent is to make this field unchangeable, suggest adding `frozen=True` to the Field args, and `Literal[True]` as the type annotation:

```python
evpn: Literal[True] = Field(default=True, frozen=True, description=("Enable BGP EVPN as the control plane and VXLAN as the data plane for this fabric"))
```

plugins/module_utils/models/manage_fabric/manage_fabric_base.py
```diff
@@ -0,0 +1,158 @@
...
+            self.telemetry_settings = TelemetrySettingsModel()
+
+        # Subclass hook for additional validation
+        self._post_validate_consistency()
+
+        return self
+
+    def _post_validate_consistency(self) -> None:
+        """Hook for subclass-specific post-validation logic. Default is a no-op."""
+        pass
```
allenrobel, line 157: The two current subclasses (`FabricEbgpModel`, `FabricIbgpModel`) each define their own version and never call `super()._post_validate_consistency()`. That's harmless today because the base body is `pass`, but the moment a new shared post-validation step is added to the base, every existing override silently swallows it. Worth either:

- documenting the contract ("subclasses must call `super()._post_validate_consistency()` first/last"); or
- collapsing the hook into the base validator with a small dispatch table keyed off `_fabric_type` — which is what the discriminator was added for.

plugins/module_utils/models/manage_fabric/manage_fabric_base.py
```diff
@@ -0,0 +1,158 @@
...
+    LocationModel,
+    TelemetrySettingsModel,
+)
+
+
+class FabricBaseModel(NDBaseModel):
+    """
+    # Summary
+
+    Abstract base for all fabric models (eBGP, iBGP, External, Campus, etc.).
```
allenrobel, line 52: The docstring says "Abstract base for all fabric models" and `_fabric_type` (line 75) is declared as a `ClassVar` without a value, but nothing prevents a developer from instantiating `FabricBaseModel` directly — they just get an `AttributeError` deep inside the model_validator when it tries to read `self._fabric_type`. Same for forgetting to redeclare `management` in a subclass. Either:

- add an `__init_subclass__` check that subclasses define both `_fabric_type` and a `management` field, raising a clear error at class-definition time; or
- raise from `__init__` with `if type(self) is FabricBaseModel: raise TypeError(...)`.

Not a bug today, but the abstraction is load-bearing for future fabric types.

plugins/module_utils/models/manage_fabric/manage_fabric_base.py
```diff
@@ -0,0 +1,158 @@
...
+    # Subclass must set this to the appropriate FabricTypeEnum member
+    _fabric_type: ClassVar[FabricTypeEnum]
+
+    # ── Basic Fabric Properties ──
+    category: Literal["fabric"] = Field(description="Resource category", default="fabric")
+    fabric_name: NdFabricName
+    location: LocationModel | None = Field(description="Geographic location of the fabric", default=None)
+
+    # ── License, Telemetry, and Operations ──
+    license_tier: LicenseTierEnum = Field(alias="licenseTier", description="License Tier for ffabric.", default=LicenseTierEnum.ESSENTIALS)
```
allenrobel: Typo ffabric -> fabric

allenrobel (changes requested):
Left a five comments.  Thanks!

tests/integration/targets/nd_manage_fabric/tasks/fabric_ebgp.yaml
samiib, line 17: Since these modules are only applicable on ND 4.2 we may need to skip the tests unless ND version is at least 4.2.

We do this in most of our collections so that tests can run across multiple versions. For example the nd_backup tests have a version check inside a block. See https://github.com/CiscoDevNet/ansible-nd/blob/9558784766c9bac49a6c1bccadb093071a91dd3a/tests/integration/targets/nd_backup/tasks/main.yml#L24-L26

samiib (changes requested):

plugins/module_utils/models/manage_fabric/manage_fabric_base.py
mtarking: Fixed typo.

plugins/module_utils/models/manage_fabric/manage_fabric_base.py
mtarking: Went with the ```__init_subclass__``` option. Please review again.

plugins/module_utils/models/manage_fabric/manage_fabric_base.py
mtarking: Went with option 1. Please review again.

plugins/module_utils/models/manage_fabric/manage_fabric_ebgp_vxlan.py
mtarking: Good catch. Updated.

plugins/module_utils/models/types.py
mtarking: Went with suggested fix. Thanks!

tests/integration/targets/nd_manage_fabric/tasks/fabric_ebgp.yaml
mtarking: Comment added to issue https://github.com/CiscoDevNet/ansible-nd/issues/293

plugins/module_utils/models/manage_fabric/manage_fabric_ebgp_vxlan.py
```diff
@@ -135,19 +124,23 @@ class VxlanEbgpManagementModel(NDNestedModel):
...
+    border_bgp_as: str | None = Field(
+        alias="borderBgpAs", description="BGP Autonomous System Number for Borders 1-4294967295 | 1-65535[.0-65535]", default=None
+    )
+    super_spine_bgp_as: str | None = Field(
+        alias="superSpineBgpAs", description="BGP Autonomous System Number for Super Spines 1-4294967295 | 1-65535[.0-65535]", default=None
+    )
 
     # Propagated from FabricEbgpModel
-    name: Optional[str] = Field(description="Fabric name", min_length=1, max_length=64, default="")
+    name: str | None = Field(description="Fabric name", min_length=1, max_length=64, default="")
```
akinross: How can the default be "" when the min_length=1?

plugins/module_utils/models/manage_fabric/manage_fabric_ebgp_vxlan.py
```diff
@@ -686,9 +674,9 @@ class VxlanEbgpManagementModel(NDNestedModel):
     ai_load_sharing: bool = Field(
         alias="aiLoadSharing", description=("Enable IP load sharing using source and destination address for AI workloads"), default=False
     )
-    priority_flow_control_watch_interval: Optional[int] = Field(
+    priority_flow_control_watch_interval: int | None = Field(
```
akinross, line 677: default is different as ibgp is this intended?

plugins/module_utils/models/manage_fabric/manage_fabric_ebgp_vxlan.py
```diff
@@ -657,7 +645,7 @@ class VxlanEbgpManagementModel(NDNestedModel):
     dlb_mixed_mode_default: DlbMixedModeDefaultEnum = Field(
         alias="dlbMixedModeDefault", description="Default load balancing mode for policy driven mixed mode DLB", default=DlbMixedModeDefaultEnum.ECMP
     )
-    flowlet_aging: Optional[int] = Field(
+    flowlet_aging: int | None = Field(
```
akinross, line 648: default is different as ibgp is this intended?

plugins/module_utils/models/manage_fabric/manage_fabric_ebgp_vxlan.py
```diff
@@ -249,7 +242,7 @@ class VxlanEbgpManagementModel(NDNestedModel):
     mvpn_vrf_route_import_id: bool = Field(
         alias="mvpnVrfRouteImportId", description="Enable MVPN VRI ID Generation For Tenant Routed Multicast With IPv4 Underlay", default=True
     )
-    mvpn_vrf_route_import_id_range: Optional[str] = Field(
+    mvpn_vrf_route_import_id_range: str | None = Field(
```
akinross, line 245: default is different as ibgp is this intended?

plugins/module_utils/models/manage_fabric/manage_fabric_ebgp_vxlan.py
```diff
@@ -407,10 +400,10 @@ class VxlanEbgpManagementModel(NDNestedModel):
     netflow_settings: NetflowSettingsModel = Field(alias="netflowSettings", description="Netflow configuration", default_factory=NetflowSettingsModel)
 
     # Backup / Restore
-    real_time_backup: Optional[bool] = Field(
+    real_time_backup: bool | None = Field(
```
akinross, line 403: default is different as ibgp is this intended?

plugins/module_utils/models/manage_fabric/manage_fabric_ebgp_vxlan.py
```diff
@@ -458,35 +451,35 @@ class VxlanEbgpManagementModel(NDNestedModel):
...
     )
-    ntp_server_vrf_collection: List[str] = Field(
+    ntp_server_vrf_collection: list[str] = Field(
         default_factory=lambda: ["string"],
         alias="ntpServerVrfCollection",
         description=("NTP Server VRFs. One VRF for all NTP servers or a list of VRFs, one per NTP server"),
     )
-    dns_collection: List[str] = Field(default_factory=lambda: ["5.192.28.174"], alias="dnsCollection", description="List of IPv4 and IPv6 DNS addresses")
-    dns_vrf_collection: List[str] = Field(
+    dns_collection: list[str] = Field(default_factory=lambda: ["5.192.28.174"], alias="dnsCollection", description="List of IPv4 and IPv6 DNS addresses")
```
akinross: is this default ip intended to be set?

plugins/module_utils/models/manage_fabric/manage_fabric_ebgp_vxlan.py
mtarking: Updated. `_validate_fabric_name` handles validation.

plugins/module_utils/models/manage_fabric/manage_fabric_ebgp_vxlan.py
mtarking: Fixed.

plugins/module_utils/models/manage_fabric/manage_fabric_ebgp_vxlan.py
mtarking: Fixed.

plugins/module_utils/models/manage_fabric/manage_fabric_ebgp_vxlan.py
mtarking: Fixed.

plugins/module_utils/models/manage_fabric/manage_fabric_ebgp_vxlan.py
mtarking: Update all fabrics to be the same.

plugins/module_utils/models/manage_fabric/manage_fabric_ebgp_vxlan.py
mtarking: Update all fabrics to be the same.

allenrobel (approved):
LGTM

akinross (approved):

samiib (approved):
LGTM

gmicol (approved):
LGTM

mikewiebe (approved):

PR #265, opened 2026-04-22
nd_interface_ethernet_trunk_host

allenrobel:
## Related Issue(s)

- CiscoDevNet/ansible-nd#318

## Proposed Changes

Adds the `nd_interface_ethernet_trunk_host` module for managing Cisco Nexus Dashboard 4.2 ethernet interfaces with the `trunkHost` policy type. Parallels `nd_interface_ethernet_access`: composite `(switch_ip, interface_name)` identifier, bulk create / bulk normalize-as-delete, per-interface deploy queueing, and the full merged / replaced / overridden / deleted state lifecycle.

- New `EthernetTrunkHostInterfaceModel` with trunk-specific fields — `allowed_vlans` (regex-validated, accepts `none`, `all`, or comma-separated IDs/ranges), `native_vlan`, `vlan_mapping`, and nested `vlan_mapping_entries` with `customer_vlan_id` as `List[str]` to match the ND API.
- `TrunkHostPolicyTypeEnum` added to shared interface enums.
- `EthernetTrunkHostInterfaceOrchestrator` filters `query_all` to exclude interfaces matching the unconfigured `int_trunk_host` default signature, so `state: overridden` is idempotent on the second run. Normalizing a trunkHost interface produces another trunkHost interface, so without this filter every fabric-default interface would appear in `before` and be re-queued for normalization on every run.
- `InterfaceDefaultPolicyModel` gains `description: ""` and `nativeVlan: 1` so `interfaceActions/normalize` actually clears user-set descriptions and native VLANs on deleted / overridden interfaces instead of leaving them behind. This affects all ethernet modules that delete via normalize, which is the desired behavior.
- `allowed_vlans` validator coerces bare-integer ND responses (e.g. a single VLAN returned as `999`) to string before regex validation.
- File-based logging integrated via `setup_logging()` in `main()`. Module-level `nd.nd_interface_ethernet_trunk_host` logger emits debug records at `expand_config`, `manage_state` begin/end, and on `NDStateMachineError` (via `log.exception` so tracebacks reach the log file regardless of `output_level`). Logging activates only when `ND_LOGGING_CONFIG` points at a `logging.config.dictConfig` JSON file; otherwise the calls are no-ops.
- Integration tests cover all four states. `selectattr('interface_name', ...)` lookups are scoped by `switch_ip` so assertions hold on multi-switch fabrics. The VLAN-mapping test block is gated on a `supports_vlan_mapping` variable because Nexus 9000v virtual switches reject `switchport vlan mapping ... dot1q-tunnel ...` at the NX-OS layer. The test target's `tasks/main.yaml` wraps its `include_tasks` in a block that forwards `nd_logging_config` (set in `inventory.networking [nd:vars]`) into the module subprocess as `ND_LOGGING_CONFIG`, since `ansible-test` strips the controller's shell environment.

## Test Notes

- Full integration suite passes against an ND 4.2 testbed with two Nexus 9000v switches in the target fabric: `ansible-test network-integration nd_interface_ethernet_trunk_host -v`.
- Merged, replaced, overridden (including idempotent re-run), and deleted states all pass.
- VLAN-mapping block is skipped by default; to exercise it on a hardware testbed add `supports_vlan_mapping=true` to the `[nd:vars]` section of `tests/integration/inventory.networking`. (`ansible-test` does not accept `--extra-vars` on the command line.)
- Unit tests cover the two branch-new source files with 142 tests and ~99% line coverage:
  - `tests/unit/module_utils/models/test_ethernet_trunk_host_interface.py` exercises every nested Pydantic model (`VlanMappingEntry`, `Policy`, `NetworkOS`, `ConfigData`, `Interface`) with field validation, range and enum constraints, validators, serializers, round-trip `to_payload` / `to_config` / `from_response` / `from_config`, composite identifier, diff, merge, and argument_spec shape.
  - `tests/unit/module_utils/orchestrators/test_ethernet_trunk_host_interface.py` wires a real `RestSend` with the file-based `Sender` from `tests/unit/module_utils/sender_file.py` and `ResponseHandler` to verify `_managed_policy_types`, the `_is_unconfigured_default` truth table, and `query_all` filtering (policy-type + unconfigured-default) across multiple switches.
  - Parametrize collapses all value-table tests (range checks, enum choice checks, truth table, input/output tables) into single functions with descriptive `ids=` labels so pytest output still names each case.
  - Run with: `python -m pytest tests/unit/module_utils/models/test_ethernet_trunk_host_interface.py tests/unit/module_utils/orchestrators/test_ethernet_trunk_host_interface.py`

4.2

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist


plugins/module_utils/models/interfaces/ethernet_trunk_host_interface.py
```diff
@@ -0,0 +1,513 @@
...
+
+        None
+        """
+        if not isinstance(value, str) or not value:
+            return value
+        match = _INTERFACE_NAME_PREFIX_RE.match(value)
+        if not match:
+            return value
+        prefix, rest = match.groups()
+        return prefix[0].upper() + prefix[1:].lower() + rest
```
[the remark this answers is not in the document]
allenrobel: Addressed in 41d0590f — `normalize_interface_name` now expands any case-insensitive prefix of `Ethernet` (e.g. `e1/1`, `eth1/1` -> `Ethernet1/1`) before falling back to Title case, matching the ethernet_access fix. Lab-verified the idempotency path on ND 4.2.1.

tests/unit/module_utils/models/test_ethernet_trunk_host_interface.py
```diff
@@ -0,0 +1,1792 @@
...
+
+@pytest.mark.parametrize(
+    "value,expected",
+    [
+        ("ethernet1/1", "Ethernet1/1"),
+        ("Ethernet1/1", "Ethernet1/1"),
+        ("e1/1", "E1/1"),
+        ("eth1/1/1", "Eth1/1/1"),
+        ("", ""),
+    ],
```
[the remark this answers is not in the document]
allenrobel: Addressed in 41d0590f — unit test `...00550` now asserts the expanded forms (`e1/1`/`eth1/1` -> `Ethernet1/1`, `eth1/1/1` -> `Ethernet1/1/1`), plus a Title-case fallback case for unrecognized prefixes.

tests/integration/targets/nd_interface_ethernet_trunk_host/tasks/merged.yaml
```diff
@@ -0,0 +1,314 @@
...
+      - eth43_after.config_data.network_os.policy.admin_state == false
+
+# --- MERGED CREATE: VLAN MAPPING ---
+# Skipped by default because Nexus 9000v virtual switches reject
+# `switchport vlan mapping ... dot1q-tunnel ...` ("not supported in the
+# current switchport mode"), which surfaces as HTTP 400 from ND on PUT.
+# Enable on a hardware testbed with:
+#   ansible-test network-integration nd_interface_ethernet_trunk_host \
+#       --extra-vars supports_vlan_mapping=true
+
```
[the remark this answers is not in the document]
allenrobel: Fixed in 3d364ed3 — the comment now points to the `[nd:vars]` section of `tests/integration/inventory.networking` instead of `--extra-vars`, matching the mechanism documented in `tasks/main.yaml`.

plugins/modules/nd_interface_ethernet_trunk_host.py
```diff
@@ -0,0 +1,643 @@
...
+ANSIBLE_METADATA = {"metadata_version": "1.1", "status": ["preview"], "supported_by": "community"}
+
+DOCUMENTATION = r"""
+---
+module: nd_interface_ethernet_trunk_host
+version_added: "2.0.0"
+short_description: Manage ethernet trunkHost interfaces on Cisco Nexus Dashboard
+description:
+- Manage ethernet trunkHost interfaces on Cisco Nexus Dashboard.
+- It supports creating, updating, querying, and deleting trunkHost interface configurations on switches within a fabric.
```
akinross:
```suggestion
- It supports creating, updating, and deleting trunkHost interface configurations on switches within a fabric.
```

tests/integration/targets/nd_interface_ethernet_trunk_host/tasks/overridden.yaml
```diff
@@ -0,0 +1,129 @@
...
+
+# --- SETUP ---
+# At this point Ethernet1/41, Ethernet1/42-43, Ethernet1/44, Ethernet1/45-47
+# exist from prior tests. Override will reduce the set to only what is specified.
+#
+# Note: trunkHost is the fabric-default policyType, so "removed" interfaces
+# are normalized back to default trunkHost configuration. They remain visible
+# to this module's query filter but as defaults rather than the prior config.
+
+# --- OVERRIDDEN: REDUCE TO A FEW INTERFACES ---
```
akinross, line 15: do we allow overriden to empty list, thus removing all the interfaces? should we add tests for this?

plugins/modules/nd_interface_ethernet_trunk_host.py
allenrobel: Good catch — removed "querying" from the module description. This module exposes only the state choices `[merged, replaced, overridden, deleted]`; there is no `query` state. Fixed in 959b92a5.

tests/integration/targets/nd_interface_ethernet_trunk_host/tasks/overridden.yaml
allenrobel: Yes — an empty `config` under `state: overridden` normalizes every trunkHost interface in the fabric back to its fabric default, since the overridden scope is fabric-wide. We decided to keep that behavior and cover it explicitly. 75514170 adds an integration test (check + normal + idempotency, then restores the fan-out config so the deleted-state tests keep their preconditions) and documents the fabric-wide empty-config behavior in the `state` option description and the module notes. Integration suite is green.

plugins/modules/nd_interface_ethernet_trunk_host.py
```diff
@@ -0,0 +1,648 @@
...
+                    - Whether traffic storm control is enabled on the interface.
+                    type: bool
+                  storm_control_action:
+                    description:
+                    - Storm control action on threshold violation.
+                    type: str
+                    choices: [ shutdown, trap, default ]
+                  storm_control_broadcast_level:
+                    description:
+                    - Broadcast storm control level in percentage (format V(whole.decimal), range 0.00-100.00).
```
mikewiebe, line 213: **Impact**

A playbook can supply both the percentage and PPS variant for the same storm-control traffic class, even though the module docs say those fields are mutually exclusive. That leaves the conflict to be resolved by ND, serialization order, or API-side validation instead of failing early with a clear Ansible/Pydantic error.

**Evidence**

- [Module docs](https://github.com/CiscoDevNet/ansible-nd/pull/265/files#diff-f7581ec3bfcae3e122df059e557bd92dafdb2f4b31e8339187987a8d14c337baR211) document the broadcast percentage/PPS pair as mutually exclusive.
- [Module docs](https://github.com/CiscoDevNet/ansible-nd/pull/265/files#diff-f7581ec3bfcae3e122df059e557bd92dafdb2f4b31e8339187987a8d14c337baR222) do the same for multicast.
- [Module docs](https://github.com/CiscoDevNet/ansible-nd/pull/265/files#diff-f7581ec3bfcae3e122df059e557bd92dafdb2f4b31e8339187987a8d14c337baR233) do the same for unicast.
- [The model fields](https://github.com/CiscoDevNet/ansible-nd/pull/265/files#diff-aec2c04ae468fb9c9f05bee0765635325601113fce4945ad5c37b32fd8076340R265) only apply range constraints; there is no model-level mutual-exclusion validator.
- [The argspec](https://github.com/CiscoDevNet/ansible-nd/pull/265/files#diff-aec2c04ae468fb9c9f05bee0765635325601113fce4945ad5c37b32fd8076340R507) lists each field independently.
- [The tests](https://github.com/CiscoDevNet/ansible-nd/pull/265/files#diff-cba2af0e708741da60f5f7d054b9f93c1b381f408ac09239e0de977ef2f26812R721) cover range boundaries for the PPS fields, but I did not find a test proving that percentage plus PPS is rejected for the same class.

**Recommendation**

Add a Pydantic model-level validator on `EthernetTrunkHostPolicyModel` that rejects these pairs:

- `storm_control_broadcast_level` with `storm_control_broadcast_level_pps`
- `storm_control_multicast_level` with `storm_control_multicast_level_pps`
- `storm_control_unicast_level` with `storm_control_unicast_level_pps`

Add one focused unit test per pair.

plugins/module_utils/models/interfaces/ethernet_trunk_host_interface.py
```diff
@@ -0,0 +1,537 @@
...
+    )
+    storm_control_unicast_level_pps: int | None = Field(
+        default=None,
+        alias="stormControlUnicastLevelPps",
+        ge=0,
+        le=200000000,
+        description="Unicast storm control level in packets per second",
+    )
+    vlan_mapping: bool | None = Field(default=None, alias="vlanMapping", description="Enable VLAN mapping on the interface")
+    vlan_mapping_entries: list[EthernetTrunkHostVlanMappingEntryModel] | None = Field(
```
mikewiebe, line 309: **Impact**

Users can enable VLAN mapping without supplying any mapping entries, even though the module docs and model description say entries are required. That creates a confusing runtime contract: the docs say the module validates the relationship, but the model accepts an incomplete policy and lets a later layer decide what happens.

**Evidence**

- [Module docs](https://github.com/CiscoDevNet/ansible-nd/pull/265/files#diff-f7581ec3bfcae3e122df059e557bd92dafdb2f4b31e8339187987a8d14c337baR244) say `vlan_mapping_entries` must be provided when `vlan_mapping` is true.
- [The model](https://github.com/CiscoDevNet/ansible-nd/pull/265/files#diff-aec2c04ae468fb9c9f05bee0765635325601113fce4945ad5c37b32fd8076340R307) makes both `vlan_mapping` and `vlan_mapping_entries` optional.
- [The argspec](https://github.com/CiscoDevNet/ansible-nd/pull/265/files#diff-aec2c04ae468fb9c9f05bee0765635325601113fce4945ad5c37b32fd8076340R513) also marks both fields independently.
- [The current unit coverage](https://github.com/CiscoDevNet/ansible-nd/pull/265/files#diff-cba2af0e708741da60f5f7d054b9f93c1b381f408ac09239e0de977ef2f26812R385) proves nested entries are coerced when provided, but I did not find a negative test for `vlan_mapping=True` without entries.

**Recommendation**

Add a model-level validator that rejects `vlan_mapping=True` when `vlan_mapping_entries` is missing or empty. Add a unit test for that exact case.

plugins/modules/nd_interface_ethernet_trunk_host.py
```diff
@@ -0,0 +1,648 @@
...
+                      dot1q_tunnel:
+                        description:
+                        - Whether selective dot1q-tunnel is enabled for this mapping.
+                        type: bool
+                      provider_vlan_id:
+                        description:
+                        - Provider VLAN that customer VLAN(s) are mapped to.
+                        - Valid range is 1-4094.
+                        type: int
+  deploy:
```
mikewiebe: Allen, we discussed this in our Ansible team meeting but I think I forgot to relay this to you.

New modules should converge on `config_actions` instead of top-level `deploy`

**Impact**

The module exposes the older interface-family top-level `deploy` option. The current review rule prefers `config_actions.save`, `config_actions.deploy`, and `config_actions.type` so save/deploy behavior has a consistent shape across new modules.

This is not necessarily a blocker for this PR because the existing interface family already uses `deploy`, and the PR may have been opened before this newer convention was established. It is still an outlier worth tracking for API consistency.

**Evidence**

- [The docs](https://github.com/CiscoDevNet/ansible-nd/pull/265/files#diff-f7581ec3bfcae3e122df059e557bd92dafdb2f4b31e8339187987a8d14c337baR277) expose top-level `deploy`.
- [The wrapper argspec](https://github.com/CiscoDevNet/ansible-nd/pull/265/files#diff-f7581ec3bfcae3e122df059e557bd92dafdb2f4b31e8339187987a8d14c337baR574) adds `deploy=dict(type="bool", default=True)`.
- [The orchestrator bridge](https://github.com/CiscoDevNet/ansible-nd/pull/265/files#diff-f7581ec3bfcae3e122df059e557bd92dafdb2f4b31e8339187987a8d14c337baR612) maps that top-level option to `model_orchestrator.deploy`.

**Recommendation**

Either migrate to the standard `config_actions` shape or explicitly document that interface modules are retaining `deploy` for family compatibility until the shared interface base is migrated.

Example from the inventory module

https://github.com/CiscoDevNet/ansible-nd/blob/develop/plugins/modules/nd_manage_switches.py#L47

mikewiebe (changes requested):

plugins/modules/nd_interface_ethernet_trunk_host.py
```diff
@@ -0,0 +1,648 @@
...
+notes:
+- This module is only supported on Nexus Dashboard.
+- This module manages NX-OS ethernet trunkHost interfaces only (interface_type C(ethernet), mode C(trunk),
+  network_os_type C(nx-os), policy_type C(trunkHost)). These values are hardcoded by the module and are not user-configurable.
+- Interfaces that are port-channel members have restricted mutability.
+- O(state=overridden) operates fabric-wide. An empty O(config) list resets every trunkHost
+  interface in the fabric to its fabric default configuration.
+"""
+
+EXAMPLES = r"""
```
mikewiebe, line 322: **Impact**

The integration tests exercise all four supported mutation states, but the module examples only show `merged` and `deleted`. Users reading the module docs do not get copyable examples for the more dangerous state transitions, especially fabric-wide `overridden`.

**Evidence**

- [Examples](https://github.com/CiscoDevNet/ansible-nd/pull/265/files#diff-f7581ec3bfcae3e122df059e557bd92dafdb2f4b31e8339187987a8d14c337baR317) cover `merged` and `deleted`, but not `replaced` or `overridden`.
- Integration coverage exists for [merged](https://github.com/CiscoDevNet/ansible-nd/pull/265/files#diff-af4fe85076eace279d4bb0778fd50bedaf7c312ee9cb6c5d5637b0f497db9071R15), [replaced](https://github.com/CiscoDevNet/ansible-nd/pull/265/files#diff-cf37418d4353f5bec2549cb6edae1f7669c7d2f12b6184908e8505ccd7a3b713R28), [overridden](https://github.com/CiscoDevNet/ansible-nd/pull/265/files#diff-6bdd779c8b7ce01c1eeb7919a05e3ee6d7f793268732b8b2f0e740f66aa51f74R42), and [deleted](https://github.com/CiscoDevNet/ansible-nd/pull/265/files#diff-8cf7fb39deecffaaa32947658f4d41037f09ad854f89e57e35706d661083545eR25).

**Recommendation**

Add one `replaced` example and one `overridden` example to the module docs. The `overridden` example should make the fabric-wide scope obvious, since the module already correctly warns about empty config resetting all trunk-host interfaces in the fabric.

plugins/modules/nd_interface_ethernet_trunk_host.py
allenrobel: Fixed in 9b05ed88 — added a model-level validator on `EthernetTrunkHostPolicyModel` that rejects setting both the percentage and the PPS variant of the same storm-control class (broadcast/multicast/unicast), with a parametrized unit test covering the both-set reject and the single/none accept cases per class.

plugins/module_utils/models/interfaces/ethernet_trunk_host_interface.py
allenrobel: Fixed in 13a63562 — added a model-level validator that rejects `vlan_mapping=true` with missing or empty `vlan_mapping_entries`, plus a unit test for the missing/empty reject and the enabled-with-entry / disabled / unset accept cases.

plugins/modules/nd_interface_ethernet_trunk_host.py
allenrobel: Fixed in ec82a6f2 — migrated to `config_actions.deploy`, mirroring the landed `nd_interface_port_channel_access` shape. `save` and `type` are intentionally omitted: verified against the ND 4.2.1 OpenAPI spec that `interfaceActions/deploy` is interface-type agnostic (body is `{interfaces:[{interfaceName,switchId}]}`), has no separate save/recalculate step, and exposes no `switch|global` scope. A global scope would also deploy unrelated pending fabric config, which the interface family deliberately avoids for resource-module safety. `version_added` is 2.0.0 (unreleased), so this is not a breaking change.

plugins/modules/nd_interface_ethernet_trunk_host.py
allenrobel: Fixed in fc3ac974 — added a `replaced` example and a fabric-wide `overridden` example. The overridden example carries an inline warning making the fabric-wide scope explicit: any trunkHost interface not listed is reset to its fabric default, and an empty config resets all of them.

plugins/modules/nd_interface_ethernet_trunk_host.py
allenrobel: Resolving after 1 week.

tests/integration/targets/nd_interface_ethernet_trunk_host/tasks/overridden.yaml
allenrobel: Resolving after 1 week.

plugins/modules/nd_interface_ethernet_trunk_host.py
```diff
@@ -0,0 +1,697 @@
...
+    default: merged
+    choices: [ merged, replaced, overridden, deleted ]
+extends_documentation_fragment:
+- cisco.nd.modules
+- cisco.nd.check_mode
+notes:
+- This module is only supported on Nexus Dashboard.
+- This module manages NX-OS ethernet trunkHost interfaces only (interface_type C(ethernet), mode C(trunk),
+  network_os_type C(nx-os), policy_type C(trunkHost)). These values are hardcoded by the module and are not user-configurable.
+- Interfaces that are port-channel members have restricted mutability.
```
mikewiebe, line 317: **Impact**

The integration tests exercise all four supported mutation states, but the module examples only show `merged` and `deleted`. Users reading the module docs do not get copyable examples for the more dangerous state transitions, especially fabric-wide `overridden`.

**Evidence**

- [Examples](https://github.com/CiscoDevNet/ansible-nd/pull/265/files#diff-f7581ec3bfcae3e122df059e557bd92dafdb2f4b31e8339187987a8d14c337baR317) cover `merged` and `deleted`, but not `replaced` or `overridden`.
- Integration coverage exists for [merged](https://github.com/CiscoDevNet/ansible-nd/pull/265/files#diff-af4fe85076eace279d4bb0778fd50bedaf7c312ee9cb6c5d5637b0f497db9071R15), [replaced](https://github.com/CiscoDevNet/ansible-nd/pull/265/files#diff-cf37418d4353f5bec2549cb6edae1f7669c7d2f12b6184908e8505ccd7a3b713R28), [overridden](https://github.com/CiscoDevNet/ansible-nd/pull/265/files#diff-6bdd779c8b7ce01c1eeb7919a05e3ee6d7f793268732b8b2f0e740f66aa51f74R42), and [deleted](https://github.com/CiscoDevNet/ansible-nd/pull/265/files#diff-8cf7fb39deecffaaa32947658f4d41037f09ad854f89e57e35706d661083545eR25).

**Recommendation**

Add one `replaced` example and one `overridden` example to the module docs. The `overridden` example should make the fabric-wide scope obvious, since the module already correctly warns about empty config resetting all trunk-host interfaces in the fabric.

plugins/modules/nd_interface_ethernet_trunk_host.py
```diff
@@ -0,0 +1,697 @@
...
+          network_os:
+            policy:
+              admin_state: true
+              allowed_vlans: "100-200"
+    config_actions:
+      deploy: false
+    state: merged
+"""
+
+RETURN = r"""
```
mikewiebe, line 460: **Issue**

The module returns structured state-machine output, but `RETURN` is empty. Generated `ansible-doc` output will not tell users what keys to expect from normal, check-mode, or failed runs.

**Evidence**

`RETURN` is currently defined as an empty block here, while the module exits with `nd_state_machine.output.format()` after state handling.

**Existing PR overlap**

**Impact**

Users reading the module docs do not know whether they can rely on keys such as `before`, `after`, `response`, `result`, or `diff`, which makes playbook assertions and automation around the module harder to write safely.

**Suggested fix**

Document the common state-machine return keys emitted by this module, including `before`, `after`, `response`, `result`, `diff`, and the behavior in check mode and failure paths. If the interface family has a shared return contract, mirror that wording here or move it into a reusable documentation fragment.

plugins/modules/nd_interface_ethernet_trunk_host.py
allenrobel: Already addressed via the [earlier comment for the same changes](https://github.com/CiscoDevNet/ansible-nd/pull/265#discussion_r3460584420).  Adding the reply to that comment here:

Fixed in https://github.com/CiscoDevNet/ansible-nd/commit/fc3ac9749559b218a78e3333a0dbb9f41a5e833d — added a replaced example and a fabric-wide overridden example. The overridden example carries an inline warning making the fabric-wide scope explicit: any trunkHost interface not listed is reset to its fabric default, and an empty config resets all of them.

plugins/modules/nd_interface_ethernet_trunk_host.py
allenrobel: Fixed in 39b4e7e — the previously-empty `RETURN` block now documents the keys actually emitted by `NDOutput.format()`: `changed`, `output_level`, `before`, `after`, `diff` (always), `proposed` (when `output_level` is `info`/`debug`), `logs` (when `output_level` is `debug`), and `msg` (on failure), with check-mode behavior noted on `after`.

On the reusable-documentation-fragment suggestion: I looked into it and Ansible can't share a `RETURN` block via `doc_fragments`. In ansible-core, `get_docstring()` runs `add_fragments()` against the DOCUMENTATION block only — `returndocs` is never fragment-merged, and `add_fragments` requires each fragment to carry `options`/`attributes`. A Python-constant approach is out too: `RETURN` is extracted by AST literal-eval (`ansible.parsing.plugin_docs`), so the module is never executed and the block can't be composed from an imported variable. It has to be a literal `r"""..."""` in each module.

The good news is the output shape is identical across the interface family — every `nd_interface_*` module exits through the same shared `NDOutput.format()` / `NDStateMachine`, with no interface-specific keys. So I authored this block canonically: the key/description/returned/type lines are reusable verbatim, and only the `sample:` values are interface-specific. I'll roll the same block out to the other `nd_interface_*` branches (svi, etc.) so the family has a consistent return contract.

plugins/modules/nd_interface_ethernet_trunk_host.py
allenrobel: Note: this branch was rebased onto the latest `develop` during the RETURN-block rollout, so the commit referenced above (`fc3ac97`) is now orphaned. The same change is now `35af8f1d`. Resolving as addressed.

plugins/modules/nd_interface_ethernet_trunk_host.py
allenrobel: Note: this branch was rebased onto the latest `develop` during the RETURN-block rollout, so the commit referenced above (`fc3ac97`) is now orphaned. The same change is now `35af8f1d`. Resolving as addressed.

plugins/modules/nd_interface_ethernet_trunk_host.py
allenrobel: Note: rebased onto the latest `develop` after this reply, so `39b4e7e` is now orphaned; the RETURN-block change is now `2ac0af5e`. Resolving as addressed.

mikewiebe (approved):
LGTM

gmicol (approved):
LGTM

akinross (approved):

allenrobel:
Merging on 3 approvals.  Thank you @akinross @gmicol @mikewiebe

PR #266, opened 2026-04-23
Add ACL manage module for Routing Policies

skaszlik:
New module to support Routing Policies - available in ND 4.1 and above.

Example playbook:
```yaml
  - name: TEST4 - Create ACL with TCP port options
      cisco.nd.nd_acl:
        fabric: "{{ test_fabric }}"
        state: merged
        config:
          - name: "{{ ipv4_acl_tcp }}"
            type: ipv4
            entries:
              - sequence_number: 10
                action: permit
                protocol: tcp
                src: any
                dst: "{{ test_network_ipv4 }}"
                dst_port_action: equal_to
                dst_port: 80
              - sequence_number: 20
                action: permit
                protocol: tcp
                src: any
                dst: "{{ test_network_ipv4 }}"
                dst_port_action: equal_to
                dst_port: 443
              - sequence_number: 30
                action: permit
                protocol: tcp
                src: any
                dst: "{{ test_network_ipv4 }}"
                dst_port_action: port_range
                dst_port_range_start: 8080
                dst_port_range_end: 8090
              - sequence_number: 40
                action: permit
                protocol: tcp
                src: any
                dst: any
                src_port_action: greater_than
                src_port: 1024
```                

## Related Issue(s)
#240 

## Proposed Changes
New module cisco.nd.nd_acl

## Test Notes
  All test phases passed:
    [OK] Merged state  - create, merge, check mode
    [OK] Replaced state - replace, check mode
    [OK] Query state   - all, specific, missing ACL
    [OK] Deleted state - specific, bulk, check mode

ND4.2

* [ ] analyze
* [ ] infa
* [ x] manage
* [ ] onemanage
* [ ] other

## Checklist

* [ x] Latest commit is rebased from develop with merge conflicts resolved


plugins/module_utils/endpoints/v1/manage/manage_acl.py
```diff
@@ -0,0 +1,173 @@
...
+  (POST /api/v1/manage/fabrics/{fabricName}/accessControlLists)
+- EpManageAclGet    - Get a specific ACL
+  (GET /api/v1/manage/fabrics/{fabricName}/accessControlLists/{aclName})
+- EpManageAclPut    - Update a specific ACL
+  (PUT /api/v1/manage/fabrics/{fabricName}/accessControlLists/{aclName})
+- EpManageAclDelete - Delete a specific ACL
+  (DELETE /api/v1/manage/fabrics/{fabricName}/accessControlLists/{aclName})
+"""
+
+from __future__ import absolute_import, annotations, division, print_function
```
allenrobel: No longer needed.  Please remove.

Instead replace with the following so that modern annotations are accepted:

```python
```

Comment applies to all files introduced in this PR.

plugins/module_utils/endpoints/v1/manage/manage_acl.py
```diff
@@ -0,0 +1,173 @@
...
+
+from typing import ClassVar, Literal, Optional
```
allenrobel: Remove Optional i.e.:

```python
from typing import ClassVar, Literal
```

We should be using modern annotations throughout i.e.:

- `Dict` -> `dict`
- `List` -> `list`
- `Set` -> `set`

Please replace all legacy annotions (if any) with the above across all files commited in this PR.

plugins/module_utils/endpoints/v1/manage/manage_acl.py
```diff
@@ -0,0 +1,173 @@
...
+
+class _EpManageAclBase(FabricNameMixin, NDEndpointBaseModel):
+    """
+    Base class for ND Manage ACL endpoints.
+
+    All ACL endpoints require a fabric_name path parameter.
+    Item-level endpoints also require an acl_name path parameter.
+    """
+
+    acl_name: Optional[str] = Field(default=None, description="ACL name")
```
allenrobel: Should be:

```python
    acl_name: str | None = Field(default=None, description="ACL name")
```

plugins/module_utils/endpoints/v1/manage/manage_acl.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def set_identifiers(self, identifier: IdentifierKey = None):
```
allenrobel: If a method has no return value signal that with an annotation.

```python
    def set_identifiers(self, identifier: IdentifierKey = None) -> None:
```

Comment applies to all methods.

plugins/modules/nd_acl.py
```diff
@@ -0,0 +1,425 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2026, Slawomir Kaszlikowski
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
```
allenrobel: `__metaclass__ = type` is no longer needed.  Remove.

Same comment applies to all files in this PR where this is present.

allenrobel (changes requested):
Good work Sławomir!  Just a few comments, please:

1. Replace legacy annotations with modern equivalents
2. Remove `__metaclass__ = type` where present
3. Remove  `from __future__ import absolute_import, annotations, division, print_function` and for files that need it, replace with `from __future__ import annotations`
4. Add return type annotations to all methods that lack them

plugins/module_utils/endpoints/v1/manage/manage_acl.py
skaszlik: removed

plugins/module_utils/endpoints/v1/manage/manage_acl.py
skaszlik: Done — Optional removed from imports, all annotations use modern X | None syntax.

plugins/module_utils/endpoints/v1/manage/manage_acl.py
skaszlik: Fixed — now acl_name: str | None = Field(default=None, ...).

plugins/module_utils/endpoints/v1/manage/manage_acl.py
skaszlik: Fixed — all methods now have explicit return type annotations.

plugins/modules/nd_acl.py
skaszlik: Removerd

skaszlik:
The `nd_acl` module has been **replaced** by `nd_manage_acl`, fully rebuilt on the
`NDStateMachine`/`NDBaseOrchestrator` framework to align with the canonical
manage-module pattern established in PR #286 (prefix lists, by gmicol).

### Files added

- `plugins/modules/nd_manage_acl.py` — new module (replaces `nd_acl.py`)
- `plugins/module_utils/models/acl/enums.py` — `AclTypeEnum`, `AclActionEnum`, `AclProtocolEnum`, `AclPortActionEnum`
- `tests/unit/module_utils/{endpoints,models,orchestrators}/test_manage_acl.py` — 49 unit tests
- `tests/unit/module_utils/fixtures/fixture_data/test_manage_acl.json` — test fixture data
- `tests/integration/targets/nd_manage_acl/` — new integration target (IPv4 + IPv6 + bulk-delete)

### Files modified

- `plugins/module_utils/models/acl/acl.py` — rebuilt with Pydantic `AclModel`/`AclEntryModel`; `exclude_from_diff = {"last_update_timestamp", "description"}`
- `plugins/module_utils/endpoints/v1/manage/manage_acl.py` — endpoint classes (`Get`, `ListGet`, `Post`, `Put`, `Delete`, `BulkDelete`)
- `plugins/module_utils/endpoints/mixins.py` — added `AclNameMixin`
- `plugins/module_utils/orchestrators/manage_acl.py` — rebuilt on `NDBaseOrchestrator`
- `plugins/module_utils/nd_state_machine.py` — minor, shared framework

### Files removed

- `plugins/modules/nd_acl.py` — old module
- `tests/integration/targets/nd_acl/` — old integration target

---

## Why

1. **Pattern alignment** — The old `nd_acl` module predated the `NDStateMachine`
   framework and used ad-hoc REST calls, custom diff logic, and non-standard
   argument specs. All new manage modules (`nd_manage_prefix_list`,
   `nd_manage_l3out`, etc.) use the same base classes, making the codebase
   consistent and easier to maintain.

2. **Pydantic data models** — Entry validation, camelCase↔snake_case translation,
   and wire-format serialization are now handled declaratively. Enums accept
   snake_case input and store ND-native camelCase via
   `field_validator(mode="before")`.

3. **Idempotency correctness** — The old module had edge cases where
   `description` caused false diffs (ND never returns it). The new model
   explicitly excludes `description` and `last_update_timestamp` from diff
   comparison.

4. **Consistent state semantics** — `merged` with entries now REPLACES the entry
   list (same as prefix-list); `state: overridden, config: []` removes all ACLs
   in a fabric; `state: deleted` accepts name-only config.

---

## Known ND API behaviors (documented in integration tests)

| Behavior | Detail |
|----------|--------|
| `type` field ignored for IPv6 | ND always persists and returns `type: ipv4` regardless of address family. The module echoes the user-specified type in `after`. |
| Port-range bounds returned as strings | `dstPortRangeStart: "80"` not `80`. Model handles coercion. |
| `description` never returned | ND silently drops it on create/update; excluded from diff to prevent false changes. |
| Bulk-delete body key | `{"accessControlListNames": [...]}` (not `accessControlLists`) |

---

## Testing status

| Layer | Result |
|-------|--------|
| Unit tests (49) | All pass — endpoints, models, orchestrators |
| Live playbooks (5 files, 28+ tasks) | All pass against ND 198.18.133.100 / NAC-SiteA |
| Integration target (IPv4 + IPv6 + overridden) | `ok=34 changed=9 failed=0` live |
| Dead-code scan (pyflakes) | Clean — no unused imports/vars |
| Stale reference scan | Zero `nd_acl` refs remaining |

---

## Migration notes for consumers

- Module FQCN changes from `cisco.nd.nd_acl` → `cisco.nd.nd_manage_acl`
- `fabric_name` is now a top-level parameter (not nested under `config`)
- `config` is always a list of ACL objects (even for single-ACL operations)
- `state: gathered` is removed; use `state: merged, config: []` for read-only preflight
- Return keys: `before`, `after`, `changed` (no `diff` output — always `[]`)

skaszlik:
Hi @gmicol,

Could you please do the final peer review? All requested changes have been addressed.

gmicol (approved):
LGTM

mikewiebe:
### High: PR currently reports merge conflicts

**Issue**

GitHub currently reports this PR as conflicting with the target branch.

**Evidence**

- PR metadata reports `mergeable: CONFLICTING`.
- The PR base reviewed was `0c34665128cf6df64bc9dae20175a1d5c119c3a6`.
- The PR changes `plugins/module_utils/endpoints/mixins.py`, which has also changed on the target branch.

**Existing PR overlap**

**Impact**

The PR cannot merge cleanly, and CI may not represent the final merged code until the branch is rebased or the conflicts are resolved.

**Suggested fix**

Please rebase this branch on current `develop`, resolve the merge conflicts, and re-run CI.

plugins/module_utils/models/acl/acl.py
```diff
@@ -0,0 +1,340 @@
...
+
+    # --- Serialization Configuration ---
+
+    exclude_from_diff: ClassVar[Set[str]] = {"last_update_timestamp", "description"}
+    payload_exclude_fields: ClassVar[Set[str]] = {"last_update_timestamp"}
+    unwanted_keys: ClassVar[List] = []
+
+    # --- Fields ---
+
+    name: str = Field(
```
mikewiebe, line 288: ### High: Tenant-scoped ACL names are valid in OpenAPI but rejected by the model

**Issue**

The ACL model only accepts names matching `[a-zA-Z0-9_-]` with max length 63, but the Manage OpenAPI schema allows tenant-qualified ACL names such as `tenant1~acl3` and max length 115. The schema also includes `tenantName`, which is not modeled here.

**Evidence**

- `plugins/module_utils/models/acl/acl.py` sets `name` max length to 63.
- `plugins/module_utils/models/acl/acl.py` rejects names outside `^[a-zA-Z0-9_-]+$`.
- The Manage OpenAPI `accessControlListName` schema allows `~` and max length 115 for non-default tenant qualified names.
- The OpenAPI base ACL schema includes `tenantName`.

I also reproduced this locally:

```python
AclModel.from_config({
    "name": "tenant1~acl3",
    "type": "ipv4",
    "entries": [
        {"sequence_number": 10, "action": "permit", "protocol": "ip", "src": "any", "dst": "any"}
    ],
})
```

This raises a validation error.

**Existing PR overlap**

**Impact**

A fabric containing any tenant-qualified ACL can break module initialization when `query_all()` parses existing controller state, even if the user is managing a different ACL. The module also cannot create, update, or delete tenant-scoped ACLs.

**Suggested fix**

Allow `~` and 115-character ACL names per OpenAPI, add `tenant_name` / `tenantName` support if tenant scope is intended, and add unit/integration coverage for tenant-qualified ACLs.

plugins/module_utils/models/acl/acl.py
```diff
@@ -0,0 +1,340 @@
...
+        description="Destination match (address/CIDR or 'any').",
+    )
+
+    src_port_action: Optional[PortActionEnum] = Field(
+        default=None,
+        alias="srcPortAction",
+        description="Source port operator (tcp/udp).",
+    )
+
+    src_port: Optional[int] = Field(default=None, alias="srcPort", description="Source port value.")
```
mikewiebe: ### High: ACL port fields reject valid service-name ports

**Issue**

The ACL port fields are modeled as integers, but the Manage OpenAPI schema defines them as strings and allows service names such as `www` and `ftp`.

**Evidence**

- `plugins/module_utils/models/acl/acl.py` models `src_port`, `dst_port`, and range endpoints as `int`.
- The argspec exposes the same fields as `type: int`.
- The Manage OpenAPI schema defines `srcPort`, `dstPort`, `srcPortRangeStart`, `srcPortRangeEnd`, `dstPortRangeStart`, and `dstPortRangeEnd` as strings.
- OpenAPI examples include values such as `srcPort: "www"` and `dstPort: "ftp"`.

I also reproduced this locally:

```python
AclModel.from_config({
    "name": "ACL1",
    "type": "ipv4",
    "entries": [
        {
            "sequence_number": 10,
            "action": "permit",
            "protocol": "tcp",
            "src": "any",
            "dst": "any",
            "dst_port_action": "equal_to",
            "dst_port": "www",
        }
    ],
})
```

This raises a validation error.

**Existing PR overlap**

**Impact**

Users cannot configure valid ACLs using service-name ports. Existing controller state with service-name ports can also make `query_all()` fail before the module acts on the requested object.

**Suggested fix**

Model the port fields as strings, or accept `str | int` input and serialize to API strings while preserving service names. Please update docs, examples, fixtures, and tests to include at least one service-name port.

tests/integration/targets/nd_manage_acl/tasks/acl_ipv6.yaml
```diff
@@ -0,0 +1,159 @@
...
+    status_code:
+      - 200
+  register: ipv6_acl_query
+  delegate_to: localhost
+
+- name: "VALIDATION: Parse the IPv6 ACL response"
+  ansible.builtin.set_fact:
+    ipv6_acl_wire: "{{ ipv6_acl_query.json }}"
+
+# NOTE on the persisted wire payload (verified against live ND):
```
mikewiebe, line 112: ### Medium: Module `after` can disagree with controller state for IPv6 ACLs

**Issue**

The state machine returns `after` from locally mutated desired state rather than from a post-write controller read. For this ACL module, the integration test itself notes that ND stores and returns IPv6 ACLs as `type: ipv4`, while module output still echoes the requested `ipv6`.

**Evidence**

- `plugins/module_utils/nd_state_machine.py` mutates the local `existing` collection during create/update handling.
- `plugins/module_utils/nd_state_machine.py` then assigns that local collection to output `after`.
- `tests/integration/targets/nd_manage_acl/tasks/acl_ipv6.yaml` notes that direct ND readback always returns `type: ipv4`, while the module `after` echoes `ipv6`.

**Existing PR overlap**

**Impact**

Users may treat `after` as persisted controller state, but for IPv6 ACLs it can be predicted state instead. That can hide controller normalization and make troubleshooting harder.

**Suggested fix**

Re-query changed ACLs after successful create/update/delete, or explicitly document `after` as predicted state and add a warning/note for the ND IPv6 `type` readback behavior.

NOTE: @skaszlik please compare notes on this one with @gmicol to see if the other route control modules have the same issue.  Today `after` does represent predicted state not persisted controller state and is supposed to be controlled [by these additional module options](https://github.com/CiscoDevNet/ansible-nd/blob/develop/plugins/modules/nd_manage_vpc_pair.py#L70) which are not supported across all modules today.

Let me know what you and @gmicol decide.  If the decision is to bring in the verify capabilities or just document the behavior.  If the decision is to just document for now I will open a new issue to track bringing in verify across the modules so someone can work this as a separate bolt on capability.

plugins/module_utils/models/acl/acl.py
```diff
@@ -0,0 +1,340 @@
...
+    # --- Fields ---
+
+    name: str = Field(
+        alias="name",
+        min_length=1,
+        max_length=63,
+        description="Name of the ACL (pattern: ^[a-zA-Z0-9_-]+$).",
+    )
+
+    type: Optional[AclTypeEnum] = Field(
```
mikewiebe: ### Medium: State-aware required fields are documented but not enforced locally

**Issue**

The module documentation says `type` and `entries` are required for `merged`, `replaced`, and `overridden`, but the model and argspec allow both fields to be omitted for every state.

**Evidence**

- `plugins/modules/nd_manage_acl.py` documents `type` and `entries` as required for `state=merged`, `state=replaced`, and `state=overridden`.
- `plugins/module_utils/models/acl/acl.py` makes `type` optional.
- `plugins/module_utils/models/acl/acl.py` makes `entries` optional.
- The generated argspec exposes the same optional shape.

I also reproduced this locally:

```python
AclModel.from_config({"name": "ACL1"}).to_payload()
```

This returns:

```python
{"name": "ACL1"}
```

**Existing PR overlap**

**Impact**

For create/update states, a name-only config can pass local validation and reach the controller as an invalid or ambiguous payload. The module should fail fast before making API calls.

**Suggested fix**

Add state-aware validation before building proposed models, or pass validation context into the model so identifier-only entries are accepted only for `state: deleted`.

plugins/module_utils/models/acl/enums.py
```diff
@@ -0,0 +1,89 @@
...
+
+class AclProtocolEnum(str, Enum):
+    """
+    Layer-3/4 protocol matched by an ACL entry.
+
+    Ansible value equals the API wire value for every member, so no
+    translation is required.
+    """
+
+    IP = "ip"
```
mikewiebe, line 49: ### Medium: Protocol validation is not family-specific and accepts unsupported values

**Issue**

The module uses one protocol choice list for both IPv4 and IPv6 ACL entries, and that list includes values that are not advertised by the ACL OpenAPI schema.

**Evidence**

- `plugins/module_utils/models/acl/enums.py` includes `ahp`, `gre`, `nos`, and `esp`.
- The OpenAPI IPv4 ACL protocol enum is `icmp`, `ip`, `tcp`, `udp`, `eigrp`, `ospf`, `pim`, `igmp`, and `custom`.
- The OpenAPI IPv6 ACL protocol enum is `icmp`, `ipv6`, `tcp`, `udp`, `eigrp`, `pim`, and `custom`.

**Existing PR overlap**

**Impact**

The module accepts protocol values the ACL API does not advertise, and it allows IPv4-only protocols on IPv6 ACLs or IPv6-only protocol values on IPv4 ACLs. Those errors will be caught late by the controller instead of by Ansible validation.

**Suggested fix**

Validate protocol choices against the ACL `type`. Keep only OpenAPI-supported values, or document and test any intentionally accepted values that the live controller supports beyond the schema.

plugins/module_utils/orchestrators/manage_acl.py
```diff
@@ -0,0 +1,179 @@
...
+        """Retrieve a single ACL by name."""
+        try:
+            api_endpoint = self.query_one_endpoint()
+            api_endpoint.fabric_name = self.fabric_name
+            api_endpoint.set_identifiers(model_instance.get_identifier_value())
+            return self._request(path=api_endpoint.path, verb=api_endpoint.verb)
+        except Exception as e:
+            raise Exception(f"Query failed for {model_instance.get_identifier_value()}: {e}") from e
+
+    def query_all(self, model_instance: AclModel = None, **kwargs) -> ResponseType:
```
mikewiebe, line 147: ### Medium: ACL collection reads should be evaluated with scale testing

**Issue**

`query_all()` performs a single bare collection request even though the ACL list endpoint supports pagination and returns metadata. This may be fine for small fabrics, but we should validate the behavior with a larger ACL dataset before deciding whether code changes are required.

**Evidence**

- `plugins/module_utils/orchestrators/manage_acl.py` performs one list request and returns `raw.get("accessControlLists", [])`.
- The endpoint model exposes `max` and `offset`.
- The OpenAPI list response includes `meta`, which indicates collection pagination data is available.

**Existing PR overlap**

**Impact**

If the controller applies a default page size, idempotency, delete, and overridden behavior may only see the first page of ACLs on larger fabrics. That could cause false creates, missed deletes, or incomplete `overridden` cleanup. This needs scale testing before we treat it as a confirmed defect.

**Suggested fix**

Please add or run scale testing with enough ACLs to confirm whether the default collection response includes all ACLs. If ND paginates by default, implement pagination for ACL collection reads or explicitly request a safe `max` and document any remaining scale limit.

mikewiebe:
### Medium: NaC route-control ACL feature coverage is only partial

**Issue**

`nd_manage_acl` covers basic ACL entries, but current NaC route-control ACL templates can express several ACL constructs that are not represented by this module.

**Evidence**

- NaC IPv4 route-control ACL templates support top-level and entry-level `fragments`, `ignore routable`, `statistics per-entry`, `addrgroup`, and rich filtering options.
- NaC IPv6 route-control ACL templates support `extension-header`, `addrgroup`, and rich filtering options.
- `nd_manage_acl` currently models basic `src`, `dst`, ports, `icmp_option`, and `tcp_option`.

**Existing PR overlap**

**Impact**

A future NaC migration from generated `dcnm_policy` / freeform ACL templates to this module cannot represent several current route-control ACL constructs. Some of these may be ND Manage API limitations rather than module omissions.

**Suggested fix**

Please document the current ACL feature boundary. If ND Manage ACL supports these features through additional API properties, add model/docs/tests. If it does not, we should track this as a NaC migration limitation until ND exposes equivalent capability.

plugins/modules/nd_manage_acl.py
```diff
@@ -0,0 +1,356 @@
...
+
+- name: Delete an ACL (identifier only)
+  cisco.nd.nd_manage_acl:
+    fabric_name: my-fabric
+    state: deleted
+    config:
+      - name: ACL-IPV4-1
+"""
+
+RETURN = r"""
```
mikewiebe, line 312: ### Low: `RETURN` documentation is empty

**Issue**

The module has an empty `RETURN` block even though it returns structured state-machine output.

**Evidence**

- `plugins/modules/nd_manage_acl.py` defines `RETURN = r""" """`.
- `NDOutput` returns `output_level`, `changed`, `after`, `before`, `diff`, and optionally `proposed` and `logs`.

**Existing PR overlap**

**Impact**

`ansible-doc` does not describe the module's result contract, including the important predicted-vs-controller-state nuance for `after`.

**Suggested fix**

Populate `RETURN` with the common NDStateMachine fields and ACL-specific examples. A good current pattern is the `RETURN` block in `plugins/modules/nd_interface_port_channel_access.py` on `develop`, starting at [line 348](https://github.com/CiscoDevNet/ansible-nd/blob/develop/plugins/modules/nd_interface_port_channel_access.py#L348). That example documents each result key with `description`, `returned`, `type`, and `sample`/`elements` where applicable:

```yaml
changed:
  description: Whether the module changed, or in check mode would change, the fabric configuration.
  returned: always
  type: bool
  sample: true
output_level:
  description: The output verbosity level in effect for the run, echoing the output_level parameter.
  returned: always
  type: str
  sample: normal
before:
  description: The existing configuration before the module ran.
  returned: always
  type: list
  elements: dict
after:
  description: The configuration after the module ran, or predicted configuration in check mode.
  returned: always
  type: list
  elements: dict
diff:
  description: The difference between before and after.
  returned: always
  type: list
  elements: dict
proposed:
  description: The configuration the module proposed to apply.
  returned: when output_level is info or debug
  type: list
  elements: dict
logs:
  description: Internal diagnostic log messages collected during the run.
  returned: when output_level is debug
  type: list
  elements: str
msg:
  description: A human-readable error message, present only when the module fails.
  returned: on failure
  type: str
```

For this module, the samples should be ACL-specific rather than port-channel-specific, but the same documented result keys and structure should apply.

plugins/modules/nd_manage_acl.py
```diff
@@ -0,0 +1,356 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2026, Slawomir Kaszlikowski (@skaszlik)
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
```
mikewiebe: ### Low: Previously resolved style concerns still appear in current files

**Issue**

Earlier review comments asked to remove legacy future imports, `__metaclass__ = type`, and legacy typing aliases. Those original threads are resolved and outdated, but the same style concerns still appear in the current `nd_manage_acl` files after the module rewrite.

**Evidence**

- `plugins/modules/nd_manage_acl.py` still uses `from __future__ import absolute_import, division, print_function`.
- `plugins/modules/nd_manage_acl.py` still has `__metaclass__ = type`.
- `plugins/module_utils/models/acl/acl.py` still imports legacy typing aliases such as `Dict`, `List`, `Optional`, and `Set`.
- `plugins/module_utils/orchestrators/manage_acl.py` imports `List`.

**Existing PR overlap**

Related to prior resolved/outdated comments:
- https://github.com/CiscoDevNet/ansible-nd/pull/266#discussion_r3171541199
- https://github.com/CiscoDevNet/ansible-nd/pull/266#discussion_r3171544315
- https://github.com/CiscoDevNet/ansible-nd/pull/266#discussion_r3171559049

**Impact**

This is not a runtime blocker, but it means the earlier style cleanup was only partially carried forward after the `nd_acl` to `nd_manage_acl` rewrite.

**Suggested fix**

Apply the earlier resolved style cleanup consistently to the current PR files.

skaszlik:
> ### High: PR currently reports merge conflicts
> 
> **Issue**
> 
> GitHub currently reports this PR as conflicting with the target branch.
> 
> **Evidence**
> 
>     * PR metadata reports `mergeable: CONFLICTING`.
> 
>     * The PR base reviewed was `0c34665128cf6df64bc9dae20175a1d5c119c3a6`.
> 
>     * The PR changes `plugins/module_utils/endpoints/mixins.py`, which has also changed on the target branch.
> 
> 
> **Existing PR overlap**
> 
> No matching existing PR comment found.
> 
> **Impact**
> 
> The PR cannot merge cleanly, and CI may not represent the final merged code until the branch is rebased or the conflicts are resolved.
> 
> **Suggested fix**
> 
> Please rebase this branch on current `develop`, resolve the merge conflicts, and re-run CI.

Conflict resolved

plugins/module_utils/models/acl/acl.py
skaszlik: Fixed. The model now accepts tenant-qualified ACL names: the name pattern is ^[a-zA-Z0-9_~-]+$ (added ~) and the max length is raised from 63 to 115, matching the OpenAPI accessControlListName schema. I also bumped the acl_name endpoint path parameter (AclNameMixin) from 63 to 115

plugins/module_utils/models/acl/acl.py
skaszlik: Done. `str | int ` input accepted

plugins/module_utils/models/acl/acl.py
skaszlik: Addressed. type and entries are now enforced as required for the write states (merged, replaced, overridden), while identifier-only items (name only) remain valid for deleted — using the validation-context approach you suggested, which matches the existing SwitchConfigModel.apply_state_defaults precedent.

Added unit coverage (test_manage_acl_00135) for the write-state raise, the deleted/no-context accept paths, and the full-config pass.

plugins/module_utils/models/acl/enums.py
skaszlik: dropped AHP, GRE, NOS and ESP

plugins/module_utils/models/acl/acl.py
```diff
@@ -0,0 +1,394 @@
...
+        """
+        Validate semantic correctness of every entry and enforce unique
+        sequence numbers within the ACL.
+        """
+        entries = self.entries or []
+        seen_sequence_numbers: Set[int] = set()
+
+        for idx, entry in enumerate(entries):
+            if entry.sequence_number in seen_sequence_numbers:
+                raise ValueError(f"entries[{idx}].sequenceNumber '{entry.sequence_number}' is duplicated. Sequence numbers must be unique within an ACL.")
```
akinross: I think we should error with the complete list of duplicate sequence numbers, instead of error on the first duplication found. Same applies for all the individual entry validations and all other value errors raised. If the ACL is long and user needs to fix this, that user might need to go through a long iterative process to get errors returned.

plugins/module_utils/models/acl/acl.py
```diff
@@ -0,0 +1,394 @@
...
+                raise ValueError(f"ACL '{self.name}': '{', '.join(missing)}' {verb} required for state '{state}'.")
+        return self
+
+    @model_validator(mode="after")
+    def validate_entries(self) -> "AclModel":
+        """
+        Validate semantic correctness of every entry and enforce unique
+        sequence numbers within the ACL.
+        """
+        entries = self.entries or []
```
akinross: i think you can already return here when not entries?

plugins/module_utils/models/acl/acl.py
```diff
@@ -0,0 +1,394 @@
...
+        for idx, entry in enumerate(entries):
+            if entry.sequence_number in seen_sequence_numbers:
+                raise ValueError(f"entries[{idx}].sequenceNumber '{entry.sequence_number}' is duplicated. Sequence numbers must be unique within an ACL.")
+            seen_sequence_numbers.add(entry.sequence_number)
+            self._validate_entry(idx, entry)
+
+        return self
+
+    @staticmethod
+    def _validate_entry(idx: int, entry: "AclEntryModel") -> None:
```
akinross: currently you validate and error per entry, why in that case not define a model_validator or field_validator on AclEntryModel when you are not accumulating all errors??

plugins/module_utils/models/acl/acl.py
```diff
@@ -0,0 +1,394 @@
...
+        state = (info.context or {}).get("state") if info else None
+        if state in ("merged", "replaced", "overridden"):
+            missing = [name for name in ("type", "entries") if getattr(self, name) is None]
+            if missing:
+                verb = "is" if len(missing) == 1 else "are"
+                raise ValueError(f"ACL '{self.name}': '{', '.join(missing)}' {verb} required for state '{state}'.")
+        return self
+
+    @model_validator(mode="after")
+    def validate_entries(self) -> "AclModel":
```
akinross: why is this a model_validator and not a field_validator when it touches only entries field?

plugins/module_utils/models/acl/acl.py
```diff
@@ -0,0 +1,394 @@
...
+            AclModel._validate_port_options(idx, entry, "src")
+            AclModel._validate_port_options(idx, entry, "dst")
+
+        if entry.icmp_option is not None and protocol != "icmp":
+            raise ValueError(f"entries[{idx}]: 'icmp_option' is only valid when protocol is 'icmp'.")
+        if entry.tcp_option is not None and protocol != "tcp":
+            raise ValueError(f"entries[{idx}]: 'tcp_option' is only valid when protocol is 'tcp'.")
+
+    @staticmethod
+    def _validate_port_options(idx: int, entry: "AclEntryModel", prefix: str) -> None:
```
akinross: why in that case not define a model_validator on AclEntryModel when it is specific to that model and you are not accumulating all errors?

akinross (changes requested):

plugins/module_utils/orchestrators/manage_acl.py
skaszlik: Fixed in orchestrators/manage_acl.py: query_all now walks the collection with max/offset (page size 100) until it hits a short page (< page_size) or an empty page. Guards: a per-name seen set stops the walk if a full page returns no new names (controller ignoring offset), and a query_all_max_pages cap is a final safety net. Page size / cap are ClassVars so unit tests can shrink them.

plugins/modules/nd_manage_acl.py
skaszlik: I've documented the feature boundary directly in the module's notes.  Populated RETURN following the nd_interface_port_channel_access.py pattern you referenced, with ACL-specific samples: changed, output_level, before, after, diff, proposed, logs, and msg, each documented with description/returned/type and sample/elements.

plugins/modules/nd_manage_acl.py
skaszlik: Applied the style cleanup consistently across the nd_manage_acl files:
- plugins/modules/nd_manage_acl.py: removed from __future__ import absolute_import, division, print_function and __metaclass__ = type.
- plugins/module_utils/models/acl/acl.py: dropped the legacy typing aliases; now Any, ClassVar, Literal only, with builtin generics and X | None throughout.
- plugins/module_utils/orchestrators/manage_acl.py: List → list.
- Also swept plugins/module_utils/endpoints/v1/manage/manage_acl.py for the same Optional → X | None consistency.

Full unit suite and ansible-test sanity pass across all touched files.

plugins/module_utils/models/acl/acl.py
skaszlik: validate_entries now aggregates: it collects every entry's errors and raises them together in a single message instead of failing on the first. For duplicate sequence numbers it reports the complete set, e.g. `sequenceNumber values 10, 20 are duplicated. Sequence numbers must be unique within an ACL.`

plugins/module_utils/models/acl/acl.py
skaszlik: Good call — added the explicit early return:
```python
entries = self.entries or []
if not entries:
    return self
```

plugins/module_utils/models/acl/acl.py
skaszlik: This ties directly into your comment above about accumulating all errors — I went the accumulation route, and that's exactly why the semantic checks stay on AclModel rather than moving to AclEntryModel:

- The duplicate-sequence-number check is cross-entry (it needs the whole list), so it can't live on AclEntryModel at all.
- If the per-entry checks were an AclEntryModel model_validator, a failing entry would short-circuit the parent — pydantic aborts nested validation before AclModel's cross-entry validator runs, so you'd never get the per-entry errors and the duplicate-sequence error in the same failure.
- Field-local normalization does live on AclEntryModel already (normalize_port, normalize_port_action field validators)

plugins/module_utils/models/acl/acl.py
skaszlik: Good point — converted validate_entries to a field_validator("entries") since it only inspects the entries field. The field validator errors now aggregate with other field-level errors (e.g. an invalid ACL name and a bad entry now surface together in one ValidationError instead of the name error masking the entry errors).

plugins/module_utils/models/acl/acl.py
skaszlik: moved the per-entry semantic validation onto AclEntryModel as a model_validator (it accumulates all of that entry's errors), leaving only the cross-entry unique-sequence check on AclModel. pydantic aggregates errors across entries automatically. One trade-off: when entries are individually invalid and have a duplicate sequence number, you'll see the entry errors first and the duplicate on the next run (pydantic can't run the parent's cross-entry check until each entry validates). Updated the entry-construction unit tests accordingly.

tests/integration/targets/nd_manage_acl/tasks/acl_ipv6.yaml
skaszlik: I've adjust the documentation notes: now state after is predicted post-write state (not a controller re-read), plus the IPv6 single-object-GET caveat.

mikewiebe (approved):
LGTM

akinross (approved):

PR #268, opened 2026-04-27
nd_interface_svi

allenrobel:
## Related Issue(s)

- CiscoDevNet/ansible-nd#319

- #349 — query_all per-switch fan-out / no fabric-wide interface endpoint (cross-cutting)
- #352 — SVI phase 2: HSRP sub-block modeling + OSPF/ISIS/BFD interface fields

## Proposed Changes

Adds `nd_interface_svi` to manage SVI (switched virtual interface) configurations on Cisco Nexus Dashboard via the Manage Interfaces API.

- `plugins/module_utils/models/interfaces/svi_interface.py` — `SviInterfaceModel` plus nested `SviPolicyModel`, `SviNetworkOSModel`, `SviConfigDataModel`, `SviOperDataModel`. Composite identifier `(switch_ip, interface_name)` where `interface_name` is normalized to lowercase `vlan<id>`.
- `plugins/module_utils/orchestrators/svi_interface.py` — `SviInterfaceOrchestrator` inheriting from `NDBaseInterfaceOrchestrator` (not the ethernet base; SVIs have no port-channel concerns and support real DELETE via `interfaceActions/remove` + `interfaceActions/deploy`).
- `plugins/modules/nd_interface_svi.py` — module entry point. Each config item targets a single SVI via `interface_name` (e.g. `vlan333`); to configure multiple SVIs, list multiple config items.
- `plugins/module_utils/models/interfaces/enums.py` — adds `SviPolicyTypeEnum.SVI = "svi"`.

Phase 1 exposes the policy fields the ND GUI sends on create: `admin_state`, `description`, `ip`/`prefix`, `ipv6`/`v6prefix`, `ip_redirects`, `pim_sparse`, `pim_dr_priority`, `hsrp_group`, `hsrp_version`, `preempt`, `advertise_subnet_in_underlay`, `mtu`, `extra_config`, `netflow`. OSPF / ISIS / BFD / proper HSRP sub-block modeling deferred to a follow-up release.

One SVI-specific deviation from the loopback / ethernet-access pattern, validated by Bruno testing against a live testbed:

1. `from_response()` strips `hsrpVersion`. ND's GET returns `hsrpVersion: 1` (integer) as a server-side default even when HSRP is unconfigured; re-emitting that on PUT — as either int `1` or string `"1"` — triggers a generic 500 ("unexpected error during policy execution"). `hsrpGroup: 1` round-trips fine, so only `hsrpVersion` is poisoned. Proper HSRP sub-block modeling is deferred to phase 2.

## Test Notes

Full create / update / query / idempotency / delete lifecycle verified against an ND 4.2 testbed (`fabric_1`, switch `LE1` at `192.168.12.151`, VLANs 333/334/335).

- Unit tests: `tests/unit/.../test_svi_interface*.py` (commit `434b221`).
- Integration tests: `tests/integration/targets/nd_interface_svi/` (commit `f8b2223`).
- POST: 201 with single and multi-VLAN config items.
- PUT: 200 on partial body (only `description` + `policyType`) — server merges.
- GET: round-trips through model with `oper_data` populated.
- Idempotency: second-run reports `changed: false` with empty diff.
- DELETE: `interfaceActions/remove` (204) followed by `interfaceActions/deploy` confirmed via post-delete check_mode query (`before: []`).

4.2

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

* [ ] Latest commit is rebased from develop with merge conflicts resolved (this branch is stacked on `nd_interface_ethernet_trunk_host`; rebase will follow that branch's merge to develop)
* [x] New or updates to documentation has been made accordingly (module DOCUMENTATION + EXAMPLES populated)

### DIFF
--- a/plugins/module_utils/models/interfaces/svi_interface.py
+++ b/plugins/module_utils/models/interfaces/svi_interface.py
@@ -42,6 +42,7 @@ from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat im
 from ansible_collections.cisco.nd.plugins.module_utils.models.base import NDBaseModel
 from ansible_collections.cisco.nd.plugins.module_utils.models.interfaces.enums import SviPolicyTypeEnum
 from ansible_collections.cisco.nd.plugins.module_utils.models.nested import NDNestedModel
+from ansible_collections.cisco.nd.plugins.module_utils.models.types import AsciiDescription
 
 
 class SviPolicyModel(NDNestedModel):
@@ -60,7 +61,7 @@ class SviPolicyModel(NDNestedModel):
 
     policy_type: SviPolicyTypeEnum = Field(default=SviPolicyTypeEnum.SVI, alias="policyType", description="Interface policy type")
     admin_state: bool | None = Field(default=None, alias="adminState", description="Enable or disable the interface")
-    description: str | None = Field(default=None, alias="description", max_length=254, description="Interface description")
+    description: AsciiDescription = Field(default=None, alias="description", max_length=254, description="Interface description")
     extra_config: str | None = Field(default=None, alias="extraConfig", description="Additional CLI for the interface")
     mtu: int | None = Field(default=None, alias="mtu", ge=576, le=9216, description="Interface MTU")
     ip: str | None = Field(default=None, alias="ip", description="IPv4 address of the SVI")
@@ -79,34 +80,6 @@ class SviPolicyModel(NDNestedModel):
     netflow: bool | None = Field(default=None, alias="netflow", description="Enable Netflow on the interface")
 
     # --- Validators ---
-
-    @field_validator("description")
-    @classmethod
-    def description_must_be_ascii(cls, value):
-        """
-        # Summary
-
-        Reject non-ASCII characters in `description`. Some NX-OS / ND backend code paths pipe interface descriptions
-        through CLI generators that do not handle UTF-8 cleanly and return a generic 500 ("unexpected error during
-        policy execution") instead of a meaningful validation error. Catching this client-side gives the user a clear
-        message instead of a confusing server fault. Remove or relax this check once Cisco fixes the backend.
-
-        ## Raises
-
-        ### ValueError
-
-        - If `value` contains any non-ASCII character.
-        """
-        if value is None:
-            return value
-        try:
-            value.encode("ascii")
-        except UnicodeEncodeError as e:
-            raise ValueError(
-                f"description must contain only ASCII characters; got non-ASCII at position {e.start}: {value[e.start]!r}. "
-                "The ND backend currently returns a generic 500 for non-ASCII descriptions."
-            ) from None
-        return value
 
     @field_validator("policy_type", mode="before")
     @classmethod
--- a/plugins/module_utils/models/interfaces/svi_interface.py
+++ b/plugins/module_utils/models/interfaces/svi_interface.py
@@ -1,0 +1,356 @@
+# Copyright: (c) 2026, Allen Robel (@allenrobel)
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+"""
+SVI (switched virtual interface) Pydantic models for Nexus Dashboard.
+
+This module defines nested Pydantic models that mirror the ND Manage Interfaces API payload structure for SVI
+interfaces (`interfaceType: "svi"`, `policyType: "svi"`, `mode: "managed"`). The playbook config uses the same
+nesting so that `to_payload()` and `from_response()` work via standard Pydantic serialization with no custom
+wrapping or flattening.
+
+## Model Hierarchy
+
+- `SviInterfaceModel` (top-level, `NDBaseModel`)
+    - `interface_name` (identifier, e.g. `vlan333`)
+    - `interface_type` (default: "svi")
+    - `config_data` -> `SviConfigDataModel`
+        - `mode` (default: "managed")
+        - `network_os` -> `SviNetworkOSModel`
+            - `network_os_type` (default: "nx-os")
+            - `policy` -> `SviPolicyModel`
+                - `policy_type` (default: "svi"), `admin_state`, `ip`, `prefix`, etc.
+    - `oper_data` -> `SviOperDataModel` (read-only, returned on GET, excluded from payload)
+
+## Phase 1 field set
+
+The fields in `SviPolicyModel` cover the SVI options that the ND GUI sends on create. OSPF / ISIS / BFD /
+routing-protocol / replication-mode fields are deferred to phase 2 once their create/update flows have been
+captured from the GUI.
+"""
+
+import copy
+from typing import ClassVar, Literal
+
+from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import (
+    Field,
+    FieldSerializationInfo,
+    field_serializer,
+    field_validator,
+)
+from ansible_collections.cisco.nd.plugins.module_utils.models.base import NDBaseModel
+from ansible_collections.cisco.nd.plugins.module_utils.models.interfaces.enums import SviPolicyTypeEnum
+from ansible_collections.cisco.nd.plugins.module_utils.models.nested import NDNestedModel
+
+
+class SviPolicyModel(NDNestedModel):
+    """
+    # Summary
+
+    Policy fields for an SVI interface. Maps directly to the `configData.networkOS.policy` object in the ND API.
+
+    `policy_type` is required by the API as a discriminator on both POST and PUT, so it carries a default of
+    `SviPolicyTypeEnum.SVI` and is always serialized.
+
+    ## Raises
+
+    None
+    """
+
+    policy_type: SviPolicyTypeEnum = Field(default=SviPolicyTypeEnum.SVI, alias="policyType", description="Interface policy type")
+    admin_state: bool | None = Field(default=None, alias="adminState", description="Enable or disable the interface")
+    description: str | None = Field(default=None, alias="description", max_length=254, description="Interface description")
+    extra_config: str | None = Field(default=None, alias="extraConfig", description="Additional CLI for the interface")
+    mtu: int | None = Field(default=None, alias="mtu", ge=576, le=9216, description="Interface MTU")
+    ip: str | None = Field(default=None, alias="ip", description="IPv4 address of the SVI")
+    prefix: int | None = Field(default=None, alias="prefix", ge=1, le=31, description="IPv4 netmask length used with `ip`")
+    ipv6: str | None = Field(default=None, alias="ipv6", description="IPv6 address of the SVI")
+    v6prefix: int | None = Field(default=None, alias="v6prefix", ge=1, le=127, description="IPv6 netmask length used with `ipv6`")
+    ip_redirects: bool | None = Field(default=None, alias="ipRedirects", description="Disable both IPv4/IPv6 redirects on the interface")
+    pim_sparse: bool | None = Field(default=None, alias="pimSparse", description="Enable PIM sparse-mode on the interface")
+    pim_dr_priority: int | None = Field(default=None, alias="pimDrPriority", ge=1, le=4294967295, description="Priority for PIM DR election on the interface")
+    hsrp_group: int | None = Field(default=None, alias="hsrpGroup", description="HSRP group number")
+    hsrp_version: str | None = Field(default=None, alias="hsrpVersion", description="HSRP version")
+    preempt: bool | None = Field(default=None, alias="preempt", description="Enable HSRP preemption")
+    advertise_subnet_in_underlay: bool | None = Field(
+        default=None, alias="advertiseSubnetInUnderlay", description="Advertise the SVI subnet into the underlay routing protocol"
+    )
+    netflow: bool | None = Field(default=None, alias="netflow", description="Enable Netflow on the interface")
+
+    # --- Validators ---
+
+    @field_validator("description")
+    @classmethod
+    def description_must_be_ascii(cls, value):
+        """
+        # Summary
+
+        Reject non-ASCII characters in `description`. Some NX-OS / ND backend code paths pipe interface descriptions
+        through CLI generators that do not handle UTF-8 cleanly and return a generic 500 ("unexpected error during
+        policy execution") instead of a meaningful validation error. Catching this client-side gives the user a clear
+        message instead of a confusing server fault. Remove or relax this check once Cisco fixes the backend.
+
+        ## Raises
+
+        ### ValueError
+
+        - If `value` contains any non-ASCII character.
+        """
+        if value is None:
+            return value
+        try:
+            value.encode("ascii")
+        except UnicodeEncodeError as e:
+            raise ValueError(
+                f"description must contain only ASCII characters; got non-ASCII at position {e.start}: {value[e.start]!r}. "
+                "The ND backend currently returns a generic 500 for non-ASCII descriptions."
+            ) from None
+        return value
+
+    @field_validator("policy_type", mode="before")
+    @classmethod
+    def normalize_policy_type(cls, value):
+        """
+        # Summary
+
+        Accept `policy_type` in either Ansible (`svi`) or API (`svi`) format. They are identical for SVI so this is a
+        no-op today, but the validator is kept for symmetry with the ethernet policy models in case the API ever
+        diverges from the Ansible-side name.
+
+        ## Raises
+
+        None
+        """
+        if value is None:
+            return value
+        ansible_to_api = {e.name.lower(): e.value for e in SviPolicyTypeEnum}
+        return ansible_to_api.get(value, value)
+
+    # --- Serializers ---
+
+    @field_serializer("policy_type")
+    def serialize_policy_type(self, value: str | None, info: FieldSerializationInfo) -> str | None:
+        """
+        # Summary
+
+        Serialize `policy_type` to the API's value in payload mode, or the Ansible-friendly name in config mode. With
+        `use_enum_values=True`, the stored value is the enum's `.value` string (e.g. `"svi"`).
+
+        ## Raises
+
+        None
+        """
+        if value is None:
+            return None
+        mode = (info.context or {}).get("mode", "payload")
+        if mode == "config":
+            reverse = {e.value: e.name.lower() for e in SviPolicyTypeEnum}
+            return reverse.get(value, value)
+        return value
+
+
+class SviNetworkOSModel(NDNestedModel):
+    """
+    # Summary
+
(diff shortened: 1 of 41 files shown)


plugins/modules/nd_interface_svi.py
```diff
@@ -0,0 +1,474 @@
...
+                  netflow_monitor:
+                    description:
+                    - Layer 3 netflow monitor name.
+                    - Required when O(config[].config_data.network_os.policy.netflow=true).
+                    type: str
+                  netflow_sampler:
+                    description:
+                    - Netflow sampler name (applicable to N7K only).
+                    type: str
+  deploy:
```
mikewiebe: ### Medium: Replace top-level `deploy` with `config_actions`

**Issue**

**Evidence**

- [`plugins/modules/nd_interface_svi.py` lines 209-230](https://github.com/CiscoDevNet/ansible-nd/pull/268/files#diff-de731073b24ba71852b07e66406cadcfecd308e915581429604cfded12b2ad4dR209) documents `deploy` as a standalone option and lists states without the standard `config_actions` option.
- [`plugins/modules/nd_interface_svi.py` line 417](https://github.com/CiscoDevNet/ansible-nd/pull/268/files#diff-de731073b24ba71852b07e66406cadcfecd308e915581429604cfded12b2ad4dR417) adds `deploy=dict(type="bool", default=True)` directly in the wrapper argspec.

**Existing PR overlap**

**Impact**

**Suggested fix**

---

plugins/modules/nd_interface_svi.py
```diff
@@ -0,0 +1,474 @@
...
+    description:
+    - Whether to deploy interface changes after mutations are complete.
+    - When V(true), all queued interface changes are deployed in a single bulk API call at the end of module execution
+      via the C(interfaceActions/deploy) API. Only the interfaces modified by this task are deployed.
+    - When V(false), changes are staged but not deployed. Use a separate deploy module or task to deploy later.
+    - Setting O(deploy=false) is useful when batching changes across multiple interface tasks before a single deploy.
+    type: bool
+    default: true
+  state:
```
mikewiebe, line 229: ### Low: Track `gathered` support against issue #340 (non-blocking)

**Issue**

**Evidence**

- [`plugins/modules/nd_interface_svi.py` lines 218-230](https://github.com/CiscoDevNet/ansible-nd/pull/268/files#diff-de731073b24ba71852b07e66406cadcfecd308e915581429604cfded12b2ad4dR218) documents state choices without `gathered`.
- [`plugins/module_utils/models/interfaces/svi_interface.py` lines 316-320](https://github.com/CiscoDevNet/ansible-nd/pull/268/files#diff-44eaf9433f4623a48c4570528b419ceb2dba0daef47a1242bbb074f1a2cbc5d1R316) returns the same state choices from the model argspec.

**Existing PR overlap**

**Impact**

**Suggested fix**

---

mikewiebe, line 229: ### Medium: Add examples for every supported state

**Issue**

The module advertises `replaced` and `overridden`, but the `EXAMPLES` block only shows `merged`, `deleted`, deploy-disabled `merged`, and feature-specific merged examples.

**Evidence**

- [`plugins/modules/nd_interface_svi.py` lines 218-230](https://github.com/CiscoDevNet/ansible-nd/pull/268/files#diff-de731073b24ba71852b07e66406cadcfecd308e915581429604cfded12b2ad4dR218) lists `merged`, `replaced`, `overridden`, and `deleted`.
- [`plugins/modules/nd_interface_svi.py` lines 273-321](https://github.com/CiscoDevNet/ansible-nd/pull/268/files#diff-de731073b24ba71852b07e66406cadcfecd308e915581429604cfded12b2ad4dR273) contains examples for `merged`, `deleted`, and `deploy: false`.
- [`plugins/modules/nd_interface_svi.py` lines 343-382](https://github.com/CiscoDevNet/ansible-nd/pull/268/files#diff-de731073b24ba71852b07e66406cadcfecd308e915581429604cfded12b2ad4dR343) contains additional HSRP and DHCP relay examples, but they are also `merged`.

**Existing PR overlap**

**Impact**

Users cannot tell from the docs how destructive or broad `replaced` and `overridden` are for SVIs. Those states are where examples matter most because they define the scope of replacement and deletion.

**Suggested fix**

Add one clear `replaced` example and one clear `overridden` example. The `overridden` example should make the deletion scope obvious, especially whether it applies to all SVIs in the fabric, all on a switch, or only those matching the provided config.

plugins/modules/nd_interface_svi.py
allenrobel: Addressed in d5afacfb.

Replaced the top-level `deploy` bool with a `config_actions` dict carrying a nested `deploy` suboption, mirroring the `nd_interface_ethernet_trunk_host` migration (ec82a6f2) exactly:

- **EXAMPLES + merged integration test** — the no-deploy/batching cases now use the nested `config_actions: {deploy: false}` form.

On `save`/`type`: I intentionally omitted them here (same as the ethernet module). Verified against the ND 4.2.1 OpenAPI spec that `interfaceActions/deploy` is interface-type agnostic (body is `{interfaces:[{interfaceName,switchId}]}`), has no separate save/recalculate step, and exposes no switch|global scope parameter — a global scope would also deploy unrelated pending fabric config, which the interface resource modules deliberately avoid for safety. Happy to add `save`/`type` if you'd prefer full parity with `nd_manage_switches`.

Verification: `validate-modules` sanity clean, `black`/`isort` clean, SVI unit tests 84 passed.

plugins/modules/nd_interface_svi.py
allenrobel: Addressed in 642baee0.

Added a `replaced` example and an `overridden` example to the `EXAMPLES` block, mirroring the `nd_interface_ethernet_trunk_host` change (fc3ac974):

- **`replaced`** — reprovisions a specific SVI (`vlan333`) to a new IP/description.
- **`overridden`** — carries an inline comment making the fabric-wide scope explicit: every SVI **managed by this module** that is not listed is **deleted**, and an empty `config` list deletes **all** of them.

Note the deletion scope differs from the ethernet trunk_host example: SVIs are logical interfaces, so unlisted ones are *deleted* (matching this module's documented `overridden` behavior), whereas physical ethernet interfaces are *reset to fabric default*. The example comment reflects that.

Verification: `validate-modules` sanity clean (parses the EXAMPLES YAML), `black` clean.

allenrobel: Thanks @mikewiebe   Looks like your mention of this PR in #340 is enough to link them for tracking.  I'll go ahead and resolve this comment.  Your other two comments are addressed so feel free to resolve them when you get a minute.

plugins/modules/nd_interface_svi.py
allenrobel: Note: this branch was rebased onto the latest `develop` during the RETURN-block rollout, so the commit referenced above (`d5afacf`) is now orphaned. The same change is now `ca089eb9`. Resolving as addressed.

plugins/modules/nd_interface_svi.py
allenrobel: Note: this branch was rebased onto the latest `develop` during the RETURN-block rollout, so the commit referenced above (`642baee`) is now orphaned. The same change is now `cb3a2dfe`. Resolving as addressed.

allenrobel:
## Automated code-review pass (high effort) — outcome

Ran a multi-agent code review scoped to this branch's own commits (diff vs `nd_interface_ethernet_trunk_host`). Verified findings were triaged against the already-merged loopback/ethernet siblings.

### Fixed in this PR (commit `22cdc0fe`)

- **ip/prefix pairing** — added `_validate_ip_prefix_paired`: `ip`/`prefix` and `ipv6`/`prefixv6` are now required together, failing a partial address early instead of serializing half a pair.
- **netflow_monitor dependency** — added `_validate_netflow_monitor_present`: `netflow_monitor` is required when `netflow: true`, matching the documented dependency and the `trunk_host` `@model_validator` convention.
- **query_all guard** — tightened the per-switch guard from `if not result` to `if not isinstance(result, dict)`, matching the loopback sibling so a truthy non-dict response is skipped rather than raising `AttributeError`.
- Added unit tests for both validators; updated the prefix/prefixv6 range test to supply a paired address.

### Deferred (intentionally not changed here)

These are either shared patterns with the **already-merged** loopback/ethernet modules (fixing on one branch would diverge the siblings) or need ND-domain confirmation. Tracking them as collection-wide items rather than one-off changes:

- **Capability preflight opt-in** — SVI omits the `interface_type`/`interface_mode` ClassVars loopback sets, so it skips preflight. The base class documents opting out as legitimate; enabling it first needs confirmation that ND exposes an `svi` `capableSwitches` endpoint.
- **Read-back enforces write-side constraints** (`prefix le=31`/`le=127`, `description max_length=254`) — the same model parses both user input and ND GET responses, so a non-conforming pre-existing SVI could abort a run. This is a model-architecture pattern shared with loopback, not SVI-specific.
- **`delete()` queues remove + deploy** — byte-identical to the merged loopback `delete()`; shared behavior.
- **`policyType == "svi"` filter** — matches loopback's documented "this module manages only its own policy type" design.
- **HSRP fields not gated by `hsrp: true`** — flagged as plausible; enforcing it risks being over-strict for staged configs.
- **interface-name normalizer** — minor robustness on non-canonical tokens (e.g. `vlan 333`); low severity.

Verification: 86 unit tests pass; black/isort clean; pylint/mypy show no new issues vs the sibling baseline; `validate-modules` sanity clean.

allenrobel:
Appended commit `654d5765`: removed a stray blank line before the EXAMPLES closing delimiter that `ansible-lint` flagged (`yaml[empty-lines]`). Surfaced while reviewing the downstream `nd_interface_subinterface_managed` branch (#296) with `ansible-lint --offline`; the original run missed it because it hit a galaxy-cache error before linting. The branch was force-pushed (fast-forward append onto the existing tip). `plugins/` now passes ansible-lint at the production profile.

mikewiebe (dismissed):
LGTM

gmicol (dismissed):
LGTM

plugins/modules/nd_interface_svi.py
```diff
@@ -0,0 +1,596 @@
...
+                  vrf_dhcp3:
+                    description:
+                    - VRF used to reach DHCP server 3.
+                    - Use V(default) for the default VRF; leave blank to use the interface VRF.
+                    type: str
+                  advertise_subnet_in_underlay:
+                    description:
+                    - Advertise the SVI subnet into the underlay routing protocol.
+                    type: bool
+                  netflow:
```
akinross, line 202: do we need this boolean input? netflow_monitor appears to be required input netflow is enabled, did you consider deducing this from netflow_monitor input?

plugins/modules/nd_interface_svi.py
```diff
@@ -0,0 +1,596 @@
...
+                  pim_sparse:
+                    description:
+                    - Enable PIM sparse-mode on the interface.
+                    type: bool
+                  pim_dr_priority:
+                    description:
+                    - Priority for PIM DR election on the interface.
+                    - Valid range is 1-4294967295.
+                    type: int
+                  hsrp:
```
akinross, line 125: when HSRP is enabled are there some attributes required? UI seems to pre-populate hsrp_group ( also has a marking required ) and hsrp_version

plugins/modules/nd_interface_svi.py
```diff
@@ -0,0 +1,596 @@
...
+    suboptions:
+      switch_ip:
+        description:
+        - The management IP address of the switch on which to manage the SVI interface.
+        - This is resolved to the switch serial number (switchId) internally.
+        type: str
+        required: true
+      interface_name:
+        description:
+        - The name of the SVI interface, in the form C(vlan<id>) (e.g. C(vlan333)).
```
akinross, line 45: in https://github.com/CiscoDevNet/ansible-nd/pull/268/changes#diff-44eaf9433f4623a48c4570528b419ceb2dba0daef47a1242bbb074f1a2cbc5d1R270 it states bare integers are accepted, should we add this to the description/example?

plugins/module_utils/models/interfaces/svi_interface.py
```diff
@@ -0,0 +1,366 @@
...
+
+    switch_ip: str = Field(alias="switchIp")
+    interface_name: str = Field(alias="interfaceName")
+    interface_type: Literal["svi"] = Field(default="svi", alias="interfaceType", frozen=True)
+    config_data: SviConfigDataModel | None = Field(default=None, alias="configData")
+    oper_data: SviOperDataModel | None = Field(default=None, alias="operData")
+
+    @field_validator("interface_name", mode="before")
+    @classmethod
+    def normalize_interface_name(cls, value):
```
akinross, line 265: Did you consider exposing vlan_id as UI does instead if interface_name. This would make the validation easier to have int input where range can be checked. Then normalisation would prepend f"vlan{vlan_id}".

plugins/modules/nd_interface_svi.py
```diff
@@ -0,0 +1,596 @@
...
+                    type: str
+                  routing_tag:
+                    description:
+                    - Routing tag associated with the interface IP address.
+                    type: str
+                  pim_sparse:
+                    description:
+                    - Enable PIM sparse-mode on the interface.
+                    type: bool
+                  pim_dr_priority:
```
akinross, line 119: Ui claims default of 1, is api handling this default? should this behaviour be documented?

akinross (changes requested):

plugins/modules/nd_interface_svi.py
allenrobel: Agreed — documented in 5a34e489. The `interface_name` description now notes that bare VLAN IDs (e.g. `333`) and mixed-case prefixes (`Vlan333`, `VLAN333`) are accepted and normalized to the `vlan<id>` form, matching `normalize_interface_name`.

plugins/modules/nd_interface_svi.py
allenrobel: Good question. In the `int_vlan.template` (`policyType: svi`) schema none of the HSRP sub-options are in `required`; `hsrpGroup` and `hsrpVersion` carry schema defaults of `1`, so the GUI pre-populating them is just surfacing those API defaults rather than enforcing a hard requirement. So no model-level `required` constraint is warranted — but the docs were silent on it. Documented in 47642d7e: clarified that no HSRP sub-options are strictly required and that `hsrp_group`/`hsrp_version` default to `1` on the controller.

plugins/modules/nd_interface_svi.py
allenrobel: Yes — the controller handles the default: `pimDrPriority` has a schema default of `1` in `int_vlan.template`, and the module omits the field from the payload when unset, so ND applies `1`. Documented in a5654e36.

plugins/module_utils/models/interfaces/svi_interface.py
allenrobel: I considered it, but I would prefer to keep `interface_name` as the exposed field since it is the actual ND API identifier key — the module uses `(switch_ip, interface_name)` as the composite identifier, mirroring the wire shape, so swapping to `vlan_id` would diverge from that and be a breaking interface change. Your validation point is well taken though: I added range validation to `normalize_interface_name` in e3e8f064. When a numeric VLAN ID is extractable (bare int, digit string, or `vlan`-prefixed), it is now range-checked to 1-4094 with a clear error, plus boundary and out-of-range tests. That captures the int-range-check benefit without changing the public interface.

plugins/modules/nd_interface_svi.py
allenrobel: Thanks @akinross. I considered deducing it, but kept the explicit boolean to mirror the API/UI contract: in the `int_vlan.template` (`policyType: svi`) schema `netflow` is a standalone boolean (`default: false`) and `netflowMonitor` is a separate, API-optional string — they are independent fields on the wire. The "monitor required when netflow=true" rule is a module-side convenience validator, not an API constraint. Keeping the boolean also lets users express `netflow: false` explicitly (useful for `replaced`/`overridden` idempotency) instead of inferring intent from monitor presence. Happy to revisit if you feel strongly.

mikewiebe (approved):

plugins/modules/nd_interface_svi.py
```diff
@@ -0,0 +1,601 @@
...
+
+        module.exit_json(**nd_state_machine.output.format())
+
+    except NDStateMachineError as e:
+        module_log.exception("NDStateMachineError during module execution")
+        output = nd_state_machine.output.format() if nd_state_machine else {}
+        error_msg = f"Module execution failed: {str(e)}"
+        if module.params.get("output_level") == "debug":
+            error_msg += f"\nTraceback:\n{traceback.format_exc()}"
+        module.fail_json(msg=error_msg, **output)
```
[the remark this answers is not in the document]
allenrobel: Opened issue #379 to cover this.

plugins/module_utils/models/interfaces/svi_interface.py
```diff
@@ -0,0 +1,382 @@
...
+    )
+    netflow: bool | None = Field(default=None, alias="netflow", description="Enable Netflow on the interface")
+    netflow_monitor: str | None = Field(default=None, alias="netflowMonitor", description="Layer 3 Netflow monitor name (required when `netflow=true`)")
+    netflow_sampler: str | None = Field(default=None, alias="netflowSampler", description="Netflow sampler name (applicable to N7K only)")
+
+    # --- Validators ---
+
+    # TODO(4.2.1) GET returns `routingTag` as an integer though the OpenAPI spec declares it as string and POST/PUT accept string.
+    # Remove this validator once ND 4.2 reaches end-of-support. Cisco has confirmed this is fixed in ND 4.3, so once 4.2 is
+    # deprecated this coercion is dead weight. Tracked in `project_svi_hsrp_phase2.md`.
```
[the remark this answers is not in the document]
allenrobel: Opened issue #380 to track

tests/integration/targets/nd_interface_ethernet_trunk_host/tasks/setup.yaml
```diff
@@ -0,0 +1,47 @@
...
+          network_os:
+            policy:
+    state: merged
+  check_mode: true
+  register: debug_requery
+
+- name: "DEBUG: Show re-query result (check before list)"
+  ansible.builtin.debug:
+    msg: "Before count: {{ debug_requery.before | length }}, Eth1/41 in before: {{ debug_requery.before | selectattr('switch_ip', 'equalto', test_switch_ip) | selectattr('interface_name', 'equalto', 'Ethernet1/41') |
+  list | length }}"
```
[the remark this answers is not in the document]
allenrobel: Not a valid issue.  Closing.

PR #269, opened 2026-05-02
nd_interface_port_channel_access

allenrobel:
## Related Issue(s)

- CiscoDevNet/ansible-nd#320

- #348 — interface orchestrator CRUD + policy-type serializer dedup (cross-cutting)
- #349 — query_all per-switch fan-out / no fabric-wide interface endpoint (cross-cutting)
- #350 — create accepts a policy-less config item (config_data optional, no create-time validator)
- #351 — storm-control percentage vs pps not mutually exclusive
- #295 — HTTP 207 multi-status silent success (shared RestSend infra; 207 guard deferred on this branch)

## Proposed Changes

Adds `nd_interface_port_channel_access` for managing accessPoHost port-channel interfaces on ND 4.2 fabrics. The PC policy is the single source of truth for members (the `ports` list); ND propagates access-mode configuration to member interfaces — users do not pre-configure members.

- `PortChannelAccessInterfaceModel` — Pydantic policy/payload model targeting `accessPoHost`, composite identifier `(switch_ip, interface_name)`, all access-mode policy fields plus port-channel specifics (`ports`, `copy_description`, `port_channel_mode`, `lacp_rate`, `lacp_port_priority`, `lacp_suspend`).
- `PortChannelBaseOrchestrator` — shared CRUD base extending `NDBaseInterfaceOrchestrator`. Uses `EpManageInterfacesRemove` for delete (PCs are virtual/deletable like loopbacks). Filters `query_all` by `interfaceType="portChannel"` and per-subclass policy-type set.
- `PortChannelAccessInterfaceOrchestrator` — concrete subclass binding `model_class` and the `accessPoHost` policy-type filter.
- `plugins/module_utils/models/interfaces/enums.py` — first introduction on this branch; contains `AccessPoHostPolicyTypeEnum`, `PortChannelModeEnum`, `LacpRateEnum`, plus access-mode enums that will be shared with future trunk-host work (`BpduFilter`, `BpduGuard`, `DuplexMode`, `Mtu`, `Speed`, `StormControlAction`).

Branch base: stacked on `nd_interface_loopback` for the base interface infrastructure (`NDBaseInterfaceOrchestrator`, `FabricContext`, `manage_interfaces` endpoints, base interface models).

### Unit tests

Adapted from the ethernet_trunk_host unit-test trio:

- `tests/unit/module_utils/models/test_port_channel_access_interface.py` — 114 cases covering policy/network_os/config_data/interface model defaults, alias population, `normalize_policy_type`, `normalize_ports`, range/enum validation, `AsciiDescription`, serializer modes, `to_payload` / `to_config` / `from_response` / `from_config` round-trips, composite identifier, `get_diff`, `merge`, and `get_argument_spec`.
- `tests/unit/module_utils/orchestrators/test_port_channel_access_interface.py` — 8 cases covering `model_class`, bulk-support flags, `_managed_policy_types == {"accessPoHost"}`, and `PortChannelBaseOrchestrator.query_all` filtering by `interfaceType` and `policyType`, plus fabric-not-found and per-switch 404 paths.
- `tests/unit/module_utils/fixtures/fixture_data/test_port_channel_access_interface.json` — fabric summary, deploymentFreeze, switches, and per-switch interfaces fixtures for the orchestrator tests.

To run:

```
python -m pytest tests/unit/module_utils/models/test_port_channel_access_interface.py tests/unit/module_utils/orchestrators/test_port_channel_access_interface.py
```

122/122 pass. Full unit suite (558 tests) green; black/isort clean.

### Integration tests

Cover: create / multi-create across switches, idempotency, update access_vlan and description, add member to PC, remove member from PC, `state: replaced`, `state: overridden` (incl. interfaceType filter check), `state: deleted` with member-revert path, `deploy: false` staging, and non-existent delete no-op.

Tests passed end-to-end on a 9000v testbed against ND 4.2 (192.168.7.8 / fabric_1 / leaf at 192.168.12.151).

The test target uses `module_defaults: { cisco.nd.nd_interface_port_channel_access: { timeout: 300 } }` to scope a 300s timeout to PC tests — port-channel deploys (build bundle, pull members, apply LACP) routinely exceed the default 30s on slower testbeds. Scoped at the test target level so we don't touch shared `nd_argument_spec` defaults.

To run:

```
ansible-test network-integration nd_interface_port_channel_access -v
```

Required `[nd:vars]` entries (in `tests/integration/inventory.networking`):

```
nd_test_fabric_name=<fabric>
nd_test_switch_ip=<switch IP>
nd_test_pc_member_a=Ethernet1/10
nd_test_pc_member_b=Ethernet1/11
nd_test_pc_member_c=Ethernet1/12
nd_test_pc_member_d=Ethernet1/13
```

The four member ethernets must exist on the target switch and not currently belong to any other port-channel (the SETUP cleanup at the start of each state file resets them).

4.2

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

* [x] Latest commit is rebased from base branch with merge conflicts resolved

### DIFF
--- a/plugins/module_utils/models/interfaces/port_channel_access_interface.py
+++ b/plugins/module_utils/models/interfaces/port_channel_access_interface.py
@@ -30,6 +30,7 @@ interfaces inherit access-mode settings from the port-channel; users do not pre-
 
 from __future__ import annotations
 
+import re
 from typing import ClassVar, Literal
 
 from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import (
@@ -52,6 +53,14 @@ from ansible_collections.cisco.nd.plugins.module_utils.models.interfaces.enums i
 )
 from ansible_collections.cisco.nd.plugins.module_utils.models.nested import NDNestedModel
 from ansible_collections.cisco.nd.plugins.module_utils.models.types import AsciiDescription
+
+# Splits an interface name into its leading alphabetic prefix and the rest (digits/separators).
+_INTERFACE_NAME_PREFIX_RE = re.compile(r"^([A-Za-z]+)(.*)$")
+
+# Member interfaces of a port-channel are always physical ethernet ports, so the canonical wire prefix
+# is always "Ethernet". Any case-insensitive NX-OS abbreviation (e.g. "e", "eth", "ether") expands to the
+# full form so user input matches the wire key and idempotency holds.
+_CANONICAL_MEMBER_TYPE = "Ethernet"
 
 
 class PortChannelAccessPolicyModel(NDNestedModel):
@@ -151,23 +160,50 @@ class PortChannelAccessPolicyModel(NDNestedModel):
         """
         # Summary
 
-        Normalize each member interface name to ND API convention (e.g. `ethernet1/1` -> `Ethernet1/1`).
+        Normalize each member interface name to ND's canonical `Ethernet` form so that any user-supplied casing or
+        NX-OS abbreviation round-trips against the wire form. Members are always physical ethernet ports. Examples:
+
+        - `ethernet1/1` -> `Ethernet1/1`
+        - `ETHERNET1/1` -> `Ethernet1/1`
+        - `eth1/1` -> `Ethernet1/1` (abbreviation expanded)
+        - `e1/1` -> `Ethernet1/1` (abbreviation expanded)
+        - `Ethernet1/1` -> `Ethernet1/1` (idempotent)
+
+        Because the wire key is matched exactly, an abbreviated prefix that is not expanded would never match ND's
+        `Ethernet...` form, silently breaking idempotency (the port-channel re-deploys on every run). Any
+        case-insensitive prefix of `Ethernet` (`e`, `et`, `eth`, ...) is therefore expanded to the full canonical
+        name; an unrecognized prefix falls back to Title case so it still round-trips. Only the leading alphabetic
+        run is rewritten; digits and separators are preserved verbatim.
 
         ## Raises
 
         None
         """
-        if value is None:
-            return value
         if not isinstance(value, list):
             return value
-        normalized = []
-        for name in value:
-            if isinstance(name, str) and name:
-                normalized.append(name[0].upper() + name[1:])
-            else:
-                normalized.append(name)
-        return normalized
+        return [cls._normalize_member_name(name) for name in value]
+
+    @staticmethod
+    def _normalize_member_name(name):
+        """
+        # Summary
+
+        Normalize a single member interface name to ND's canonical `Ethernet` form (see `normalize_ports`).
+        Non-string or empty values are returned unchanged.
+
+        ## Raises
+
+        None
+        """
+        if not isinstance(name, str) or not name:
+            return name
+        match = _INTERFACE_NAME_PREFIX_RE.match(name)
+        if not match:
+            return name
+        prefix, rest = match.groups()
+        if _CANONICAL_MEMBER_TYPE.lower().startswith(prefix.lower()):
+            return _CANONICAL_MEMBER_TYPE + rest
+        return prefix[0].upper() + prefix[1:].lower() + rest
 
     @model_serializer(mode="wrap")
     def _strip_policy_type_in_config(self, handler, info: SerializationInfo):
--- a/plugins/module_utils/models/interfaces/port_channel_access_interface.py
+++ b/plugins/module_utils/models/interfaces/port_channel_access_interface.py
@@ -1,0 +1,379 @@
+# Copyright: (c) 2026, Allen Robel (@allenrobel)
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+"""
+Port-channel access (accessPoHost) interface Pydantic models for Nexus Dashboard.
+
+This module defines nested Pydantic models that mirror the ND Manage Interfaces API payload
+structure for port-channel accessPoHost interfaces. The playbook config uses the same nesting
+so that `to_payload()` and `from_response()` work via standard Pydantic serialization with no
+custom wrapping or flattening.
+
+The port-channel policy is the single source of truth for member configuration. Member ethernet
+interfaces inherit access-mode settings from the port-channel; users do not pre-configure members.
+
+## Model Hierarchy
+
+- `PortChannelAccessInterfaceModel` (top-level, `NDBaseModel`)
+    - `switch_ip` (composite identifier)
+    - `interface_name` (composite identifier; e.g. `port-channel501`)
+    - `interface_type` (default: "portChannel")
+    - `config_data` -> `PortChannelAccessConfigDataModel`
+        - `mode` (default: "access")
+        - `network_os` -> `PortChannelAccessNetworkOSModel`
+            - `network_os_type` (default: "nx-os")
+            - `policy` -> `PortChannelAccessPolicyModel`
+                - `admin_state`, `access_vlan`, `ports`, `port_channel_mode`,
+                  `lacp_rate`, `bpdu_guard`, `description`, `policy_type`, etc.
+"""
+
+from __future__ import annotations
+
+from typing import ClassVar, Literal
+
+from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import (
+    Field,
+    FieldSerializationInfo,
+    field_serializer,
+    field_validator,
+)
+from ansible_collections.cisco.nd.plugins.module_utils.models.base import NDBaseModel
+from ansible_collections.cisco.nd.plugins.module_utils.models.interfaces.enums import (
+    AccessPoHostPolicyTypeEnum,
+    BpduFilterEnum,
+    BpduGuardEnum,
+    DuplexModeEnum,
+    LacpRateEnum,
+    MtuEnum,
+    PortChannelModeEnum,
+    SpeedEnum,
+    StormControlActionEnum,
+)
+from ansible_collections.cisco.nd.plugins.module_utils.models.nested import NDNestedModel
+from ansible_collections.cisco.nd.plugins.module_utils.models.types import AsciiDescription
+
+
+class PortChannelAccessPolicyModel(NDNestedModel):
+    """
+    # Summary
+
+    Policy fields for a port-channel accessPoHost interface. Maps directly to the `configData.networkOS.policy`
+    object in the ND API.
+
+    The `ports` field carries the list of member interface names (e.g. `Ethernet1/1`). Member interfaces inherit
+    access-mode configuration from this policy; modifying a member's standalone configuration while it is a
+    port-channel member is restricted by the ethernet orchestrators.
+
+    ## Raises
+
+    None
+    """
+
+    admin_state: bool | None = Field(default=None, alias="adminState", description="Enable or disable the interface")
+    access_vlan: int | None = Field(default=None, alias="accessVlan", ge=1, le=4094, description="VLAN for this access port-channel")
+    bpdu_filter: BpduFilterEnum | None = Field(default=None, alias="bpduFilter", description="Configure spanning-tree BPDU filter")
+    bpdu_guard: BpduGuardEnum | None = Field(default=None, alias="bpduGuard", description="Enable spanning-tree BPDU guard")
+    cdp: bool | None = Field(default=None, alias="cdp", description="Enable CDP on the interface")
+    copy_description: bool | None = Field(
+        default=None, alias="copyDescription", description="Propagate the port-channel description to all member interfaces"
+    )
+    description: AsciiDescription = Field(default=None, alias="description", max_length=254, description="Interface description")
+    duplex_mode: DuplexModeEnum | None = Field(default=None, alias="duplexMode", description="Port duplex mode")
+    extra_config: str | None = Field(default=None, alias="extraConfig", description="Additional CLI for the interface")
+    lacp_port_priority: int | None = Field(
+        default=None, alias="lacpPortPriority", ge=1, le=65535, description="LACP port priority (1-65535, default 32768)"
+    )
+    lacp_rate: LacpRateEnum | None = Field(default=None, alias="lacpRate", description="LACP rate (normal=30s, fast=1s)")
+    lacp_suspend: bool | None = Field(default=None, alias="lacpSuspend", description="Suspend port if LACP PDUs not received")
+    monitor: bool | None = Field(default=None, alias="monitor", description="Enable switchport monitor for SPAN/ERSPAN")
+    mtu: MtuEnum | None = Field(default=None, alias="mtu", description="Interface MTU")
+    netflow: bool | None = Field(default=None, alias="netflow", description="Enable Netflow on the interface")
+    netflow_monitor: str | None = Field(default=None, alias="netflowMonitor", description="Layer 2 Netflow monitor name")
+    netflow_sampler: str | None = Field(default=None, alias="netflowSampler", description="Netflow sampler name")
+    policy_type: AccessPoHostPolicyTypeEnum | None = Field(default=None, alias="policyType", description="Interface policy type")
+    port_channel_mode: PortChannelModeEnum | None = Field(default=None, alias="portChannelMode", description="Port-channel mode (on/active/passive)")
+    ports: list[str] | None = Field(default=None, alias="ports", description="Member interface names (e.g. ['Ethernet1/1', 'Ethernet1/2'])")
+    qos: bool | None = Field(default=None, alias="qos", description="Enable QoS configuration for this interface")
+    qos_policy: str | None = Field(default=None, alias="qosPolicy", description="Custom QoS policy name")
+    queuing_policy: str | None = Field(default=None, alias="queuingPolicy", description="Custom queuing policy name")
+    speed: SpeedEnum | None = Field(default=None, alias="speed", description="Interface speed")
+    storm_control: bool | None = Field(default=None, alias="stormControl", description="Enable traffic storm control")
+    storm_control_action: StormControlActionEnum | None = Field(
+        default=None, alias="stormControlAction", description="Storm control action on threshold violation"
+    )
+    storm_control_broadcast_level: float | None = Field(
+        default=None,
+        alias="stormControlBroadcastLevel",
+        ge=0.0,
+        le=100.0,
+        description="Broadcast storm control level in percentage (0.00-100.00)",
+    )
+    storm_control_broadcast_level_pps: int | None = Field(
+        default=None,
+        alias="stormControlBroadcastLevelPps",
+        ge=0,
+        le=200000000,
+        description="Broadcast storm control level in packets per second",
+    )
+    storm_control_multicast_level: float | None = Field(
+        default=None,
+        alias="stormControlMulticastLevel",
+        ge=0.0,
+        le=100.0,
+        description="Multicast storm control level in percentage (0.00-100.00)",
+    )
+    storm_control_multicast_level_pps: int | None = Field(
+        default=None,
+        alias="stormControlMulticastLevelPps",
+        ge=0,
+        le=200000000,
+        description="Multicast storm control level in packets per second",
+    )
+    storm_control_unicast_level: float | None = Field(
+        default=None,
+        alias="stormControlUnicastLevel",
+        ge=0.0,
+        le=100.0,
+        description="Unicast storm control level in percentage (0.00-100.00)",
+    )
+    storm_control_unicast_level_pps: int | None = Field(
+        default=None,
+        alias="stormControlUnicastLevelPps",
+        ge=0,
+        le=200000000,
+        description="Unicast storm control level in packets per second",
+    )
+
+    # --- Validators ---
+
+    @field_validator("policy_type", mode="before")
+    @classmethod
+    def normalize_policy_type(cls, value):
+        """
+        # Summary
+
+        Accept `policy_type` in either Ansible (`access_po_host`) or API (`accessPoHost`) format, normalizing to the
+        API value for enum validation.
+
+        ## Raises
+
+        None
+        """
+        if value is None:
+            return value
+        ansible_to_api = {e.name.lower(): e.value for e in AccessPoHostPolicyTypeEnum}
+        return ansible_to_api.get(value, value)
+
+    @field_validator("ports", mode="before")
+    @classmethod
+    def normalize_ports(cls, value):
+        """
+        # Summary
+
+        Normalize each member interface name to ND API convention (e.g. `ethernet1/1` -> `Ethernet1/1`).
+
+        ## Raises
+
+        None
+        """
+        if value is None:
+            return value
+        if not isinstance(value, list):
+            return value
+        normalized = []
+        for name in value:
+            if isinstance(name, str) and name:
+                normalized.append(name[0].upper() + name[1:])
+            else:
+                normalized.append(name)
+        return normalized
+
+    # --- Serializers ---
+
+    @field_serializer("policy_type")
+    def serialize_policy_type(self, value: str | None, info: FieldSerializationInfo) -> str | None:
+        """
+        # Summary
+
+        Serialize `policy_type` to the API's camelCase value in payload mode, or the Ansible-friendly name in config mode.
+
+        With `use_enum_values=True`, the stored value is the enum's `.value` string (e.g. `"accessPoHost"`).
+
+        ## Raises
+
+        None
+        """
+        if value is None:
+            return None
+        mode = (info.context or {}).get("mode", "payload")
+        if mode == "config":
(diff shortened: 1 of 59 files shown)


plugins/modules/nd_interface_port_channel_access.py
```diff
@@ -0,0 +1,384 @@
...
+
+    except NDStateMachineError as e:
+        module_log.exception("NDStateMachineError during module execution")
+        output = nd_state_machine.output.format() if nd_state_machine else {}
+        error_msg = f"Module execution failed: {str(e)}"
+        if module.params.get("output_level") == "debug":
+            error_msg += f"\nTraceback:\n{traceback.format_exc()}"
+        module.fail_json(msg=error_msg, **output)
+
+
```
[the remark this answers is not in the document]
allenrobel: Addressed in f3cd3dc. Added a broad `except Exception` handler to `main()` matching the `nd_interface_loopback` reference pattern, so the docstring contract ("catches all exceptions and calls `module.fail_json`") now holds and unexpected exceptions produce structured Ansible output instead of a raw traceback.

plugins/modules/nd_interface_port_channel_access.py
```diff
@@ -0,0 +1,384 @@
...
+module: nd_interface_port_channel_access
+version_added: "1.4.0"
+short_description: Manage port-channel accessPoHost interfaces on Cisco Nexus Dashboard
+description:
+- Manage port-channel accessPoHost interfaces on Cisco Nexus Dashboard.
+- It supports creating, updating, querying, and deleting accessPoHost port-channel configurations on switches within a fabric.
+- Each config item represents one port-channel interface. Member ethernet interfaces are listed in
+  O(config[].config_data.network_os.policy.ports) and inherit access-mode configuration from the port-channel policy.
+- Member interface field mutability is restricted while members of a port-channel; only description, admin_state, and
+  extra_config can be modified on members via the C(nd_interface_ethernet_access) module.
```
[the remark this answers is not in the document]
allenrobel: This module is in a different branch so the comment will be moot one all branches are merged into develop.

tests/unit/module_utils/fixtures/fixture_data/test_port_channel_access_interface.json
```diff
@@ -0,0 +1,270 @@
+{
+    "TEST_NOTES": [
+        "Fixture data for tests/unit/module_utils/orchestrators/test_port_channel_access_interface.py",
+        "Each key matches a test function + suffix (a/b/c/...) per CLAUDE.md unit-test conventions.",
+        "The fixtures simulate ND responses for:",
+        "  - GET /api/v1/manage/fabrics/{fabric_name}/summary  (fabric_summary via FabricContext)",
+        "  - GET /api/v1/manage/fabrics/{fabric_name}/deploymentFreeze (deploymentFreeze via FabricContext)",
+        "  - GET /api/v1/manage/fabrics/{fabric_name}/switches (switch_map via FabricContext)",
+        "  - GET /api/v1/manage/fabrics/{fabric_name}/switches/{sn}/interfaces (EpManageInterfacesListGet per switch)"
```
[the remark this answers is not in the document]
allenrobel: Good catch — fixed in 363dcb8.

You're right: `FabricContext.fabric_is_deployment_frozen()` reads the `fabricStatus` field from the cached `/summary` response (`fabric_context.py:169`), not a separate `/deploymentFreeze` endpoint. The note was misleading and the three `*_freeze` fixture entries (`00400_freeze`, `00410_freeze`, `00430_freeze`) were never yielded by any test in `test_port_channel_access_interface.py` — confirmed dead.

- Rewrote the `/summary` TEST_NOTES line to state that freeze status comes from the `fabricStatus` field of that response, with no `/deploymentFreeze` call.
- Removed the three unused `*_freeze` entries.

All 8 orchestrator tests still pass.

plugins/modules/nd_interface_port_channel_access.py
```diff
@@ -0,0 +1,392 @@
...
+                    type: int
+                  storm_control_unicast_level:
+                    description:
+                    - Unicast storm control level in percentage (0.00-100.00).
+                    type: float
+                  storm_control_unicast_level_pps:
+                    description:
+                    - Unicast storm control level in packets per second (0-200000000).
+                    type: int
+  deploy:
```
mikewiebe: ### Medium: Replace top-level `deploy` with `config_actions`

**Issue**

**Evidence**

- [`plugins/modules/nd_interface_port_channel_access.py` lines 207-229](https://github.com/CiscoDevNet/ansible-nd/pull/269/files#diff-e005d60b69237b2686dd7ff0a46429a31e70c9fde798ae7bd213452bd90be5e4R207) documents `deploy` as a standalone option and lists states without the standard `config_actions` option.

**Existing PR overlap**

**Impact**

**Suggested fix**

---

plugins/modules/nd_interface_port_channel_access.py
```diff
@@ -0,0 +1,392 @@
...
+    description:
+    - Whether to deploy port-channel changes after mutations are complete.
+    - When V(true), all queued port-channel changes are deployed in a single bulk API call at the end of module
+      execution via the C(interfaceActions/deploy) API. Only the port-channels modified by this task are deployed.
+    - When V(false), changes are staged but not deployed. Use a separate deploy module or task to deploy later.
+    - Setting O(deploy=false) is useful when batching changes across multiple interface tasks before a single deploy.
+    type: bool
+    default: true
+  state:
```
mikewiebe, line 221: ### Low: Track `gathered` support against issue #340 (non-blocking)

**Issue**

**Evidence**

- [`plugins/modules/nd_interface_port_channel_access.py` lines 216-229](https://github.com/CiscoDevNet/ansible-nd/pull/269/files#diff-e005d60b69237b2686dd7ff0a46429a31e70c9fde798ae7bd213452bd90be5e4R216) documents state choices without `gathered`.
- [`plugins/module_utils/models/interfaces/port_channel_access_interface.py` lines 358-362](https://github.com/CiscoDevNet/ansible-nd/pull/269/files#diff-69ba8563713299543a337708ba4cd27c127e05e8166838910c04c87f0bf5fff4R358) returns the same state choices from the model argspec.

**Existing PR overlap**

**Impact**

**Suggested fix**

---

mikewiebe, line 221: ### Medium: Add examples for every supported state

**Issue**

The module advertises `replaced` and `overridden`, but the `EXAMPLES` block only shows `merged`, `deleted`, and deploy-disabled `merged` usage.

**Evidence**

- [`plugins/modules/nd_interface_port_channel_access.py` lines 216-229](https://github.com/CiscoDevNet/ansible-nd/pull/269/files#diff-e005d60b69237b2686dd7ff0a46429a31e70c9fde798ae7bd213452bd90be5e4R216) lists `merged`, `replaced`, `overridden`, and `deleted`.
- [`plugins/modules/nd_interface_port_channel_access.py` lines 258-298](https://github.com/CiscoDevNet/ansible-nd/pull/269/files#diff-e005d60b69237b2686dd7ff0a46429a31e70c9fde798ae7bd213452bd90be5e4R258) contains examples for `merged`, `deleted`, and `deploy: false`, but no `replaced` or `overridden` example.

**Existing PR overlap**

**Impact**

Users cannot tell from the docs how destructive or broad `replaced` and `overridden` are for port-channel access interfaces. Those states are where examples matter most because they define the scope of replacement and deletion.

**Suggested fix**

Add one clear `replaced` example and one clear `overridden` example. The `overridden` example should make the deletion scope obvious, especially whether it applies to all port-channel access interfaces in the fabric, all on a switch, or only those matching the provided config.

plugins/modules/nd_interface_port_channel_access.py
allenrobel: Implemented in c0fa91c4. Replaced the top-level `deploy` boolean with a `config_actions` dict.

Per our follow-up discussion, this module exposes only `config_actions.deploy` (deploy-only container). The interface deploy API (`interfaceActions/deploy`, verified against the ND 4.2.1 OpenAPI) has no separate save step and no switch/global scope — it takes only an explicit `{interfaceName, switchId}` list — so `save`/`type` are intentionally omitted here rather than shipped as no-ops. The deploy-disabled example and the no-deploy integration task were updated to the nested `config_actions: {deploy: false}` shape.

plugins/modules/nd_interface_port_channel_access.py
allenrobel: Added in c0fa91c4. EXAMPLES now includes a `state=replaced` example and a `state=overridden` example.

The overridden example carries a WARNING comment making the deletion scope explicit: it is fabric-wide — every `accessPoHost` port-channel on any switch in the fabric that is not present in `config` is deleted (member interfaces revert to their fabric default), and the listed port-channels are reconciled to exactly match.

allenrobel: Acknowledged.  This is tracked in #430 with a link to this PR.

plugins/modules/nd_interface_port_channel_access.py
allenrobel: Note: this branch was rebased onto the latest `develop` during the RETURN-block rollout, so the commit referenced above (`c0fa91c`) is now orphaned. The same change is now `0c0f6cab`. Resolving as addressed.

plugins/modules/nd_interface_port_channel_access.py
allenrobel: Note: this branch was rebased onto the latest `develop` during the RETURN-block rollout, so the commit referenced above (`c0fa91c`) is now orphaned. The same change is now `0c0f6cab`. Resolving as addressed.

mikewiebe (approved):
LGTM

gmicol (approved):
LGTM

plugins/modules/nd_interface_port_channel_access.py
```diff
@@ -0,0 +1,530 @@
...
+ANSIBLE_METADATA = {"metadata_version": "1.1", "status": ["preview"], "supported_by": "community"}
+
+DOCUMENTATION = r"""
+---
+module: nd_interface_port_channel_access
+version_added: "2.0.0"
+short_description: Manage port-channel accessPoHost interfaces on Cisco Nexus Dashboard
+description:
+- Manage port-channel accessPoHost interfaces on Cisco Nexus Dashboard.
+- It supports creating, updating, querying, and deleting accessPoHost port-channel configurations on switches within a fabric.
```
akinross, line 18:
```suggestion
- It supports creating, updating, and deleting (accessPoHost) port-channel configurations on switches within a fabric.
```

plugins/module_utils/models/interfaces/port_channel_access_interface.py
```diff
@@ -0,0 +1,399 @@
...
+    identifier_strategy: ClassVar[Literal["single", "composite", "hierarchical", "singleton"] | None] = "composite"
+
+    # --- Serialization Configuration ---
+
+    payload_exclude_fields: ClassVar[set[str]] = {"switch_ip"}
+
+    # --- Fields ---
+
+    switch_ip: str = Field(alias="switchIp")
+    interface_name: str = Field(alias="interfaceName")
```
akinross, line 303: did you consider exposing port channel id instead and adding validation of Port Channel Id should be in the range of 1-4096?

plugins/modules/nd_interface_port_channel_access.py
```diff
@@ -0,0 +1,530 @@
...
+                    type: str
+                  netflow_sampler:
+                    description:
+                    - The netflow Layer-2 sampler name for the port-channel.
+                    type: str
+                  port_channel_mode:
+                    description:
+                    - The port-channel mode.
+                    type: str
+                    choices: [ 'on', active, passive ]
```
akinross, line 148: is on quoted because of ansible doc restrictions?

plugins/modules/nd_interface_port_channel_access.py
```diff
@@ -0,0 +1,530 @@
...
+                    type: bool
+                  netflow_monitor:
+                    description:
+                    - The netflow Layer-2 monitor name for the port-channel.
+                    type: str
+                  netflow_sampler:
+                    description:
+                    - The netflow Layer-2 sampler name for the port-channel.
+                    type: str
+                  port_channel_mode:
```
akinross, line 144: in UI this is required input for access

plugins/modules/nd_interface_port_channel_access.py
```diff
@@ -0,0 +1,530 @@
...
+                    description:
+                    - The port-channel mode.
+                    type: str
+                    choices: [ 'on', active, passive ]
+                  ports:
```
akinross, line 149: did you consider ranged input like UI exposes eth1-10, or does user input 10 interfaces in this case?

plugins/modules/nd_interface_port_channel_access.py
```diff
@@ -0,0 +1,530 @@
...
+                    description:
+                    - The access VLAN for the port-channel.
+                    - Valid range is 1-4094.
+                    type: int
+                  bpdu_filter:
+                    description:
+                    - BPDU filter setting for the port-channel.
+                    type: str
+                    choices: [ enable, disable, default ]
+                  bpdu_guard:
```
akinross, line 81: this seems required in UI

plugins/modules/nd_interface_port_channel_access.py
```diff
@@ -0,0 +1,530 @@
...
+                    choices: [ normal, fast ]
+                  lacp_suspend:
+                    description:
+                    - Whether to suspend the port if LACP PDUs are not received.
+                    type: bool
+                  monitor:
+                    description:
+                    - Whether the port-channel is configured as a SPAN/ERSPAN monitor source.
+                    type: bool
+                  mtu:
```
akinross, line 127: this seems required in UI

plugins/modules/nd_interface_port_channel_access.py
```diff
@@ -0,0 +1,530 @@
...
+                  admin_state:
+                    description:
+                    - The administrative state of the port-channel.
+                    type: bool
+                  access_vlan:
+                    description:
+                    - The access VLAN for the port-channel.
+                    - Valid range is 1-4094.
+                    type: int
+                  bpdu_filter:
```
akinross, line 76: UI defaults to default, is this actual API default and should we document this behaviour?

plugins/modules/nd_interface_port_channel_access.py
```diff
@@ -0,0 +1,530 @@
...
+        - The management IP address of the switch on which to manage the port-channel.
+        - This is resolved to the switch serial number (switchId) internally.
+        type: str
+        required: true
+      interface_name:
+        description:
+        - The port-channel interface name (e.g. C(port-channel501)).
+        type: str
+        required: true
+      config_data:
```
akinross, line 52: in UI some part are exposed which i do not see in your module (checked on iBGP fabric leaf device, might be different per fabric/device type):

- Spanning-tree Link-type
- Enable Port Type Fast
- Enable Auto-Negotiation
- Enable vPC Orphan Port
- Bandwidth in kilobits
- Inherit Bandwidth in kilobits
- Enable priority flow control

How do we currently verify that we expose all the options in our modules? Do we do manual UI verification? Especially when we have to deal with multiple types of fabric / device / interface combinations.

plugins/modules/nd_interface_port_channel_access.py
```diff
@@ -0,0 +1,530 @@
...
+                    type: bool
+                  qos_policy:
+                    description:
+                    - Custom QoS policy name associated with the port-channel.
+                    type: str
+                  queuing_policy:
+                    description:
+                    - Custom queuing policy name associated with the port-channel.
+                    type: str
+                  speed:
```
akinross, line 169: UI defaults to auto, is this actual API default and should we document this behaviour?

plugins/modules/nd_interface_port_channel_access.py
```diff
@@ -0,0 +1,530 @@
...
+                    description:
+                    - BPDU filter setting for the port-channel.
+                    type: str
+                    choices: [ enable, disable, default ]
+                  bpdu_guard:
+                    description:
+                    - BPDU guard setting for the port-channel.
+                    type: str
+                    choices: [ enable, disable, default ]
+                  cdp:
```
akinross, line 86: UI defaults to true, is this actual API default and should we document this behaviour?

plugins/modules/nd_interface_port_channel_access.py
```diff
@@ -0,0 +1,530 @@
...
+                  mtu:
+                    description:
+                    - The MTU setting for the port-channel.
+                    type: str
+                    choices: [ default, jumbo ]
+                  netflow:
+                    description:
+                    - Whether netflow is enabled on the port-channel.
+                    type: bool
+                  netflow_monitor:
```
akinross, line 136: missing required doc part when netflow is enabled. 

also see link for same deduce question: https://github.com/CiscoDevNet/ansible-nd/pull/268#discussion_r3496420565

plugins/modules/nd_interface_port_channel_access.py
```diff
@@ -0,0 +1,530 @@
...
+                  speed:
+                    description:
+                    - The speed setting for the port-channel.
+                    type: str
+                    choices: [ auto, 10Mb, 100Mb, 1Gb, 2.5Gb, 5Gb, 10Gb, 25Gb, 40Gb, 50Gb, 100Gb, 200Gb, 400Gb, 800Gb ]
+                  storm_control:
+                    description:
+                    - Whether traffic storm control is enabled on the port-channel.
+                    type: bool
+                  storm_control_action:
```
akinross, line 178: UI defaults to default, is this actual API default and should we document this behaviour?

akinross (changes requested):

plugins/modules/nd_interface_port_channel_access.py
```diff
@@ -0,0 +1,530 @@
...
+
+ANSIBLE_METADATA = {"metadata_version": "1.1", "status": ["preview"], "supported_by": "community"}
+
+DOCUMENTATION = r"""
+---
+module: nd_interface_port_channel_access
+version_added: "2.0.0"
+short_description: Manage port-channel accessPoHost interfaces on Cisco Nexus Dashboard
+description:
+- Manage port-channel accessPoHost interfaces on Cisco Nexus Dashboard.
```
akinross, line 17:
```suggestion
short_description: Manage port-channel (accessPoHost) interfaces on Cisco Nexus Dashboard
description:
- Manage port-channel (accessPoHost) interfaces on Cisco Nexus Dashboard.
```

plugins/modules/nd_interface_port_channel_access.py
```diff
@@ -0,0 +1,530 @@
...
+- Allen Robel (@allenrobel)
+options:
+  fabric_name:
+    description:
+    - The name of the fabric containing the target switches.
+    type: str
+    required: true
+  config:
+    description:
+    - The list of port-channel accessPoHost interfaces to configure.
```
akinross, line 33:
```suggestion
    - The list of port-channel (accessPoHost) interfaces to configure.
```

plugins/modules/nd_interface_port_channel_access.py
```diff
@@ -0,0 +1,530 @@
...
+        type: dict
+        suboptions:
+          network_os:
+            description:
+            - Network OS specific configuration.
+            type: dict
+            suboptions:
+              policy:
+                description:
+                - The policy configuration for the accessPoHost port-channel.
```
akinross, line 64:
```suggestion
                - The policy configuration for the (accessPoHost) port-channel.
```

plugins/modules/nd_interface_port_channel_access.py
allenrobel: Thanks @akinross. We considered ranged input (e.g. `eth1-10`) but decided against implementing it. A port-channel supports at most 16 active members (32 total, with members beyond 16 in hot-standby), so the typical use-case is <= 16 member interfaces — a manageable number to list explicitly in YAML. Explicit lists also keep the module input identical to the wire shape (`ports` is a plain list on the API side) and unambiguous in diffs. If users ask for range expansion we can revisit.

plugins/modules/nd_interface_port_channel_access.py
allenrobel: Yes — unquoted `on` is a YAML 1.1 boolean (like `yes`/`true`), so it must be quoted to keep the choice a string; ansible's doc parsing/sanity requires it.

plugins/module_utils/models/interfaces/port_channel_access_interface.py
allenrobel: Thanks @akinross. As with the SVI discussion, we're keeping `interface_name` as the exposed field since it's the ND API identifier key (composite identifier `switch_ip`/`interface_name`), but adopting the same resolution: accept bare integers and range-check the extracted ID against 1-4096 inside the normalizer. Tracked in #378.

plugins/modules/nd_interface_port_channel_access.py
allenrobel: Confirmed — all seven fields (`linkType`, `portTypeEdgeTrunk`, `negotiateAuto`, `orphanPort`, `bandwidth`, `inheritBandwidth`, `pfc`) exist in the accessPoHost template schema and are already implemented in the trunk-host sibling model; only this module is missing them. Tracked in #381, which also picks up your coverage-verification question (proposal: a unit test diffing model field aliases against the template schema properties).

plugins/modules/nd_interface_port_channel_access.py
allenrobel: Confirmed against the ND 4.2.1 OpenAPI schema: `bpduFilter` has schema default `"default"`, matching the UI. The module omits the field when unset so ND applies it server-side. Documenting these server-side defaults collection-wide is tracked in #382.

plugins/modules/nd_interface_port_channel_access.py
allenrobel: Confirmed: `speed` has schema default `"auto"`, matching the UI. Documenting server-side defaults collection-wide is tracked in #382.

plugins/modules/nd_interface_port_channel_access.py
allenrobel: Confirmed: `cdp` has schema default `true`, matching the UI. Documenting server-side defaults collection-wide is tracked in #382.

plugins/modules/nd_interface_port_channel_access.py
allenrobel: Confirmed: `stormControlAction` has schema default `"default"`, matching the UI. Documenting server-side defaults collection-wide is tracked in #382.

plugins/modules/nd_interface_port_channel_access.py
allenrobel: Checked the accessPoHost template schema: `portChannelMode` is not in the schema's `required` list (only `policyType` is) — the UI is pre-populating the schema default `"active"`, same situation as the SVI HSRP fields in #268. When unset the module omits it and ND applies the default. Documenting this collection-wide is tracked in #382.

plugins/modules/nd_interface_port_channel_access.py
allenrobel: Same as `portChannelMode`: `bpduGuard` is not API-required; the UI pre-populates the schema default `"enable"`. Documented via #382.

plugins/modules/nd_interface_port_channel_access.py
allenrobel: Same pattern: `mtu` is not API-required; the UI pre-populates the schema default `"jumbo"`. Documented via #382.

plugins/modules/nd_interface_port_channel_access.py
allenrobel: Applied in #383.

plugins/modules/nd_interface_port_channel_access.py
allenrobel: Applied in #383.

plugins/modules/nd_interface_port_channel_access.py
allenrobel: Applied in #383.

plugins/modules/nd_interface_port_channel_access.py
allenrobel: Applied in #383.

plugins/modules/nd_interface_port_channel_access.py
allenrobel: Both parts addressed in #383: added the "Required when netflow=true" doc line, and — matching the resolution of the #268 discussion — kept the explicit `netflow` boolean (it's an independent field on the wire) while adding the same model-layer validator `nd_interface_svi` uses, so `netflow=true` without `netflow_monitor` now fails early with a clear error.

PR #270, opened 2026-05-04
nd_interface_port_channel_trunk_host

allenrobel:
## Related Issue(s)

- CiscoDevNet/ansible-nd#321

- CiscoDevNet/ansible-nd#347 — consolidate the per-branch VLAN-parsing types (`AllowedVlans`, `CustomerVlanIdList`, `_validate_vlan_id_or_range`) into shared `models/types.py` once the interface stacks land (captured by the in-code TODO in this module).
- CiscoDevNet/ansible-nd#348 — lift the duplicated interface orchestrator CRUD and the `_strip_policy_type_in_config` policy-type serializer into shared bases (this module inherits the same `PortChannelBaseOrchestrator` and now carries the same ported serializer).
- CiscoDevNet/ansible-nd#349 — `query_all` fans out one GET per switch; no fabric-wide interface-list endpoint exists (config-scoped for non-`overridden` states, fabric-wide for `overridden`).
- CiscoDevNet/ansible-nd#353 — consolidate the per-model interface-name normalizers (`normalize_ports`/`_normalize_member_name`/`normalize_interface_name`) into shared type-specific helpers; this module carries another copy of the same logic.

## Proposed Changes

- Add `nd_interface_port_channel_trunk_host` Ansible module managing `trunkPoHost` port-channel interfaces on Cisco Nexus Dashboard 4.2.
- Mirrors `nd_interface_port_channel_access` structure: composite identifier (`switch_ip`, `interface_name`), `policyType` hardcoded to `"trunkPoHost"` and not exposed in argspec, full ND OpenAPI `intPortChannelTrunkHostTemplate` field set (42 policy fields including `bandwidth`, `link_type`, `negotiate_auto`, `vlan_mapping` with selective dot1q-tunnel, storm-control float fields).
- New enums: `TrunkPoHostPolicyTypeEnum`, `LinkTypeEnum` in `plugins/module_utils/models/interfaces/enums.py`.
- New orchestrator inherits from `PortChannelBaseOrchestrator`; managed policy types filter to `{"trunkPoHost"}` only.
- `state: overridden` deletes via `interfaceActions/remove` (mirrors access PC); `state: query` deferred consistent with current parity stance.
- Comprehensive unit test coverage (148 model + 8 orchestrator tests). Covers field defaults, snake/camelCase construction, validators (including `allowed_vlans` regex with `int -> str` coercion for ND's single-id-as-int responses), AsciiDescription, storm-control float-from-string coercion, vlan_mapping entries, composite identifier round-trips, and `query_all` filtering.
- Integration tests cover merged/replaced/overridden/deleted state transitions, member-uniqueness ordering, and a non-9000v `vlan_mapping` block guarded by `nd_test_platform`.

## Test Notes

- Unit tests pass: `pytest tests/unit/module_utils/ -k port_channel_trunk_host -v` (156 passed). Full suite (`pytest tests/unit/`): 705 passed, no regressions.
- `black` and `pylint` clean (matching access PC scoring).
- Integration tests pass against ND 4.2 fabric on N9K-C9300v testbed (single-fabric, single-switch). `vlan_mapping` block skipped on 9000v as expected.
- Live-API verification: ND returns `policyType: "trunkPoHost"` and accepts the model's payload shape (verified with POST/GET/PUT/DELETE round-trip during development).
- Note: ND echoes single-id `allowedVlans` as a JSON int rather than string; the model coerces `int -> str` in a `mode='before'` validator and a unit test locks the behavior.

4.2

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

## Stacking

This PR is stacked on top of #269 (`nd_interface_port_channel_access`). Base branch is `nd_interface_port_channel_access`; merge after #269 lands.

### DIFF
--- a/plugins/module_utils/models/interfaces/port_channel_trunk_host_interface.py
+++ b/plugins/module_utils/models/interfaces/port_channel_trunk_host_interface.py
@@ -37,7 +37,9 @@ from typing import Annotated, ClassVar, Literal, Optional  # Optional needed for
 from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import (
     BeforeValidator,
     Field,
+    SerializationInfo,
     field_validator,
+    model_serializer,
 )
 from ansible_collections.cisco.nd.plugins.module_utils.models.base import NDBaseModel
 from ansible_collections.cisco.nd.plugins.module_utils.models.interfaces.enums import (
@@ -59,6 +61,14 @@ from ansible_collections.cisco.nd.plugins.module_utils.models.types import Ascii
 _ALLOWED_VLANS_SHAPE = re.compile(r"^(none|all|(\d+(-\d+)?)(,\d+(-\d+)?)*)$")
 # Single VLAN id or range token (e.g. "100" or "100-200"). Range bounds are validated separately.
 _VLAN_ID_OR_RANGE_SHAPE = re.compile(r"^\d+(-\d+)?$")
+
+# Splits an interface name into its leading alphabetic prefix and the rest (digits/separators).
+_INTERFACE_NAME_PREFIX_RE = re.compile(r"^([A-Za-z]+)(.*)$")
+
+# Member interfaces of a port-channel are always physical ethernet ports, so the canonical wire prefix
+# is always "Ethernet". Any case-insensitive NX-OS abbreviation (e.g. "e", "eth", "ether") expands to the
+# full form so user input matches the wire key and idempotency holds.
+_CANONICAL_MEMBER_TYPE = "Ethernet"
 
 
 def _validate_vlan_id_or_range(token: str, field_name: str) -> None:
@@ -113,6 +123,10 @@ def _validate_allowed_vlans(value):
     """
     if value is None or value == "":
         return value
+    # TODO(4.2.1) interface-get-field-normalization
+    # ND echoes a single-id allowed_vlans as a JSON int (e.g. 250) on GET even though POST/PUT accept (and the
+    # spec types it as) a string. We coerce int -> str here so a GET response round-trips against the str the
+    # user supplied and idempotency comparisons stay stable. Remove when the GET-side retype is fixed.
     if isinstance(value, int) and not isinstance(value, bool):
         value = str(value)
     if not isinstance(value, str):
@@ -209,6 +223,10 @@ class PortChannelTrunkHostPolicyModel(NDNestedModel):
     ### ValueError
 
     - If `allowed_vlans` is set and does not match `none`, `all`, or comma-separated VLAN ranges.
+
+    ### AssertionError
+
+    - If the wrapped model serializer receives a non-`dict` from the handler (see `_strip_policy_type_in_config`).
     """
 
     admin_state: bool | None = Field(default=None, alias="adminState", description="Enable or disable the interface")
@@ -246,7 +264,6 @@ class PortChannelTrunkHostPolicyModel(NDNestedModel):
     policy_type: TrunkPoHostPolicyTypeEnum = Field(
         default=TrunkPoHostPolicyTypeEnum.TRUNK_PO_HOST, alias="policyType", frozen=True, description="Interface policy type (hardcoded for this module)"
     )
-    port_channel_id: str | None = Field(default=None, alias="portChannelId", description="Port-channel id (response-only echo of interface_name)")
     port_channel_mode: PortChannelModeEnum | None = Field(default=None, alias="portChannelMode", description="Port-channel mode (on/active/passive)")
     port_type_edge_trunk: bool | None = Field(default=None, alias="portTypeEdgeTrunk", description="Configure as edge trunk port (PortFast on trunk)")
     ports: list[str] | None = Field(default=None, alias="ports", description="Member interface names (e.g. ['Ethernet1/1', 'Ethernet1/2'])")
@@ -314,23 +331,85 @@ class PortChannelTrunkHostPolicyModel(NDNestedModel):
         """
         # Summary
 
-        Normalize each member interface name to ND API convention (e.g. `ethernet1/1` -> `Ethernet1/1`).
+        Normalize each member interface name to ND's canonical `Ethernet` form so that any user-supplied casing or
+        NX-OS abbreviation round-trips against the wire form. Members are always physical ethernet ports. Examples:
+
+        - `ethernet1/1` -> `Ethernet1/1`
+        - `ETHERNET1/1` -> `Ethernet1/1`
+        - `eth1/1` -> `Ethernet1/1` (abbreviation expanded)
+        - `e1/1` -> `Ethernet1/1` (abbreviation expanded)
+        - `Ethernet1/1` -> `Ethernet1/1` (idempotent)
+
+        Because the wire key is matched exactly, an abbreviated prefix that is not expanded would never match ND's
+        `Ethernet...` form, silently breaking idempotency (the port-channel re-deploys on every run). Any
+        case-insensitive prefix of `Ethernet` (`e`, `et`, `eth`, ...) is therefore expanded to the full canonical
+        name; an unrecognized prefix falls back to Title case so it still round-trips. Only the leading alphabetic
+        run is rewritten; digits and separators are preserved verbatim.
 
         ## Raises
 
         None
         """
-        if value is None:
-            return value
         if not isinstance(value, list):
             return value
-        normalized = []
-        for name in value:
-            if isinstance(name, str) and name:
-                normalized.append(name[0].upper() + name[1:])
-            else:
-                normalized.append(name)
-        return normalized
+        return [cls._normalize_member_name(name) for name in value]
+
+    @staticmethod
+    def _normalize_member_name(name):
+        """
+        # Summary
+
+        Normalize a single member interface name to ND's canonical `Ethernet` form (see `normalize_ports`).
+        Non-string or empty values are returned unchanged.
+
+        ## Raises
+
+        None
+        """
+        if not isinstance(name, str) or not name:
+            return name
+        match = _INTERFACE_NAME_PREFIX_RE.match(name)
+        if not match:
+            return name
+        prefix, rest = match.groups()
+        if _CANONICAL_MEMBER_TYPE.lower().startswith(prefix.lower()):
+            return _CANONICAL_MEMBER_TYPE + rest
+        return prefix[0].upper() + prefix[1:].lower() + rest
+
+    @model_serializer(mode="wrap")
+    def _strip_policy_type_in_config(self, handler, info: SerializationInfo):
+        """
+        # Summary
+
+        Omit `policy_type` from `to_config()` output while leaving payload and diff modes untouched.
+
+        The field is hardcoded by the model (frozen at `TrunkPoHostPolicyTypeEnum.TRUNK_PO_HOST`), is excluded
+        from the Ansible argspec, and is therefore not something the user supplies or needs surfaced back.
+        The wire form `"trunkPoHost"` would otherwise appear under the `policy_type` key in
+        `before`/`after`/`gathered` output and confuse playbooks that compare against the Ansible
+        snake_case convention. Payload and diff serialization still emit the wire value so the POST/PUT
+        body and the round-trip diff comparison line up with what ND returns.
+
+        Implemented as a wrap-mode model serializer because `exclude_none=True` on `to_config()` evaluates
+        the field value before serialization runs — returning None from a field_serializer is too late to
+        drop the key.
+
+        ## Raises
+
+        ### AssertionError
+
+        - If the wrapped handler returns a non-`dict`. A model-level serializer always serializes to a `dict`,
+          so this is an invariant check that fails loudly rather than silently leaving `policy_type` in the
+          config output.
+        """
+        result = handler(self)
+        if not isinstance(result, dict):
+            raise AssertionError(f"Expected dict from model serialization, got {type(result).__name__}")
+        mode = (info.context or {}).get("mode", "payload")
+        if mode == "config":
+            result.pop("policy_type", None)
+            result.pop("policyType", None)
+        return result
 
 
 class PortChannelTrunkHostNetworkOSModel(NDNestedModel):
@@ -360,7 +439,7 @@ class PortChannelTrunkHostConfigDataModel(NDNestedModel):
     """
 
     mode: Literal["trunk"] = Field(default="trunk", alias="mode", frozen=True)
-    network_os: PortChannelTrunkHostNetworkOSModel = Field(alias="networkOS")
+    network_os: PortChannelTrunkHostNetworkOSModel = Field(default_factory=PortChannelTrunkHostNetworkOSModel, alias="networkOS")
 
 
 class PortChannelTrunkHostInterfaceModel(NDBaseModel):
--- a/plugins/module_utils/models/interfaces/port_channel_trunk_host_interface.py
+++ b/plugins/module_utils/models/interfaces/port_channel_trunk_host_interface.py
@@ -243,7 +243,7 @@ class PortChannelTrunkHostPolicyModel(NDNestedModel):
         default=None, alias="orphanPort", description="Configure as a vPC orphan port (suspended by secondary peer on vPC failure)"
     )
     pfc: bool | None = Field(default=None, alias="pfc", description="Enable Priority Flow Control")
-    policy_type: TrunkPoHostPolicyTypeEnum = Field(default=TrunkPoHostPolicyTypeEnum.TRUNK_PO_HOST, alias="policyType", description="Interface policy type")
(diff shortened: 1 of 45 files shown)


plugins/module_utils/models/interfaces/port_channel_trunk_host_interface.py
```diff
@@ -0,0 +1,424 @@
...
+    PortChannelModeEnum,
+    SpeedEnum,
+    StormControlActionEnum,
+    TrunkPoHostPolicyTypeEnum,
+)
+from ansible_collections.cisco.nd.plugins.module_utils.models.nested import NDNestedModel
+from ansible_collections.cisco.nd.plugins.module_utils.models.types import AsciiDescription
+
+# Accepts: "none", "all", or comma-separated VLAN ids/ranges (e.g. "1-200,500-2000,3000").
+ALLOWED_VLANS_PATTERN = re.compile(r"^(none|all|(\d+(-\d+)?)(,\d+(-\d+)?)*)$")
```
akinross: should this be made stricter to only allow valid VLAN values? Similar as you do for provider_vlan_id

plugins/modules/nd_interface_port_channel_trunk_host.py
```diff
@@ -0,0 +1,492 @@
+#!/usr/bin/python
+
+# Copyright: (c) 2026, Allen Robel (@allenrobel)
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+"""Ansible module for managing port-channel trunkPoHost interfaces on Cisco Nexus Dashboard."""
```
akinross: Did you consider providing a more human readable description of trunkPoHost with (trunkPoHost) for instance between brackets to make clear it is a named this way in ND? Similar to something you do here: https://github.com/CiscoDevNet/ansible-nd/pull/270/changes#diff-1bc3ea994025d9c3d17dfc41a4b4e62aba250b851cd388081f17ada991dc838bR44

plugins/modules/nd_interface_port_channel_trunk_host.py
```diff
@@ -0,0 +1,492 @@
...
+              policy:
+                description:
+                - The policy configuration for the trunkPoHost port-channel.
+                type: dict
+                suboptions:
+                  admin_state:
+                    description:
+                    - The administrative state of the port-channel.
+                    type: bool
+                  allowed_vlans:
```
akinross, line 71: should you also provide valid range of VLAN accepted here?

plugins/modules/nd_interface_port_channel_trunk_host.py
```diff
@@ -0,0 +1,492 @@
...
+                  port_channel_mode:
+                    description:
+                    - The port-channel mode.
+                    type: str
+                    choices: [ on, active, passive ]
+                  port_type_edge_trunk:
+                    description:
+                    - Configure the port-channel as an edge trunk port (PortFast on trunk).
+                    type: bool
+                  ports:
```
akinross, line 187: did you consider allowing range input, and creating the list of interfaces from a range?

plugins/modules/nd_interface_port_channel_trunk_host.py
```diff
@@ -0,0 +1,492 @@
...
+extends_documentation_fragment:
+- cisco.nd.modules
+- cisco.nd.check_mode
+notes:
+- This module is only supported on Nexus Dashboard.
+- This module manages NX-OS port-channel trunkPoHost interfaces only.
+- The port-channel policy is the source of truth for member interface configuration.
+"""
+
+EXAMPLES = r"""
```
akinross, line 319: this is more a generic example question, is it on purpose to only show examples for merged and deleted states?

tests/integration/targets/nd_interface_port_channel_trunk_host/tasks/deleted.yaml
```diff
@@ -0,0 +1,151 @@
...
+
+- name: "DELETED IDEMPOTENT: Delete port-channel502 again"
+  cisco.nd.nd_interface_port_channel_trunk_host:
+    output_level: "{{ nd_info.output_level }}"
+    fabric_name: "{{ test_fabric_name }}"
+    config:
+      - switch_ip: "{{ test_switch_ip }}"
+        interface_name: port-channel502
+    state: deleted
+  register: nm_deleted_502_idem
```
akinross, line 38: Did you consider anchor usage to shorten the test and avoid small typo's in the configuration that used across tests?

```suggestion
- name: "DELETED: Delete port-channel502 (check mode)"
  cisco.nd.nd_interface_port_channel_trunk_host: &delete_pc502
    output_level: "{{ nd_info.output_level }}"
    fabric_name: "{{ test_fabric_name }}"
    config:
      - switch_ip: "{{ test_switch_ip }}"
        interface_name: port-channel502
    state: deleted
  check_mode: true
  register: cm_deleted_502

- name: "DELETED: Delete port-channel502 (normal mode)"
  cisco.nd.nd_interface_port_channel_trunk_host: *delete_pc502
  register: nm_deleted_502

- name: "DELETED: Verify port-channel502 was deleted"
  ansible.builtin.assert:
    that:
      - cm_deleted_502 is changed
      - nm_deleted_502 is changed
      - nm_deleted_502.after | selectattr('interface_name', 'equalto', 'port-channel502') | list | length == 0

- name: "DELETED IDEMPOTENT: Delete port-channel502 again"
  cisco.nd.nd_interface_port_channel_trunk_host: *delete_pc502
  register: nm_deleted_502_idem
```

tests/integration/targets/nd_interface_port_channel_trunk_host/tasks/merged.yaml
```diff
@@ -0,0 +1,256 @@
...
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+# Test order: single-PC tests first, then multi-PC. The MEMBER ADD test pulls
+# `test_pc_member_c` into port-channel501, so it must run BEFORE PC 502 (which
+# claims member_c) is created — ND rejects adding a port to a PC if it's
+# already a member of another. After MEMBER REMOVE, all members except a are
+# free, so the multi-create that follows succeeds.
+
+# --- CLEANUP ---
```
akinross: did you consider having setup/cleanup into a separate file?

plugins/modules/nd_interface_port_channel_trunk_host.py
allenrobel: Thanks Akini.  Addressed with commit 24e0364d0853e4fdaeefbb07e61dcd25ef045e6b

plugins/module_utils/models/interfaces/port_channel_trunk_host_interface.py
allenrobel: Good catch Akini.  Addressed with commit 5bb9514 using shareable Annotated types `AllowedVlans`.  Also gave the same treatment for `customer_vlan_id` with Annotated type `CustomerVlanIdList`.

I added TODO comments to move these types into module_utils/models/types.py AFTER merging, since `nd_interface_ethernet_trunk_host ` requires the same treatment, but the fixes are inappropriate for the common branch `nd_interface_loopback` (since `nd_interface_loopback` has no concept of vlans).  I've added these same fixes to `nd_interface_ethernet_trunk_host` and we can consolidate post-merge.

I also added `provider_vlan_id` to the same TODO comment, since it (and several other fields in this PR, and many other fields across the ND OpenAPI spec) would also benefit from a shared ` VlanId` Annotated type that constrains the vlan to the 1-4094 range.  This way, `VlanId` would be the single source of truth across the codebase for the `ge=1, le=4094` field constraint.

plugins/modules/nd_interface_port_channel_trunk_host.py
allenrobel: Yes, thanks Akini.  Addressed with commit ea08a6f.  Also addresses the same isse for customer_vlan_id.

plugins/modules/nd_interface_port_channel_trunk_host.py
allenrobel: Thanks for the suggestion Akini. I did consider this.  A few reasons I went with a plain list:

- Current cisco.dcnm.dcnm_interface behavior: that module also takes members as a plain list of strings (no range expansion).  Though, I concede that we're not constrained by the DCNM collection's behavior.
- Ansible does already provide a generic way to do this using Jinja, e.g. members: "{{ range(1, 5) | map('regex_replace', '^', 'Ethernet1/') | list }}" or a `with_sequence` loop.
- NX-OS max port-channel membership is 32 interfaces, so the list isn't horribly unwieldy, and a parser would have to handle edge cases like multi-module ranges like `Eth1/1-4,2/1-4`, mixed lists+ranges, and validation messages for not a lot of UX improvement (IMHO).

Happy to revisit if it comes up as a real pain point — easy to add later as an additive change without breaking the current contract.

plugins/modules/nd_interface_port_channel_trunk_host.py
allenrobel: Thanks, addressed with commit 726738b.

tests/integration/targets/nd_interface_port_channel_trunk_host/tasks/deleted.yaml
allenrobel: Thanks Akini.  I modified all integration tests to use anchors.  Commit: e21c2b3

I'll work on the same for the other interface PRs.

tests/integration/targets/nd_interface_port_channel_trunk_host/tasks/merged.yaml
allenrobel: Thanks for the suggestion.  Addressed with commit 22ea587.

plugins/modules/nd_interface_port_channel_trunk_host.py
akinross: Ok makes sense!

plugins/modules/nd_interface_port_channel_trunk_host.py
```diff
@@ -0,0 +1,510 @@
...
+                      dot1q_tunnel:
+                        description:
+                        - Use selective dot1q-tunnel mode for this entry.
+                        type: bool
+                      provider_vlan_id:
+                        description:
+                        - Provider VLAN id.
+                        - Valid range is 1-4094.
+                        type: int
+  deploy:
```
mikewiebe: ### Medium: Replace top-level `deploy` with `config_actions`

**Issue**

**Evidence**

- [`plugins/modules/nd_interface_port_channel_trunk_host.py` lines 281-303](https://github.com/CiscoDevNet/ansible-nd/pull/270/files#diff-1bc3ea994025d9c3d17dfc41a4b4e62aba250b851cd388081f17ada991dc838bR281) documents `deploy` as a standalone option and lists states without the standard `config_actions` option.

**Existing PR overlap**

**Impact**

**Suggested fix**

---

plugins/modules/nd_interface_port_channel_trunk_host.py
```diff
@@ -0,0 +1,510 @@
...
+    description:
+    - Whether to deploy port-channel changes after mutations are complete.
+    - When V(true), all queued port-channel changes are deployed in a single bulk API call at the end of module
+      execution via the C(interfaceActions/deploy) API. Only the port-channels modified by this task are deployed.
+    - When V(false), changes are staged but not deployed. Use a separate deploy module or task to deploy later.
+    - Setting O(deploy=false) is useful when batching changes across multiple interface tasks before a single deploy.
+    type: bool
+    default: true
+  state:
```
mikewiebe, line 295: ### Low: Track `gathered` support against issue #340 (non-blocking)

**Issue**

**Evidence**

- [`plugins/modules/nd_interface_port_channel_trunk_host.py` lines 290-303](https://github.com/CiscoDevNet/ansible-nd/pull/270/files#diff-1bc3ea994025d9c3d17dfc41a4b4e62aba250b851cd388081f17ada991dc838bR290) documents state choices without `gathered`.
- [`plugins/module_utils/models/interfaces/port_channel_trunk_host_interface.py` lines 504-508](https://github.com/CiscoDevNet/ansible-nd/pull/270/files#diff-d2ab524cc8cd4727b8449727f59773a79668d8ac08e5617de6638ef13c4c74a4R504) returns the same state choices from the model argspec.

**Existing PR overlap**

**Impact**

**Suggested fix**

---

plugins/modules/nd_interface_port_channel_trunk_host.py
allenrobel: Addressed in fde71850.

Replaced the top-level `deploy` bool with a `config_actions` dict carrying a nested `deploy` suboption, mirroring the same migration across the interface module family (`nd_interface_ethernet_trunk_host` ec82a6f2, `nd_interface_svi` d5afacfb, `nd_interface_subinterface_managed` 8806d87d, `nd_interface_subinterface_unmanaged` ace59646):

- **EXAMPLES + merged integration test** — the no-deploy/batching cases now use the nested `config_actions: {deploy: false}` form.

plugins/modules/nd_interface_port_channel_trunk_host.py
allenrobel: Acknowledged — no change in this PR. As noted, `gathered` is tracked collection-wide in #340 for all `NDStateMachine` + orchestrator modules, so `nd_interface_port_channel_trunk_host` will pick it up consistently when that lands (with documented filter semantics, a `gathered` example, and a read-only test asserting no create/update/delete/deploy call is made). Keeping this thread open as the tracking marker for this module.

plugins/modules/nd_interface_port_channel_trunk_host.py
allenrobel: Note: this branch was rebased onto the latest `develop` during the RETURN-block rollout, so the commit referenced above (`fde7185`) is now orphaned. The same change is now `39dbfba1`. Resolving as addressed.

mikewiebe (approved):
LGTM

gmicol (approved):
LGTM

plugins/modules/nd_interface_port_channel_trunk_host.py
```diff
@@ -0,0 +1,617 @@
...
+- Allen Robel (@allenrobel)
+options:
+  fabric_name:
+    description:
+    - The name of the fabric containing the target switches.
+    type: str
+    required: true
+  config:
+    description:
+    - The list of port-channel trunkPoHost interfaces to configure.
```
akinross, line 33:
```suggestion
    - The list of port-channel (trunkPoHost) interfaces to configure.
```

plugins/modules/nd_interface_port_channel_trunk_host.py
```diff
@@ -0,0 +1,617 @@
...
+        type: dict
+        suboptions:
+          network_os:
+            description:
+            - Network OS specific configuration.
+            type: dict
+            suboptions:
+              policy:
+                description:
+                - The policy configuration for the trunkPoHost port-channel.
```
akinross, line 64:
```suggestion
                - The policy configuration for the (trunkPoHost) port-channel.
```

plugins/modules/nd_interface_port_channel_trunk_host.py
```diff
@@ -0,0 +1,617 @@
...
+description:
+- Manage port-channel (trunkPoHost) interfaces on Cisco Nexus Dashboard.
+- It supports creating, updating, and deleting (trunkPoHost) port-channel configurations on switches within a fabric.
+- Each config item represents one port-channel interface. Member ethernet interfaces are listed in
+  O(config[].config_data.network_os.policy.ports) and inherit trunk-mode configuration from the port-channel policy.
+- Member interface field mutability is restricted while members of a port-channel; only description, admin_state, and
+  extra_config can be modified on members via the C(nd_interface_ethernet_trunk_host) module.
+author:
+- Allen Robel (@allenrobel)
+options:
```
akinross, line 25: cross module wide question, how do we document UI default behaviour? 

are these values tested when set empty, does the API also default in that case or could you get errors? 

see https://github.com/CiscoDevNet/ansible-nd/pull/269 where I dropped comments on defaults from UI

akinross (changes requested):

plugins/modules/nd_interface_port_channel_trunk_host.py
allenrobel: Opened issue  #382  to track.

allenrobel: Answered collection-wide in #382: the UI defaults are template-schema defaults (verified against the ND 4.2.1 OpenAPI spec); the modules omit unset fields so ND applies the schema default server-side — no errors when set empty. #382 adds a doc line to every interface module option whose schema declares a default, stating the server-side default applied on create.

plugins/modules/nd_interface_port_channel_trunk_host.py
allenrobel: Applied in #383.

plugins/modules/nd_interface_port_channel_trunk_host.py
allenrobel: Applied in #383.

PR #275, opened 2026-05-11
interface capability preflight (#273)

### LINKED ISSUE #273: Add interface capability pre-flight check as a shared class
## Summary

Add a shared validator class that uses the unpublished `/api/v1/manage/fabrics/{fabric}/capableSwitches` endpoint to pre-flight which switches in a fabric can host a given interface type before issuing per-switch configuration calls. This turns confusing API 400s on incapable switches into a clear, single aggregate error naming the offending switches and the requested interface type.

The validator is intended to be consumed by every per-policy interface module (`nd_interface_loopback`, `nd_interface_ethernet`, `nd_interface_port_channel`, `nd_interface_svi`, `nd_interface_vpc`, `nd_interface_breakout`, ...) via the shared `BaseInterfaceOrchestrator`.

## Endpoint

```text
GET /api/v1/manage/fabrics/{fabric_name}/capableSwitches?interfaceType={type}&mode={mode}
```

Example response:

```json
{
    "switches": [
        {"model": "N9K-C9300v", "switchId": "9F8KHEUX7M7", "switchName": "BG1"},
        {"model": "N9K-C9300v", "switchId": "90U5ECLHWJ7", "switchName": "BG2"},
        {"model": "N9K-C9300v", "switchId": "9GOX969OCIA", "switchName": "LE1"},
        {"model": "N9K-C9300v", "switchId": "9FX6HABUHYX", "switchName": "LE2"},
        {"model": "N9K-C9300v", "switchId": "9Y5F9PLMR3H", "switchName": "SP1"},
        {"model": "N9K-C9300v", "switchId": "9S0Q153TUEX", "switchName": "SP2"}
    ]
}
```

## Stability and risk

This endpoint is **not published** in the ND OpenAPI spec. Confirmed with ND developers, but will not be removed until ND 4.2.1 is deprecated -- likely two to three years out given Nexus Dashboard's deprecation and release cadence.

We accept the risk of consuming an unpublished endpoint in exchange for the UX win, and will revisit if/when published generalized replacements endpoints are introduced. Any code touching this endpoint should be isolated behind the validator class so the future migration is a single-file change.

## Open data we still need

The following must be captured experimentally (Safari Web Inspector / browser devtools against the GUI, then verified with curl/Bruno) before implementation:

1. **Valid `interfaceType` values** — try at least: `loopback`, `ethernet`, `portChannel`, `subinterface`, `svi`, `vlanInterface`, `vpc`, `loopbackInterface`, `breakout`. Note which return 200 vs. 400.
2. **Valid `mode` values per `interfaceType`** — try `routed`, `access`, `trunk`, `trunkHost`, `host`, `managed`, `unmanaged`, and omitting `mode` entirely. Determine: required or optional, accepted set, behavior on unknown value (400 vs. empty `switches`).
3. **Error response shape** — capture body for bad fabric, bad `interfaceType`, bad `mode`.
4. **Multi-mode** — does `mode=access,trunk` work (union) or 400?

Document the findings in the validator's class docstring so future readers do not need to re-run the experiments.

## Proposed implementation

Three logical pieces, each in its own file. Loopback is the natural reference impl (smallest, already on `nd_interface_loopback` branch); other interface modules pick up the behavior via a `ClassVar` override on their orchestrator subclass.

### 1. Endpoint model

New file: `plugins/module_utils/endpoints/v1/manage/manage_fabric_capable_switches.py`

- `EpManageFabricsCapableSwitchesGet` — Pydantic, `FabricNameMixin`, `interface_type` and `mode` required before `.path` is read, URL-encoded query string.
- New unit test file under `tests/unit/module_utils/endpoints/`.

### 2. Validator

New file: `plugins/module_utils/capable_switches.py`, mirroring the shape of `fabric_context.py`:

- `__init__(rest_send, fabric_name)`; lazy fetch on first access.
- `get_capable_switch_ids(interface_type, mode) -> set[str]` — cached per `(interface_type, mode)` tuple for the duration of one task run (parallel to `fabric_context`).
- `get_capable_switches(interface_type, mode) -> list[dict]` — full records for richer error messages (model / switchName).
- `validate(interface_type, mode, switch_ids)` — raises `RuntimeError` with offending IDs, the capable set, and the endpoint URL so the user can verify drift.
- `invalidate()` — drop cache.
- Optional injection of `FabricContext.switch_map_by_id` to enrich errors with `switch_ip` / `switch_name`.
- Unit tests at `tests/unit/module_utils/test_capable_switches.py` plus fixtures `test_capable_switches.json`.

### 3. Orchestrator wiring

Edit: `plugins/module_utils/orchestrators/base_interface.py`

- `interface_type: ClassVar[str] = ""` — subclasses override (e.g. loopback orchestrator sets `interface_type = "loopback"`).
- Lazy `capable_switches` property, parallel to the existing `fabric_context` property.
- `_resolve_mode(model_instance) -> str` — defaults to `model_instance.mode`, override per-orchestrator if needed.
- New `validate_switches_capable(model_instances)` method — groups by `(interface_type, mode)`, issues one GET per group, raises a single aggregate error.
- **Do not** fold this into the existing `validate_prerequisites()` (that takes no args); add it as a separate preflight step.

Module entry points (e.g. `plugins/modules/nd_interface_loopback.py`) call `validate_switches_capable(model_instances)` immediately after `validate_prerequisites()`.

## Design choices already settled

- Hard-fail, no soft-warn. If anyone needs a bypass later, add `validate_capability: bool = True` then; do not preempt.
- Cache scope = one task run, not cross-task (matches `fabric_context`).
- The validator runs in `check_mode` — it is a GET, and the user wants the error before play execution.
- Per-endpoint file rather than appending to `manage_fabrics.py` (already large; `capableSwitches` is a distinct sub-resource).
- Argued and rejected: letting ND's API reject the request to keep our code simpler. The UX win of naming the offending `switch_ip` / `switch_name` outweighs the extra GET.

## Acceptance

- [ ] `EpManageFabricsCapableSwitchesGet` endpoint model + unit tests
- [ ] `CapableSwitchesValidator` (or chosen name) class + unit tests with fixtures
- [ ] `BaseInterfaceOrchestrator.validate_switches_capable()` wired
- [ ] `nd_interface_loopback` overrides `interface_type` ClassVar and calls the validator from the module entry point (reference impl)
- [ ] Class docstring records the experimentally-determined `interfaceType`/`mode` taxonomy and error shapes
- [ ] Negative integration test (deferred only if the testbed cannot reproduce an incapable switch for any supported interface type)

## Context

Originally surfaced and parked during work on `nd_interface_loopback` when the endpoint was found to be absent from the ND 4.2 OpenAPI spec. Cisco's ND developers have since confirmed the endpoint is stable for the lifetime of ND 4.2.x, unblocking the implementation.

### LINKED ISSUE #300: NDBaseInterfaceOrchestrator.validate_switches_capable consumes its input; bulk callers iterate it twice
## Summary

`NDBaseInterfaceOrchestrator.validate_switches_capable(model_instances: Iterable[ModelType])` consumes its argument in a set comprehension (`switch_ids = {self._resolve_switch_id(m.switch_ip) for m in model_instances}`). Bulk callers — e.g. `LoopbackInterfaceOrchestrator.create_bulk` — then iterate the SAME parameter a second time (`for model_instance in model_instances:`) to build their per-switch payload groups. If a caller ever passes a one-pass iterable (generator, `filter(...)`, `map(...)`), the second loop sees an exhausted iterator: `groups` stays empty, no POSTs fire, and the state machine records a silent successful no-op.

## Reproduction

```python
# In a hypothetical refactor or new caller:
gen = (m for m in inventory_models)
loopback_orchestrator.create_bulk(gen)
# create_bulk -> validate_switches_capable(gen) consumes the generator
# -> for model_instance in gen: <iterates nothing>
# -> POST /interfaces is never called; module reports changed=True with empty results.
```

Today's sole caller (`NDStateMachine`) passes a list, so the bug is latent — but the public `Iterable` signature openly invites future misuse.

## Fix sketch

- Narrow the parameter type on `validate_switches_capable` (and on every bulk method that iterates its input twice) from `Iterable[ModelType]` to `Sequence[ModelType]`. Mypy will then catch generator-passing callers at static-check time.
- Optionally add `model_instances = list(model_instances)` at the top of `validate_switches_capable` as runtime defense-in-depth.

## Files

- `plugins/module_utils/orchestrators/base_interface.py` (`validate_switches_capable`, line 142)
- `plugins/module_utils/orchestrators/loopback_interface.py` (`create_bulk`, line 165)

## Context

- Surfaced in code review of the capability-preflight PR on branch \`nd_interface_capability_preflight\` (issue #273).
- Both files originate in develop (PR #220); tracked here separately so the fix isn't bundled into an unrelated PR.

### LINKED ISSUE #301: validate_switches_capable: a single unresolvable switch_ip masks capability problems on other switches
## Summary

`NDBaseInterfaceOrchestrator.validate_switches_capable` resolves every `switch_ip` to a `switch_id` inside a single set comprehension:

```python
switch_ids = {self._resolve_switch_id(m.switch_ip) for m in model_instances}
```

`_resolve_switch_id` (via `FabricContext.get_switch_id`) raises `RuntimeError` on the first unknown IP. In a bulk operation with N target switches, a single typo'd IP short-circuits the whole preflight before any capability check runs — the user gets one "no switch matches IP" error and never sees capability problems on the remaining switches. The method's docstring promises "A single GET covers every target switch ... naming every offending switch", which isn't delivered today.

## Reproduction

```yaml
- cisco.nd.nd_interface_loopback:
    fabric: my_fabric
    config:
      - switch_ip: 10.1.1.1            # valid
        interface_name: loopback10
      - switch_ip: 10.1.1.99            # typo, not in fabric
        interface_name: loopback11
      - switch_ip: 10.1.1.2             # valid, but incapable
        interface_name: loopback12
```

Today: module fails with "no switch matches 10.1.1.99". After fixing the typo, module fails again with "switch 10.1.1.2 not capable". Round-trip per error.

## Fix sketch

1. Resolve all IPs first, collecting unresolved ones into an `unresolved: list[str]`.
2. If `unresolved` is non-empty, raise a single `RuntimeError` listing all of them — same aggregate-error shape the capability check uses.
3. Otherwise, run the capability check on the resolved subset (current behavior).

## File

- `plugins/module_utils/orchestrators/base_interface.py` (`validate_switches_capable`, line 168)

## Context

- Surfaced in code review of the capability-preflight PR on branch \`nd_interface_capability_preflight\` (issue #273).
- File originates in develop (PR #220); tracked here separately so the fix isn't bundled into an unrelated PR.

### LINKED ISSUE #302: Capability preflight issues a live GET in --check mode and can fail dry-runs that previously succeeded
## Summary

The new `validate_switches_capable` call sites in `LoopbackInterfaceOrchestrator.create`, `update`, and `create_bulk` drive a GET against `/api/v1/manage/fabrics/{fabric}/capableSwitches`. `RestSend.commit()` only short-circuits writes in `--check` mode; GETs fall through and hit the live controller. This is a behavior change versus develop:

1. Every `--check` run now issues at least one extra live GET per orchestrator call (mitigated by per-(interface_type, mode) cache after the first call, but never zero).
2. Once GitHub issue #273's fix lands (404 → `RuntimeError` with a clear "endpoint unavailable" message), a `--check` run on an ND release that has retired/relocated the unpublished `capableSwitches` endpoint will fail the dry-run with an error that did not exist before this PR.

Existing `validate_prerequisites` already does live GETs in check mode, so option (b) below is consistent with current design — but the regression risk in (a) merits an explicit decision.

## Reproduction

```bash
ansible-playbook loopback.yml --check
# Before this PR: succeeds without contacting capableSwitches.
# After this PR:
#   - Issues GET /api/v1/manage/fabrics/{fabric}/capableSwitches?interfaceType=loopback&mode=managed
#   - If endpoint returns 404 on this ND release -> RuntimeError; dry-run fails.
```

## Fix sketch

Two viable options; team should pick one:

- **Skip preflight in check mode.** Early-return in `validate_switches_capable` when `self.rest_send.check_mode` is `True`. Quieter dry-runs, but check mode no longer flags incapable switches.
- **Run preflight but soften endpoint errors in check mode.** Catch the `RuntimeError` from `validate` inside `validate_switches_capable`, and in check mode log a warning instead of re-raising. Dry-runs stay green when the endpoint is down; capability problems still surface as informational output.

## Files

- `plugins/module_utils/orchestrators/loopback_interface.py` (`create`/`update`/`create_bulk` — lines 102, 128, 172)
- `plugins/module_utils/orchestrators/base_interface.py` (`validate_switches_capable`, line 142) if the gate lives in the base method

## Context

- Surfaced in code review of the capability-preflight PR on branch \`nd_interface_capability_preflight\` (issue #273).
- Both files originate in develop (PR #220); tracked here separately so the fix isn't bundled into an unrelated PR.

allenrobel:
## Related Issue(s)

Closes #273, #300, #301, #302

## Proposed Changes

Pre-flights ND's unpublished `/api/v1/manage/fabrics/{fabric}/capableSwitches` endpoint before per-switch configuration calls, converting confusing HTTP 400s on incapable switches into a single aggregate error that names the offending switches.

- **Endpoint model** `EpManageFabricsCapableSwitchesGet`, inheriting `_EpManageFabricsBase` for path construction, with required `interfaceType` / `mode` query params.
- **`InterfaceCapabilityPreflight`** class (`plugins/module_utils/interface_capability_preflight.py`) mirroring `FabricContext` shape:
  - `(interface_type, mode)` cache (switch records + derived switch-id set); one GET per pair per task run
  - Client-side taxonomy enforcement (captured from the ND 4.2 GUI 2026-05-11) — invalid combos raise `ValueError` without consuming a round trip
  - Optional `FabricContext` injection enriches error messages with `switch_ip`
  - `invalidate()` for cache reset
  - Treats HTTP 404 from `capableSwitches` as a hard failure (rather than caching an empty capable set that would mislabel every requested switch); returns defensive copies from `get_capable_switches` / `get_capable_switch_ids` so callers can't corrupt cached state.
- **Orchestrator wiring** on `NDBaseInterfaceOrchestrator`:
  - `interface_type` / `interface_mode` ClassVars (both default `""`) — subclasses opt in by setting **both** to a valid taxonomy pair
  - Lazy `capability_preflight` property
  - `validate_switches_capable(model_instances)` — resolves the target switch-id set, issues a single `capableSwitches` GET, and raises one `RuntimeError` naming every incapable switch; also raises a clear `RuntimeError` if a subclass sets `interface_type` without `interface_mode`. Parameter typed `Sequence[ModelType]` so static analysis rejects generator callers (commit `0157f2f`, fixes #300). Resolves all `switch_ip` values up front and aggregates unresolvable ones into a single `RuntimeError` so a typo on one entry no longer masks problems on the rest (commit `8835915`, fixes #301). In `--check` mode, downgrades capability-endpoint failures and capability mismatches to warnings via `AnsibleModule.warn` so dry-runs stay green when the unpublished endpoint is unavailable (commit `215601a`, fixes #302).
- **State-machine wiring** — `NDStateMachine.manage_state()` invokes the preflight once per run, over the proposed (desired-config) set, **before** the create/update operations (which are skipped in `--check` mode). This is what makes preflight fire during a dry-run; see "Modifies existing files on `develop`" below.
- **Loopback as reference impl** — `interface_type = "loopback"`, `interface_mode = "managed"`. Preflight is driven centrally by the state machine (above), not from loopback's per-op CRUD methods. Delete and read paths intentionally skip preflight.

### Taxonomy embedded in `InterfaceCapabilityPreflight` docstring

| interface_type | valid modes                                                              |
|----------------|--------------------------------------------------------------------------|
| `loopback`     | `managed`                                                                |
| `svi`          | `managed`                                                                |
| `tunnel`       | `managed`                                                                |
| `ethernet`     | `trunk`, `access`, `routed`, `fex`, `pvlan`, `dot1qTunnel`, `unmanaged`  |
| `portChannel`  | `trunk`, `access`, `routed`, `fex`, `pvlan`, `dot1qTunnel`, `unmanaged`  |

VPC uses a different endpoint (`/vpcPairs?view=intendedPairs`) and is handled separately (follow-up). Breakout and subinterface are out of scope (no pre-check available / absent from taxonomy).

## Modifies existing files on `develop` (merge-conflict surface)

This PR is **not purely additive** — it modifies several files that already exist on `develop` (from PRs #220 / loopback). They are the merge-conflict surface; anyone rebasing an in-flight branch onto this work should expect touch-points here, and conflict resolution should preserve the intent noted per file:

**Core base files (shared by every Gen-3 / interface module):**

- `plugins/module_utils/orchestrators/base.py` — adds a no-op `preflight(model_instances)` hook on `NDBaseOrchestrator`. Polymorphic seam so `NDStateMachine` can call preflight on any orchestrator; no-op for non-interface orchestrators (local_user, fabric_*), so behavior there is unchanged. **Conflict intent:** keep the new method; it has no interaction with existing CRUD.
- `plugins/module_utils/nd_state_machine.py` — `manage_state()` now calls `self.model_orchestrator.preflight(list(self.proposed))` before `_manage_create_update_state()` (i.e. ahead of the check-mode skip in `_execute_operation`). The `deleted` branch is intentionally left without a preflight call. **Conflict intent:** the preflight call must stay above the create/update path and out of the delete path.
- `plugins/module_utils/orchestrators/base_interface.py` — overrides `preflight()` to delegate to the existing `validate_switches_capable()`, plus the `interface_type`/`interface_mode` ClassVars and `validate_switches_capable()` itself (from the original feature commits).

**Loopback reference impl (already merged):**

- `plugins/module_utils/orchestrators/loopback_interface.py` — sets the `interface_type`/`interface_mode` ClassVars. The earlier per-op `validate_switches_capable()` calls in `create`/`update`/`create_bulk` were **removed** once preflight moved to the state machine.
- `tests/unit/module_utils/orchestrators/test_loopback_interface.py` and `tests/unit/module_utils/fixtures/fixture_data/test_loopback_interface.json` — mutating tests no longer consume a per-op `capableSwitches` GET (those fixture responses were removed; capability is now covered at the state-machine and base-interface level).

No change to `plugins/modules/nd_interface_loopback.py`, the loopback model, or loopback's user-facing behavior — preflight is internal and runs ahead of the existing API calls.

## Code-review follow-ups resolved in this PR

A high-effort `/code-review` pass on this branch surfaced three additional findings whose fixes touch files originating in `develop` (`base_interface.py`, `loopback_interface.py`, both from PR #220). All three are now resolved in this PR rather than deferred:

- **#300** — Resolved (commit `0157f2f`). `validate_switches_capable` accepted `Iterable` but bulk callers iterate the same parameter twice; a generator caller would have become a silent no-op. Narrowed the base-class signature to `Sequence[ModelType]` so mypy rejects generator callers at the boundary. The method body and all existing call sites (`list[...]`) are unchanged.
- **#301** — Resolved (commit `8835915`). A single unresolvable `switch_ip` masked aggregate capability problems for the rest of the batch. Now resolves every `switch_ip` up front, collects the unresolvable ones, and raises a single aggregate `RuntimeError` naming every unknown IP; the capability GET only runs on the resolved subset.
- **#302** — Resolved (commit `215601a`). Capability preflight issues a live GET in `--check` mode; a 404 (post-#273-merge) would now fail a dry-run that previously succeeded. In `--check` mode, capability-endpoint failures and capability mismatches are downgraded to warnings via `AnsibleModule.warn` so dry-runs stay green and the information still surfaces. Unresolvable `switch_ip` values still raise even in check mode because they reflect user input errors.

Two further in-branch findings in `interface_capability_preflight.py`: the 404-silently-cached-as-empty-set and an initial collection-level defensive copy are in commit `7a130c4`. A later review (gmicol, discussion_r3429267935) found that copy was only shallow — `get_capable_switches()` still shared the cached record dicts by reference, so a caller could mutate a returned record and corrupt the cache (and the derived id set). Completed in commit `8c1ce044`: `get_capable_switches()` now returns a `copy.deepcopy` of the cached list, with a regression test that mutates returned record dicts and asserts the cache is unaffected.

## Adopting preflight in downstream interface modules

Once this lands, each new interface orchestrator (svi, ethernet access / trunk-host, port-channel, ...) opts in with no boilerplate beyond two ClassVars:

1. **Inherit `NDBaseInterfaceOrchestrator`** — already standard for interface orchestrators; nothing new.
2. **Set both ClassVars** to a valid taxonomy pair from the table above:
   - `interface_type: ClassVar[str] = "..."` (e.g. `"ethernet"`, `"portChannel"`, `"svi"`)
   - `interface_mode: ClassVar[str] = "..."` (e.g. `"access"`, `"trunk"`, `"routed"`, `"managed"`)

   One orchestrator = one `(interface_type, interface_mode)` pair; per-mode variants (e.g. ethernet access vs. trunk-host) each set their own `interface_mode`.
3. **Nothing to call per-op.** Preflight is driven centrally by `NDStateMachine.manage_state()` over the proposed set, before create/update; orchestrators no longer call `validate_switches_capable(...)` from their CRUD methods. Setting the two ClassVars is the entire opt-in.
4. **Nothing else to override.** `_resolve_mode()` no longer exists — mode is the `interface_mode` ClassVar. If a subclass sets `interface_type` but omits `interface_mode`, `validate_switches_capable` raises a clear `RuntimeError` naming the class.
5. **Opt out** by leaving `interface_type` as `""` (the base default) — for interface types with no `capableSwitches` coverage (breakout, subinterface).
6. **VPC is separate** — vpc orchestrators use a different endpoint (`/vpcPairs?view=intendedPairs`) and must not set these ClassVars; vpc preflight is a follow-up.
7. **Tests** — no per-op fixture changes needed; capability behavior is covered once at the state-machine level (`test_nd_state_machine.py`) and base-interface level (`test_base_interface.py`). Orchestrator tests no longer carry a per-op `capableSwitches` response.

## Test Notes

- 50 new unit tests for the preflight feature pass:
  - `tests/unit/module_utils/endpoints/test_endpoints_api_v1_manage_fabric_capable_switches.py` — 9 tests
  - `tests/unit/module_utils/test_interface_capability_preflight.py` — 35 tests (incl. 24 parametrized taxonomy cases, plus 404-failure and defensive-copy tests added in commit `7a130c4`)
  - `tests/unit/module_utils/orchestrators/test_base_interface.py` — 6 tests covering `validate_switches_capable`: opted-out no-op, half-configured `RuntimeError`, aggregate unresolved-IP error (#301), check-mode soften on endpoint failure (#302), check-mode soften on capability mismatch (#302), and the regression guard ensuring mismatches still raise outside check mode (#302)
- **State-machine wiring coverage** — new `tests/unit/module_utils/test_nd_state_machine.py` drives the full `manage_state()` path and proves: check-mode `merged` calls `preflight` while the mutation (`create_bulk`) is skipped; the normal-mode contrast calls both (preflight first); `deleted` does not preflight; `overridden` preflights the proposed set. Plus a `preflight` delegation test in `test_base_interface.py` and generic no-op tests in `test_base_orchestrator.py`.
- `test_loopback_interface.py` / `test_loopback_interface.json` — the per-op `capableSwitches` responses were **removed** (preflight no longer runs inside loopback CRUD); fixtures renumbered to keep the documented contiguous order. Loopback suite green.
- `MockAnsibleModule` gains a `warn()` method that records to `self.warnings` so the check-mode soften tests can assert on emitted warnings.
- Full local unit suite: **1210 passed**, no regressions
- `nd_interface_loopback` integration tests green (happy path unchanged)
- `ansible-test sanity --docker` passes on all changed files (compile / import / pep8 / pylint / validate-modules / empty-init / yamllint / ... across Python 3.8–3.13)
- `black` + `isort` clean
- Integration smoke against the ND 4.2 lab is deferred to a follow-up — depends on having a reproducer for an incapable switch.

4.2.1

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

### DIFF
--- a/plugins/module_utils/endpoints/v1/manage/manage_fabric_capable_switches.py
+++ b/plugins/module_utils/endpoints/v1/manage/manage_fabric_capable_switches.py
@@ -24,12 +24,9 @@ from __future__ import annotations
 from typing import ClassVar, Literal
 
 from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import Field
-from ansible_collections.cisco.nd.plugins.module_utils.endpoints.base import NDEndpointBaseModel
-from ansible_collections.cisco.nd.plugins.module_utils.endpoints.mixins import FabricNameMixin
 from ansible_collections.cisco.nd.plugins.module_utils.endpoints.query_params import EndpointQueryParams
-from ansible_collections.cisco.nd.plugins.module_utils.endpoints.v1.manage.base_path import BasePath
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.v1.manage.manage_fabrics import _EpManageFabricsBase
 from ansible_collections.cisco.nd.plugins.module_utils.enums import HttpVerbEnum
-from ansible_collections.cisco.nd.plugins.module_utils.types import IdentifierKey
 
 
 class CapableSwitchesEndpointParams(EndpointQueryParams):
@@ -72,7 +69,7 @@ class CapableSwitchesEndpointParams(EndpointQueryParams):
     )
 
 
-class EpManageFabricsCapableSwitchesGet(FabricNameMixin, NDEndpointBaseModel):
+class EpManageFabricsCapableSwitchesGet(_EpManageFabricsBase):
     """
     # Summary
 
@@ -115,41 +112,34 @@ class EpManageFabricsCapableSwitchesGet(FabricNameMixin, NDEndpointBaseModel):
         description="Class name for backward compatibility",
     )
 
-    _require_fabric_name: ClassVar[bool] = True
-    _path_suffix: ClassVar[str] = "capableSwitches"
+    _path_suffix: ClassVar[str | None] = "capableSwitches"
 
     endpoint_params: CapableSwitchesEndpointParams = Field(
         default_factory=CapableSwitchesEndpointParams,
         description="Endpoint-specific query parameters (interface_type, mode).",
     )
 
-    def set_identifiers(self, identifier: IdentifierKey = None):
-        """Set the fabric_name from the identifier value."""
-        self.fabric_name = identifier
-
     @property
     def path(self) -> str:
         """
         # Summary
 
-        Build the endpoint path with fabric name, capableSwitches suffix, and required query string.
+        Validate the required query parameters, then delegate to `_EpManageFabricsBase.path` to build the path with the
+        fabric name, `capableSwitches` suffix, and query string.
 
         ## Raises
 
         ### ValueError
 
-        - If `fabric_name` is not set
         - If `interface_type` is not set
         - If `mode` is not set
+        - If `fabric_name` is not set
         """
-        if self.fabric_name is None:
-            raise ValueError(f"{type(self).__name__}.path: fabric_name must be set before accessing path.")
         if self.endpoint_params.interface_type is None:
             raise ValueError(f"{type(self).__name__}.path: endpoint_params.interface_type must be set before accessing path.")
         if self.endpoint_params.mode is None:
             raise ValueError(f"{type(self).__name__}.path: endpoint_params.mode must be set before accessing path.")
-        base_path = BasePath.path("fabrics", self.fabric_name, self._path_suffix)
-        return f"{base_path}?{self.endpoint_params.to_query_string()}"
+        return super().path
 
     @property
     def verb(self) -> HttpVerbEnum:
--- a/plugins/module_utils/endpoints/v1/manage/manage_fabric_capable_switches.py
+++ b/plugins/module_utils/endpoints/v1/manage/manage_fabric_capable_switches.py
@@ -21,7 +21,7 @@ stable for the lifetime of ND 4.2.x. See GitHub issue #273 for context and risk 
 
 from __future__ import annotations
 
-from typing import ClassVar, Literal, Optional
+from typing import ClassVar, Literal
 
 from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import Field
 from ansible_collections.cisco.nd.plugins.module_utils.endpoints.base import NDEndpointBaseModel
@@ -60,12 +60,12 @@ class CapableSwitchesEndpointParams(EndpointQueryParams):
     None
     """
 
-    interface_type: Optional[str] = Field(
+    interface_type: str | None = Field(
         default=None,
         min_length=1,
         description="Interface type (e.g. loopback, ethernet, portChannel, svi, tunnel). Sent as interfaceType.",
     )
-    mode: Optional[str] = Field(
+    mode: str | None = Field(
         default=None,
         min_length=1,
         description="Interface mode (e.g. managed, trunk, access, routed).",
--- a/plugins/module_utils/endpoints/v1/manage/manage_fabric_capable_switches.py
+++ b/plugins/module_utils/endpoints/v1/manage/manage_fabric_capable_switches.py
@@ -1,0 +1,157 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2026, Allen Robel (@allenrobel) <arobel@cisco.com>
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+"""
+ND Manage Fabrics capableSwitches endpoint model.
+
+Endpoint definition for the unpublished but stable
+``GET /api/v1/manage/fabrics/{fabric_name}/capableSwitches`` resource.
+
+## Endpoints
+
+- ``EpManageFabricsCapableSwitchesGet`` - Get switches in a fabric capable of hosting a given
+  ``interfaceType`` and ``mode``
+  (``GET /api/v1/manage/fabrics/{fabric_name}/capableSwitches?interfaceType={type}&mode={mode}``)
+
+This endpoint is not present in the ND OpenAPI spec. Cisco ND developers have confirmed it is
+stable for the lifetime of ND 4.2.x. See GitHub issue #273 for context and risk discussion.
+"""
+
+from __future__ import annotations
+
+from typing import ClassVar, Literal, Optional
+
+from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import Field
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.base import NDEndpointBaseModel
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.mixins import FabricNameMixin
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.query_params import EndpointQueryParams
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.v1.manage.base_path import BasePath
+from ansible_collections.cisco.nd.plugins.module_utils.enums import HttpVerbEnum
+from ansible_collections.cisco.nd.plugins.module_utils.types import IdentifierKey
+
+
+class CapableSwitchesEndpointParams(EndpointQueryParams):
+    """
+    # Summary
+
+    Endpoint-specific query parameters for the capableSwitches endpoint.
+
+    ## Parameters
+
+    - interface_type: Interface type to query for capable switches (e.g. `loopback`, `ethernet`, `portChannel`, `svi`, `tunnel`).
+      Sent to the API as `interfaceType` (camelCase).
+    - mode: Mode for the given interface type (e.g. `managed`, `trunk`, `access`, `routed`).
+
+    Valid `(interface_type, mode)` pairs are enforced by `InterfaceCapabilityPreflight`, not by this model — the endpoint layer
+    is intentionally a thin wire-format adapter.
+
+    ## Usage
+
+    ```python
+    params = CapableSwitchesEndpointParams(interface_type="loopback", mode="managed")
+    query_string = params.to_query_string()
+    # Returns: "interfaceType=loopback&mode=managed"
+    ```
+
+    ## Raises
+
+    None
+    """
+
+    interface_type: Optional[str] = Field(
+        default=None,
+        min_length=1,
+        description="Interface type (e.g. loopback, ethernet, portChannel, svi, tunnel). Sent as interfaceType.",
+    )
+    mode: Optional[str] = Field(
+        default=None,
+        min_length=1,
+        description="Interface mode (e.g. managed, trunk, access, routed).",
+    )
+
+
+class EpManageFabricsCapableSwitchesGet(FabricNameMixin, NDEndpointBaseModel):
+    """
+    # Summary
+
+    ND Manage Fabrics capableSwitches GET Endpoint
+
+    ## Description
+
+    Endpoint to retrieve the list of switches in a fabric that are capable of hosting interfaces of a given `interface_type` and
+    `mode`. Used by `InterfaceCapabilityPreflight` to fail fast with a clear aggregate error before per-switch configuration calls.
+
+    ## Path
+
+    - `/api/v1/manage/fabrics/{fabric_name}/capableSwitches?interfaceType={type}&mode={mode}`
+
+    ## Verb
+
+    - GET
+
+    ## Raises
+
+    ### ValueError
+
+    - If `fabric_name`, `interface_type`, or `mode` is not set when accessing `path`
+
+    ## Usage
+
+    ```python
+    ep = EpManageFabricsCapableSwitchesGet()
+    ep.fabric_name = "my-fabric"
+    ep.endpoint_params.interface_type = "loopback"
+    ep.endpoint_params.mode = "managed"
+    rest_send.path = ep.path
+    rest_send.verb = ep.verb
+    # Path: /api/v1/manage/fabrics/my-fabric/capableSwitches?interfaceType=loopback&mode=managed
+    ```
+    """
+
+    class_name: Literal["EpManageFabricsCapableSwitchesGet"] = Field(
+        default="EpManageFabricsCapableSwitchesGet",
+        description="Class name for backward compatibility",
+    )
+
+    _require_fabric_name: ClassVar[bool] = True
+    _path_suffix: ClassVar[str] = "capableSwitches"
+
+    endpoint_params: CapableSwitchesEndpointParams = Field(
+        default_factory=CapableSwitchesEndpointParams,
+        description="Endpoint-specific query parameters (interface_type, mode).",
+    )
+
+    def set_identifiers(self, identifier: IdentifierKey = None):
+        """Set the fabric_name from the identifier value."""
+        self.fabric_name = identifier
+
+    @property
+    def path(self) -> str:
+        """
+        # Summary
+
+        Build the endpoint path with fabric name, capableSwitches suffix, and required query string.
+
+        ## Raises
+
+        ### ValueError
+
+        - If `fabric_name` is not set
+        - If `interface_type` is not set
+        - If `mode` is not set
+        """
+        if self.fabric_name is None:
+            raise ValueError(f"{type(self).__name__}.path: fabric_name must be set before accessing path.")
+        if self.endpoint_params.interface_type is None:
+            raise ValueError(f"{type(self).__name__}.path: endpoint_params.interface_type must be set before accessing path.")
+        if self.endpoint_params.mode is None:
+            raise ValueError(f"{type(self).__name__}.path: endpoint_params.mode must be set before accessing path.")
+        base_path = BasePath.path("fabrics", self.fabric_name, self._path_suffix)
+        return f"{base_path}?{self.endpoint_params.to_query_string()}"
+
+    @property
+    def verb(self) -> HttpVerbEnum:
+        """Return the HTTP verb for this endpoint."""
+        return HttpVerbEnum.GET
--- a/plugins/module_utils/interface_capability_preflight.py
+++ b/plugins/module_utils/interface_capability_preflight.py
@@ -1,0 +1,272 @@
+# Copyright: (c) 2026, Allen Robel (@allenrobel) <arobel@cisco.com>
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+"""
+Interface capability preflight for per-policy interface orchestrators.
+
+Provides `InterfaceCapabilityPreflight`, a lazy-loaded cache + validator that pre-flights
+which switches in a fabric are capable of hosting a given `interface_type` and `mode`. Used
+by `NDBaseInterfaceOrchestrator.validate_switches_capable` to convert confusing ND HTTP 400s
+on incapable switches into a single aggregate error naming the offending switches.
+
+Uses the unpublished `/api/v1/manage/fabrics/{fabric_name}/capableSwitches?interfaceType=...&mode=...`
+endpoint. See GitHub issue #273 for stability/risk discussion.
+"""
+
+from __future__ import annotations
+
+from typing import ClassVar
+
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.v1.manage.manage_fabric_capable_switches import (
+    EpManageFabricsCapableSwitchesGet,
+)
+from ansible_collections.cisco.nd.plugins.module_utils.enums import HttpVerbEnum
+from ansible_collections.cisco.nd.plugins.module_utils.fabric_context import FabricContext
+from ansible_collections.cisco.nd.plugins.module_utils.rest.rest_send import RestSend
+
+
+class InterfaceCapabilityPreflight:
+    """
+    # Summary
+
+    Cached interface capability validator for per-policy interface orchestrators.
+
+    Lazily fetches the set of switches capable of hosting a given `(interface_type, mode)` pair from
+    the ND Manage `capableSwitches` endpoint, caches by `(interface_type, mode)`, and validates a
+    user-supplied set of switch IDs against that capable set. On failure, raises `RuntimeError` with a
+    single aggregate message naming the offending switches (`switch_ip` / `switch_name` / `model` when
+    `fabric_context` is injected).
+
+    The `(interface_type, mode)` taxonomy is captured experimentally (ND 4.2, 2026-05-11) and enforced
+    client-side before any API call — invalid combos raise `ValueError` without consuming a round trip.
+
+    ## Taxonomy (captured ND 4.2, 2026-05-11)
+
+    | interface_type | valid modes                                                              |
+    |----------------|--------------------------------------------------------------------------|
+    | `loopback`     | `managed`                                                                |
+    | `svi`          | `managed`                                                                |
+    | `tunnel`       | `managed`                                                                |
+    | `ethernet`     | `trunk`, `access`, `routed`, `fex`, `pvlan`, `dot1qTunnel`, `unmanaged`  |
+    | `portChannel`  | `trunk`, `access`, `routed`, `fex`, `pvlan`, `dot1qTunnel`, `unmanaged`  |
+
+    `vpc` interfaces use a separate endpoint (`/api/v1/manage/fabrics/{fabric}/vpcPairs?view=intendedPairs`)
+    and are handled by a sibling validator, not this class. `breakout` and `subinterface` have no
+    pre-check endpoint and are out of scope.
+
+    ## Raises
+
+    ### RuntimeError
+
+    - Via `validate` when one or more switch IDs are not in the capable set for `(interface_type, mode)`.
+    - Via `get_capable_switches` if the underlying GET request fails (other than 404).
+
+    ### ValueError
+
+    - Via any method that takes `(interface_type, mode)` when the pair is not in the supported taxonomy.
+    """
+
+    _TAXONOMY: ClassVar[dict[str, frozenset[str]]] = {
+        "loopback": frozenset({"managed"}),
+        "svi": frozenset({"managed"}),
+        "tunnel": frozenset({"managed"}),
+        "ethernet": frozenset({"trunk", "access", "routed", "fex", "pvlan", "dot1qTunnel", "unmanaged"}),
+        "portChannel": frozenset({"trunk", "access", "routed", "fex", "pvlan", "dot1qTunnel", "unmanaged"}),
+    }
+
+    def __init__(self, rest_send: RestSend, fabric_name: str, fabric_context: FabricContext | None = None):
+        """
+        # Summary
+
+        Initialize the preflight with a `RestSend` instance, fabric name, and optional `FabricContext` for error enrichment.
+
+        ## Parameters
+
+        - rest_send: configured `RestSend` (sender, response handler, params already wired).
+        - fabric_name: target fabric for the capability query.
+        - fabric_context: optional `FabricContext` used to enrich error messages with `switch_ip` / `switch_name` / `model`.
+
+        ## Raises
+
+        None
+        """
+        self._rest_send = rest_send
+        self._fabric_name = fabric_name
+        self._fabric_context = fabric_context
+        self._cache: dict[tuple[str, str], list[dict]] = {}
+
+    @classmethod
+    def supported_modes(cls, interface_type: str) -> frozenset[str]:
+        """
+        # Summary
+
+        Return the set of valid modes for a given `interface_type` from the taxonomy.
+
+        ## Raises
+
+        ### ValueError
+
+        - If `interface_type` is not present in the taxonomy.
+        """
+        try:
+            return cls._TAXONOMY[interface_type]
+        except KeyError as e:
+            supported = ", ".join(sorted(cls._TAXONOMY))
+            raise ValueError(f"Unsupported interface_type '{interface_type}'. Supported interface types: {supported}.") from e
+
+    @classmethod
+    def _check_taxonomy(cls, interface_type: str, mode: str) -> None:
+        """
+        # Summary
+
+        Validate that `(interface_type, mode)` is a supported pair in the taxonomy. Raises before any API call.
+
+        ## Raises
+
+        ### ValueError
+
+        - If `interface_type` is not in the taxonomy.
+        - If `mode` is not valid for the given `interface_type`.
+        """
+        valid_modes = cls.supported_modes(interface_type)
+        if mode not in valid_modes:
+            modes = ", ".join(sorted(valid_modes))
+            raise ValueError(f"Unsupported mode '{mode}' for interface_type '{interface_type}'. Supported modes: {modes}.")
+
+    @property
+    def fabric_name(self) -> str:
+        """
+        # Summary
+
+        Return the fabric name this preflight was created for.
+
+        ## Raises
+
+        None
+        """
+        return self._fabric_name
+
+    def _query_get(self, path: str) -> dict:
+        """
+        # Summary
+
+        Issue a GET request via `RestSend` and return the `DATA` dict. Returns `{}` on HTTP 404.
+
+        ## Raises
+
+        ### RuntimeError
+
+        - If the request fails with any non-success status other than 404.
+        """
+        self._rest_send.path = path
+        self._rest_send.verb = HttpVerbEnum.GET
+        self._rest_send.commit()
+        if self._rest_send.return_code == 404:
+            return {}
+        if not self._rest_send.success:
+            raise RuntimeError(f"GET {path} failed {self._rest_send.error_summary}")
+        return self._rest_send.response_current.get("DATA", {})
+
+    def get_capable_switches(self, interface_type: str, mode: str) -> list[dict]:
+        """
+        # Summary
+
+        Return the list of switch records capable of hosting `(interface_type, mode)` in this fabric. Each record contains
+        at least `switchId`, `switchName`, and `model`. Lazily fetched and cached per `(interface_type, mode)`.
+
+        ## Raises
+
+        ### ValueError
+
+        - If `(interface_type, mode)` is not a supported pair in the taxonomy.
+
+        ### RuntimeError
+
+        - If the underlying GET request fails (other than 404).
+        """
+        self._check_taxonomy(interface_type, mode)
+        key = (interface_type, mode)
+        if key in self._cache:
+            return self._cache[key]
+        ep = EpManageFabricsCapableSwitchesGet()
+        ep.fabric_name = self._fabric_name
+        ep.endpoint_params.interface_type = interface_type
+        ep.endpoint_params.mode = mode
+        result = self._query_get(ep.path)
+        switches = (result.get("switches") or []) if result else []
+        self._cache[key] = switches
+        return switches
+
+    def get_capable_switch_ids(self, interface_type: str, mode: str) -> set[str]:
+        """
+        # Summary
+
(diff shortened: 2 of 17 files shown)


plugins/module_utils/endpoints/v1/manage/manage_fabric_capable_switches.py
```diff
@@ -0,0 +1,147 @@
...
+        - If `interface_type` is not set
+        - If `mode` is not set
+        - If `fabric_name` is not set
+        """
+        if self.endpoint_params.interface_type is None:
+            raise ValueError(f"{type(self).__name__}.path: endpoint_params.interface_type must be set before accessing path.")
+        if self.endpoint_params.mode is None:
+            raise ValueError(f"{type(self).__name__}.path: endpoint_params.mode must be set before accessing path.")
+        return super().path
+
```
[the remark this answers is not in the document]
allenrobel: Will open an issue for this.

plugins/module_utils/orchestrators/loopback_interface.py
```diff
@@ -97,6 +98,7 @@ def create(self, model_instance: LoopbackInterfaceModel, **kwargs) -> ResponseTy
         - If the create API request fails.
         """
         try:
+            self.validate_switches_capable([model_instance])
             switch_id = self._resolve_switch_id(model_instance.switch_ip)
             api_endpoint = self._configure_endpoint(self.create_endpoint(), switch_sn=switch_id)
```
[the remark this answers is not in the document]
allenrobel: The comment is stale — most likely written against an intermediate state of the PR before _resolve_mode() was committed, or the override at lines 87-98 was missed.

plugins/module_utils/orchestrators/base_interface.py
```diff
@@ -108,6 +118,80 @@ def _resolve_switch_id(self, switch_ip: str) -> str:
...
+        ## Raises
+
+        ### AttributeError
+
+        - If the model has no `mode` attribute and the subclass has not overridden this method.
+        """
+        # getattr (not direct access): ModelType is bound to NDBaseModel, which has no `mode`. The
+        # attribute is duck-typed -- present on most interface models, absent on some (e.g. loopback).
+        return getattr(model_instance, "mode")
+
```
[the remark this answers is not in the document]
allenrobel: Good thing to flag, though the specifics need a correction.

`_resolve_mode()` doesn't crash for loopback: `LoopbackInterfaceOrchestrator` overrides it (`return "managed"`), so the base `getattr(model_instance, "mode")` is never reached. `validate_switches_capable()` is also a no-op unless an orchestrator opts in via `interface_type`, and loopback is the only orchestrator that does today.

But checking the other interface branches (`nd_interface_svi`, `nd_interface_ethernet_access`) showed the problem is actually broader than "some models." `mode` is nested under `config_data` on **every** interface model — `LoopbackConfigDataModel.mode`, `SviConfigDataModel.mode`, `EthernetAccessConfigDataModel.mode` — and none expose a top-level `mode`. So the base `getattr(model_instance, "mode")` default worked for zero real models, and the inline comment claiming `mode` is "present on most interface models" was simply wrong.

It's also a per-interface-type constant, not per-instance data: each is a frozen `Literal` (`"managed"`, `"managed"`, `"access"`). The loopback override already reflected that — it ignored its `model_instance` argument entirely and returned a hardcoded string.

Given that, a nested-field fallback in the base (option 1) isn't the right fix: `config_data` is `Optional` on every model, so a probe would `NoneType`-crash, and it would be resolving per-instance what is really a class-level constant.

Fixed in 806a39a — promoted `mode` to a ClassVar alongside `interface_type`:

- `base_interface.py`: added `interface_mode: ClassVar[str] = ""`; deleted `_resolve_mode()`. `validate_switches_capable()` reads `self.interface_mode` directly and raises a clear `RuntimeError` if an orchestrator sets `interface_type` without `interface_mode` — the override contract is now enforced structurally instead of by docstring.
- `loopback_interface.py`: sets `interface_mode = "managed"`; override removed.
- Added base-level tests for `validate_switches_capable` (opted-out no-op; half-configured error) — it had no coverage before.

svi/ethernet orchestrators will just set both ClassVars when they adopt capability preflight; there's no `_resolve_mode()` to override and no `AttributeError` surface left.

plugins/module_utils/orchestrators/loopback_interface.py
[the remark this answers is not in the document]
allenrobel: 806a39a removes `_resolve_mode()` entirely. `mode` is now an `interface_mode` ClassVar set per orchestrator (`"managed"` for loopback), so there's no `model_instance.mode` access and no `AttributeError` surface left. Verifying against the svi/ethernet branches showed `mode` is nested under `config_data` on every interface model, not just loopback — full rationale is in the thread on `base_interface.py` (discussion_r3290895317).

plugins/module_utils/interface_capability_preflight.py
```diff
@@ -0,0 +1,300 @@
...
+
+    - Via `validate` when one or more switch IDs are not in the capable set for `(interface_type, mode)`.
+    - Via `get_capable_switches` if the underlying GET request fails.
+
+    ### ValueError
+
+    - Via any method that takes `(interface_type, mode)` when the pair is not in the supported taxonomy.
+    """
+
+    _TAXONOMY: ClassVar[dict[str, frozenset[str]]] = {
```
samiib, line 71: Since this captured experimentally from an undocumented API, should we add the ability to allow unknown types and modes? Additional values could be supported in future versions but this validation would still block it.

plugins/module_utils/orchestrators/base_interface.py
```diff
@@ -108,6 +119,94 @@ def _resolve_switch_id(self, switch_ip: str) -> str:
...
+        # Summary
+
+        Emit a warning through the underlying `AnsibleModule` so it surfaces in `ansible-playbook` output. Used to soften
+        non-fatal preflight failures (see `validate_switches_capable`) without converting them into module errors.
+
+        ## Raises
+
+        None
+        """
+        self.rest_send.sender.ansible_module.warn(message)  # type: ignore[attr-defined]
```
samiib: Are there any contexts where `ansible_module` may not be defined? eg in tests?
Perhaps it would be good to use a getattr guard or a method in RestSend instead?

allenrobel: Good catch — surfacing the seam was the right call. Reworked in a3c502a8 to route through a new `RestSend.warn(message)` method.

Walking through the trade-offs we considered:

- **Direct call (what's here now in 215601a)** — minimal code, loud failure if `ansible_module` is missing. But it reaches three levels deep (`rest_send.sender.ansible_module.warn`) and needs the `# type: ignore` because `SenderProtocol` doesn't declare `ansible_module`.
- **`getattr` guard on `_warn`** — the most defensive, but silently swallowing warnings would defeat the whole #302 fix (the point of the soften-in-check-mode path is to surface the information, not hide it). It would also hide future test-wiring bugs.
- **Encapsulate on `RestSend.warn`** — best Law-of-Demeter posture; the orchestrator just talks to `RestSend`. One localized `# type: ignore` lives inside `RestSend.warn` instead of being repeated at every future call site. Reusable from any orchestrator. `RestSend` is already the boundary between orchestrator and connection, so adding `warn` there is on-theme.

Went with option 3. Behavior is unchanged — the call still routes to `AnsibleModule.warn`, and production always has a valid `ansible_module` (sender_nd.py raises on access if it's unset, so a misconfiguration surfaces loudly rather than being swallowed). Tests use `MockAnsibleModule` which already has `warn()` for the check-mode assertion.

If we ever need to tolerate a Sender without `ansible_module` (e.g. a hypothetical async sender), the right place to add that fallback is inside `RestSend.warn`, not at every call site — but I'd leave that until there's a real second consumer.

plugins/module_utils/interface_capability_preflight.py
allenrobel: Real concern, worth thinking through. Sketching where I land and the options, but happy to swing the other way if you'd rather.

**Why I left it strict by default:** the taxonomy isn't gating end-user input — it's gating what the *orchestrator author* sets as ClassVars. Each per-policy orchestrator subclass sets `interface_type` + `interface_mode` itself; the user never types these. So strict validation is really catching typos and half-configured subclasses at construction time, before the first round trip to ND. The same fail-fast philosophy is already in `validate_switches_capable` (raises if `interface_type` is set without `interface_mode`, etc.).

A new ND release that adds a new `(interface_type, mode)` combo also needs a new orchestrator subclass to expose it — so the taxonomy update naturally rides along in the same PR as the orchestrator. There's no scenario today where a user is "blocked" from a new ND feature: they'd be blocked by the absence of the orchestrator, not by the taxonomy.

**Where strict bites:** the case you're flagging — ND ships a new value, the taxonomy hasn't been updated yet, and someone wants to ship an out-of-tree orchestrator (or hot-patch in their playbook environment) before the next collection release. Today they'd have to monkey-patch `_TAXONOMY`.

**Three options if we want to soften it:**

1. **Drop the taxonomy entirely.** Let the `capableSwitches` GET be the source of truth: if ND returns 400, we surface it. Loses the no-round-trip rejection of typos, but is the simplest forward-compat story.
2. **Class-level escape hatch.** Add `extra_modes: ClassVar[frozenset[str]] = frozenset()` (and matching `extra_types`) that subclasses can extend without touching `_TAXONOMY`. Strict by default; opt-in widening per orchestrator. Keeps typo protection for the common path.
3. **Warn instead of raise on unknown pairs.** `_check_taxonomy` would emit a deprecation-style warning via the new `RestSend.warn` and proceed with the GET. Best of both worlds for end users but blurs the contract — typos in well-known orchestrators would now slip past local validation.

My default is to keep it strict (current behavior) — every interface module we've shipped or planned has fit the taxonomy, and the orchestrator-PR cadence handles new ND values cleanly. If you'd like a softer stance, option 2 is the smallest blast radius and the easiest to reverse later; option 3 would be my second choice.

What's your read?

samiib: That sounds good to me. Without the extra context, I was thinking this gets exposed at a user level. If it requires a new class to support it then this is good.

samiib (approved):
LGTM

plugins/module_utils/orchestrators/base_interface.py
```diff
@@ -108,6 +119,95 @@ def _resolve_switch_id(self, switch_ip: str) -> str:
...
+
+        Emit a warning through `RestSend.warn`, which routes to the underlying `AnsibleModule.warn` so it surfaces in
+        `ansible-playbook` output. Used to soften non-fatal preflight failures (see `validate_switches_capable`) without
+        converting them into module errors.
+
+        ## Raises
+
+        None
+        """
+        self.rest_send.warn(message)
```
akinross: what is the value add of _warn?
```suggestion
             self.rest_send.warn(f"Capability preflight skipped in check mode: {e}")
```

akinross (changes requested):

plugins/module_utils/orchestrators/base_interface.py
allenrobel: Good call — `_warn` was a one-line pass-through with a single internal call site, and the encapsulation boundary already lives at `RestSend.warn` (where the `ansible_module` access / `# type: ignore` got localized when samiib raised the seam earlier in this file). The orchestrator-level wrapper was a redundant second layer.

Applied your suggestion in dcdaac8e: inlined `self.rest_send.warn(...)` in `validate_switches_capable` and removed the method. Also dropped the two stale `_warn()` lines from the test docstrings — the tests already drive the public `validate_switches_capable()` path and assert on `rest_send.sender.ansible_module.warnings`, so they pass unchanged (26/26).

plugins/module_utils/interface_capability_preflight.py
```diff
@@ -0,0 +1,300 @@
...
+        ### RuntimeError
+
+        - If the underlying GET request fails.
+        """
+        self._check_taxonomy(interface_type, mode)
+        key = (interface_type, mode)
+        if key not in self._cache:
+            result = self._query_get(self._build_endpoint(interface_type, mode).path)
+            self._cache[key] = result.get("switches") or []
+        return list(self._cache[key])
```
gmicol: `get_capable_switches()` returns `list(self._cache[key])`, which only copies the outer list. The cached switch record dicts are still returned by reference, so a caller can mutate a returned record and corrupt the cached state, e.g. changing `records[0]["switchId"]` before `get_capable_switch_ids()` is built.

That also means the defensive-copy test only covers list/set mutation, not mutation of the records themselves. Can we return copied records here, such as `[dict(sw) for sw in self._cache[key]]` or `copy.deepcopy(...)` if nested response data is expected, and add a regression test that mutates a returned switch dict?

gmicol (changes requested):

plugins/module_utils/orchestrators/base_interface.py
```diff
@@ -108,6 +119,81 @@ def _resolve_switch_id(self, switch_ip: str) -> str:
...
+        `RuntimeError` naming every offending switch.
+
+        When `interface_type` is `""` (default on the base class) this method is a no-op — subclasses opt in by setting
+        both the `interface_type` and `interface_mode` ClassVars.
+
+        All `switch_ip` values are resolved before the capability check runs; if any are unresolvable, a single aggregate
+        `RuntimeError` names every unknown IP so a typo on one entry does not mask resolution or capability problems on
+        the remaining entries (issue #301).
+
+        In `--check` mode the capability GET is still issued, but a failure of the capability check itself — whether an
```
mikewiebe, line 174: **Issue**

The new base-interface method says the capability GET still runs in `--check` mode and downgrades failures to warnings, but the loopback wiring calls `validate_switches_capable()` only inside `create`, `update`, and `create_bulk`. The ND state machine skips those operations entirely in check mode, so the capability preflight is skipped too.

The delete paths also do not call `validate_switches_capable()`, so `state: deleted` and overridden delete work do not get the same preflight behavior.

**Evidence**

- [`NDStateMachine._execute_operation`](https://github.com/CiscoDevNet/ansible-nd/blob/ed0aa6395d969617607f8fedce75e02a5550c341/plugins/module_utils/nd_state_machine.py#L111) returns without invoking the operation when `check_mode` is true.
- [`LoopbackInterfaceOrchestrator.create`](https://github.com/CiscoDevNet/ansible-nd/pull/275/files#diff-bfbfe608d8199f5d7c947a889c0a9b2acfa286b562db44bee753e947e6707316R102), [`update`](https://github.com/CiscoDevNet/ansible-nd/pull/275/files#diff-bfbfe608d8199f5d7c947a889c0a9b2acfa286b562db44bee753e947e6707316R128), and [`create_bulk`](https://github.com/CiscoDevNet/ansible-nd/pull/275/files#diff-bfbfe608d8199f5d7c947a889c0a9b2acfa286b562db44bee753e947e6707316R172) are the only loopback methods that call `validate_switches_capable()`.
- [`delete`](https://github.com/CiscoDevNet/ansible-nd/pull/275/files#diff-bfbfe608d8199f5d7c947a889c0a9b2acfa286b562db44bee753e947e6707316R140) and [`delete_bulk`](https://github.com/CiscoDevNet/ansible-nd/pull/275/files#diff-bfbfe608d8199f5d7c947a889c0a9b2acfa286b562db44bee753e947e6707316R193) resolve switches and queue removals without the capability preflight.
- [`nd_interface_loopback.py`](https://github.com/CiscoDevNet/ansible-nd/blob/ed0aa6395d969617607f8fedce75e02a5550c341/plugins/modules/nd_interface_loopback.py#L311) calls `manage_state()` directly; there is no module-level preflight call between state-machine initialization and state execution.
- [`NDBaseInterfaceOrchestrator.validate_switches_capable()`](https://github.com/CiscoDevNet/ansible-nd/pull/275/files#diff-1bdd24be418d1be85dcf48e8bd16da0b84e1c8aaaf0d001caf00e693ed4f6322R157) documents check-mode GET behavior, and [the tests](https://github.com/CiscoDevNet/ansible-nd/pull/275/files#diff-96f8720d140c351bc1b40917120644a7a4117109b27c0524839be8266880f092R892) call the method directly in check mode. They do not exercise the full module/state-machine path where operations are skipped.

**Impact**

This undercuts one of issue #273’s design choices: the user should get capability feedback during dry-run before play execution. In the current wiring, a playbook can pass `--check` without a capability warning, then fail later in normal execution. It also makes the preflight behavior state-dependent in a way users will not expect.

**Suggested fix**

Move the capability preflight to a state-machine-visible preflight point that runs before mutation operations are skipped in check mode. For example, add an orchestrator hook that validates `self.proposed` after proposed models are built, or explicitly call `validate_switches_capable()` from the loopback module/state-machine path before `manage_state()`. Include module/state-machine-level tests proving check mode emits the warning and that deleted/overridden delete paths get the intended behavior or are explicitly documented as out of scope.

plugins/module_utils/orchestrators/base_interface.py
allenrobel: Resolving conversation after two weeks.

plugins/module_utils/orchestrators/base_interface.py
allenrobel: Good catch — confirmed and fixed in 8a096389.

You were right that the preflight never fired in check mode: it lived inside the loopback `create`/`update`/`create_bulk` methods, and `NDStateMachine._execute_operation` returns early in check mode, so those were never invoked. Rather than wire it per-module, I moved it to a state-machine-visible hook:

- `NDBaseOrchestrator.preflight(model_instances)` — new no-op pre-mutation hook (no-op for non-interface orchestrators).
- `NDBaseInterfaceOrchestrator.preflight()` overrides it to delegate to `validate_switches_capable()`.
- `NDStateMachine.manage_state()` calls `self.model_orchestrator.preflight(list(self.proposed))` **before** `_manage_create_update_state()`, i.e. ahead of the check-mode skip. Every interface module inherits this the moment it sets the `interface_type`/`interface_mode` ClassVars — no per-module duplication.
- Removed the now-redundant per-op calls in loopback; the capability GET runs once, up front.

On the delete paths: I left `state: deleted` (and the delete half of `overridden`) **out of scope** — removing configuration doesn't depend on a switch's capability to host the interface type. That's documented in `manage_state()` and asserted by a test.

New module/state-machine-level coverage in `tests/unit/module_utils/test_nd_state_machine.py` drives the full `manage_state()` path and proves: check-mode `merged` calls `preflight` while `create_bulk` is skipped; the normal-mode contrast calls both (preflight first); `deleted` does not preflight; `overridden` preflights the proposed set. Full unit suite (1210) green; loopback integration green.

plugins/module_utils/interface_capability_preflight.py
allenrobel: Good catch — fixed in 8c1ce044.

You're right: `list(self._cache[key])` only copied the outer list, so the record dicts were shared by reference and a caller mutating `records[0]["switchId"]` would corrupt the cache (and the `switchId` set `get_capable_switch_ids()` derives from it afterward). Switched to `copy.deepcopy(self._cache[key])` — went with deepcopy rather than `[dict(sw) for sw in ...]` so we're robust to any nested response fields ND might add, and the records are small/cached so the cost is negligible. Docstring updated to say the returned list *and its records* are safe to mutate.

Added a regression test (`test_interface_capability_preflight_00320`) that mutates returned record dicts in place (`["switchId"] = "HACKED"`, `del [...]["model"]`, etc.) and asserts both a subsequent `get_capable_switches()` and the post-mutation `get_capable_switch_ids()` still return the original cached values. Confirmed it fails on the old shallow copy (cache reads back `HACKED`) and passes with the fix. Full unit suite green (1211).

mikewiebe (approved):
LGTM

akinross (approved):

akinross (approved):

gmicol (approved):

allenrobel:
Merged after 3 approvals.  Thanks!

PR #278, opened 2026-05-13
nd_vpc_pair

allenrobel:
## Related Issue(s)

Sivakami has an open PR [197 - VPC Pair 4.x Implementation](https://github.com/CiscoDevNet/ansible-nd/pull/197) for VPC pairing.  I needed this to implement VPC interface modules, but this has not been merged into develop, and it does not currently leverage Gaspard's framework.  Due to this, I asked Claude to write a VPC pair module that I could stack my VPC interface PRs (both will be created today) onto.  I want to acknowledge Sivakami's work and did ask Claude to ensure nd_vpc_pair is functionally equivalent to it (so the PR did serve as a baseline reference from a functionality perspective).  I'll  reach out to Sivakami to explain this situation and ask if this is OK with her.

## Proposed Changes

- Adds the `nd_vpc_pair` module, which manages vPC pair lifecycle (pair / unpair) between two switches in a fabric via the ND `PUT /api/v1/manage/fabrics/{fabric}/switches/{sn}/vpcPair` endpoint.
- Adds the shared `VpcBaseOrchestrator` (`plugins/module_utils/orchestrators/vpc_base.py`) that every `nd_interface_vpc_*` module will reuse for CRUD, peer-serial resolution, per-interface DELETE, and `query_all` dedupe.
- Adds endpoint models: `EpManageFabricSwitchVpcPairGet`, `EpManageFabricVpcPairsListGet`, `EpManageFabricSwitchActionsDeploy`, and `EpManageInterfacesDelete`.
- Adds the `VpcPairModel` Pydantic model with `vpcPairDetails` subtree and `identifier_strategy = "single"` on `switch_ip`.

Three ND 4.2.1 wire quirks are accommodated and tagged `TODO(4.2.1)`:

- `interfaceActions/remove` returns "Invalid Interface" for vPC entries; vPC delete uses the per-interface DELETE endpoint instead.
- ND echoes each vPC interface from BOTH peer switches in their per-switch `/interfaces` GET; the orchestrator dedupes by `interfaceName`, keeping the alphabetically-lower `switchId` as canonical.
- These are documented inline at each workaround site and audited by `/nd-workaround-audit`.

## Test Notes

- 69 new unit tests across model, orchestrator, and 3 endpoint files: \`python -m pytest tests/unit/module_utils/{models,orchestrators,endpoints}/ -q\` (926 total pass, no regressions).
[...]

PR #197, opened 2026-03-09
VPC Pair 4.x Implementation

sivakasi-cisco:
https://github.com/CiscoDevNet/ansible-nd/issues/230

vPC Pair support for ND 4.x

- nd_manage_vpc_pair can create, update, delete, override, and gather vPC pairs.

- The big module logic was split into focused helper files (query, validation, actions, deploy, etc.)

- We added dedicated vPC Pair endpoints/models/schemas and connected them through the orchestrator + state machine flow.

- Integration tests were added across all main scenarios (merge/replace/override/delete/gather).

PR #279, opened 2026-05-13
nd_interface_vpc_access

allenrobel:
## Related Issue(s)

Stacked on #334 (`nd_interface_vpc_base`), which carries the shared vPC-interface infrastructure that previously lived on the now-closed #278 (`nd_vpc_pair`). #278 was closed in favor of #197 (`nd_manage_vpc_pair`); the shared interface base was re-homed onto #334 so this module can reach `develop` independently.

- CiscoDevNet/ansible-nd#322

Code-review findings (`/code-review high`) deferred from this PR, each tracked as a GitHub issue:

- #355 — vPC `peer1`/`peer2` member-port orientation may not track the user-supplied `switch_ip` (needs ND lab verification; shared with #280).
- #356 — vPC name-only identifier breaks when two pairs reuse the same vPC id in one fabric (shared with #280).
- #350 — `config_data` is optional on create, so a policy-less config item yields a late `RuntimeError` rather than a clear validation error (matches the sibling interface models).
- #348 — Lift duplicated interface-orchestrator CRUD + policy-type serializer into shared bases (this module's `normalize_member_ports` / `serialize_policy` are sibling-parity copies).
- #349 — `query_all` fans out one GET per switch; no fabric-wide interface-list endpoint.
- #354 — vPC interface delete queues a deploy even when the per-interface DELETE was a 404 no-op (inherited from `VpcInterfaceBaseOrchestrator`).

## Proposed Changes

- Adds the `nd_interface_vpc_access` module, which manages vPC `accessVpcHost` interfaces (template `int_vpc_access_host`) on Cisco Nexus Dashboard.
- Adds `AccessVpcHostPolicyTypeEnum` to `plugins/module_utils/models/interfaces/enums.py` so the model can frozen-pin its `policy_type`.
- Inherits all CRUD, peer-serial resolution, per-interface DELETE, and `query_all` dedupe from `VpcInterfaceBaseOrchestrator` (introduced in the stack-parent PR #334 `nd_interface_vpc_base`). Peer-serial resolution uses develop's `EpVpcPairGet` endpoint.

Access-specific design (lab-verified on ND 4.2.1):

- Single user-facing `access_vlan` field fans out to `peer1AccessVlan` / `peer2AccessVlan` at payload-serialization time via a `@model_serializer`. ND collapses both per-peer keys back to a single `accessVlan` in the GET response, so the round-trip matches the wire echo without further normalization. Tagged `TODO(4.2.1)`.
- Per-peer `peer1/peer2` `member_ports` / `port_channel_id` / `port_channel_description` / `port_channel_configuration` are exposed (the OpenAPI schema preserves these on read).
- `identifier_strategy = "single"` on `interface_name`; `switch_ip` is routing-only and excluded from diff/payload because ND echoes the same vPC record from BOTH peers' `/interfaces` GETs.

## Test Notes

- Unit tests across model + orchestrator + fixture JSON: `python -m pytest tests/unit/module_utils/models/test_vpc_access_interface.py tests/unit/module_utils/orchestrators/test_vpc_access_interface.py -q`. Full repo suite stays green (no regressions).
- Integration target `tests/integration/targets/nd_interface_vpc_access/` with `merged` / `replaced` / `overridden` / `deleted` state coverage, exercised against ND 4.2.1 on the SITE1 lab (S1_LE1 / S1_LE2): `ansible-test integration --docker nd_interface_vpc_access -vv`.
- The vPC pair substrate is created via `nd_manage_vpc_pair` using a **virtual** peer-link. Physical peer-link is currently broken at the ND controller (`configSave` 500 "Member port … is a member of VPC_PAIR"); see bug-vault `vpc-physical-peer-link-configsave-fails`. A `TODO(4.2.1)` marker in the test vars tracks reverting to physical once ND fixes it.
- `ansible-test sanity --docker default --python 3.11` passes on the changed files.
- Wire round-trip verified live: writing `peer1AccessVlan` + `peer2AccessVlan` returns 207 success; GET echoes a single `accessVlan` matching what the model exposes.

4.2.1

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

### DIFF
--- a/plugins/modules/nd_interface_vpc_access.py
+++ b/plugins/modules/nd_interface_vpc_access.py
@@ -341,6 +341,50 @@ EXAMPLES = r"""
                 - Ethernet1/2
     state: merged
 
+- name: Replace the configuration of a specific vPC interface
+  cisco.nd.nd_interface_vpc_access:
+    fabric_name: my_fabric
+    config:
+      - switch_ip: 192.168.1.1
+        interface_name: vpc100
+        config_data:
+          network_os:
+            policy:
+              admin_state: true
+              access_vlan: 20
+              peer1_port_channel_id: 100
+              peer1_member_ports:
+                - Ethernet1/3
+              peer2_port_channel_id: 100
+              peer2_member_ports:
+                - Ethernet1/3
+              port_channel_mode: active
+              peer1_port_channel_description: Reprovisioned Server-A on peer1
+              peer2_port_channel_description: Reprovisioned Server-A on peer2
+    state: replaced
+
+# state=overridden is fabric-wide: every vPC interface in the fabric that is managed by this module and is NOT
+# listed below is deleted. An empty config list deletes ALL such vPC interfaces in the fabric. Use with caution.
+- name: Enforce vPC interfaces fabric-wide, deleting all others managed by this module
+  cisco.nd.nd_interface_vpc_access:
+    fabric_name: my_fabric
+    config:
+      - switch_ip: 192.168.1.1
+        interface_name: vpc100
+        config_data:
+          network_os:
+            policy:
+              admin_state: true
+              access_vlan: 10
+              peer1_port_channel_id: 100
+              peer1_member_ports:
+                - Ethernet1/1
+              peer2_port_channel_id: 100
+              peer2_member_ports:
+                - Ethernet1/1
+              port_channel_mode: active
+    state: overridden
+
 - name: Delete a vPC interface
   cisco.nd.nd_interface_vpc_access:
     fabric_name: my_fabric
--- a/plugins/modules/nd_interface_vpc_access.py
+++ b/plugins/modules/nd_interface_vpc_access.py
@@ -264,15 +264,20 @@ options:
                     description:
                     - Unicast storm control level in packets per second (0-200000000).
                     type: int
-  deploy:
+  config_actions:
     description:
-    - Whether to deploy vPC interface changes after mutations are complete.
-    - When V(true), all queued vPC interface changes are deployed in a single bulk API call at the end of module
-      execution via the C(interfaceActions/deploy) API. Only the vPC interfaces modified by this task are deployed.
-    - When V(false), changes are staged but not deployed. Use a separate deploy module or task to deploy later.
-    - Setting O(deploy=false) is useful when batching changes across multiple interface tasks before a single deploy.
-    type: bool
-    default: true
+    - Controls deploy behavior after interface mutations are complete.
+    type: dict
+    suboptions:
+      deploy:
+        description:
+        - Whether to deploy vPC interface changes after mutations are complete.
+        - When V(true), all queued vPC interface changes are deployed in a single bulk API call at the end of module
+          execution via the C(interfaceActions/deploy) API. Only the vPC interfaces modified by this task are deployed.
+        - When V(false), changes are staged but not deployed. Use a separate deploy module or task to deploy later.
+        - Setting O(config_actions.deploy=false) is useful when batching changes across multiple interface tasks before a single deploy.
+        type: bool
+        default: true
   state:
     description:
     - The desired state of the network resources on the Cisco Nexus Dashboard.
@@ -357,7 +362,8 @@ EXAMPLES = r"""
               access_vlan: 10
               peer1_port_channel_id: 100
               peer2_port_channel_id: 100
-    deploy: false
+    config_actions:
+      deploy: false
     state: merged
 
 """
@@ -398,7 +404,12 @@ def main():
     argument_spec = nd_argument_spec()
     argument_spec.update(AccessVpcHostInterfaceModel.get_argument_spec())
     argument_spec.update(
-        deploy={"type": "bool", "default": True},
+        config_actions={
+            "type": "dict",
+            "options": {
+                "deploy": {"type": "bool", "default": True},
+            },
+        },
     )
 
     module = AnsibleModule(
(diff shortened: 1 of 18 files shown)


plugins/modules/nd_interface_vpc_access.py
```diff
@@ -0,0 +1,448 @@
...
+                    type: int
+                  storm_control_unicast_level:
+                    description:
+                    - Unicast storm control level in percentage (0.00-100.00).
+                    type: float
+                  storm_control_unicast_level_pps:
+                    description:
+                    - Unicast storm control level in packets per second (0-200000000).
+                    type: int
+  deploy:
```
mikewiebe: ### Medium: Replace top-level `deploy` with `config_actions`

**Issue**

**Evidence**

- [`plugins/modules/nd_interface_vpc_access.py` lines 267-289](https://github.com/CiscoDevNet/ansible-nd/pull/279/files#diff-40bfaf3d5e8818668dc14938b001b647d46bd8d9b37cee5ad7d5934e209f45f7R267) documents `deploy` as a standalone option and lists states without the standard `config_actions` option.

**Existing PR overlap**

**Impact**

**Suggested fix**

---

plugins/modules/nd_interface_vpc_access.py
```diff
@@ -0,0 +1,448 @@
...
+    description:
+    - Whether to deploy vPC interface changes after mutations are complete.
+    - When V(true), all queued vPC interface changes are deployed in a single bulk API call at the end of module
+      execution via the C(interfaceActions/deploy) API. Only the vPC interfaces modified by this task are deployed.
+    - When V(false), changes are staged but not deployed. Use a separate deploy module or task to deploy later.
+    - Setting O(deploy=false) is useful when batching changes across multiple interface tasks before a single deploy.
+    type: bool
+    default: true
+  state:
```
mikewiebe, line 281: ### Low: Track `gathered` support against issue #340 (non-blocking)

**Issue**

**Evidence**

- [`plugins/modules/nd_interface_vpc_access.py` lines 276-289](https://github.com/CiscoDevNet/ansible-nd/pull/279/files#diff-40bfaf3d5e8818668dc14938b001b647d46bd8d9b37cee5ad7d5934e209f45f7R276) documents state choices without `gathered`.
- [`plugins/module_utils/models/interfaces/vpc_access_interface.py` lines 431-435](https://github.com/CiscoDevNet/ansible-nd/pull/279/files#diff-fc0d76021a429b07196ccfc093d05609d28284084411ce37e2d92c36fcd0fe33R431) returns the same state choices from the model argspec.

**Existing PR overlap**

**Impact**

**Suggested fix**

---

mikewiebe, line 281: ### Medium: Add examples for every supported state

**Issue**

The module advertises `replaced` and `overridden`, but the `EXAMPLES` block only shows `merged`, `deleted`, and deploy-disabled `merged` usage.

**Evidence**

- [`plugins/modules/nd_interface_vpc_access.py` lines 276-289](https://github.com/CiscoDevNet/ansible-nd/pull/279/files#diff-40bfaf3d5e8818668dc14938b001b647d46bd8d9b37cee5ad7d5934e209f45f7R276) lists `merged`, `replaced`, `overridden`, and `deleted`.
- [`plugins/modules/nd_interface_vpc_access.py` lines 322-361](https://github.com/CiscoDevNet/ansible-nd/pull/279/files#diff-40bfaf3d5e8818668dc14938b001b647d46bd8d9b37cee5ad7d5934e209f45f7R322) contains examples for `merged`, `deleted`, and `deploy: false`, but no `replaced` or `overridden` example.

**Existing PR overlap**

**Impact**

Users cannot tell from the docs how destructive or broad `replaced` and `overridden` are for vPC access interfaces. Those states are where examples matter most because they define the scope of replacement and deletion.

**Suggested fix**

Add one clear `replaced` example and one clear `overridden` example. The `overridden` example should make the deletion scope obvious, especially whether it applies to all vPC access interfaces in the fabric, all on a switch, or only those matching the provided config.

plugins/modules/nd_interface_vpc_access.py
allenrobel: Addressed in b3adb8b2.

Replaced the top-level `deploy` bool with a `config_actions` dict carrying a nested `deploy` suboption, mirroring the same migration across the interface module family (`nd_interface_ethernet_trunk_host` ec82a6f2, `nd_interface_svi` d5afacfb, `nd_interface_subinterface_managed` 8806d87d, `nd_interface_subinterface_unmanaged` ace59646, `nd_interface_port_channel_trunk_host` fde71850):

- **EXAMPLES** — the no-deploy/batching example now uses the nested `config_actions: {deploy: false}` form.

plugins/modules/nd_interface_vpc_access.py
allenrobel: Addressed in 65a7d43a.

Added a `replaced` example and an `overridden` example to the `EXAMPLES` block, mirroring the same change across the interface module family:

- **`replaced`** — reprovisions a specific vPC interface (`vpc100`) with a new access VLAN, member ports, and per-peer descriptions.
- **`overridden`** — carries an inline comment making the fabric-wide scope explicit: every vPC interface **managed by this module** that is not listed is **deleted**, and an empty `config` list deletes **all** of them (matching this module's documented `overridden` behavior, where ND-side resources absent from the config are removed).

Verification: `validate-modules` sanity clean (parses the EXAMPLES YAML), `black` clean.

allenrobel: Acknowledged — no change in this PR. As noted, `gathered` is tracked collection-wide in #340 for all `NDStateMachine` + orchestrator modules, so `nd_interface_vpc_access` will pick it up consistently when that lands (with documented filter semantics, a `gathered` example, and a read-only test asserting no create/update/delete/deploy call is made). Keeping this thread open as the tracking marker for this module.

plugins/modules/nd_interface_vpc_access.py
allenrobel: Note: this branch was rebased onto the latest `develop` during the RETURN-block rollout, so the commit referenced above (`b3adb8b`) is now orphaned. The same change is now `5990f345`. Resolving as addressed.

plugins/modules/nd_interface_vpc_access.py
allenrobel: Note: this branch was rebased onto the latest `develop` during the RETURN-block rollout, so the commit referenced above (`65a7d43`) is now orphaned. The same change is now `cb1874ea`. Resolving as addressed.

mikewiebe (approved):
LGTM

gmicol (approved):
LGTM

PR #280, opened 2026-05-13
nd_interface_vpc_trunk_host

allenrobel:
## Related Issue(s)

Stacked on #279 (`nd_interface_vpc_access`) → #334 (`nd_interface_vpc_base`), which carries the shared vPC-interface infrastructure that previously lived on the now-closed #278 (`nd_vpc_pair`). #278 was closed in favor of #197 (`nd_manage_vpc_pair`); the shared interface base was re-homed onto #334 so this module can reach `develop` independently.

- CiscoDevNet/ansible-nd#323

Items raised by the code review of this branch (or inherited from the shared vPC infrastructure) that are tracked as GitHub issues rather than fixed in this PR:

- **#347** — Consolidate the trunk VLAN-parsing types (`AllowedVlans`, `CustomerVlanIdList`, validators/regexes) **and** the duplicated VLAN-mapping entry model (`TrunkVpcHostVlanMappingEntryModel` ≡ `PortChannelTrunkHostVlanMappingEntryModel`) into shared `models/types.py` / `models/nested.py` once the per-policy interface stacks land. Each trunk model intentionally carries its own copy to keep the stacks self-contained (branch-ordering hazard); in-code TODO marks the sites.
- **#348** — Lift the duplicated interface CRUD + policy-type/per-peer wrap-serializer into shared bases. This module's `expand_per_peer_fields` (per-peer fan-out + `mode="config"` `policy_type` strip) is a sibling-parity copy of `nd_interface_vpc_access`'s `serialize_policy`.
- **#349** — `query_all` fans out per-switch because ND exposes no fabric-wide vPC-interface endpoint (inherited from `VpcInterfaceBaseOrchestrator`).
- **#354** — vPC delete queues a deploy after a 404 no-op (ND can't distinguish 204 from 404 by body for vPC; inherited shared behavior).
- **#355** — vPC `peer1`/`peer2` member-port orientation may not track the user-supplied `switch_ip` (needs ND lab verification; cross-cutting #279/#280).
- **#356** — The name-only vPC identifier breaks if two pairs reuse the same vPC id in one fabric (cross-cutting #279/#280).

## Proposed Changes

- Adds the `nd_interface_vpc_trunk_host` module, which manages vPC `trunkVpcHost` interfaces (template `int_vpc_trunk_host`) on Cisco Nexus Dashboard.
- Adds `TrunkVpcHostPolicyTypeEnum` to `plugins/module_utils/models/interfaces/enums.py` so the model can frozen-pin its `policy_type`.
- Inherits all CRUD, peer-serial resolution, per-interface DELETE, and `query_all` dedupe from `VpcInterfaceBaseOrchestrator` (introduced in the stack-grandparent PR #334 `nd_interface_vpc_base`). Peer-serial resolution uses develop's `EpVpcPairGet` endpoint.

Trunk-specific design (lab-verified on ND 4.2.1):

- Single user-facing `allowed_vlans` and `native_vlan` fields fan out to `peer1AllowedVlans` / `peer2AllowedVlans` and `peer1NativeVlan` / `peer2NativeVlan` at payload-serialization time via a `@model_serializer`. ND collapses both per-peer pairs back to single `allowedVlans` / `nativeVlan` on GET, AND wire-side rejects divergent peer values with HTTP 400 `"…should be consistent"` — so there is no use case for exposing the per-peer keys. Tagged `TODO(4.2.1)`.
- Adds `vlan_mapping` and `vlan_mapping_entries` with shared `AllowedVlans` / `CustomerVlanIdList` validators (int→str coercion for the ND single-id echo, 1–4094 bounds, reversed-range rejection).
- Per-peer `peer1/peer2` `member_ports` / `port_channel_id` / `port_channel_description` / `port_channel_configuration` exposed (OpenAPI preserves these on read).
- `identifier_strategy = "single"` on `interface_name`; `switch_ip` is routing-only and excluded from diff/payload.

## Test Notes

- Unit tests (model + orchestrator) plus a fixture JSON: `python -m pytest tests/unit/module_utils/models/test_vpc_trunk_host_interface.py tests/unit/module_utils/orchestrators/test_vpc_trunk_host_interface.py -q`. Full repo suite stays green (no regressions).
- Integration target `tests/integration/targets/nd_interface_vpc_trunk_host/` with `merged` / `replaced` / `overridden` / `deleted` state coverage, exercised against ND 4.2.1 on the SITE1 lab (S1_LE1 / S1_LE2): `ansible-test integration --docker nd_interface_vpc_trunk_host -vv`. `vlan_mapping` cases gated on `supports_vlan_mapping` (Nexus 9000v rejects `dot1q-tunnel`).
- The vPC pair substrate is created via `nd_manage_vpc_pair` using a **virtual** peer-link. Physical peer-link is currently broken at the ND controller (`configSave` 500 "Member port … is a member of VPC_PAIR"); see bug-vault `vpc-physical-peer-link-configsave-fails`. A `TODO(4.2.1)` marker in the test vars tracks reverting to physical once ND fixes it. The test vPC interface uses PO 200/201 (not 500) because the virtual peer-link's MCT auto-takes `port-channel500`.
- `ansible-test sanity --docker default --python 3.11` passes on the changed files.
- Wire shape verified live: created `vpc200` on S1_LE1/S1_LE2 with distinct then matching per-peer VLANs to confirm ND's consistency-enforcement + collapse-on-GET; clean per-interface DELETE cascades to both peers.

4.2.1

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

### DIFF
--- a/plugins/module_utils/models/interfaces/vpc_trunk_host_interface.py
+++ b/plugins/module_utils/models/interfaces/vpc_trunk_host_interface.py
@@ -1,0 +1,601 @@
+# Copyright: (c) 2026, Allen Robel (@allenrobel)
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+"""
+vPC trunk (trunkVpcHost) interface Pydantic models for Nexus Dashboard.
+
+This module defines nested Pydantic models that mirror the ND Manage Interfaces API payload
+structure for vPC trunkVpcHost interfaces (`int_vpc_trunk_host.template`). The playbook
+config uses the same nesting so that `to_payload()` and `from_response()` work via standard
+Pydantic serialization with no custom wrapping or flattening.
+
+A vPC trunk-host interface spans two switches in a vPC pair. The user supplies one peer's
+management IP as `switch_ip`; the orchestrator auto-resolves the peer serial via the
+`vpcPair` endpoint and injects it as `peerSwitchId` in the payload. Per-peer policy fields
+use the ND-native `peer1_*` / `peer2_*` naming where `peer1` corresponds to `switch_ip`
+(the switch in the URL path) and `peer2` corresponds to the auto-resolved peer.
+
+## Model Hierarchy
+
+- `TrunkVpcHostInterfaceModel` (top-level, `NDBaseModel`)
+    - `switch_ip` (routing field; excluded from diff/payload)
+    - `interface_name` (identifier; e.g. `vpc100`)
+    - `interface_type` (frozen: "vpc")
+    - `config_data` -> `TrunkVpcHostConfigDataModel`
+        - `mode` (frozen: "trunk")
+        - `network_os` -> `TrunkVpcHostNetworkOSModel`
+            - `network_os_type` (frozen: "nx-os")
+            - `policy` -> `TrunkVpcHostPolicyModel`
+                - `policy_type` (frozen: "trunkVpcHost")
+                - `peer_switch_id` (orchestrator-injected; not in argspec)
+                - single user-facing `allowed_vlans`, `native_vlan` (fanned out to per-peer keys on write)
+                - per-peer fields: `peer1_*` / `peer2_*` (member_ports, port_channel_id, descriptions, configuration)
+                - `vlan_mapping`, `vlan_mapping_entries`
+                - shared fields: `admin_state`, `cdp`, `lacp_*`, `mtu`, etc.
+"""
+
+from __future__ import annotations
+
+import re
+from typing import Annotated, ClassVar, Literal, Optional  # Optional needed for Annotated runtime expr (see types.py)
+
+from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import (
+    BeforeValidator,
+    Field,
+    field_validator,
+    model_serializer,
+)
+from ansible_collections.cisco.nd.plugins.module_utils.models.base import NDBaseModel
+from ansible_collections.cisco.nd.plugins.module_utils.models.interfaces.enums import (
+    BpduFilterEnum,
+    BpduGuardEnum,
+    DuplexModeEnum,
+    LacpRateEnum,
+    LinkTypeEnum,
+    MtuEnum,
+    PortChannelModeEnum,
+    SpeedEnum,
+    StormControlActionEnum,
+    TrunkVpcHostPolicyTypeEnum,
+)
+from ansible_collections.cisco.nd.plugins.module_utils.models.nested import NDNestedModel
+from ansible_collections.cisco.nd.plugins.module_utils.models.types import AsciiDescription
+
+# Shape regex: "none", "all", or comma-separated VLAN ids/ranges. Range bounds are validated separately.
+_ALLOWED_VLANS_SHAPE = re.compile(r"^(none|all|(\d+(-\d+)?)(,\d+(-\d+)?)*)$")
+# Single VLAN id or range token (e.g. "100" or "100-200"). Range bounds are validated separately.
+_VLAN_ID_OR_RANGE_SHAPE = re.compile(r"^\d+(-\d+)?$")
+
+
+def _validate_vlan_id_or_range(token: str, field_name: str) -> None:
+    """
+    # Summary
+
+    Validate a single VLAN id or range token (e.g. `"100"` or `"100-200"`) and confirm every id is in 1..4094 and any range has start <= end.
+
+    ## Raises
+
+    ### ValueError
+
+    - If `token` does not match `\\d+(-\\d+)?`.
+    - If any VLAN id is outside 1..4094.
+    - If a range has start greater than end.
+    """
+    if not _VLAN_ID_OR_RANGE_SHAPE.match(token):
+        raise ValueError(f"{field_name} entry {token!r} must be a VLAN id or range (e.g. '100' or '100-200')")
+    if "-" in token:
+        start_str, end_str = token.split("-", 1)
+        start, end = int(start_str), int(end_str)
+        if not 1 <= start <= 4094 or not 1 <= end <= 4094:
+            raise ValueError(f"{field_name} range {token!r} is out of bounds; VLAN ids must be in 1..4094")
+        if start > end:
+            raise ValueError(f"{field_name} range {token!r} has start greater than end")
+    else:
+        vid = int(token)
+        if not 1 <= vid <= 4094:
+            raise ValueError(f"{field_name} id {vid} is out of bounds; VLAN ids must be in 1..4094")
+
+
+def _validate_allowed_vlans(value):
+    """
+    # Summary
+
+    Validate `allowed_vlans` matches `"none"`, `"all"`, or a comma-separated list of VLAN ids/ranges where every id is in 1..4094 and every range start <= end.
+    ND returns single-id values as JSON ints (e.g. `250`) but accepts both forms on input; this validator coerces int -> str so round-trips and idempotency
+    comparisons are stable.
+
+    ## Raises
+
+    ### ValueError
+
+    - If `value` is a non-empty string that does not match the expected shape.
+    - If any VLAN id is outside 1..4094.
+    - If any range has start greater than end.
+    """
+    if value is None or value == "":
+        return value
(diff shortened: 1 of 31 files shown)


plugins/modules/nd_interface_vpc_trunk_host.py
```diff
@@ -0,0 +1,559 @@
...
+                      dot1q_tunnel:
+                        description:
+                        - Use selective dot1q-tunnel mode for this entry.
+                        type: bool
+                      provider_vlan_id:
+                        description:
+                        - Provider VLAN id.
+                        - Valid range is 1-4094.
+                        type: int
+  deploy:
```
mikewiebe: ### Medium: Replace top-level `deploy` with `config_actions`

**Issue**

**Evidence**

- [`plugins/modules/nd_interface_vpc_trunk_host.py` lines 310-332](https://github.com/CiscoDevNet/ansible-nd/pull/280/files#diff-12fe8523db7ac503ab0f0fe981d6fd477fbe22c3e6842abeb0ba8b2e2f475dc1R310) documents `deploy` as a standalone option and lists states without the standard `config_actions` option.

**Existing PR overlap**

**Impact**

**Suggested fix**

plugins/modules/nd_interface_vpc_trunk_host.py
```diff
@@ -0,0 +1,559 @@
...
+    description:
+    - Whether to deploy vPC interface changes after mutations are complete.
+    - When V(true), all queued vPC interface changes are deployed in a single bulk API call at the end of module
+      execution via the C(interfaceActions/deploy) API. Only the vPC interfaces modified by this task are deployed.
+    - When V(false), changes are staged but not deployed. Use a separate deploy module or task to deploy later.
+    - Setting O(deploy=false) is useful when batching changes across multiple interface tasks before a single deploy.
+    type: bool
+    default: true
+  state:
```
mikewiebe, line 324: ### Low: Track `gathered` support against issue #340 (non-blocking)

**Issue**

**Evidence**

- [`plugins/modules/nd_interface_vpc_trunk_host.py` lines 319-332](https://github.com/CiscoDevNet/ansible-nd/pull/280/files#diff-12fe8523db7ac503ab0f0fe981d6fd477fbe22c3e6842abeb0ba8b2e2f475dc1R319) documents state choices without `gathered`.
- [`plugins/module_utils/models/interfaces/vpc_trunk_host_interface.py` lines 596-600](https://github.com/CiscoDevNet/ansible-nd/pull/280/files#diff-d1d2f61f4360961f191c824d5a3da29c8459b747986de7c63626cf96a343f47dR596) returns the same state choices from the model argspec.

**Existing PR overlap**

**Impact**

**Suggested fix**

---

plugins/module_utils/models/interfaces/vpc_trunk_host_interface.py
```diff
@@ -0,0 +1,601 @@
...
+    MtuEnum,
+    PortChannelModeEnum,
+    SpeedEnum,
+    StormControlActionEnum,
+    TrunkVpcHostPolicyTypeEnum,
+)
+from ansible_collections.cisco.nd.plugins.module_utils.models.nested import NDNestedModel
+from ansible_collections.cisco.nd.plugins.module_utils.models.types import AsciiDescription
+
+# Shape regex: "none", "all", or comma-separated VLAN ids/ranges. Range bounds are validated separately.
```
mikewiebe, line 66: ### Medium: Move reusable VLAN parsing into shared model types

**Issue**

The vPC trunk model carries local VLAN parsing helpers and a TODO to move them later, but this logic is also needed by sibling trunk-style interface models.

**Evidence**

- [`plugins/module_utils/models/interfaces/vpc_trunk_host_interface.py` lines 65-161](https://github.com/CiscoDevNet/ansible-nd/pull/280/files#diff-d1d2f61f4360961f191c824d5a3da29c8459b747986de7c63626cf96a343f47dR65) defines local VLAN range regexes, validators, `AllowedVlans`, and `CustomerVlanIdList`, then notes that the types should later move to `models/types.py`.

**Existing PR overlap**

**Impact**

If every trunk-like interface module carries its own VLAN parser, fixes and edge cases can drift between modules. That makes the user contract inconsistent and increases the chance that one module accepts or rejects VLAN syntax differently from another.

**Suggested fix**

Move the reusable VLAN ID, VLAN range, allowed-VLAN, and customer-VLAN-list types into `plugins/module_utils/models/types.py` or another shared interface-family helper before the branch stack lands, then import those types from each trunk-like model. If branch ordering prevents that in this PR, keep a tracked follow-up tied to the stack merge.

plugins/modules/nd_interface_vpc_trunk_host.py
allenrobel: Addressed in 4af2ad34.

Replaced the top-level `deploy` bool with a `config_actions` dict carrying a nested `deploy` suboption, mirroring the same migration across the interface module family (`nd_interface_ethernet_trunk_host`, `nd_interface_svi`, `nd_interface_subinterface_managed`, `nd_interface_subinterface_unmanaged`, `nd_interface_port_channel_trunk_host`, `nd_interface_vpc_access`):

- **EXAMPLES** — the no-deploy/batching example now uses the nested `config_actions: {deploy: false}` form.

plugins/modules/nd_interface_vpc_trunk_host.py
allenrobel: Acknowledged — no change in this PR. As noted, `gathered` is tracked collection-wide in #340 for all `NDStateMachine` + orchestrator modules, so `nd_interface_vpc_trunk_host` will pick it up consistently when that lands (with documented filter semantics, a `gathered` example, and a read-only test asserting no create/update/delete/deploy call is made). Keeping this thread open as the tracking marker for this module.

plugins/module_utils/models/interfaces/vpc_trunk_host_interface.py
allenrobel: Taking the tracked-follow-up route you offered ("If branch ordering prevents that in this PR, keep a tracked follow-up tied to the stack merge"), rather than moving the types in this PR.

Rationale — this is a deliberate, documented decision, not an oversight. The model already carries an in-code TODO at the type definitions:

```python
# TODO: After all per-policy interface modules merge to develop, consolidate AllowedVlans, CustomerVlanIdList,
# and _validate_vlan_id_or_range into models/types.py so siblings share a single source of truth. This module
# carries its own copy to keep the vpc stack self-contained.
```

The same VLAN parser is duplicated in `port_channel_trunk_host_interface.py` (and the pattern originated in `ethernet_trunk_host`), and those siblings live on **separate stacked branches** (`stack.json` for ethernet, the vpc stack for port-channel → vpc). Promoting the types into shared `models/types.py` now would make this PR's branch the owner of a symbol that the other in-flight branches depend on, coupling independently-reviewed branches and creating merge-ordering hazards (whichever lands first defines the shared type; the others then conflict). Keeping each trunk model self-contained lets the branches merge in any order.

The consolidation is intended as a single coordinated change once the per-policy interface modules are all on develop — at which point there's one obvious place to define the types and every trunk model can import them in the same commit, with no cross-branch coupling. The in-code TODO is the tracker; happy to also open a dedicated GitHub issue tied to the stack merge if you'd prefer a GitHub-tracked item over the in-code marker — just say the word and I'll file it and link it here.

allenrobel: Filed the GitHub-tracked follow-up: #347 — "Consolidate trunk VLAN-parsing types (AllowedVlans, CustomerVlanIdList) into shared models/types.py after interface stacks land". It captures the duplication sites, the branch-ordering rationale for deferring, and the proposed consolidation steps + acceptance criteria, and links back to this thread.

plugins/modules/nd_interface_vpc_trunk_host.py
allenrobel: Note: this branch was rebased onto the latest `develop` during the RETURN-block rollout, so the commit referenced above (`4af2ad3`) is now orphaned. The same change is now `5c00aa37`. Resolving as addressed.

mikewiebe (approved):
LGTM

PR #281, opened 2026-05-14
Vrf_lite module for Ansible ND 4.x

sivakasi-cisco:
## Related Issue(s)

- CiscoDevNet/ansible-nd#235

This PR adds a new `nd_manage_vrf_lite` module for managing VRF Lite attachments on Nexus Dashboard / NDFC 4.x fabrics.

The module supports:

- `gathered`
- `merged`
- `replaced`
- `overridden`
- `deleted`

The implementation uses the existing `NDStateMachine` pattern, but keeps the VRF Lite-specific API details in a dedicated orchestrator and helper files under `plugins/module_utils/manage_vrf_lite`.

Main things included in this PR:

- New `cisco.nd.nd_manage_vrf_lite` module
- VRF Lite models for nested playbook input and flat runtime attachment entries
- Query, attach, detach, save, deploy, and verification helpers
- Support for `config_actions.save` and `config_actions.deploy`
- Support for check mode previews
- Guardrails for switch lookup, VRF existence, and VRF Lite support validation
- Unit and integration tests for merge, replace, override, delete, gather, deploy, and check mode flows

**Note:** `issubset()` list matching changed from greedy to bipartite matching. This affects all callers (including `nd.py` idempotency checks), strictly in the "fewer false `changed`" direction.


allenrobel (commented):
### Code review

Found 6 issues, posted as inline comments below so each can be resolved individually.
Highlights: the module is built on the abandoned `nd_v2` `NDModule` transport instead of the Gen-3 orchestrator `RestSend` path; the switch inventory is fetched twice per run, bypassing `FabricContext`; and validation issues a per-attachment GET (N+1).

Also, a couple general comments.

1. It would be helpful for reviewers if the PR Description template was completed so that we have a summary of what the PR contains, design decisions, caveats, workarounds required, unit test coverage, etc.
2. The title of the PR implies this is a draft.  Since, it's now labeled "Ready for Review" you might want to edit the title to remove "initial draft"

plugins/module_utils/orchestrators/manage_vrf_lite.py
```diff
@@ -0,0 +1,500 @@
...
+    VrfLiteEndpoints,
+)
+from ansible_collections.cisco.nd.plugins.module_utils.manage_vrf_lite.validation import (
+    validate_vrf_lite_write_guardrails,
+)
+from ansible_collections.cisco.nd.plugins.module_utils.models.base import NDBaseModel
+from ansible_collections.cisco.nd.plugins.module_utils.models.manage_vrf_lite.vrf_lite_attachment_entry import (
+    VrfLiteAttachmentEntry,
+)
+from ansible_collections.cisco.nd.plugins.module_utils.nd_v2 import NDModule as NDModuleV2
```
allenrobel: `nd_v2.py` is dead code in this collection — the abandoned Gen-2 `NDModule` rewrite has no other importers, and the project convention is to not build new modules on it ("if you find yourself reaching for `NDModule` from `nd_v2.py`, stop and use an orchestrator instead"). This PR instantiates `NDModuleV2` as a REST transport here (lines 218, 240, 421) and in `manage_vrf_lite/query.py` (lines 23, 219) and `manage_vrf_lite/validation.py` (lines 22, 67, 118). Please drive these requests through the orchestrator's injected `RestSend` (`NDBaseOrchestrator._request()`, https://github.com/CiscoDevNet/ansible-nd/blob/bee38b72bf9c3f92e0739fc34d7c29b1f9d730d0/plugins/module_utils/orchestrators/base.py) as the other Gen-3 modules do.

plugins/module_utils/manage_vrf_lite/query.py
```diff
@@ -0,0 +1,408 @@
...
+        return data
+    if isinstance(data, dict):
+        for key in keys:
+            value = data.get(key)
+            if isinstance(value, list):
+                return value
+    return []
+
+
+def _query_fabric_switches(module: Any, nd_v2: Any, fabric_name: str) -> dict[str, str]:
```
allenrobel: This fetches the switch inventory directly. `FabricContext` (https://github.com/CiscoDevNet/ansible-nd/blob/bee38b72bf9c3f92e0739fc34d7c29b1f9d730d0/plugins/module_utils/fabric_context.py) already lazily fetches and caches the fabric summary and switch IP/serial lookups — the convention is to never fetch the switch inventory directly. The TODO two lines below treats this as future work, but `FabricContext` exists today. Compounding it, `_load_switch_inventory()` in `validation.py` (line 61) fetches the same `/fabrics/{name}/switches` endpoint again with a separate cache (`_sn_to_ip_mapping` vs `_fabric_switch_inventory`), so a single write run fetches the inventory twice. Please reuse `FabricContext`, or at minimum share one cache between the two paths.

plugins/module_utils/manage_vrf_lite/validation.py
```diff
@@ -0,0 +1,214 @@
...
+
+        vrf_lite_entries = getattr(attach, "extensions", None) or getattr(attach, "vrf_lite", None)
+        if not vrf_lite_entries:
+            continue
+
+        switch_data = inventory[serial_number]
+        role = switch_data.get("role", "")
+        support = None
+        try:
+            support = _query_vrf_lite_support(module, fabric_name, vrf_name, serial_number)
```
allenrobel: Per-item GET inside the `for attach in attachments:` loop (line 159): `_query_vrf_lite_support()` issues one GET per attachment with no caching, so N attachments means N sequential requests (with repeats for the same switch across entries). Module performance is the top source of user complaints for this collection, and a per-item GET in a loop is treated as a defect in review. Prefer one bulk query plus in-memory filtering, or cache results per (fabric, vrf, serial) across the loop.

plugins/modules/nd_manage_vrf_lite.py
```diff
@@ -0,0 +1,334 @@
...
+ANSIBLE_METADATA = {"metadata_version": "1.1", "status": ["preview"], "supported_by": "community"}
+
+DOCUMENTATION = r"""
+---
+module: nd_manage_vrf_lite
+version_added: "1.4.0"
+short_description: Manage VRF Lite attachments on Cisco Nexus Dashboard
+
+description:
+- Manage VRF Lite attachment configuration in ND/NDFC fabrics.
```
allenrobel: This collection targets ND 4.2+ only — new code and docs should not reference NDFC. Besides this description ("ND/NDFC fabrics"), NDFC appears in `manage_vrf_lite/actions.py` docstrings (lines 35, 51, 226, 274) and, most visibly, in the runtime warning string in `manage_vrf_lite/validation.py` line 133 ("...but NDFC did not return an explicit VRF Lite support flag..."), which is shown to end users. Please replace with "ND" or "the controller".

plugins/modules/nd_manage_vrf_lite.py
```diff
@@ -0,0 +1,334 @@
...
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import annotations
+
+ANSIBLE_METADATA = {"metadata_version": "1.1", "status": ["preview"], "supported_by": "community"}
+
+DOCUMENTATION = r"""
+---
+module: nd_manage_vrf_lite
+version_added: "1.4.0"
```
allenrobel: `version_added: "1.4.0"` is a past release (galaxy.yml on develop is already at 1.5.0). New modules use the next unreleased collection version, currently `2.0.0` — see `nd_fabric_update_group` and `nd_maintenance_mode` on develop: https://github.com/CiscoDevNet/ansible-nd/blob/develop/plugins/modules/nd_fabric_update_group.py

plugins/modules/nd_manage_vrf_lite.py
```diff
@@ -0,0 +1,334 @@
...
+        type: bool
+        default: true
+      retries:
+        description:
+        - Number of polling retries during verification.
+        type: int
+        default: 5
+      timeout:
+        description:
+        - Seconds to wait between verification retries.
```
allenrobel: DOCUMENTATION says `verify.timeout` is "Seconds to wait between verification retries", but `request_with_verify_settings()` (`manage_vrf_lite/common.py` line 136) assigns this value to `rest_send.timeout`, which `RestSend` treats as the *total per-request retry budget* (decremented by `send_interval` each attempt — see `_commit_normal_mode` in https://github.com/CiscoDevNet/ansible-nd/blob/bee38b72bf9c3f92e0739fc34d7c29b1f9d730d0/plugins/module_utils/rest/rest_send.py). A user setting `timeout: 60` gets a 60-second total budget per request, not 60-second pauses between retries. Please align the docs (or the wiring) with the actual semantics.

sivakasi-cisco:
> ### Code review
> Found 6 issues, posted as inline comments below so each can be resolved individually. Highlights: the module is built on the abandoned `nd_v2` `NDModule` transport instead of the Gen-3 orchestrator `RestSend` path; the switch inventory is fetched twice per run, bypassing `FabricContext`; and validation issues a per-attachment GET (N+1).
> 
> Also, a couple general comments.
> 
> 1. It would be helpful for reviewers if the PR Description template was completed so that we have a summary of what the PR contains, design decisions, caveats, workarounds required, unit test coverage, etc.
> 2. The title of the PR implies this is a draft.  Since, it's now labeled "Ready for Review" you might want to edit the title to remove "initial draft"
> 
> 🤖 Generated with [Claude Code](https://claude.ai/code)
> 
> - If this code review was useful, please react with 👍. Otherwise, react with 👎.

That is true. I had missed it. Now added

plugins/module_utils/orchestrators/manage_vrf_lite.py
sivakasi-cisco: Thanks, I updated this. VRF Lite requests now go through the orchestrator RestSend flow instead of creating NDModuleV2. I added a shared helper for the custom VRF Lite requests and reused it for query, attach, save, and deploy actions.

plugins/module_utils/manage_vrf_lite/query.py
sivakasi-cisco: I have changed the query path to use FabricContext for switch inventory and shared that inventory through the module params cache now. validation.py can now reuse the same cached inventory instead of doing another fabric switches call.

plugins/module_utils/manage_vrf_lite/validation.py
sivakasi-cisco: Addressed. I have added a small cache for VRF Lite support checks with fabric, VRF, and switch serial number as key. So repeated checks for the same switch/VRF during one run will reuse the cached result instead of calling the API again

plugins/modules/nd_manage_vrf_lite.py
sivakasi-cisco: Corrected

plugins/modules/nd_manage_vrf_lite.py
sivakasi-cisco: Corrected

plugins/modules/nd_manage_vrf_lite.py
sivakasi-cisco: I have updated the option description so it describes the total per-request timeout used by verification queries.

allenrobel:
Re-reviewed against head `450043c4` (the "Addressing review comments" commit). All six of my earlier comments are addressed in code — verified file-by-file, not just by thread state:

1. **`nd_v2`/`NDModuleV2` removed** — VRF Lite requests now go through the orchestrator's `RestSend` via a shared `request_with_verify_settings(...)` helper. ✅
2. **`FabricContext` reused** — query and validation both resolve switch inventory through `FabricContext`, sharing a single `_fabric_switch_inventory` cache instead of two separate fetches. ✅
3. **N+1 GET resolved** — `_query_vrf_lite_support` now caches per `(fabric, vrf, serial)`, so repeat checks in the attach loop don't re-hit the API. ✅
4. **NDFC references removed** — no `NDFC` left in docs, docstrings, or the user-facing warning string (now "the controller…"). ✅
5. **`version_added`** — bumped from `1.4.0` to `2.0.0`. ✅
6. **`verify.timeout` docs** — now describe the total per-request timeout budget, matching the RestSend wiring. ✅

Thanks for the thorough turnaround.

allenrobel (dismissed):
Approving. All six of my earlier review comments are addressed in code (verified against head 450043c4) — see https://github.com/CiscoDevNet/ansible-nd/pull/281#issuecomment-4721069415 for the per-point breakdown.

plugins/module_utils/manage_vrf_lite/query.py
```diff
@@ -0,0 +1,392 @@
...
+
+def _query_fabric_switches(module: Any, rest_send: Any, fabric_name: str) -> dict[str, str]:
+    """Return serial->mgmt-ip mapping from fabric switch inventory."""
+    fabric_context = FabricContext(rest_send=rest_send, fabric_name=fabric_name)
+    inventory = fabric_context.switch_inventory_by_id
+    module.params["_fabric_switch_inventory"] = inventory
+    return {str(serial).strip(): str(switch.get("fabricManagementIp")).strip() for serial, switch in inventory.items() if switch.get("fabricManagementIp")}
+
+
+def _query_vrfs(module: Any, rest_send: Any, fabric_name: str) -> list[dict[str, Any]]:
```
mikewiebe, line 94: **Issue:** VRF and attachment queries ignore pagination and filtering

**NOTE: I am not sure this is actually an issue but we should keep this in mind when we do scale testing for this resource.  Not suggesting you do anything now but adding here for informational purposes during scale testing**

**PR code links**

- [`query.py` lines 68-80](https://github.com/CiscoDevNet/ansible-nd/pull/281/files#diff-3be2d9b08bbba67015aca2b4ea5a446741e5b1cb053820738be8886c0f7dcd42R68-R80)
- [`query.py` lines 205-235](https://github.com/CiscoDevNet/ansible-nd/pull/281/files#diff-3be2d9b08bbba67015aca2b4ea5a446741e5b1cb053820738be8886c0f7dcd42R205-R235)

**Impact**

Large fabrics can return only a default or first page of VRFs and attachments.
If the state machine builds current state from a partial response, it can make
incorrect idempotency decisions, fail to preserve existing attachments, or plan
deletes because resources outside the returned page look absent. It also forces
fabric-wide scans even when the playbook is scoped to a small set of VRFs.

**Details**

`_query_vrfs()` calls `GET /fabrics/{fabricName}/vrfs` without `offset`, `max`,
or `filter`. `_query_vrf_attachments()` calls
`POST /fabrics/{fabricName}/vrfAttachments/query` once without `offset`, `max`,
`filter`, or `includeAll`. `query_vrf_lite_state()` then fetches all VRFs first,
filters client-side, and queries attachments only for the resulting in-memory
map.

The bundled OpenAPI spec advertises pagination and filtering for both relevant
read paths:

- `GET /fabrics/{fabricName}/vrfs`: `offset`, `max`, `sort`, `filter`
- `POST /fabrics/{fabricName}/vrfAttachments/query`: `offset`, `max`, `sort`,
  `filter`, `includeAll`

The unit tests do not appear to cover paginated responses or API-side filters
for this module.

**Suggested Investigation / Possible Resolution if its an issue**

The current state query path looks risky for large fabrics. `_query_vrfs()` and
`_query_vrf_attachments()` each make a single unpaginated request, even though
the OpenAPI spec exposes `offset`/`max` and filtering controls for these APIs.
For a fabric with more VRFs or attachments than the controller returns by
default, the state machine may build `current` from only part of the real
controller state. That can turn into false diffs, missed preservation, or
incorrect deletes.

Can we update the query helpers to use API-side filtering for the requested VRFs
where possible and page until the response is exhausted? I would also add unit
coverage for multi-page VRF and attachment responses so this does not regress.

---

plugins/modules/nd_manage_vrf_lite.py
```diff
@@ -0,0 +1,334 @@
...
+    default: merged
+    choices: [ merged, replaced, deleted, overridden, gathered ]
+
+  fabric_name:
+    description:
+    - Target fabric name.
+    type: str
+    required: true
+
+  config_actions:
```
mikewiebe, line 40: **Issue**

The module exposes the standard `config_actions.type` option, but the docs and
runtime apply it to config save rather than to deploy scope.

This is what we discussed during the standup earlier in the week.

1. `config_actions.save` - Should only control calling the config-save global API
2. `config_actions.deploy` - Controls if a deploy is executed
3. `config_actions.type` - Controls if the deploy is global level or switch level

**Evidence**

- [`plugins/modules/nd_manage_vrf_lite.py` lines 40-60](https://github.com/CiscoDevNet/ansible-nd/pull/281/files#diff-1afcf411e0de32f4e06eede34d24dbd50de8f5c327692da711123c1e9a441597R40)
  documents `config_actions.type` as the "Scope of the config save operation."
- [`plugins/module_utils/orchestrators/manage_vrf_lite.py` lines 389-408](https://github.com/CiscoDevNet/ansible-nd/pull/281/files#diff-f809bb3eef17163bbd8b0a46efee709d7bc389d1a53265f1b343cb6c676fa202R389)
  computes target VRFs and target switch IDs independently of
  `config_actions.type`.
- [`plugins/module_utils/orchestrators/manage_vrf_lite.py` lines 424-464](https://github.com/CiscoDevNet/ansible-nd/pull/281/files#diff-f809bb3eef17163bbd8b0a46efee709d7bc389d1a53265f1b343cb6c676fa202R424)
  sends `{"type": ...}` to `configSave`, then builds the VRF deploy payload
  separately with `vrfNames` and optional `switchIds`.
- OpenAPI lookup for `POST /fabrics/{fabricName}/actions/configSave` did not
  show a request body.
- OpenAPI lookup for `deployVrfsRequest` shows required `vrfNames` and optional
  `switchIds` and `switchFabricNames`.

**Impact**

Users who set `config_actions.type: global` can reasonably expect deploy scope
to change, but the module still appears to build deploy payloads from changed
VRFs and target switch IDs. That makes the user-facing contract inconsistent
with the standard `config_actions` pattern and may send an unsupported body to
`configSave`.

**Suggested fix**

Make `config_actions.type` control deploy behavior, or remove/document the
option if VRF Lite cannot support the `switch` versus `global` distinction. For
the standard contract, `switch` should deploy only affected switches and
`global` should omit `switchIds` or use the appropriate global deploy behavior.
Update documentation, examples, and tests for whichever behavior is intended.

---

plugins/modules/nd_manage_vrf_lite.py
```diff
@@ -0,0 +1,334 @@
...
+                description:
+                - Name of the peer VRF for the VRF Lite extension.
+                type: str
+
+extends_documentation_fragment:
+- cisco.nd.modules
+- cisco.nd.check_mode
+"""
+
+EXAMPLES = r"""
```
mikewiebe, line 186: **Issue**

The module supports five states and exposes save/deploy controls, but the
public examples only show `gathered`, `merged`, and `deleted`.

**Evidence**

- [`plugins/modules/nd_manage_vrf_lite.py` lines 173-203](https://github.com/CiscoDevNet/ansible-nd/pull/281/files#diff-1afcf411e0de32f4e06eede34d24dbd50de8f5c327692da711123c1e9a441597R173)
  shows only gathered, merged, and deleted examples.
- [`plugins/modules/nd_manage_vrf_lite.py` lines 27-60](https://github.com/CiscoDevNet/ansible-nd/pull/281/files#diff-1afcf411e0de32f4e06eede34d24dbd50de8f5c327692da711123c1e9a441597R27)
  documents state choices `merged`, `replaced`, `deleted`, `overridden`, and
  `gathered`, plus `config_actions`.
- [`tests/integration/targets/nd_manage_vrf_lite/tasks/nd_manage_vrf_lite_replace.yaml` lines 73-74](https://github.com/CiscoDevNet/ansible-nd/pull/281/files#diff-dac02789c1f3b173bf7ff2bc30492f2cecb477b05c9d3219f5f3c5438d52729bR73)
  exercises `state: replaced` with `config_actions`.
- [`tests/integration/targets/nd_manage_vrf_lite/tasks/nd_manage_vrf_lite_override.yaml` lines 73-74](https://github.com/CiscoDevNet/ansible-nd/pull/281/files#diff-c20d187e0e3c6fe183d85a02db2392ac2eed6f150b97434ca3435a3fc3c43f43R73)
  exercises `state: overridden` with `config_actions`.

**Impact**

Users do not get a documented example for the difference between `replaced` and
`overridden`, or for the standard save/deploy control surface. That is a
noticeable gap for a new module because these are the higher-risk state
transitions and the controls that affect when staged intent is pushed.

**Suggested fix**

Add documentation examples for `replaced` and `overridden`, and examples that show the deploy patterns and outcomes

```yaml
config_actions:
  save: true
  deploy: true
  type: switch
```

Keep the examples aligned with the tested integration workflows.

---

plugins/modules/nd_manage_vrf_lite.py
sivakasi-cisco: Right. Now the 'type' controls the deploy scope instead of being attached to config-save

save: true just triggers the config-save API (no body).
deploy: true runs the VRF deploy.
type: switch deploys only the switches affected by the changed VRFs 
type: global does a fabric-wide deploy

Updated the docs, the orchestrator logic, and added UT and IT assertions that check the actual wire payload for both the switch and global cases.

plugins/modules/nd_manage_vrf_lite.py
sivakasi-cisco: Fixed. Expanded the block from 3 to 5 tasks and added inline comments explaining the following outcomes.

merged with save: true, deploy: false to show staging without deploying.
replaced with save/deploy, type: switch — rewrites only the listed VRFs and deploys just the affected switches.
overridden with save/deploy, type: global — makes the fabric match the config exactly (unlisted attachments are removed) and deploys fabric-wide.

sivakasi-cisco (commented):
Addressed the review comments

plugins/module_utils/manage_vrf_lite/query.py
sivakasi-cisco: Sure, Mike. I have not added changes for this. I had done a scale of vrflite in single switch covered. I have one more scale test with a scaled set of preprovisioned switches to be tested. I can check this while testing them

sivakasi-cisco (commented):
Addressed the review comments

AKDRG:
Hi Sivakami,
Adding a note here as per our discussion. Everything else looks good. 
Pending:
Need to use the common fabric details utility for the type from fabric_details_cache once https://github.com/CiscoDevNet/ansible-nd/pull/405 is merged.

sivakasi-cisco:
Have tested the nd_vrf_lite module with the latest changes in [PR294](https://github.com/CiscoDevNet/ansible-nd/pull/294).

plugins/module_utils/orchestrators/manage_vrf_lite.py
```diff
@@ -0,0 +1,527 @@
...
+            logger.debug("_post_grouped_rows: posting %d attachment row(s) for VRF %s", len(rows), vrf_name)
+            responses.append(
+                {
+                    "vrf_name": vrf_name,
+                    "response": _post_attachment_payload(module, self.rest_send, module.params.get("fabric_name"), vrf_name, rows),
+                }
+            )
+        return {"response": responses}
+
+    def _post_attach_entries(self, entries: list[Any]) -> dict[str, Any]:
```
mikewiebe, line 333: ### High: Check mode skips attachment write preflight

**Issue**

The attach path returns its planned entries before verifying that the VRF exists or
that the switch/fabric supports the requested VRF Lite operation.

**Evidence**

- [`manage_vrf_lite.py` lines 231-243](https://github.com/CiscoDevNet/ansible-nd/pull/281/files#diff-f809bb3eef17163bbd8b0a46efee709d7bc389d1a53265f1b343cb6c676fa202R231)
  returns on `module.check_mode` before calling `_validate_attach_entries()`.
- The generic state machine does not invoke write methods in its check-mode path, so
  the validation is not reached elsewhere.
- `_validate_attach_entries()` is the code that calls VRF existence and switch-role /
  capability guardrails (`manage_vrf_lite.py` lines 206-217).
- A practical failure is a deployment pipeline that uses check mode as its validation
  gate:

  ```yaml
  - name: Validate VRF Lite attachment before deployment
    cisco.nd.nd_manage_vrf_lite:
      fabric_name: site-1
      state: merged
      config:
        - vrf_name: TENANT-PROD
          vlan_id: 500
          attach:
            - ip_address: 10.122.84.191
              vrf_lite:
                - interface: Ethernet1/20
                  dot1q: 500
                  ipv4_addr: 10.33.0.2/30
                  neighbor_ipv4: 10.33.0.1
    check_mode: true
  ```

  If `TENANT-PROD` does not exist, check mode reports the attachment and deployment
  as a viable plan because it returns before preflight. The same task then fails in
  the production run when `_validate_attach_entries()` finally checks the VRF. An
  unsupported switch role or capability produces the same false-green validation.

**Existing PR overlap**

No matching open issue found.

**Impact**

Check mode can approve a nonexistent VRF or unsupported switch and report a viable
plan that normal execution immediately rejects.

**Suggested fix**

Move non-mutating preflight ahead of the check-mode gate and share it between planning
and execution. Add parity tests for nonexistent VRF, unsupported role/capability, and
valid input in both modes.

---

plugins/module_utils/manage_vrf_lite/query.py
```diff
@@ -0,0 +1,392 @@
...
+from ansible_collections.cisco.nd.plugins.module_utils.manage_vrf_lite.runtime_endpoints import (
+    VrfLiteEndpoints,
+)
+from ansible_collections.cisco.nd.plugins.module_utils.manage_vrf_lite.runtime_payloads import (
+    parse_instance_values,
+    parse_vrf_lite_extension_values,
+)
+
+
+def _parse_vrf_template_vlan(vrf_object: dict[str, Any]) -> int | None:
```
mikewiebe, line 35: ### Medium: VRF VLAN discovery ignores the current top-level API field

**Issue**

The query helper obtains VRF VLAN only from the legacy serialized
`vrfTemplateConfig.vrfVlanId`, while OpenAPI also returns top-level `vlanId`.

**Evidence**

- [`query.py` lines 26-46](https://github.com/CiscoDevNet/ansible-nd/pull/281/files#diff-3be2d9b08bbba67015aca2b4ea5a446741e5b1cb053820738be8886c0f7dcd42R26)
  parses only `vrfTemplateConfig`.
- [`query.py` lines 218-226](https://github.com/CiscoDevNet/ansible-nd/pull/281/files#diff-3be2d9b08bbba67015aca2b4ea5a446741e5b1cb053820738be8886c0f7dcd42R218)
  stores only that result.
- OpenAPI's VRF object exposes top-level `vlanId`.
- Action/payload helpers require the discovered VLAN for several attachment operations.
- For example, ND can return an existing VRF in the current documented shape:

  ```json
  {
    "vrfName": "TENANT-A",
    "vlanId": 500,
    "vrfTemplateConfig": "{}"
  }
  ```

  A user can then add a VRF Lite attachment without redundantly supplying the
  existing VRF VLAN:

  ```yaml
  - name: Add a VRF Lite attachment
    cisco.nd.nd_manage_vrf_lite:
      fabric_name: site-1
      state: merged
      config:
        - vrf_name: TENANT-A
          attach:
            - ip_address: 10.122.84.191
              vrf_lite:
                - interface: Ethernet1/20
                  dot1q: 500
                  ipv4_addr: 10.33.0.2/30
                  neighbor_ipv4: 10.33.0.1
  ```

  Although ND supplied `vlanId: 500`, the helper examines only
  `vrfTemplateConfig.vrfVlanId`. Because that legacy field is absent, it records
  the VRF VLAN as `None`, so the otherwise valid attachment can fail or be built
  with incomplete payload data.

**Existing PR overlap**

No matching open issue found.

**Impact**

Valid VRFs can appear to have no VLAN, causing an otherwise valid attach/detach action
to fail or use incomplete payload data.

**Suggested fix**

Prefer top-level `vlanId`, retain `vrfTemplateConfig.vrfVlanId` as a compatibility
fallback, and test both response shapes and disagreement handling.

---

plugins/modules/nd_manage_vrf_lite.py
```diff
@@ -0,0 +1,397 @@
...
+      type:
+        description:
+        - Scope of the VRF deploy operation.
+        - C(switch) deploys only the switches affected by the changed VRFs.
+        - C(global) performs a fabric-wide deploy of the changed VRFs.
+        type: str
+        default: switch
+        choices: [ switch, global ]
+
+  verify:
```
mikewiebe, line 68: **Medium: `verify.retries` does not currently retry state convergence**

The documentation says verification polls until the controller state matches the desired state or the retry limit is reached. However, this path performs one immediate re-query after the actions and assigns that result to `after/current`. The shared retry helper only retries transport or 5xx failures.

For example, if ND accepts an attachment but does not expose it until several seconds later, `verify.enabled: true` can return successfully with stale state, and `retries: 5` will not perform five convergence checks.

The cross-module implementation is being tracked by [#385](https://github.com/CiscoDevNet/ansible-nd/issues/385). Since this PR already exposes and documents `verify`, could you please either:

- implement the polling and convergence behavior in this PR; or
- resolve this thread with a note that the work is deferred to #385 and will be completed as part of that issue?

If deferred, the documentation should avoid promising retry-based convergence until #385 lands.

plugins/module_utils/orchestrators/manage_vrf_lite.py
```diff
@@ -0,0 +1,527 @@
...
+                "changed": False,
+                "config_actions": config_actions,
+                "response": [],
+            }
+
+        requested_deploy_vrfs = set(_target_vrfs_for_deploy(module))
+        # Deploy only VRFs that both changed and have deploy enabled.  The intersection
+        # ensures we never push a VRF that the user asked to skip (deploy: false) and
+        # also prevents deploying stale VRFs that the run did not actually touch.
+        target_vrfs = sorted(set(module.params.get("_changed_vrfs") or []) & requested_deploy_vrfs)
```
mikewiebe: **High: Replacement removals are not deployed to the removed switches**

`target_vrfs` and `target_switch_ids` are rebuilt from `module.params["config"]`, which contains only the flattened desired attachments at this point. Consequently, attachment identities removed by `replaced` or `overridden` are lost even though the state machine recorded their deletion in `sent`.

For example, assume `TENANT-PROD` is currently attached to both switches:

```text
SN1 → 10.122.84.191
SN2 → 10.122.84.192
```

The user removes SN2 by retaining only SN1:

```yaml
- name: Remove TENANT-PROD from SN2
  cisco.nd.nd_manage_vrf_lite:
    fabric_name: site-1
    state: replaced
    config:
      - vrf_name: TENANT-PROD
        vlan_id: 500
        attach:
          - ip_address: 10.122.84.191
            vrf_lite:
              - interface: Ethernet1/20
                dot1q: 500
                ipv4_addr: 10.33.0.2/30
                neighbor_ipv4: 10.33.0.1
```

The state machine correctly sends a detach for SN2 and marks `TENANT-PROD` as changed. However, this code derives switch deployment targets from the retained configuration and sends the equivalent of:

```json
{
  "vrfNames": ["TENANT-PROD"],
  "switchIds": ["SN1"]
}
```

SN2—the switch that needs the detach deployed—is omitted. Nexus Dashboard intent can show SN2 as detached while the switch retains the old VRF Lite configuration, and the module can still report success.

With `attach: []`, both detach operations can be submitted, but the flattened desired configuration is empty. That produces `target_vrfs=[]`, so no VRF deployment is sent at all.

Could the deployment scope instead be built from the state-machine operation journal, preserving the before-state VRF and switch identities for removed entries? Please add partial-removal and remove-all tests for both `replaced` and `overridden`, under switch and global deployment scopes.

This falls within the broader resource-level deployment work tracked by [#429](https://github.com/CiscoDevNet/ansible-nd/issues/429), but that issue does not currently document this concrete removal case.

plugins/module_utils/orchestrators/manage_vrf_lite.py
```diff
@@ -0,0 +1,527 @@
...
+
+    def delete_bulk(self, model_instances: list[Any], **kwargs: Any) -> dict[str, Any]:
+        return self._post_detach_entries(model_instances)
+
+    def _validate_attach_entries(self, entries: list[Any]) -> None:
+        module = self._module()
+        seen_vrfs = set()
+        for entry in entries:
+            if entry.vrf_name not in seen_vrfs:
+                _ensure_vrf_exists(module, self.rest_send, entry.vrf_name)
```
mikewiebe: **Low/non-blocking: Bulk writes repeat full state queries for every VRF**

`_ensure_vrf_exists()` invokes `query_vrf_lite_state()` once for each unique VRF. Each invocation reloads the fabric switch inventory, retrieves the VRF list, and queries attachments. Validation then performs another support request for every `(VRF, switch)` pair, and attachment writes are submitted separately per VRF.

As a concrete example, attaching 50 VRFs to two switches can generate approximately:

```text
Initial state query                         3 requests
50 repeated VRF existence queries        150 requests
100 per-(VRF, switch) support checks      100 requests
50 per-VRF attachment POSTs                50 requests
------------------------------------------------------
Before save, deploy, or verification      303 requests
```

Missing attachment details and automatic dot1q allocation can add still more calls. This can turn one bulk Ansible task into hundreds of sequential controller requests and exceed normal task or controller timeouts.

Could the initial `_known_vrfs` and switch inventory be reused for validation, with detail/support lookups and attachment writes batched where the APIs permit? Request-count tests for multi-VRF and multi-switch inputs would help prevent this from regressing.

This is related to the broader performance work in [#418](https://github.com/CiscoDevNet/ansible-nd/issues/418). If the optimization is intentionally deferred, please resolve this thread with a note that it will be handled as part of that issue.

plugins/modules/nd_manage_vrf_lite.py
```diff
@@ -0,0 +1,397 @@
...
+
+- name: Delete all attachments for a VRF
+  cisco.nd.nd_manage_vrf_lite:
+    fabric_name: my_fabric
+    state: deleted
+    config:
+      - vrf_name: TENANT_A
+"""
+
+RETURN = r"""
```
mikewiebe, line 269: **Low/non-blocking: `RETURN` omits fields emitted by the module**

The `RETURN` block documents only `changed`, the state fields, `warnings`, and `deployment`. Runtime output can also include:

- `output_level` and `diff`
- `proposed` at info/debug output levels
- `logs` at debug level
- `deployment_needed`
- `ip_to_sn_mapping`
- `response` and `result` during gathered operations

For example:

```yaml
- name: Manage VRF Lite
  cisco.nd.nd_manage_vrf_lite:
    fabric_name: site-1
    state: merged
    output_level: debug
    config:
      - vrf_name: TENANT-PROD
        attach:
          - ip_address: 10.122.84.191
  register: vrf_lite_result

- ansible.builtin.assert:
    that:
      - vrf_lite_result.output_level == "debug"
      - vrf_lite_result.diff is defined
      - vrf_lite_result.proposed is defined
      - vrf_lite_result.logs is defined
      - vrf_lite_result.deployment_needed is defined
```

These fields are available to playbooks but absent from generated `ansible-doc`, so consumers cannot discover their types or return conditions without reading the implementation.

Could the `RETURN` block document every emitted public field and its conditions? Direct wrapper assertions would also help keep the documented and runtime contracts aligned.

sivakasi-cisco:
> Hi Sivakami, Adding a note here as per our discussion. Everything else looks good. Pending: Need to use the common fabric details utility for the type from fabric_details_cache once #405 is merged.

Thank you for the comments, Akshay. 
I have added a To-Do comment on this. 

On the offline comments you had shared, I have addressed the possible items.
1. If force is not being used then we can remove it right? Additionally verify too -  - addressed, removed the force option which is unused
2. Yeah, this is where we might have issues. Let's say module is running and getting stuck somewhere in between. We won't know where - Addressed - added logging
3. "That makes me to think if we require any tests to be performed for interface changes from one vrf to other? In that case, a 2 step deploy would be required." this is what i was thinking too. Can you test this one out? 
    For the overridden part, there are no 2 deployments in the implemented approach and I'm planning to continue with the same, because of the below resons.
    It first compares the current and desired config, stage the updates/add/remove from ND and saves it. Then deploys the final config and verifies. So, I did not add any delay to it for removal and addition.
    I tried another approach which require 2 different deploy cycles.
    Both approaches work. So, I'm planning to stick with deploy that matches with our other plans of config_save as well which is through a single shot deploy.

plugins/module_utils/orchestrators/manage_vrf_lite.py
sivakasi-cisco: Thanks Mike. The validation was reached from the write path only, so the check mode could produce a plan without validating the VRF or switch support. I have added a read-only preflight before state reconciliation. It runs the same validation without making any write calls. Have added UT for the missing-VRF case and for limiting the preflight to check-mode write states as well.

plugins/module_utils/manage_vrf_lite/query.py
sivakasi-cisco: Addressed. The code now checks the top-level VLAN id, falls back to the legacy field. If both are present, it picks the top level one. Added UT for such cases

plugins/module_utils/orchestrators/manage_vrf_lite.py
sivakasi-cisco: Shall check this

plugins/modules/nd_manage_vrf_lite.py
sivakasi-cisco: I am deferring convergence polling to #385 so that it can be implemented consistently across the modules. For the current PR, I have corrected the documentation - verify.enabled performs a single post-deploy refresh, and verify.retries is currently reserved and does not imply repeated convergence checks.

plugins/modules/nd_manage_vrf_lite.py
sivakasi-cisco: Addressed the documentation gap.

- I expanded the RETURN block to document every public field emitted by the module, including output_level, diff, verbosity-controlled proposed and logs, deployment_needed, ip_to_sn_mapping, and gathered-state response and result.
- The nested deployment and config_actions fields, types, and return conditions are also documented. 
- Check-mode projection and the currently empty compatibility collections are now detailed.
- Added VRF Lite tests, ansible-test sanity, ansible-doc, black, and git diff --check.

plugins/module_utils/orchestrators/manage_vrf_lite.py
sivakasi-cisco: I have updated the validation flow to reuse the VRF and switch inventories loaded during the initial query, including when the inventory is empty. 

For bulk create and check mode, switch-detail lookups are now grouped into one request per VRF.

I have also added request-count tests covering multiple VRFs and switches.

Normal updates are still processed one at a time by the shared state machine, so fully batching update requests would require a broader framework change.

plugins/module_utils/orchestrators/manage_vrf_lite.py
sivakasi-cisco: The deployment targets are now taken from the operations recorded by the state machine, so switch details are retained for both attachments and removals.

This would ensure partial and remove-all operations with replaced and overridden are deployed to the removed switches including check mode.

Switch-scoped deployments also keep each VRF associated only with its affected switches, avoiding unrelated VRF/switch combinations. I added tests covering partial and remove-all cases, switch and global scopes, normal and check mode, mixed updates,removals, deploy: false, and switch identifier resolution.

allenrobel (commented):
Re-review after the new commits (my earlier approval was dismissed by the June pushes).

First, credit where due: I verified all 15 previously-raised review threads against the current code, and every one is genuinely addressed — deploy targets now derive from the state-machine operation journal (with real wire-payload tests proving replaced/overridden removals deploy to the removed switches, including check mode), `config_actions.type` correctly scopes deploy, the check-mode preflight exists, nd_v2 is gone, FabricContext is used with a shared inventory cache, and the bulk validation N+1 is fixed with request-count tests. The unit suite is largely real-assertion (wire payloads, request counts, exact endpoint paths). Nice work.

The inline comments below are new findings from this pass, roughly ordered by severity:

- Two framework-level concerns in `nd_state_machine.py` / `models/base.py`: the merged-state `allow_superset` diff change applies ungated to every Gen-3 module (with a concrete idempotency regression scenario on shipped interface modules), and diff/merge now disagree about list semantics.
- Two model-validation range bugs (`dot1q`, `vlan_id`) confirmed against the ND 4.2.1 OpenAPI schema, plus a `vrf_name` length that matches no schema constraint.
- A failure-detection gap that can report a failed attachment as success.
- Check-mode vs real-run validation divergence.
- Several unmarked ND-API workarounds that need `TODO(X.Y.Z) <slug>` markers per the repo convention.
- Endpoint-layer duplication/dead code, the `ignore-2.15.txt` sanity file, integration teardown leaving deployed config on switches, and a typing convention in `utils.py`.

plugins/module_utils/nd_state_machine.py
```diff
@@ -167,9 +180,11 @@ def _manage_create_update_state(self) -> None:
                 # Determine diff status
                 # For merged state, only compare fields explicitly provided by
                 # the user so that Pydantic default values do not trigger false
-                # diffs or overwrite existing configuration.
+                # diffs or overwrite existing configuration. Merged also matches
+                # list elements one-directionally (allow_superset) so an
+                # existing item with extra list entries is not seen as changed.
                 exclude_unset = self.state == "merged"
-                diff_status = self.existing.get_diff_config(proposed_item, exclude_unset=exclude_unset)
+                diff_status = self.existing.get_diff_config(proposed_item, exclude_unset=exclude_unset, allow_superset=exclude_unset)
```
allenrobel: **High: merged-state `allow_superset` changes diff semantics for every Gen-3 module, ungated.**

`allow_superset=exclude_unset` means every module using `NDStateMachine` gets one-directional list matching in merged state — not just vrf_lite. Concrete regression on a shipped module: `nd_interface_port_channel_access` with existing `ports: [Eth1/1, Eth1/2]` and a merged task proposing `ports: [Eth1/1]` — on develop this is a diff (the update replaces the list, removing Eth1/2); with this change it becomes `no_diff` and no update is sent. The same exposure applies to the vPC `peer1_member_ports`/`peer2_member_ports` lists, trunk-host `vlan_mapping_entries`, and the fabric `ntp/dns/syslog` collection lists. `merge_would_change` is not a safety net here — it only recurses into nested `NDBaseModel` fields, never lists.

The new semantic may well be the *better* definition of merge, but it silently changes idempotency results of shipped modules. Note the contrast with your other framework change in this PR: delete-tracking is properly gated behind the `track_deletes_in_sent` ClassVar that only the vrf_lite orchestrator opts into. Could we gate this the same way (orchestrator-level opt-in), or if the team decides the new semantics should apply globally, make that an explicit, tested, changelogged decision rather than a side effect of this PR?

plugins/module_utils/models/base.py
```diff
@@ -221,23 +221,26 @@ def to_diff_dict(self, **kwargs) -> Dict[str, Any]:
             **kwargs,
         )
 
-    def get_diff(self, other: "NDBaseModel", exclude_unset: bool = False) -> bool:
+    def get_diff(self, other: "NDBaseModel", exclude_unset: bool = False, allow_superset: bool = False) -> bool:
```
allenrobel, line 224: **High: `get_diff(allow_superset=True)` and `merge()` now disagree about list semantics.**

With `allow_superset`, the diff reports `no_diff` when the existing item has extra list elements (extras are "tolerated"). But `merge()` still wholesale-replaces list fields (`setattr(self, field_name, value)` — lists are not `NDBaseModel`, so there is no element-wise merge). Result: existing `ports: [Eth1/1, Eth1/2]`, merged task proposing `ports: [Eth1/1, Eth1/3]` → diff says `changed` → merge sets `ports=[Eth1/1, Eth1/3]` → `Eth1/2` is dropped, contradicting the allow_superset story that extras are preserved. Whether existing extra list elements survive a merged run currently depends on whether any *other* field on the item changed.

Whichever list semantic the team lands on (see my comment in `nd_state_machine.py`), diff and merge need to agree on it.

plugins/module_utils/models/manage_vrf_lite/vrf_lite_model.py
```diff
@@ -0,0 +1,385 @@
...
+        use_enum_values=True,
+        validate_assignment=True,
+        populate_by_name=True,
+        validate_by_alias=True,
+        validate_by_name=True,
+        extra="ignore",
+    )
+
+    interface: str = Field(alias="interface", min_length=1, max_length=128)
+    dot1q: int | None = Field(default=None, alias="dot1q", ge=1, le=4094)
```
allenrobel: **High: `dot1q` lower bound should be `ge=2`, and `vlan_id` likewise (3 sites).**

The ND 4.2.1 OpenAPI schema (`vrfExtensionCommon.dot1qId`) is `minimum: 2, maximum: 4094`, and `vrfAttachmentCommon.vlanId` is `minimum: 2, maximum: 4094` ("Must be between 2 and 4094"). The model uses `ge=1` here and for `vlan_id` at lines 160 and 254 of this file plus `vrf_lite_attachment_entry.py` line 48 — so a playbook value of `1` passes client-side validation and is rejected only by the controller.

Related: the range (and the `validate_vrf_name` validator) is duplicated verbatim between `VrfLiteModel` and `VrfLitePlaybookItemModel` — a shared mixin or field constant would keep the constraint in one place so a future fix lands everywhere.

plugins/module_utils/manage_vrf_lite/actions.py
```diff
@@ -0,0 +1,404 @@
...
+        instance_values=raw_attach.get("instance_values"),
+        freeform_config=raw_attach.get("freeform_config"),
+    )
+
+
+def _collect_attachment_failures(response: Any) -> list[str]:
+    failures: list[str] = []
+
+    if isinstance(response, str):
+        if "failed" in response.lower():
```
allenrobel: **Medium: the string branch misses failure spellings the dict branch catches, so a failed attachment can report success.**

The string branch matches only `"failed"`, but the dict branch a few lines down treats `{"failed", "failure", "error"}` as failures. A controller row value like `"Failure: peer not found"` (contains "failure", not "failed") falls through the string branch and is silently treated as success — the module reports changed/ok after a failed attachment POST. Suggest matching the same signature set in both branches.

plugins/module_utils/orchestrators/manage_vrf_lite.py
```diff
@@ -0,0 +1,650 @@
...
+        read-only ``_validate_attach_entries`` guardrails so check mode rejects
+        the same inputs the execution path would.
+        """
+        module = self._module()
+        if not module.check_mode:
+            return
+        state = module.params.get("_vrf_lite_requested_state") or module.params.get("state", "merged")
+        if state not in ("merged", "replaced", "overridden"):
+            return
+        attach_entries = [entry for entry in proposed_entries if getattr(entry, "switch_ip", None)]
```
allenrobel, line 313: **Medium: check mode validates ALL proposed entries, while a real run validates only written ones — so an idempotent playbook can pass a real run but fail check mode.**

The preflight filters only on `switch_ip` presence and validates everything, whereas the execution path reaches `_validate_attach_entries` solely for changed/created entries (inside `_post_attach_entries`). Divergence example: an already-attached row whose switch has since dropped from the fabric inventory (or whose prototype metadata is gone) — real run: no diff, green; check mode: `Switch '...' is not present in fabric` failure. It also costs support/prototype GETs on fully idempotent check runs.

Suggest running the preflight against the entries the state machine actually classified as create/update, so check mode and real runs validate the same set.

plugins/module_utils/manage_vrf_lite/deploy.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def _is_non_fatal_config_save_error(error: NDModuleError) -> bool:
+    if not isinstance(error, NDModuleError):
+        return False
+    if error.status != 500:
+        return False
+
+    message = (error.msg or "").lower()
+    signatures = (
```
allenrobel, line 84: **Medium: unmarked ND-API workaround — needs a `TODO(X.Y.Z) <slug>` marker.**

Swallowing an HTTP 500 from configSave based on message sniffing is a workaround for a controller fault (wrong status for a non-fatal condition), and per the repo convention every such site must carry a `TODO(4.2.1) <slug>` marker referencing a bug-tracker vault note, so the workaround can be audited and removed when the release cadence permits. Please add the marker (and the vault note if one doesn't exist yet).

plugins/module_utils/manage_vrf_lite/actions.py
```diff
@@ -0,0 +1,404 @@
...
+        module,
+        rest_send,
+        VrfLiteEndpoints.reserve_id(fabric_name),
+        HttpVerbEnum.POST,
+        payload,
+    )
+
+    dot1q_value = None
+    if isinstance(response, dict):
+        # Different controller versions return either a scalar DATA-equivalent
```
allenrobel: **Medium: unmarked ND-API workaround — needs a `TODO(X.Y.Z) <slug>` marker.**

The comment itself documents a wire-shape discrepancy ("Different controller versions return either a scalar DATA-equivalent or a dict-like body for reserve-id") and the code probes `("dot1q", "id", "value", "allocatedId")` with a scalar fallback. This is exactly the class of workaround the `TODO(X.Y.Z) <slug>` convention exists for — please add the marker with a vault-note slug.

plugins/module_utils/manage_vrf_lite/query.py
```diff
@@ -0,0 +1,441 @@
...
+def _query_vrf_switch_details(
+    module: Any,
+    rest_send: Any,
+    fabric_name: str,
+    vrf_name: str,
+    serial_numbers: list[str],
+) -> dict[str, dict[str, Any]]:
+    """Return per-switch VRF attachment details keyed by serial number.
+
+    This makes one additional API call per VRF that has attachments lacking
```
allenrobel: **Medium: unmarked ND-API workaround, and a per-item GET in a loop.**

Two things here:

1. The bulk attachment list silently dropping `extensionValues` for PENDING rows is a controller discrepancy — it needs a `TODO(X.Y.Z) <slug>` marker referencing a vault note per the repo convention.
2. The enrichment is one GET per affected VRF inside the loop over `attachment_objects`. The `vrf_switch` endpoint's query param is plural (`vrf-names`), which suggests the missing VRFs could be batched into a single cross-VRF request the way the validation prefetch batches serial numbers.

plugins/module_utils/manage_vrf_lite/query.py
```diff
@@ -0,0 +1,441 @@
...
+
+    return _coerce_vlan(parsed.get("vrfVlanId"))
+
+
+def _resolve_vrf_vlan(vrf_object: dict[str, Any]) -> int | None:
+    """Resolve the VRF VLAN id from a controller VRF object.
+
+    Prefer the current top-level ``vlanId`` field returned by the OpenAPI VRF
+    object and fall back to the legacy serialized ``vrfTemplateConfig.vrfVlanId``
+    for older controller response shapes. Without this, a VRF returned in the
```
allenrobel, line 56: **Medium: unmarked ND-API compatibility workaround — needs a `TODO(X.Y.Z) <slug>` marker.**

The legacy `vrfTemplateConfig.vrfVlanId` fallback "for older controller response shapes" is a version-discrepancy workaround. The same applies to the multi-spelling key probing further down (`lanAttachState`/`lanAttachedState`, `isLanAttached`/`isAttached`/`attach`, and the lowercase-l `islanAttached` variant, plus the flat-row regrouping for responses lacking `lanAttachList`). Each site should carry a marker with a vault-note slug so these can be audited against future ND releases.

tests/sanity/ignore-2.15.txt
```diff
@@ -0,0 +1,64 @@
+ansible-nd/plugins/module_utils/rest/response_strategies/nd_v1_strategy.py pylint:invalid-name
```
allenrobel: **Medium: this file should be deleted.**

Four problems: (1) `meta/runtime.yml` requires ansible >= 2.16 and the CI matrix runs stable-2.16 through stable-2.19, so no supported context ever reads a 2.15 ignore file; (2) half the lines carry an `ansible-nd/` path prefix — sanity-ignore paths are relative to the collection root, so these reference nonexistent files and would error if a 2.15 sanity run ever did execute; (3) the `shebang` ignores suppress nothing real (the modules all start with the correct `#!/usr/bin/python`); (4) it lacks the 4 `action-plugin-docs` entries every maintained ignore file (2.16–2.19) contains, so it wasn't derived from the real ignore set.

plugins/module_utils/endpoints/v1/manage/manage_fabrics_actions_config_save.py
```diff
@@ -44,8 +45,11 @@ class EpFabricConfigSavePost(
...
             raise ValueError("fabric_name is required")
-        return BasePath.path("fabrics", self.fabric_name, "actions", "configSave")
+        return BasePath.path("fabrics", quote(self.fabric_name, safe=""), "actions", "configSave")
 
     @property
     def verb(self) -> HttpVerbEnum:
         return HttpVerbEnum.POST
+
+
+EpManageFabricsActionsConfigSavePost = EpFabricConfigSavePost
```
allenrobel: **Medium: this alias collides with a different existing class of the same name.**

`manage_fabrics_actions.py` already defines a distinct `class EpManageFabricsActionsConfigSavePost(_EpManageFabricsActionsBase)` (unquoted path, different base), which is actively imported by `manage_switches/utils.py`. With this alias, the same public class name now resolves to two different implementations depending on import path. Nothing in the PR imports the alias (nor the sibling `EpManageFabricsActionsDeployPost` alias in `manage_fabrics_actions_deploy.py`) — suggest dropping both.

plugins/module_utils/endpoints/v1/manage/manage_fabrics_switches.py
```diff
@@ -460,3 +461,23 @@ def path(self) -> str:
     def verb(self) -> HttpVerbEnum:
         """Return the HTTP verb for this endpoint."""
         return HttpVerbEnum.POST
+
+
+class EpFabricSwitchesGet(FabricNameMixin, NDEndpointBaseModel):
```
allenrobel: **Medium: dead code — `EpFabricSwitchesGet` has no callers and duplicates `EpManageFabricsSwitchesGet` in this same file.**

The vrf_lite code gets its switch inventory via `FabricContext`, so nothing uses this class. Its only difference from the canonical `EpManageFabricsSwitchesGet` is URL-quoting of the fabric name — but since the quoting fix wasn't applied to the class real callers use, the two now diverge and the "fix" reaches nobody. Its name also breaks the file's `EpManageFabrics*` naming convention. Suggest removing it (or, if the quoting fix is wanted, applying it to `EpManageFabricsSwitchesGet` itself).

plugins/module_utils/manage_vrf_lite/runtime_endpoints.py
```diff
@@ -0,0 +1,82 @@
...
+    def _fabric_path(fabric_name: str, *segments: str) -> str:
+        return BasePath.path("fabrics", quote(fabric_name, safe=""), *segments)
+
+    @staticmethod
+    def _top_down_fabric_path(fabric_name: str, *segments: str) -> str:
+        suffix = "/".join(("fabrics", quote(fabric_name, safe=""), *segments))
+        return "{0}/{1}".format(VrfLiteEndpoints._TOP_DOWN_API, suffix)
+
+    @staticmethod
+    def vrfs(fabric_name: str) -> str:
```
allenrobel, line 40: **Medium: hand-built path strings duplicate existing smart-endpoint models for the Manage-API paths.**

The NDFC top-down paths have no endpoint-model equivalent, so building those here is fine. But the Manage-API ones already exist: `vrfs()` → `EpManageFabricsVrfsGet`; `vrf_deployments()` → `EpManageFabricsVrfActionsDeployPost`; `config_save()` → `EpFabricConfigSavePost` (which this PR itself modifies to add quoting). And `config_deploy()` hand-concatenates `?forceShowRun=true` even though this PR adds a `force_show_run` field to `EpFabricDeployPost` for exactly that purpose — and then never uses it. Suggest delegating the Manage-API paths to the endpoint models (the vpc_pair `runtime_endpoints.py` precedent does this) and dropping whichever of the two `forceShowRun` mechanisms loses.

Also: `fabric_switches` and `config_deploy` in this class have zero callers — dead code, along with `build_vrf_lite_extension_values_api` / `build_instance_values_api` in `runtime_payloads.py`.

plugins/module_utils/models/manage_vrf_lite/vrf_lite_model.py
```diff
@@ -0,0 +1,385 @@
...
+        str_strip_whitespace=True,
+        use_enum_values=True,
+        validate_assignment=True,
+        populate_by_name=True,
+        validate_by_alias=True,
+        validate_by_name=True,
+        extra="ignore",
+    )
+
+    vrf_name: str = Field(alias="vrf_name", min_length=1, max_length=64)
```
allenrobel: **Medium: `max_length=64` matches no ND schema constraint.**

The ND 4.2.1 OpenAPI gives the plain VRF name (`schemas-vrfName`) `maxLength: 32`, and the attachment-key form (`vrfAttachmentCommon.vrfName`, "tenantName~vrfName") `maxLength: 94`. 64 is neither: a 33-char name passes the model but a plain VRF create caps at 32, while a legitimate 70-char `tenant~vrf` attachment key would be rejected client-side. Pick the bound that matches what this field actually carries (same at line 253).

plugins/module_utils/models/manage_vrf_lite/vrf_lite_model.py
```diff
@@ -0,0 +1,385 @@
...
+            mode="json",
+            **kwargs,
+        )
+
+    @classmethod
+    def from_response(cls, response: dict[str, Any], **kwargs) -> "VrfLiteModel":
+        return cls.model_validate(response, by_alias=True, by_name=True, **kwargs)
+
+    @classmethod
+    def from_config(cls, ansible_config: dict[str, Any], **kwargs) -> "VrfLiteModel":
```
allenrobel: **Medium: these `from_response`/`from_config` overrides drop the base class's semantics — and are dead code.**

The base `NDBaseModel.from_config` runs `_strip_none_values(...)` and passes `context={"mode": "config"}` specifically so Ansible's implicit `None`s don't pollute `model_fields_set` (which merged-state diffing depends on via `exclude_unset`). These overrides skip both. No caller invokes `VrfLiteModel.from_config`/`from_response` (the only `from_config` call site is `VrfLiteAttachmentEntry.from_config` in `config_transform.py`, which correctly inherits the base) — suggest deleting the overrides; if they're kept for a future caller, they should preserve the base behavior.

plugins/module_utils/models/manage_vrf_lite/vrf_lite_model.py
```diff
@@ -0,0 +1,385 @@
...
+        str_strip_whitespace=True,
+        use_enum_values=True,
+        validate_assignment=True,
+        populate_by_name=True,
+        validate_by_alias=True,
+        validate_by_name=True,
+        extra="ignore",
+    )
+
+    interface: str = Field(alias="interface", min_length=1, max_length=128)
```
allenrobel: **Medium: every `Field(alias=...)` in these models just repeats the snake_case field name, defeating the Gen-3 alias contract.**

In the existing Gen-3 models (`models/local_user/`, `models/interfaces/`), `alias` carries the camelCase wire key so `NDBaseModel.to_payload()` (`by_alias=True`) emits the API payload directly. Here `to_payload()` emits ansible snake_case, and the real wire keys (`IF_NAME`, `DOT1Q_ID`, `IP_MASK`, `PEER_VRF_NAME`, ...) are hand-mapped in `runtime_payloads.py`. As written the aliases are pure noise — either carry the wire names in the aliases (and let `to_payload()` do its job) or drop the alias declarations.

tests/integration/targets/nd_manage_vrf_lite/tasks/nd_manage_vrf_lite_merge.yaml
```diff
@@ -0,0 +1,899 @@
...
+    vrf_lite_assert_ipv4_addr: "{{ test_ipv4_addr2 }}"
+    vrf_lite_assert_neighbor_ipv4: "{{ test_neighbor_ipv4_2 }}"
+    vrf_lite_assert_peer_vrf: "{{ test_peer_vrf }}"
+  tags: merge
+
+##############################################
+##                 CLEAN-UP                 ##
+##############################################
+
+- name: MERGE - END - ensure clean state
```
allenrobel, line 865: **Medium: teardown never deploys the detach, leaving vrf-lite CLI in switch running-config after the test run.**

TC8/TC9 in this file deploy to hardware (`save: true, deploy: true`), but this cleanup — and the equivalent END blocks in `nd_manage_vrf_lite_replace.yaml` and `nd_manage_vrf_lite_override.yaml` — runs `state: deleted` with `save: false, deploy: false`, so the detach intent is staged but never deployed. The scale target's cleanup does this correctly (`save: true, deploy: true`), as does delete TC2b. Suggest aligning the merge/replace/override teardowns so a test run leaves the switches clean.

plugins/module_utils/orchestrators/manage_vrf_lite.py
```diff
@@ -0,0 +1,650 @@
...
+        current = self._query_current_state(flat=True)
+        logger.debug("query_all: loaded %d current VRF Lite attachment row(s)", len(current))
+        self._prepare_state_machine_config(current)
+        return current
+
+    def create(self, model_instance: Any, **kwargs: Any) -> dict[str, Any]:
+        return self.create_bulk([model_instance], **kwargs)
+
+    def update(self, model_instance: Any, **kwargs: Any) -> dict[str, Any]:
+        # TODO(#418): add a post-diff bulk-update hook so multiple changed
```
allenrobel: **Medium: `TODO(#418)` points at an issue that doesn't cover this work.**

Issue #418 is "Define Performance and Scale Limits for testing" — establishing baselines and limits. It doesn't mention a post-diff bulk-update hook, so this per-entry update N+1 (N changed rows in one VRF → N attachment POSTs plus repeated validation passes, where one grouped POST per VRF would do) isn't actually tracked anywhere actionable. Suggest filing a dedicated issue for the bulk-update hook (or expanding #418's scope explicitly) and pointing this TODO at it.

plugins/module_utils/utils.py
```diff
@@ -2,11 +2,11 @@
 
 # GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
 
-from __future__ import absolute_import, annotations, division, print_function
+from __future__ import annotations
 
 import logging
 from copy import deepcopy
-from typing import Any
+from typing import Any, Dict, List, Set
```
allenrobel: **Convention: new code should use PEP 585 builtins.**

The new bipartite-matching code uses `List[List[int]]`, `Dict[int, int]`, `Set[int]` — per the repo convention, new code uses `list[list[int]]`, `dict[int, int]`, `set[int]`, and the `Dict`/`List`/`Set` imports here can be dropped (this file already has `from __future__ import annotations`). Side note: the greedy→bipartite `issubset` rewrite itself looks like a genuine correctness improvement and the new tests in `test_utils.py` are excellent — but note it changes results for all existing callers including the Gen-1 idempotency checks in `nd.py` (direction is strictly "fewer false changed"), which is worth a line in the PR description/changelog.

sivakasi-cisco: Converted them and dropped the imports.

plugins/module_utils/nd_state_machine.py
sivakasi-cisco: Hi Allen, Thank you for your in-depth review and comments. 

Here, I have added an opt-in flag merge_allow_list_superset, which is false by default, in orchestrators/base.py. Now, the nd_state_machine.py only enables superset behavior when a module opts in. Currently only the vrf_lite opts in via orchestrators/manage_vrf_lite.py. 

This would not have been kept global.  The state machine now uses allow_superset = exclude_unset and self.merge_allow_list_superset, so only vrf_lite gets the new behavior.

plugins/module_utils/models/base.py
sivakasi-cisco: Resolved by the previous reply in gating as well as the fact that vrf_lite already merges lists element-wise.

The superset comparison is now limited to VRF Lite, whose models merge attachments and extensions element by element. Existing unmentioned entries are therefore preserved by both diff and merge.

plugins/module_utils/models/manage_vrf_lite/vrf_lite_model.py
sivakasi-cisco: Updated dot1q and every vlan_id field to accept only 2–4094, matching the schema. A value of 1 now fails locally before any controller request.

sivakasi-cisco: Added validators as well

plugins/module_utils/manage_vrf_lite/actions.py
sivakasi-cisco: Made one shared set _FAILURE_STATUSES = {"failed","failure","error"} and used it in both the text and dictionary branches in actions.py, hence it could catch errors, failed and failures

plugins/module_utils/orchestrators/manage_vrf_lite.py
sivakasi-cisco: Check mode now preflights the pending_create_update_items(). It validates only the create/update subset instead of every proposed row. So --check and a real run validate the same set and idempotent no-op rows aren't rejected.

plugins/module_utils/manage_vrf_lite/query.py
sivakasi-cisco: Added the TODO(4.2.1) vrf-lite-pending-extensionvalues-dropped marker and, per your note, batched the enrichment into one cross-VRF. Also added a test asserting a single request carrying plural vrf-names

plugins/module_utils/manage_vrf_lite/deploy.py
sivakasi-cisco: Added TODO(4.2.1) vrf-lite-configsave-500-nonfatal with the reason the HTTP 500 is treated as non-fatal and the condition for removing the workaround.

plugins/module_utils/manage_vrf_lite/actions.py
sivakasi-cisco: Marked with TODO(4.2.1) vrf-lite-reserve-id-shape-drift, noting the probing is for controller-version shape drift and can be removed once the response shape is stable.

plugins/module_utils/manage_vrf_lite/query.py
sivakasi-cisco: Added TODO(4.2.1) markers for the legacy vrfVlanId fallback, the multi-spelling attachment keys, and the flat-list regrouping, each noting removal once the controller shapes stabilize.

tests/sanity/ignore-2.15.txt
sivakasi-cisco: Deleted

plugins/module_utils/endpoints/v1/manage/manage_fabrics_actions_config_save.py
sivakasi-cisco: Dropped both back-compat aliases.  Callers can use the canonical EpFabricConfigSavePost and  EpFabricDeployPost.

plugins/module_utils/endpoints/v1/manage/manage_fabrics_switches.py
sivakasi-cisco: Removed

plugins/module_utils/manage_vrf_lite/runtime_endpoints.py
sivakasi-cisco: Delegated the Manage API paths to the canonical endpoint models and removed the dead fabric_switches/config_deploy helpers plus the two unused payload builders. Only the top-down paths the Manage API doesn't expose remain hand-built; the unit test asserts their percent-encoding.

sivakasi-cisco:
I had resolved the comments here, but for the vault-note mentions on TODO(4.2.1) markers, could you pls help me with some reference?

plugins/module_utils/models/manage_vrf_lite/vrf_lite_model.py
sivakasi-cisco: Fixed. This field carries the plain VRF name with schemas-vrfName, maxLength: 32

plugins/module_utils/models/manage_vrf_lite/vrf_lite_model.py
sivakasi-cisco: Deleted both overrides.

plugins/module_utils/models/manage_vrf_lite/vrf_lite_model.py
sivakasi-cisco: I have dropped all the redundant aliases. The earlier repeated the snake_case field name and added no value: to_payload() here emits Ansible snake_case and the actual wire keys (IF_NAME, DOT1Q_ID, IP_MASK, PEER_VRF_NAME, …) are hand-mapped in runtime_payloads.py

tests/integration/targets/nd_manage_vrf_lite/tasks/nd_manage_vrf_lite_merge.yaml
sivakasi-cisco: Addressed.

All 3 teardowns now deploy the detach (save: true, deploy: true).

nd_manage_vrf_lite_merge.yaml - END clean state + secondary clean state (for test_switch2) both save: true, deploy: true

nd_manage_vrf_lite_replace.yaml — save: true, deploy: true

nd_manage_vrf_lite_override.yaml — save: true, deploy: true

plugins/module_utils/orchestrators/manage_vrf_lite.py
sivakasi-cisco: Shall create a dedicated issue on this

sivakasi-cisco: updated with https://github.com/CiscoDevNet/ansible-nd/issues/487

allenrobel:
> I had resolved the comments here, but for the vault-note mentions on TODO(4.2.1) markers, could you pls help me with some reference?

Replied via Webex with references.

allenrobel (dismissed):
LGTM after comments were addressed.

One minor nit, the TODO(4.2.1) marker should not contain 'slug: <slug>'. 

Example (correct):

`TODO(4.2.1) vrf-lite-vlan-shape-drift`

Example (incorrect):

`TODO(4.2.1) slug: vrf-lite-vlan-shape-drift`

But this is a consistency/style nit.  Functionally, 'slug: <slug>' works, but it's more verbose.

sivakasi-cisco:
> slug

Addressed them in the last changes. Confirmed that there is no "slug:" form anywhere in the code.

PR #282, opened 2026-05-17
Update httpapi login and logout to support 4.2.1+

mtarking:
## Proposed Changes

Updates the httpapi plugin (`plugins/httpapi/nd.py`) to auto-detect whether the Nexus Dashboard instance uses the new authentication endpoints (`/api/v1/infra/login`, `/api/v1/infra/logout`) introduced in ND 4.2.1+ or the legacy endpoints (`/login`, `/logout`).

- Added `_attempt_login()` helper method that encapsulates a single login attempt and returns `(token, status)`
- Updated `login()` with auto-detection logic: tries `/api/v1/infra/login` first, falls back to `/login` on failure
- Caches the detected auth base path (`self.auth_base_path`) so subsequent re-authentication skips detection
- Updated `logout()` to use the detected path (`/api/v1/infra/logout` or `/logout`)
- Token parsing now reads `jwttoken` first, falling back to `token`

No configuration changes required — detection is transparent and existing inventories/playbooks work without modification. API key auth (`ansible_httpapi_session_key`) is unaffected since it skips login entirely.

## Test Notes

- Verified syntax compiles cleanly
- Manual tested against ND < 4.2.1 (legacy `/login`) and ND >= 4.2.1 (`/api/v1/infra/login`)
- API key auth path unchanged and unaffected

- Developed against ND 4.2.1+
- Backward compatible with ND 3.2.2m

* [ ] analyze
* [x] infa
* [ ] manage
* [ ] onemanage
* [ ] other

## Checklist


gmicol (dismissed):
LGTM

plugins/httpapi/nd.py
```diff
@@ -112,15 +113,26 @@ def get_token(self):
     def get_url_connection(self):
         return self.connection._url
 
+    def _attempt_login(self, path, payload):
+        """Attempt login at the given path. Returns (token, status) on success, raises on fatal error, returns (None, status) on non-200."""
+        method = "POST"
+        full_path = self.connection.get_option("host") + path
```
akinross: what is full_path used for in this method?

plugins/httpapi/nd.py
```diff
@@ -112,15 +113,26 @@ def get_token(self):
     def get_url_connection(self):
         return self.connection._url
 
+    def _attempt_login(self, path, payload):
+        """Attempt login at the given path. Returns (token, status) on success, raises on fatal error, returns (None, status) on non-200."""
```
akinross: does "raises on fatal error" mean that it passed the raise from other functions/methods called?

plugins/httpapi/nd.py
```diff
@@ -112,15 +113,26 @@ def get_token(self):
...
+        response, response_data = self.connection.send(path, data, method=method, headers=self.headers)
+        status = response.getcode()
+        if status not in [200, 201] or (status == 200 and response_data.seek(0, 2) == 0):
+            return None, status
+        response_json = self._response_to_json(response_data)
+        token = response_json.get("jwttoken") or response_json.get("token")
+        return token, status
+
     def login(self, username, password):
         """Log in to ND"""
```
akinross: should we expand the description of login function as we do for newer functions/methods?

plugins/httpapi/nd.py
```diff
@@ -112,15 +113,26 @@ def get_token(self):
     def get_url_connection(self):
         return self.connection._url
 
+    def _attempt_login(self, path, payload):
```
akinross: for new functions/method should we introduce typehinting?

plugins/httpapi/nd.py
```diff
@@ -134,23 +146,39 @@ def login(self, username, password):
...
+                        self.connection.queue_message("info", "login() - detected new auth endpoint: /api/v1/infra")
+                    else:
+                        # Fall back to legacy endpoint
+                        path = "/login"
+                        token, self.status = self._attempt_login(path, payload)
+                        if token is not None:
+                            self.auth_base_path = ""
+                            self.connection.queue_message("info", "login() - detected legacy auth endpoint: /login")
+
+                full_path = self.connection.get_option("host") + path
```
akinross: full_path appears only used when token is None to raise ConnectionError, should we declare the variable within that scope?

plugins/httpapi/nd.py
```diff
@@ -73,6 +73,7 @@ def __init__(self, *args, **kwargs):
         self.status = -1
         self.info = {}
         self.version = None
+        self.auth_base_path = None
```
akinross: for me it is not entirely clear from this PR what the purpose is for auth_base_path and how this is set, could you perhaps explain the difference with path as an inline comment? Also why is this None and not "" like path, should they behave similar?

plugins/httpapi/nd.py
```diff
@@ -134,23 +146,39 @@ def login(self, username, password):
...
-                self.connection.queue_message("info", "login() - connection.send({0}, LOGIN_PAYLOAD_NOT_SHOWN, {1}, {2})".format(path, method, self.headers))
-                response, response_data = self.connection.send(path, data, method=method, headers=self.headers)
+                # Auto-detect auth endpoint: try new path first (ND 4.2.1+), fall back to legacy
+                if self.auth_base_path is not None:
+                    # Already detected, use cached path
+                    path = self.auth_base_path + "/login"
+                    token, self.status = self._attempt_login(path, payload)
+                else:
+                    # Try new endpoint first
+                    path = "/api/v1/infra/login"
```
akinross: should we define all path related strings as some form of constant?

akinross (changes requested):

plugins/httpapi/nd.py
mtarking: Add comment to help.

plugins/httpapi/nd.py
mtarking: I think a better comment would have been this looks stale :-) 

It was stale - removed it and refactored cleaner.

plugins/httpapi/nd.py
mtarking: updated docstring. means raises whatever comes from connection send.

akinross (approved):

gmicol (approved):
LGTM

mikewiebe (approved):

plugins/httpapi/nd.py
```diff
@@ -134,23 +182,37 @@ def login(self, username, password):
...
+                    # Previously detected new endpoint, use cached path
+                    path = self.auth_base_path + LOGIN_ENDPOINT
+                    token, self.status = self._attempt_login(path, payload)
+                else:
+                    # Try new endpoint first (ND 4.2.1+)
+                    path = AUTH_BASE_PATH_NEW + LOGIN_ENDPOINT
+                    try:
+                        token, self.status = self._attempt_login(path, payload)
+                    except Exception:
+                        token, self.status = None, -1
```
samiib: Should we log the error here on the new endpoint attempt before falling back? Otherwise any errors here would be swallowed. 
Perhaps a log like this:
```suggestion
                    except Exception as e:
                        self.connection.queue_message(
                            "debug",
                            "login() - new auth endpoint {0} failed ({1}), falling back to legacy endpoint".format(
                                AUTH_BASE_PATH_NEW + LOGIN_ENDPOINT, e
                            )
                        )
                        token, self.status = None, -1
```

samiib (changes requested):

samiib (approved):
LGTM

PR #283, opened 2026-05-18
nd_log - playbook-side logging

allenrobel:
## Related Issue(s)

N/A

## Proposed Changes

- New `cisco.nd.nd_log` module: emits a Python `logging` record under the `nd` parent logger so playbook activity can be correlated with the `nd.*` records emitted internally by other modules in this collection.
- Local-only — no HTTP/API calls. Logging is configured by the existing `ND_LOGGING_CONFIG` env var consumed by `common/log.py` (parent logger `nd`, `setup_logging()` helper).
- Mirrors `cisco.dcnm.dcnm_log` in shape and argument surface (`msg`, `severity` with choices `CRITICAL/DEBUG/ERROR/INFO/WARNING`, default `DEBUG`).
- `supports_check_mode=True` — deliberate deviation from dcnm_log parity; logging is a safe side effect during `--check` runs.
- Strict config-error behavior via `setup_logging()`: a malformed/missing `ND_LOGGING_CONFIG` file calls `module.fail_json` rather than silently swallowing the error.
- 9 unit tests under `tests/unit/modules/` covering init defaults, missing-`msg` `ValueError`, parametrized severity dispatch via `caplog`, and default-vs-explicit DEBUG equivalence. New `tests/unit/modules/__init__.py` is empty (sanity `empty-init` rule).

## Test Notes

- Unit tests: `python -m pytest tests/unit/modules/test_nd_log.py -v` — 9/9 pass.
- Lint: `black --check`, `isort --check`, `pylint`, `mypy` — clean (only the standard Ansible-module advisories that also appear on `nd_version.py`).
- `ansible-doc cisco.nd.nd_log` renders correctly.
- End-to-end smoke playbook: task completes with `changed=0`; record appended to `/tmp/nd.log` per the configured `ND_LOGGING_CONFIG`.

4.2 (module does not call ND APIs; tested against the 4.2 lab for parity with the rest of the collection).

* [ ] analyze
* [ ] infra
* [ ] manage
* [ ] onemanage
* [x] other

## Checklist

### DIFF
--- a/plugins/modules/nd_log.py
+++ b/plugins/modules/nd_log.py
@@ -1,0 +1,170 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2026, Allen Robel (@quantumonion) <arobel@cisco.com>
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import annotations
+
+ANSIBLE_METADATA = {"metadata_version": "1.1", "status": ["preview"], "supported_by": "community"}
+
+DOCUMENTATION = r"""
+---
+module: nd_log
+version_added: "1.5.0"
+short_description: Log messages from a playbook via the standard Python logging module.
+description:
+- Emit a log record under the C(nd) parent logger using the standard Python C(logging) module.
+- Logging is configured by the C(ND_LOGGING_CONFIG) environment variable, which must point to a JSON file consumable by C(logging.config.dictConfig).
+  When the variable is unset, the C(nd) logger is a no-op and the module exits successfully without writing anything.
+- Useful for correlating playbook task execution with the C(nd.*) log records emitted internally by other modules in this collection.
+options:
+  msg:
+    description:
+    - The message to log.
+    required: true
+    type: str
+  severity:
+    description:
+    - Case-sensitive logging severity (must be UPPERCASE) at which to log C(msg).
+    required: false
+    default: DEBUG
+    choices: ['CRITICAL', 'DEBUG', 'ERROR', 'INFO', 'WARNING']
+    type: str
+author:
+- Allen Robel (@quantumonion)
+"""
+
+EXAMPLES = r"""
+# This module can be used to correlate Ansible task execution with the
+# log records emitted by other cisco.nd modules when the ND_LOGGING_CONFIG
+# environment variable is set to a valid Python logging configuration file.
+
+- name: Log start of work
+  cisco.nd.nd_log:
+    msg: Starting Nexus Dashboard version query
+    severity: INFO
+
+- name: Query Nexus Dashboard version
+  cisco.nd.nd_version:
+    host: nd_host
+    username: admin
+    password: SomeSecretPassword
+    state: query
+  delegate_to: localhost
+  register: query_result
+
+- name: Log completion
+  cisco.nd.nd_log:
+    msg: Nexus Dashboard version query complete
+    severity: INFO
+"""
+
+RETURN = r"""
+"""
+
+import logging
+
+from ansible.module_utils.basic import AnsibleModule
+from ansible_collections.cisco.nd.plugins.module_utils.common.log import setup_logging
+
+
+class NDLog:
+    """
+    # Summary
+
+    Log a message via the standard Python `logging` module under the `nd` parent logger.
+
+    The severity selects which `logging.Logger` method is called (`debug`, `info`, `warning`, `error`, or `critical`).
+
+    ## Raises
+
+    ### ValueError
+
+    - If `msg` is not provided in `params`.
+    """
+
+    def __init__(self, params: dict) -> None:
+        """
+        # Summary
+
+        Initialize the `NDLog` instance from an Ansible module's `params` dict.
+
+        ## Raises
+
+        ### ValueError
+
+        - If `params["msg"]` is `None`.
+        """
+        self.class_name = self.__class__.__name__
+        self.log = logging.getLogger(f"nd.{self.class_name}")
+
+        self.params = params
+        self.message = self.params.get("msg")
+        self.severity = self.params.get("severity")
+
+        if self.message is None:
+            raise ValueError("Exiting. Missing mandatory parameter: msg")
+
+        if self.severity is None:
+            self.severity = "DEBUG"
+
+        self.result: dict = {"changed": False, "failed": False}
+
+    def msg(self) -> None:
+        """
+        # Summary
+
+        Emit `self.message` at severity `self.severity` on the `nd.NDLog` logger.
+
+        ## Raises
+
+        None
+        """
+        if self.severity == "CRITICAL":
+            self.log.critical(self.message)
+        elif self.severity == "DEBUG":
+            self.log.debug(self.message)
+        elif self.severity == "ERROR":
+            self.log.error(self.message)
+        elif self.severity == "INFO":
+            self.log.info(self.message)
+        elif self.severity == "WARNING":
+            self.log.warning(self.message)
+
+
+def main() -> None:
+    """
+    # Summary
+
+    Entry point for module execution.
+
+    ## Raises
+
+    None
+    """
+    argument_spec = {
+        "msg": {"required": True, "type": "str"},
+        "severity": {
+            "required": False,
+            "default": "DEBUG",
+            "choices": ["CRITICAL", "DEBUG", "ERROR", "INFO", "WARNING"],
+            "type": "str",
+        },
+    }
+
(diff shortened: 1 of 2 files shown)


plugins/modules/nd_log.py
```diff
@@ -0,0 +1,170 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2026, Allen Robel (@quantumonion) <arobel@cisco.com>
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import annotations
+
```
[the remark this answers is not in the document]
allenrobel: 1. All from __future__  imports, except annotations, are no longer relevant for Python 3.11.
2. __metaclass__ = type is no longer relevant for Python 3.11.

tests/unit/modules/test_nd_log.py
```diff
@@ -0,0 +1,155 @@
...
+# pylint: disable=unused-import
+# pylint: disable=redefined-outer-name
+# pylint: disable=protected-access
+# pylint: disable=unused-argument
+# pylint: disable=unused-variable
+# pylint: disable=invalid-name
+# pylint: disable=line-too-long
+# pylint: disable=too-many-lines
+
+from __future__ import annotations
```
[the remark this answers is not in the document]
allenrobel: Not relevant for Python 3.11

plugins/modules/nd_log.py
```diff
@@ -0,0 +1,170 @@
...
+ANSIBLE_METADATA = {"metadata_version": "1.1", "status": ["preview"], "supported_by": "community"}
+
+DOCUMENTATION = r"""
+---
+module: nd_log
+version_added: "1.5.0"
```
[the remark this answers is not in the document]
allenrobel: Addressed with commit: 0bc23e8050576fcc9f8a4724fa75c0b0b365565b

plugins/modules/nd_log.py
```diff
@@ -0,0 +1,170 @@
...
+- Useful for correlating playbook task execution with the C(nd.*) log records emitted internally by other modules in this collection.
+options:
+  msg:
+    description:
+    - The message to log.
+    required: true
+    type: str
+  severity:
+    description:
+    - Case-sensitive logging severity (must be UPPERCASE) at which to log C(msg).
```
akinross, line 29: shall we add some additional information to make severity more clear in combination with ND_LOGGING_CONFIG

```suggestion
    - Case-sensitive logging severity (must be UPPERCASE) at which to log C(msg).
    - The record is only written if O(severity) is greater than or equal to both the level of the C(nd) logger and the level of every
      handler attached to it, as configured in the JSON file referenced by C(ND_LOGGING_CONFIG). If either level is more restrictive,
      the call is silently discarded and the module still reports C(changed=false, failed=false).
    - For example, setting O(severity=INFO) while the JSON config has C(loggers.nd.level) set to C(WARNING) will not write anything.
```

plugins/modules/nd_log.py
```diff
@@ -0,0 +1,170 @@
...
+    required: false
+    default: DEBUG
+    choices: ['CRITICAL', 'DEBUG', 'ERROR', 'INFO', 'WARNING']
+    type: str
+author:
```
akinross, line 43: i think we can add more clarity by adding a note to the documentation and/or update the description suggestion above 

```suggestion
notes:
- The Python C(logging) module applies level checks at both the logger and the handler. A record is emitted only when its severity
  is greater than or equal to the most restrictive level along the chain (C(nd) logger, any ancestor loggers consulted for the
  effective level, and each handler). To make C(nd_log) messages visible, ensure both C(loggers.nd.level) (or the inherited level)
  and the relevant handler's C(level) in C(ND_LOGGING_CONFIG) are at or below O(severity).
author:
```

plugins/modules/nd_log.py
```diff
@@ -0,0 +1,170 @@
...
+    The severity selects which `logging.Logger` method is called (`debug`, `info`, `warning`, `error`, or `critical`).
+
+    ## Raises
+
+    ### ValueError
+
+    - If `msg` is not provided in `params`.
+    """
+
+    def __init__(self, params: dict) -> None:
```
akinross: I am kind of confused with this constructor function because if the intend it is to Initialize the `NDLog` instance from an Ansible module's `params` dict why are the is None check needed for message and severity? The current approach couples the NDLog class to ansible specific construct, which already specifies argument spec with required and defaults so they would never be None, so this could be considered as stale/dead code. 

I would argue that if we need a separate class for this we should decouple it from the Ansible module's `params` dict and pass in the attributes individually or as a NDLog configuration object.

Personally I think we should define the behaviour inside of main() without the class, which would result in way less code:

```
def main() -> None:
    argument_spec = {
        "msg": {"required": True, "type": "str"},
        "severity": {
            "required": False,
            "default": "DEBUG",
            "choices": ["CRITICAL", "DEBUG", "ERROR", "INFO", "WARNING"],
            "type": "str",
        },
    }

    module = AnsibleModule(argument_spec=argument_spec, supports_check_mode=True)
    setup_logging(module)

    getattr(logging.getLogger("nd.NDLog"), module.params["severity"].lower())(module.params["msg"])

    module.exit_json(changed=False, failed=False)
```

Do we need to support check_mode for this module? To me this does not make a lot of sense.

plugins/modules/nd_log.py
```diff
@@ -0,0 +1,170 @@
...
+
+EXAMPLES = r"""
+# This module can be used to correlate Ansible task execution with the
+# log records emitted by other cisco.nd modules when the ND_LOGGING_CONFIG
+# environment variable is set to a valid Python logging configuration file.
+
+- name: Log start of work
+  cisco.nd.nd_log:
+    msg: Starting Nexus Dashboard version query
+    severity: INFO
```
samiib, line 55: Should we use `delegate_to: localhost` with `nd_log`?

plugins/modules/nd_log.py
```diff
@@ -0,0 +1,170 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2026, Allen Robel (@quantumonion) <arobel@cisco.com>
```
samiib:
```suggestion
# Copyright: (c) 2026, Allen Robel (@allenrobel) <arobel@cisco.com>
```
Mystery user? 🤔

tests/unit/modules/test_nd_log.py
```diff
@@ -0,0 +1,155 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2026, Allen Robel (@quantumonion) <arobel@cisco.com>
```
samiib:
```suggestion
# Copyright: (c) 2026, Allen Robel (@allenrobel) <arobel@cisco.com>
```

plugins/modules/nd_log.py
allenrobel: Leveraged your suggested change.  Thanks!

plugins/modules/nd_log.py
allenrobel: Leveraged your suggested change. Thanks!

plugins/modules/nd_log.py
allenrobel: Doh!  I think this was an earlier username I chose for my Github account and later changed it.  No idea how that ended up here. 🤷

tests/unit/modules/test_nd_log.py
allenrobel: Fixed.

plugins/modules/nd_log.py
allenrobel: Good call -- agreed on all of this.  I tend to always think in terms of classes so miss opportunities to simplify things when a class is overkill 🤷   Incorporated all of your other suggestions as well and updated the unit tests to align with the changes.  Commit: d8603ca.

plugins/modules/nd_log.py
allenrobel: Good question Samita -- the answer is no, it's not needed. cisco.nd modules run through the httpapi connection plugin, which executes the module on the Ansible control node (network modules aren't shipped to the device).

So nd_log already runs on the same host as every other cisco.nd module in the play and writes to the same ND_LOGGING_CONFIG-configured log file without any delegate_to.

The delegate_to: localhost on the `nd_version` task, paired with param-style credentials (host/username/password) -- and it's actually misleading in an httpapi-style play. Fixed by dropping the inline credentials and `delegate_to` so all three example tasks are plain module calls that take their connection from inventory/play vars.

Commit: 52a56fe

samiib (approved):
LGTM

gmicol (approved):

akinross (approved):

PR #284, opened 2026-05-18
nd_maintenance_mode

allenrobel:
## Proposed Changes

Add `cisco.nd.nd_maintenance_mode`, a new Ansible module that sets the system mode (`normal` or `maintenance`) of switches in a fabric via `POST /api/v1/manage/fabrics/{fabric_name}/switchActions/changeSystemMode`.

The endpoint is action-shaped — one POST mutates a batch of switches with a single mode. The module adapts that to `NDStateMachine`'s CRUD-shaped driver via an `identifier_strategy="singleton"` model and a `config`-dict-to-list wrap in `main()`. Idempotency reads `additionalData.intendedSystemMode` per switch (from one bulk `GET /fabrics/{f}/switches`) and only POSTs switches whose intent differs from the request. Any switch currently in `migration` mode triggers a hard-fail with a remediation message asking the operator to run 'Recalculate and Deploy' in the ND UI before retrying. The 207 response body is inspected per-item; entries whose `status` is `failed`, `failure`, or `error` (case-insensitive) surface as a `RuntimeError` naming the offending switches by IP. Other tokens (`success`, missing, empty, `pending`, `warning`) are tolerated.

State support is `merged` only for now. The collection-wide `query` state is retired; `gathered` will be added once the underlying `NDStateMachine`/`NDOutput` framework support lands (the orchestrator's `query_all` already produces the per-switch snapshot it will need).

Two `TODO(4.2.1)` workaround markers flag observed ND wire/schema divergences:

- `GET /fabrics/{f}/switches/{sid}` nests `systemMode` under `additionalData`, not at the top level the OpenAPI schema documents.
- The aggregator `additionalData.systemMode` returns `"inconsistent"` (a value not in the OpenAPI enum) between an intent change and a deploy, so idempotency must compare against `intendedSystemMode`, not `systemMode`.

New / modified files:

- `plugins/modules/nd_maintenance_mode.py` — module entry point (DOCUMENTATION / EXAMPLES / argspec / `main()`). DOCUMENTATION includes a notes paragraph and a play-level `module_defaults` example covering the `ansible_command_timeout` / `persistent_command_timeout` interaction with blocking deploys.
- `plugins/module_utils/orchestrators/maintenance_mode.py` — `MaintenanceModeOrchestrator` inheriting `NDBaseInterfaceOrchestrator` for `FabricContext` plumbing. `query_all()` issues one bulk `EpManageSwitchesListGet` and caches both the IP→intendedSystemMode map and the IP→switchId map on the instance; `update()` reuses both, so each module run does one fabric-summary GET, one switches GET, and (when changes are needed) one POST.
- `plugins/module_utils/models/maintenance_mode/{__init__.py, maintenance_mode.py}` — singleton model + per-switch `get_diff`.
- `plugins/module_utils/endpoints/v1/manage/manage_fabrics_switch_actions.py` — new endpoint module for `changeSystemMode`. Path is URL-encoded; query string omits `deploy` and `blocking` when the user did not opt into them.
- `plugins/module_utils/endpoints/v1/manage/manage_switches.py` — added `EpManageSwitchesGet` (generic per-switch GET, retained as collection infrastructure though `nd_maintenance_mode` itself reads via the bulk `EpManageSwitchesListGet`). URL encoding on both new paths hardened to match the existing `EpManageSwitchActionsDeploy` convention.
- Unit tests: endpoint (9), model (15), orchestrator (29 collected). Total suite: 554/554 pass.
- Integration target `nd_maintenance_mode` with YAML-anchored single-switch, multi-switch, partial-mismatch, negative, and `check_mode` scenarios.

## Test Notes

- **Unit tests**: `source env && python -m pytest tests/unit/module_utils/` — 554/554 pass, zero regressions. Coverage: orchestrator 93%, model 97%, endpoint 97%.
- **Lint**: `black`, `isort`, `pylint`, `mypy` all clean against `pyproject.toml`. `pylint` 9.97/10 on tests, 9.78/10 on production. `yamllint` clean on integration target.
- **Integration**: Verified live on ND 4.2.1.10 @ 192.168.7.7 / SITE1 against switches S1_BG1 (192.168.12.131) and S1_SP1 (192.168.12.141). Full target green; lab state verified back to `intendedSystemMode=normal` for all test switches after the run. Re-verified green after the follow-up commits below.
- **Smoke test**: Eight standalone scenarios (baseline, set maintenance, idempotent, restore, idempotent, negative `switch_ip`, negative fabric, check_mode) all green prior to writing the integration target.
- The `radius` auth domain on this lab is effectively read-only on SITE1/SITE2; live tests use `-e ansible_httpapi_login_domain=local`. Documented in the integration target's `main.yaml`.

## Refinements since PR opened

Six follow-up commits land behavior/perf refinements on top of the initial implementation:

1. `e65143d` — Cache the `query_all()` snapshot on the orchestrator and skip the redundant re-query in `update()`. Avoids a second per-run snapshot pass when `NDStateMachine` already populated one during init.
2. `d9d09fa` — Replace the per-switch GET fan-out in `query_all()` with one bulk `EpManageSwitchesListGet` call (verified on ND 4.2.1 lab: the list response carries `additionalData.intendedSystemMode` and `discoveredSystemMode` per row). Per-run controller round-trips drop from `2 + N` to `2 + 1`; on a 32-switch fabric that is 64 GETs eliminated per task.
3. `3dde916` — URL-encode `fabric_name` (and `switch_id`) in `EpManageSwitchesGet.path` and `EpManageFabricsSwitchActionsChangeSystemModePost.path`. Matches the existing `quote(safe="")` convention used by sibling endpoints; protects against future fabric names containing reserved URL characters.
4. `8374b4b` — Only push truthy `deploy` / `blocking` onto the endpoint's query params. Previously the bool defaults (False) were assigned unconditionally and serialized as `?deploy=false&blocking=false` on every request. Functionally harmless on ND today but inconsistent with the endpoint's own `to_query_string` design.
5. `f77bd42` — 207 parser uses an explicit failure denylist (`failed` / `failure` / `error`, case-insensitive) rather than treating anything not `success` as a failure. ND's switchAction APIs occasionally return rows with missing status or progress tokens like `pending`/`warning`; the old code would surface those as spurious task failures.
6. `0efba32` — Document the `blocking=true` deploy interaction with `persistent_command_timeout`. New notes paragraph and a play-level `module_defaults` example showing the recommended `timeout: 1800` + inventory `ansible_command_timeout` pattern.

4.2.1

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

---

### DIFF
--- a/plugins/module_utils/endpoints/v1/manage/manage_switches.py
+++ b/plugins/module_utils/endpoints/v1/manage/manage_switches.py
@@ -144,6 +144,11 @@ class EpManageSwitchesListGet(_EpManageSwitchesBase):
         return HttpVerbEnum.GET
 
 
+# TODO: Remove EpManageSwitchesGet once the in-flight stacked branches have merged into develop.
+# It is currently unused here -- nd_maintenance_mode resolves per-switch intendedSystemMode from the
+# bulk EpManageSwitchesListGet snapshot, not from a per-switch GET. It is retained for now because
+# sibling branches in the current stack may still reference the per-switch endpoint; revisit after
+# the stack lands and delete it if no caller has appeared.
 class EpManageSwitchesGet(FabricNameMixin, NDEndpointBaseModel):
     """
     # Summary
--- a/plugins/module_utils/endpoints/v1/manage/manage_switches.py
+++ b/plugins/module_utils/endpoints/v1/manage/manage_switches.py
@@ -11,13 +11,15 @@ in the ND Manage API.
 
 - `EpManageSwitchesListGet` - List switches in a fabric with optional Lucene filtering
   (GET /api/v1/manage/fabrics/{fabric_name}/switches)
+- `EpManageSwitchesGet` - Get the details of a specific switch in a fabric
+  (GET /api/v1/manage/fabrics/{fabric_name}/switches/{switch_id})
 - `EpManageSwitchActionsDeploy` - Deploy all pending switch configuration
   (POST /api/v1/manage/fabrics/{fabric_name}/switchActions/deploy)
 """
 
 from __future__ import annotations
 
-from typing import Literal
+from typing import Literal, Optional
 from urllib.parse import quote
 
 from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import Field
@@ -127,6 +129,93 @@ class EpManageSwitchesListGet(_EpManageSwitchesBase):
     """
 
     class_name: Literal["EpManageSwitchesListGet"] = Field(default="EpManageSwitchesListGet", frozen=True, description="Class name for backward compatibility")
+
+    @property
+    def verb(self) -> HttpVerbEnum:
+        """
+        # Summary
+
+        Return `HttpVerbEnum.GET`.
+
+        ## Raises
+
+        None
+        """
+        return HttpVerbEnum.GET
+
+
+class EpManageSwitchesGet(FabricNameMixin, NDEndpointBaseModel):
+    """
+    # Summary
+
+    Get the details of a specific switch in a fabric.
+
+    ## Description
+
+    Endpoint to retrieve a specific switch's details from a fabric. The fabric name and switch ID (NX-OS serial number or
+    ACI node ID) are both required. The response contains switch metadata; the `systemMode` / `intendedSystemMode` /
+    `discoveredSystemMode` fields used by `nd_maintenance_mode` are nested under `additionalData` on the live wire.
+
+    ## Path
+
+    - `/api/v1/manage/fabrics/{fabric_name}/switches/{switch_id}`
+
+    ## Verb
+
+    - GET
+
+    ## Raises
+
+    ### ValueError
+
+    - If `fabric_name` or `switch_id` is not set when accessing `path`.
+
+    ## Usage
+
+    ```python
+    ep = EpManageSwitchesGet()
+    ep.fabric_name = "SITE1"
+    ep.switch_id = "9UOJ3E8A6O9"
+    rest_send.path = ep.path
+    rest_send.verb = ep.verb
+    # Path: /api/v1/manage/fabrics/SITE1/switches/9UOJ3E8A6O9
+    ```
+    """
+
+    class_name: Literal["EpManageSwitchesGet"] = Field(default="EpManageSwitchesGet", frozen=True, description="Class name for backward compatibility")
+
+    switch_id: Optional[str] = Field(default=None, description="Switch identifier (NX-OS serial number or ACI node ID)")
+
+    def set_identifiers(self, identifier: IdentifierKey = None):
+        """
+        # Summary
+
+        Set `switch_id` from `identifier`. `fabric_name` must be set separately.
+
+        ## Raises
+
+        None
+        """
+        self.switch_id = identifier
+
+    @property
+    def path(self) -> str:
+        """
+        # Summary
+
+        Build the per-switch GET endpoint path.
+
+        ## Raises
+
+        ### ValueError
+
+        - If `fabric_name` or `switch_id` is not set before accessing `path`.
+        """
+        if self.fabric_name is None:
+            raise ValueError(f"{type(self).__name__}.path: fabric_name must be set before accessing path.")
+        if self.switch_id is None:
+            raise ValueError(f"{type(self).__name__}.path: switch_id must be set before accessing path.")
+        return BasePath.path("fabrics", self.fabric_name, "switches", self.switch_id)
 
     @property
     def verb(self) -> HttpVerbEnum:
--- a/plugins/module_utils/endpoints/v1/manage/manage_switches.py
+++ b/plugins/module_utils/endpoints/v1/manage/manage_switches.py
@@ -215,7 +215,7 @@ class EpManageSwitchesGet(FabricNameMixin, NDEndpointBaseModel):
             raise ValueError(f"{type(self).__name__}.path: fabric_name must be set before accessing path.")
         if self.switch_id is None:
             raise ValueError(f"{type(self).__name__}.path: switch_id must be set before accessing path.")
-        return BasePath.path("fabrics", self.fabric_name, "switches", self.switch_id)
+        return BasePath.path("fabrics", quote(self.fabric_name, safe=""), "switches", quote(self.switch_id, safe=""))
 
     @property
     def verb(self) -> HttpVerbEnum:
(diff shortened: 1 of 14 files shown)


plugins/module_utils/orchestrators/maintenance_mode.py
```diff
@@ -0,0 +1,378 @@
...
+
+            wanted_ips = [entry["switch_ip"] for entry in user_switches if entry.get("switch_ip")]
+            wanted_ip_set = set(wanted_ips)
+
+            list_endpoint = EpManageSwitchesListGet()
+            list_endpoint.fabric_name = self.fabric_name
+            response = self._request(path=list_endpoint.path, verb=list_endpoint.verb)
+            rows = response.get("switches") if isinstance(response, dict) else None
+
+            switch_ids, switch_modes, migration_ips = self._parse_switch_rows(rows or [], wanted_ip_set)
```
akinross: why assign None and then do the or statement in _parse_switch_rows?

```suggestion
            rows = response.get("switches") if isinstance(response, dict) else []

            switch_ids, switch_modes, migration_ips = self._parse_switch_rows(rows, wanted_ip_set)
```

furthermore do we need to check that rows is of the correct type before passing it to _parse_switch_rows?

akinross (changes requested):

plugins/module_utils/models/maintenance_mode/maintenance_mode.py
```diff
@@ -0,0 +1,154 @@
...
+                options=dict(
+                    mode=dict(type="str", required=True, choices=["maintenance", "normal"]),
+                    deploy=dict(type="bool", default=False),
+                    blocking=dict(type="bool", default=False),
+                    ticket_id=dict(type="str"),
+                    switches=dict(
+                        type="list",
+                        elements="dict",
+                        required=True,
+                        options=dict(
```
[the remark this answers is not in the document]
allenrobel: Fixed in 8ffe32a — added a `model_validator` on `MaintenanceModeModel` that rejects `switches: []` when `mode` is set (`config.switches must contain at least one switch when 'mode' is set.`). The check is gated on `mode is not None` so `query_all` snapshots, which leave `mode` unset, still construct. New `test_maintenance_mode_model_00100` covers the rejection path and `_00110` covers the snapshot exemption.

plugins/module_utils/orchestrators/maintenance_mode.py
allenrobel: Good catch on both. Fixed in a5985a8 — dropped the `or []` pattern in favor of an explicit list-type guard at the boundary:

```python
rows = response.get("switches") if isinstance(response, dict) else []
if not isinstance(rows, list):
    rows = []
```

`_parse_switch_rows` is typed `list[Any]` and already skips non-dict items, but the caller previously left a non-list value (string, dict, scalar) intact. A string would iterate character-by-character (benign but wasteful); a non-iterable scalar would `TypeError` inside the helper and surface as an opaque `query_all failed:` message. With the guard, malformed shapes collapse to `[]` and the user gets the standard `No switch found with fabricManagementIp ...` error. New `test_maintenance_mode_00140` locks the behavior in with an integer-shaped `switches` fixture.

akinross (approved):

plugins/module_utils/endpoints/v1/manage/manage_switches.py
```diff
@@ -142,6 +144,93 @@ def verb(self) -> HttpVerbEnum:
         return HttpVerbEnum.GET
 
 
+class EpManageSwitchesGet(FabricNameMixin, NDEndpointBaseModel):
```
samiib, line 152: Is `EpManageSwitchesGet` used anywhere in the code currently? Should we remove it or is there a plan to use it later.

I believe the module documentation mentioned `switches/switch_id` gets used but instead the bulk `ListGet` is used instead.

plugins/modules/nd_maintenance_mode.py
```diff
@@ -0,0 +1,236 @@
...
+DOCUMENTATION = r"""
+---
+module: nd_maintenance_mode
+version_added: "2.0.0"
+short_description: Manage the system mode (normal or maintenance) of switches in a fabric on Cisco Nexus Dashboard
+description:
+- Manage the system mode of one or more switches in a fabric on Cisco Nexus Dashboard.
+- Wraps the C(POST /api/v1/manage/fabrics/{fabric_name}/switchActions/changeSystemMode) endpoint, which sets a single
+  C(mode) (C(normal) or C(maintenance)) for a batch of switches in one request.
+- For idempotency the module reads each switch's current C(intendedSystemMode) from C(GET /api/v1/manage/fabrics/{fabric_name}/switches/{switch_id})
```
samiib: This is the comment that may need updating if the individual switch_id call isn't used.

plugins/module_utils/models/maintenance_mode/maintenance_mode.py
```diff
@@ -0,0 +1,181 @@
...
+from ansible_collections.cisco.nd.plugins.module_utils.models.base import NDBaseModel
+from ansible_collections.cisco.nd.plugins.module_utils.models.nested import NDNestedModel
+
+
+class MaintenanceModeSwitchModel(NDNestedModel):
+    """
+    # Summary
+
+    Single switch entry in `config.switches`. The user supplies `switch_ip`; the orchestrator resolves
+    it to the switch serial number (`switchId`) via `FabricContext.get_switch_id()` before building the
```
samiib: This comment may also need adjusting if the individual switch_id call isn't used.

plugins/module_utils/endpoints/v1/manage/manage_switches.py
allenrobel: Confirmed: `EpManageSwitchesGet` (per-switch GET) has zero callers — `nd_maintenance_mode` reads every switch's `intendedSystemMode` from the bulk `EpManageSwitchesListGet` snapshot instead. Rather than delete it outright, I've left it in place with a `# TODO` to remove it once the in-flight stacked branches merge into develop (siblings in the stack may still reference the per-switch endpoint). Fixed in 8a5554d0.

plugins/modules/nd_maintenance_mode.py
allenrobel: Good catch — the docstring was inaccurate. The module issues a single bulk `GET /api/v1/manage/fabrics/{fabric_name}/switches` and parses every switch's `intendedSystemMode` from that snapshot; there is no per-switch `GET /switches/{switch_id}`. Corrected the comment in 06fa94ad.

plugins/module_utils/models/maintenance_mode/maintenance_mode.py
allenrobel: Adjusted. This comment was also inaccurate: the orchestrator resolves `switch_ip` to `switchId` from the bulk `EpManageSwitchesListGet` snapshot cached by `query_all()`, and deliberately avoids `FabricContext.get_switch_id()` to prevent a second list GET. Updated the docstring in ca39a4ae.

samiib (approved):
LGTM

gmicol (approved):
LGTM

PR #285, opened 2026-05-20
nd_fabric_update_group - ND 4.2 software update groups

allenrobel:
## Proposed Changes

- New module `nd_fabric_update_group` — manages Fabric Software Management **update groups** (the ND 4.2 successor to ND 3.x image policies) within a fabric.
- Gen 3 architecture: Pydantic `FabricUpdateGroupModel`, `FabricUpdateGroupOrchestrator`, smart endpoints, and `NDStateMachine` wiring.
- States supported: `merged`, `replaced`, `overridden`, `deleted`.
- **The write path uses the switch-centric `softwareUpdatePlan/actions` API**, which is ghost-safe by construction:
  - `attachGroup` creates a group and assigns its switches (requires at least one switch).
  - `detachGroup` removes switches; ND auto-deletes a group once its last switch is removed.
  - Group settings are applied via `PUT /updateGroups/{name}`, built by overlaying the user's fields onto a GET of the current group, so the PUT echoes the live membership and never moves switches.
- New `force_created` per-group option — ND raises a pre-flight warning when a group's switch set would cover all switches of a role; `force_created: true` acknowledges the warning so the change is applied. A warning otherwise fails the task, since ND applies nothing when it warns.
- Reads (`query_one` / `query_all`, used by `overridden` reconciliation) stay on the group-centric `GET` endpoints.
- New endpoint module `endpoints/v1/manage/software_update_plan_actions.py` (`attachGroup` / `detachGroup`); the now-unused group-centric `POST` endpoint is removed.
- `update_group_switches` / `installation_order_devices` accept either switch IP addresses or serial numbers (switchIds); IPs are resolved to switchIds via `FabricContext`.
- ND 4.2.1 wire workaround: `installationOrderDevices` is silently dropped by ND on write — excluded from the idempotency diff (`TODO(4.2.1)` marker). ND's system-managed default group (literally named `"None"`) is filtered out of query results so `overridden` does not try to delete it.

## Background — the "ghost group" defect this design sidesteps

The group-centric `updateGroups` CRUD endpoints have no minimum-switch guard and cannot represent a zero-switch group: reusing a switch between groups silently drained a group and left it an unreadable "ghost" (omitted from the list, HTTP 400 on the single-group GET, name still reserved). The switch-centric action API used here is ghost-safe by construction — `attachGroup` requires at least one switch and `detachGroup` cleans up emptied groups — so the module no longer depends on an ND-side fix. This module was previously postponed on that defect; the redesign removes the block.

## Test Notes

- Unit tests: 80 for this module (model + orchestrator + endpoints). Full suite 552 passing, no regressions.
- `black` / `isort` / `mypy` clean; `ansible-test sanity --docker` passes.
- Live verification on ND 4.2.1 (fabric SITE1) of behaviors the OpenAPI spec cannot confirm: `detachGroup` last-switch auto-delete, `PUT` with current membership is a membership no-op, and `attachGroup` `forceCreated` semantics (a `warning` status means ND applied nothing).
- Integration target `tests/integration/targets/nd_fabric_update_group/` — `merged`, `replaced`, `overridden`, `deleted` all pass against live ND 4.2.1.

4.2.1

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

### DIFF
--- a/plugins/module_utils/endpoints/v1/manage/fabric_update_group.py
+++ b/plugins/module_utils/endpoints/v1/manage/fabric_update_group.py
@@ -22,6 +22,7 @@ under the ND Manage Fabric Software Management API.
 from __future__ import annotations
 
 from typing import ClassVar, Literal
+from urllib.parse import quote
 
 from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import Field
 from ansible_collections.cisco.nd.plugins.module_utils.endpoints.base import NDEndpointBaseModel
@@ -66,9 +67,9 @@ class _EpFabricUpdateGroupBase(UpdateGroupNameMixin, FabricNameMixin, NDEndpoint
             raise ValueError(f"{type(self).__name__}.path: fabric_name must be set before accessing path.")
         if self._require_update_group_name and self.update_group_name is None:
             raise ValueError(f"{type(self).__name__}.path: update_group_name must be set before accessing path.")
-        segments = ["fabrics", self.fabric_name, "updateGroups"]
+        segments = ["fabrics", quote(self.fabric_name, safe=""), "updateGroups"]
         if self.update_group_name is not None:
-            segments.append(self.update_group_name)
+            segments.append(quote(self.update_group_name, safe=""))
         return BasePath.path(*segments)
 
 
--- a/plugins/module_utils/endpoints/v1/manage/fabric_update_group.py
+++ b/plugins/module_utils/endpoints/v1/manage/fabric_update_group.py
@@ -1,0 +1,216 @@
+# Copyright: (c) 2026, Allen Robel (@allenrobel) <arobel@cisco.com>
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+"""
+ND Manage Fabric Update Group endpoint models.
+
+This module contains endpoint definitions for fabric update group operations
+under the ND Manage Fabric Software Management API.
+
+## Endpoints
+
+- `EpFabricUpdateGroupListGet` - List update groups in a fabric
+  (GET /api/v1/manage/fabrics/{fabric_name}/updateGroups)
+- `EpFabricUpdateGroupPost` - Create one or more update groups in a fabric
+  (POST /api/v1/manage/fabrics/{fabric_name}/updateGroups)
+- `EpFabricUpdateGroupGet` - Get a specific update group by name
+  (GET /api/v1/manage/fabrics/{fabric_name}/updateGroups/{update_group_name})
+- `EpFabricUpdateGroupPut` - Update an existing update group
+  (PUT /api/v1/manage/fabrics/{fabric_name}/updateGroups/{update_group_name})
+- `EpFabricUpdateGroupDelete` - Delete an update group
+  (DELETE /api/v1/manage/fabrics/{fabric_name}/updateGroups/{update_group_name})
+"""
+
+from __future__ import annotations
+
+from typing import ClassVar, Literal
+
+from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import Field
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.base import NDEndpointBaseModel
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.mixins import FabricNameMixin, UpdateGroupNameMixin
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.v1.manage.base_path import BasePath
+from ansible_collections.cisco.nd.plugins.module_utils.enums import HttpVerbEnum
+from ansible_collections.cisco.nd.plugins.module_utils.types import IdentifierKey
+
+
+class _EpFabricUpdateGroupBase(UpdateGroupNameMixin, FabricNameMixin, NDEndpointBaseModel):
+    """
+    Base class for ND Manage Fabric Update Group endpoints.
+
+    Provides common functionality for all HTTP methods on the
+    /api/v1/manage/fabrics/{fabric_name}/updateGroups endpoint.
+
+    Subclasses may override:
+    - ``_require_update_group_name``: set to ``False`` for collection-level endpoints
+      (list, create) that do not include an update group name in the path.
+    """
+
+    _require_update_group_name: ClassVar[bool] = True
+
+    def set_identifiers(self, identifier: IdentifierKey = None):
+        self.update_group_name = identifier
+
+    @property
+    def path(self) -> str:
+        """
+        # Summary
+
+        Build the endpoint path with required fabric_name and optional update_group_name.
+
+        ## Raises
+
+        ### ValueError
+
+        - If `fabric_name` is not set when accessing `path`
+        - If `update_group_name` is required but not set
+        """
+        if self.fabric_name is None:
+            raise ValueError(f"{type(self).__name__}.path: fabric_name must be set before accessing path.")
+        if self._require_update_group_name and self.update_group_name is None:
+            raise ValueError(f"{type(self).__name__}.path: update_group_name must be set before accessing path.")
+        segments = ["fabrics", self.fabric_name, "updateGroups"]
+        if self.update_group_name is not None:
+            segments.append(self.update_group_name)
+        return BasePath.path(*segments)
+
+
+class EpFabricUpdateGroupListGet(_EpFabricUpdateGroupBase):
+    """
+    # Summary
+
+    ND Manage Fabric Update Group List GET endpoint.
+
+    ## Path
+
+    - `/api/v1/manage/fabrics/{fabric_name}/updateGroups`
+
+    ## Verb
+
+    - GET
+
+    ## Usage
+
+    ```python
+    ep = EpFabricUpdateGroupListGet()
+    ep.fabric_name = "SITE1"
+    rest_send.path = ep.path
+    rest_send.verb = ep.verb
+    ```
+    """
+
+    _require_update_group_name: ClassVar[bool] = False
+
+    class_name: Literal["EpFabricUpdateGroupListGet"] = Field(
+        default="EpFabricUpdateGroupListGet", frozen=True, description="Class name for backward compatibility"
+    )
+
+    @property
+    def verb(self) -> HttpVerbEnum:
+        """Return the HTTP verb for this endpoint."""
+        return HttpVerbEnum.GET
+
+
+class EpFabricUpdateGroupPost(_EpFabricUpdateGroupBase):
+    """
+    # Summary
+
+    ND Manage Fabric Update Group POST endpoint.
+
+    ## Path
+
+    - `/api/v1/manage/fabrics/{fabric_name}/updateGroups`
+
+    ## Verb
+
+    - POST
+
+    ## Usage
+
+    ```python
+    ep = EpFabricUpdateGroupPost()
+    ep.fabric_name = "SITE1"
+    rest_send.path = ep.path
+    rest_send.verb = ep.verb
+    rest_send.payload = {...}
+    ```
+    """
+
+    _require_update_group_name: ClassVar[bool] = False
+
+    class_name: Literal["EpFabricUpdateGroupPost"] = Field(default="EpFabricUpdateGroupPost", frozen=True, description="Class name for backward compatibility")
+
+    @property
+    def verb(self) -> HttpVerbEnum:
+        """Return the HTTP verb for this endpoint."""
+        return HttpVerbEnum.POST
+
+
+class EpFabricUpdateGroupGet(_EpFabricUpdateGroupBase):
+    """
+    # Summary
+
+    ND Manage Fabric Update Group GET endpoint.
+
+    ## Path
+
+    - `/api/v1/manage/fabrics/{fabric_name}/updateGroups/{update_group_name}`
+
+    ## Verb
+
+    - GET
+    """
+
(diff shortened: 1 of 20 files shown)


plugins/module_utils/orchestrators/fabric_update_group.py
```diff
@@ -0,0 +1,572 @@
...
+            current_ids = list(current_raw.get("updateGroupSwitches") or [])
+            desired_ids = [self._resolve_switch_id(s) for s in (model_instance.update_group_switches or [])]
+            if not desired_ids:
+                raise RuntimeError("update_group_switches must be non-empty; an empty update group is not permitted (use state: deleted)")
+            to_add = [s for s in desired_ids if s not in current_ids]
+            to_remove = [s for s in current_ids if s not in desired_ids]
+            if to_add:
+                self._attach([self._attach_item(model_instance, switch_ids=to_add)])
+            if to_remove:
+                self._detach([{"updateGroupName": update_group_name, "switchIds": to_remove}])
```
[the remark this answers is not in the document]
allenrobel: Fixed in ab139ea71e88c3b05e2d5a8ff6959059f2d2abb1. `update()` now distinguishes an omitted `update_group_switches` (None → membership not managed: skip attach/detach reconciliation, keep current members, and run `_apply_settings`) from an explicit empty list (still rejected — use `state: deleted`, since ND never permits an empty group). This applies uniformly across merged/replaced/overridden, so a settings-only update (e.g. `execution`/`reports`) no longer fails with the empty-group error. `_resolve_switches_in_payload` also now passes through values already present as switchIds, so the membership seeded from the current-group GET isn't run through the IP-only resolver. Covered by orchestrator tests `00270` (merged) and `00280` (replaced).

plugins/module_utils/orchestrators/fabric_update_group.py
```diff
@@ -0,0 +1,572 @@
...
+        ghost group, which ND returns HTTP 400 for).
+
+        ## Raises
+
+        None
+        """
+        try:
+            return self._get_group_raw(update_group_name)
+        except Exception:  # pylint: disable=broad-except
+            return None
```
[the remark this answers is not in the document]
allenrobel: Fixed in ab139ea71e88c3b05e2d5a8ff6959059f2d2abb1. `_get_group_raw_or_none` no longer swallows every exception: it now only returns None when `self.rest_send.return_code` is 400 (the zero-switch ghost group ND returns, flagged with a `TODO(4.2.1)` marker) or 404 (group already gone), and re-raises everything else. A transient/authorization/server failure on the preliminary GET therefore propagates instead of being turned into a destructive DELETE. Covered by orchestrator tests `00330` (500 → re-raise, no DELETE issued) and `00340` (404 → fallback DELETE).

plugins/module_utils/endpoints/v1/manage/fabric_update_group.py
```diff
@@ -0,0 +1,179 @@
...
+        - If `update_group_name` is required but not set
+        """
+        if self.fabric_name is None:
+            raise ValueError(f"{type(self).__name__}.path: fabric_name must be set before accessing path.")
+        if self._require_update_group_name and self.update_group_name is None:
+            raise ValueError(f"{type(self).__name__}.path: update_group_name must be set before accessing path.")
+        segments = ["fabrics", self.fabric_name, "updateGroups"]
+        if self.update_group_name is not None:
+            segments.append(self.update_group_name)
+        return BasePath.path(*segments)
```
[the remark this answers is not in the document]
allenrobel: Fixed in ab139ea71e88c3b05e2d5a8ff6959059f2d2abb1. Both dynamic path segments are now percent-encoded — `quote(self.fabric_name, safe="")` and `quote(self.update_group_name, safe="")` — matching the software update plan action endpoints, so a reserved character such as `/` in a fabric or update group name can no longer split the path. New endpoint test module `tests/unit/module_utils/endpoints/test_fabric_update_group.py` asserts the `%2F` encoding on the list/GET/PUT/DELETE paths.

plugins/modules/nd_fabric_update_group.py
```diff
@@ -0,0 +1,358 @@
...
+
+    _validate_report_analysis_exclusion(module)
+
+    auto_assign = module.params.get("auto_assign")
+    if auto_assign is not None:
+        state = module.params["state"]
+        if state != "merged":
+            module.fail_json(
+                msg=f"auto_assign is only valid with state 'merged', got '{state}'. "
+                "The auto-assign action already regroups every switch in the fabric, so state 'overridden' is not supported with it."
```
samiib: Should this mention any other state? The logic is just `!= merged` it could be any other state instead of overridden. ie deleted, replaced. 
Perhaps removing this extra log to avoid confusion. The description on the auto-assign option is clear.

tests/integration/targets/nd_fabric_update_group/vars/main.yaml
```diff
@@ -0,0 +1,79 @@
+---
+# Variables for nd_fabric_update_group integration tests.
+#
+# `update_group_switches` and `installation_order_devices` accept either switch IP addresses
+# or switch serial numbers (switchIds). IPs are resolved to switchIds via the fabric
```
samiib: Are switch serial numbers supported? Looks like it must be IPv4/6?

plugins/modules/nd_fabric_update_group.py
```diff
@@ -0,0 +1,358 @@
...
+      is_disruptive_update:
+        description:
+        - Whether the upgrade is allowed to be disruptive.
+        type: bool
+      update_group_switches:
+        description:
+        - The list of switches that belong to this update group.
+        - Each entry is a switch fabric management IP address (IPv4 or IPv6). Switch serial numbers are not accepted.
+        - Switch IP addresses are resolved to switchIds via the fabric inventory before the request is sent.
+        - An update group must contain at least one switch; ND does not permit a zero-switch group.
```
samiib, line 74: Should this also mention the difference between omitting and an empty list? Is omitting update_group_switches acceptable?

It looks potentially update_group_switches is silently preserved on update when omitted but omitting it on create translates it to an empty list that gets rejected by the API. Perhaps an early guard with a clear error is also useful in the module.
eg. "update_group_switches must be non-empty when creating an update group."

plugins/modules/nd_fabric_update_group.py
allenrobel: Fixed in cd85389c. You're right — the guard is `state != "merged"`, so it also rejects `replaced`, and `deleted`, not just `overridden`. I kept the rationale (which you found clear) but generalized the wording so it no longer singles out one state:

> auto_assign is only valid with state 'merged', got '{state}'. The auto-assign action already regroups every switch in the fabric, so it cannot be combined with any other state.

tests/integration/targets/nd_fabric_update_group/vars/main.yaml
allenrobel: Correct — `update_group_switches` and `installation_order_devices` accept an IPv4/IPv6 fabric management IP address only; switch serial numbers (switchIds) are not accepted. This is deliberate. An earlier revision did accept both (a heuristic: a value containing a dot was treated as an IP, anything else passed through as a switchId), but we dropped serial input in `e6d5723a` for two reasons:

1. **Idempotency.** The wire stores switchIds, but we denormalize GET responses back to fabric management IPs (`_switch_ids_to_ips`) so the user sees IP-based output that matches their IP-based input. If we also accepted serials on input, that input would no longer match the IP-denormalized read state, so the diff could compare an IP against a serial and report a spurious change. IP-only input is inherently idempotent against the read state, with no proposed-side canonicalization needed.

2. **Simpler, fail-fast resolution.** `_resolve_switch_id` no longer has to guess IP-vs-serial. It resolves IPs through `FabricContext` and rejects a non-IP value with a clear error, instead of silently passing an unvalidated serial straight to the wire where a stale/wrong serial would surface as an obscure ND error.

Documented in the module's `update_group_switches` / `installation_order_devices` descriptions (clarified further in `71bb8f7a`). Net effect: one input format, validated up front, idempotent against what ND returns.

plugins/modules/nd_fabric_update_group.py
allenrobel: Good catch — you're exactly right about the asymmetry, and it's now fixed in 7f25daad.

The update path already guarded the empty case (raising `update_group_switches must be non-empty ... use state: deleted`), but the create path didn't: `_attach_item` turned an omitted/empty list into `switchIds: []` and sent it to `attachGroup`, where ND either rejects it or leaves a zero-switch ghost group — with no clear message.

**Code:** `_attach_item` now fails fast with a clear error on the create path:

> update_group_name 'leaf_group': update_group_switches must be non-empty when creating an update group; ND does not permit a zero-switch group.

The guard sits on the `switch_ids is None` branch, which is only reached from `create` / `create_bulk`. The update path passes an explicit `switch_ids` (the to-add subset) and keeps its own empty guard, so its "omitted = membership not managed, keep current members" semantics are untouched.

**Docs:** the `update_group_switches` description now spells out the distinction you asked about:

- On **create**, it's required and must be non-empty.
- On **update**, omitting it leaves current membership unchanged (membership not managed) and applies only the other settings; providing it reconciles membership to exactly the listed switches. An explicit empty list is rejected in both cases — use `state: deleted` to remove a group.

Added regression test `00170` (parametrized over omitted `None` and explicit `[]`) asserting create fails fast with no wire request issued. Full unit suite green (663).

samiib (approved):
LGTM

akinross (approved):

plugins/modules/nd_fabric_update_group.py
```diff
@@ -0,0 +1,363 @@
...
+    argument_spec = nd_argument_spec()
+    argument_spec.update(FabricUpdateGroupModel.get_argument_spec())
+
+    module = AnsibleModule(
+        argument_spec=argument_spec,
+        supports_check_mode=True,
+        mutually_exclusive=[["config", "auto_assign"]],
+    )
+    require_pydantic(module)
+
```
gmicol, line 346: Shouldn't we set the logger here as well?

```suggestion
    require_pydantic(module)
    setup_logging(module)
    module_log = logging.getLogger("nd.nd_fabric_update_group")
```

plugins/modules/nd_fabric_update_group.py
```diff
@@ -0,0 +1,363 @@
...
+    # has the before context (after == before reflects "nothing applied yet").
+    output.assign(before=before, after=before)
+    if not module.check_mode:
+        orchestrator.propose(module.params["auto_assign"])
+    after = NDConfigCollection.from_api_response(response_data=orchestrator.query_all(), model_class=FabricUpdateGroupModel)
+
+    output.assign(before=before, after=after)
+
+
+def _validate_report_analysis_exclusion(module: AnsibleModule) -> None:
```
gmicol, line 289: quick question Allen: could we not add it as a model validator for the Pydantic  `FabricUpdateGroupModel` instead to prevent extra looping over the `config`? 

Or is the validation truly dependent on the state as we have the following if statement: 
```python
if module.params.get("state") == "deleted":
        return
```

gmicol (changes requested):

plugins/modules/nd_fabric_update_group.py
allenrobel: Good question — I looked at moving it into a `FabricUpdateGroupModel` validator and decided to keep it at the module level on purpose. Documented the rationale in afbd34b0.

The blocker is the read path: the model is built from GET responses too (`from_response` -> `model_validate` runs every validator), in NDStateMachine's `before` snapshot, `_run_auto_assign`, and the merge read in the orchestrator. A cross-field validator rejecting `analysis`+report would therefore fire on each existing group read back from ND, and could abort the whole run on a group ND itself accepted/stored — i.e. it would make us stricter than ND on read. That's the same trap we already documented on `UpdateReportCheckModel.report_check_name`.

And to your second point — yes, it's genuinely state-aware: skipped for `state: deleted` (settings ignored), and `state` isn't a model field, so a validator would need ValidationInfo-context plumbing threaded through every construction site to know it. Running it over `module.params["config"]` keeps it input-only (sidesteps the read trap) and fail-fast before any wire request. The loop is one O(n) in-memory pass where n = number of update groups in the play, so the cost is negligible.

plugins/modules/nd_fabric_update_group.py
allenrobel: Good catch — fixed in 42361f99. The module wasn't configuring logging at all, unlike the reference Gen 3 module `nd_interface_loopback`.

Added right after `require_pydantic(module)`:

```python
setup_logging(module)
module_log = logging.getLogger("nd.nd_fabric_update_group")
```

I also wired `module_log` into both execution paths (the `auto_assign` action path and the `NDStateMachine` path) so it's actually used — debug entry/exit plus `module_log.exception(...)` in the `except` blocks, mirroring the loopback pattern. Logging stays a no-op unless `ND_LOGGING_CONFIG` is set.

plugins/modules/nd_fabric_update_group.py
gmicol: It makes perfect sense, I haven't thought that it would actually make it stricter than ND on read. the cost of this loop is indeed pretty negligible, it was just a tiny optimization enhancement if that was possible.
Thank you for clarifying this point @allenrobel !

gmicol (approved):
LGTM

allenrobel:
Merged after three approvals.  Thanks!

PR #286, opened 2026-05-20
[ND 4.2+] Add module for Routing Policies Prefix Lists (NDA-27)

gmicol:
## Related Issue(s)

- CiscoDevNet/ansible-nd#239

## Proposed Changes
Add new module `nd_manage_prefix_list` to manage IPv4 and IPv6 Routing Policies prefix lists on Cisco Nexus Dashboard fabrics. A single module handles both address families via `config.ip_version`, supporting `merged`, `replaced`, `overridden`, and `deleted` states with check mode.

Full module stack added under the current Pydantic/orchestrator architecture:
- Module: `plugins/modules/nd_manage_prefix_list.py`
- Endpoints: `endpoints/v1/manage/manage_prefix_lists.py` (IPv4/IPv6 CRUD + bulk-create/bulk-delete actions, shared bases) and a new `PrefixListNameMixin` in `endpoints/mixins.py`
- Models: `models/manage_prefix_list/` (model + enums, reusing shared CIDR/ASCII validated types, with entry-level semantic guardrails)
- Orchestrator: `orchestrators/manage_prefix_list.py` (per-version routing, bulk create/delete, HTTP 207 multi-status per-item failure parsing)

Notes: bulk create/delete are issued per address family; tenant-scoped prefix lists are documented as a known idempotency limitation.

## Test Notes
- Unit tests for endpoint, model, and orchestrator (RestSend-driven CRUD/bulk/207/query_all) plus fixtures; full `module_utils` unit suite passes.
- Integration target `nd_manage_prefix_list` covering setup/merged/replaced/overridden/deleted (incl. idempotency, identifier-only delete, multi-family in one task).
- `ansible-test sanity` passes (pep8, pylint, etc.).

Nexus Dashboard 4.2.1+

* [ ] analyze
* [ ] infa
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

### DIFF
--- a/plugins/module_utils/endpoints/v1/manage/manage_prefix_lists.py
+++ b/plugins/module_utils/endpoints/v1/manage/manage_prefix_lists.py
@@ -130,11 +130,14 @@ class _EpManagePrefixListsBase(PrefixListNameMixin, FabricNameMixin, NDEndpointB
 
     def set_identifiers(self, identifier: IdentifierKey = None) -> None:
         """
-        Accept either a plain name or a composite tuple ``(ip_version, name)``
-        and assign the prefix list name.
+        Accept either a plain name, ``(ip_version, name)``, or
+        ``(ip_version, tenant_name, name)`` and assign the API prefix list name.
         """
         if isinstance(identifier, tuple):
-            self.prefix_list_name = identifier[1]
+            if len(identifier) >= 3 and identifier[1]:
+                self.prefix_list_name = f"{identifier[1]}~{identifier[2]}"
+            else:
+                self.prefix_list_name = identifier[-1]
         else:
             self.prefix_list_name = identifier
 
--- a/plugins/module_utils/endpoints/v1/manage/manage_prefix_lists.py
+++ b/plugins/module_utils/endpoints/v1/manage/manage_prefix_lists.py
@@ -44,7 +44,7 @@ from __future__ import annotations
 
 __metaclass__ = type
 
-from typing import ClassVar, Literal, Optional, Tuple, Union
+from typing import ClassVar, Literal, Optional
 
 from ansible_collections.cisco.nd.plugins.module_utils.enums import HttpVerbEnum
 from ansible_collections.cisco.nd.plugins.module_utils.endpoints.v1.manage.base_path import BasePath
@@ -179,7 +179,9 @@ class EpManageIpv4PrefixListsListGet(_EpManageIpv4PrefixListsBase):
 
     _require_prefix_list_name: ClassVar[bool] = False
 
-    class_name: Literal["EpManageIpv4PrefixListsListGet"] = Field(default="EpManageIpv4PrefixListsListGet", description="Class name for backward compatibility")
+    class_name: Literal["EpManageIpv4PrefixListsListGet"] = Field(
+        default="EpManageIpv4PrefixListsListGet", description="Class name for backward compatibility"
+    )
     endpoint_params: PrefixListsListEndpointParams = Field(default_factory=PrefixListsListEndpointParams)
 
     @property
@@ -244,7 +246,9 @@ class EpManageIpv4PrefixListsBulkDelete(FabricNameMixin, NDEndpointBaseModel):
     Request body: ``{"ipv4PrefixListNames": ["name1", ...]}``.
     """
 
-    class_name: Literal["EpManageIpv4PrefixListsBulkDelete"] = Field(default="EpManageIpv4PrefixListsBulkDelete", description="Class name for backward compatibility")
+    class_name: Literal["EpManageIpv4PrefixListsBulkDelete"] = Field(
+        default="EpManageIpv4PrefixListsBulkDelete", description="Class name for backward compatibility"
+    )
 
     @property
     def path(self) -> str:
@@ -324,7 +328,9 @@ class EpManageIpv6PrefixListsListGet(_EpManageIpv6PrefixListsBase):
 
     _require_prefix_list_name: ClassVar[bool] = False
 
-    class_name: Literal["EpManageIpv6PrefixListsListGet"] = Field(default="EpManageIpv6PrefixListsListGet", description="Class name for backward compatibility")
+    class_name: Literal["EpManageIpv6PrefixListsListGet"] = Field(
+        default="EpManageIpv6PrefixListsListGet", description="Class name for backward compatibility"
+    )
     endpoint_params: PrefixListsListEndpointParams = Field(default_factory=PrefixListsListEndpointParams)
 
     @property
@@ -389,7 +395,9 @@ class EpManageIpv6PrefixListsBulkDelete(FabricNameMixin, NDEndpointBaseModel):
     Request body: ``{"ipv6PrefixListNames": ["name1", ...]}``.
     """
 
-    class_name: Literal["EpManageIpv6PrefixListsBulkDelete"] = Field(default="EpManageIpv6PrefixListsBulkDelete", description="Class name for backward compatibility")
+    class_name: Literal["EpManageIpv6PrefixListsBulkDelete"] = Field(
+        default="EpManageIpv6PrefixListsBulkDelete", description="Class name for backward compatibility"
+    )
 
     @property
     def path(self) -> str:
--- a/plugins/module_utils/endpoints/v1/manage/manage_prefix_lists.py
+++ b/plugins/module_utils/endpoints/v1/manage/manage_prefix_lists.py
@@ -255,12 +255,15 @@ class _EpManagePrefixListsBulkDeleteBase(FabricNameMixin, NDEndpointBaseModel):
 
     # Address-family action segment, set by the per-family subclasses below.
     _action_segment: ClassVar[str] = ""
+    endpoint_params: PrefixListsEndpointParams = Field(default_factory=PrefixListsEndpointParams)
 
     @property
     def path(self) -> str:
         if self.fabric_name is None:
             raise ValueError(f"{type(self).__name__}.path: fabric_name must be set before accessing path.")
-        return BasePath.path("fabrics", quote(self.fabric_name, safe=""), self._action_segment, "remove")
+        base = BasePath.path("fabrics", quote(self.fabric_name, safe=""), self._action_segment, "remove")
+        qs = self.endpoint_params.to_query_string()
+        return f"{base}?{qs}" if qs else base
 
 
 class EpManageIpv4PrefixListsBulkDelete(_EpManagePrefixListsBulkDeleteBase):
--- a/plugins/module_utils/endpoints/v1/manage/manage_prefix_lists.py
+++ b/plugins/module_utils/endpoints/v1/manage/manage_prefix_lists.py
@@ -1,0 +1,402 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2026, Gaspard Micol (@gmicol) <gmicol@cisco.com>
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+"""
+ND Manage Prefix Lists endpoint models (IPv4 and IPv6).
+
+This module contains endpoint definitions for prefix-list operations
+in the ND Manage API, covering both IPv4 and IPv6 prefix lists.
+
+## IPv4 Endpoints
+
+- `EpManageIpv4PrefixListsGet` - Get a specific IPv4 prefix list by name
+  (GET /api/v1/manage/fabrics/{fabricName}/ipv4PrefixLists/{ipv4PrefixListName})
+- `EpManageIpv4PrefixListsListGet` - List all IPv4 prefix lists for a fabric
+  (GET /api/v1/manage/fabrics/{fabricName}/ipv4PrefixLists)
+- `EpManageIpv4PrefixListsPost` - Bulk-create IPv4 prefix lists
+  (POST /api/v1/manage/fabrics/{fabricName}/ipv4PrefixLists)
+- `EpManageIpv4PrefixListsPut` - Update a specific IPv4 prefix list
+  (PUT /api/v1/manage/fabrics/{fabricName}/ipv4PrefixLists/{ipv4PrefixListName})
+- `EpManageIpv4PrefixListsDelete` - Delete a specific IPv4 prefix list
+  (DELETE /api/v1/manage/fabrics/{fabricName}/ipv4PrefixLists/{ipv4PrefixListName})
+- `EpManageIpv4PrefixListsBulkDelete` - Bulk-delete IPv4 prefix lists
+  (POST /api/v1/manage/fabrics/{fabricName}/ipv4PrefixListActions/remove)
+
+## IPv6 Endpoints
+
+- `EpManageIpv6PrefixListsGet` - Get a specific IPv6 prefix list by name
+  (GET /api/v1/manage/fabrics/{fabricName}/ipv6PrefixLists/{ipv6PrefixListName})
+- `EpManageIpv6PrefixListsListGet` - List all IPv6 prefix lists for a fabric
+  (GET /api/v1/manage/fabrics/{fabricName}/ipv6PrefixLists)
+- `EpManageIpv6PrefixListsPost` - Bulk-create IPv6 prefix lists
+  (POST /api/v1/manage/fabrics/{fabricName}/ipv6PrefixLists)
+- `EpManageIpv6PrefixListsPut` - Update a specific IPv6 prefix list
+  (PUT /api/v1/manage/fabrics/{fabricName}/ipv6PrefixLists/{ipv6PrefixListName})
+- `EpManageIpv6PrefixListsDelete` - Delete a specific IPv6 prefix list
+  (DELETE /api/v1/manage/fabrics/{fabricName}/ipv6PrefixLists/{ipv6PrefixListName})
+- `EpManageIpv6PrefixListsBulkDelete` - Bulk-delete IPv6 prefix lists
+  (POST /api/v1/manage/fabrics/{fabricName}/ipv6PrefixListActions/remove)
+"""
+
+from __future__ import annotations
+
+__metaclass__ = type
+
+from typing import ClassVar, Literal, Optional, Tuple, Union
+
+from ansible_collections.cisco.nd.plugins.module_utils.enums import HttpVerbEnum
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.v1.manage.base_path import BasePath
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.mixins import FabricNameMixin, PrefixListNameMixin
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.base import NDEndpointBaseModel
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.query_params import EndpointQueryParams
+from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import Field
+from ansible_collections.cisco.nd.plugins.module_utils.types import IdentifierKey
+
+
+class PrefixListsEndpointParams(EndpointQueryParams):
+    """
+    Query parameters shared by single-item prefix list endpoints.
+
+    - cluster_name: Name of the target Nexus Dashboard cluster (multi-cluster deployments)
+    """
+
+    cluster_name: Optional[str] = Field(
+        default=None,
+        min_length=1,
+        description="Name of the target Nexus Dashboard cluster to execute this API, in a multi-cluster deployment",
+    )
+
+
+class PrefixListsListEndpointParams(EndpointQueryParams):
+    """
+    Query parameters for prefix list collection (list) endpoints.
+
+    - cluster_name: multi-cluster target cluster name
+    - filter: Lucene-format filter string
+    - max: maximum number of records
+    - offset: records to skip for pagination
+    - sort: sort field with optional ``:desc`` suffix
+    """
+
+    cluster_name: Optional[str] = Field(
+        default=None,
+        min_length=1,
+        description="Name of the target Nexus Dashboard cluster to execute this API, in a multi-cluster deployment",
+    )
+
+    filter: Optional[str] = Field(
+        default=None,
+        description="Lucene format filter - Filter the response based on this filter field",
+    )
+
+    max: Optional[int] = Field(
+        default=None,
+        ge=1,
+        description="Number of records to return",
+    )
+
+    offset: Optional[int] = Field(
+        default=None,
+        ge=0,
+        description="Number of records to skip for pagination",
+    )
+
+    sort: Optional[str] = Field(
+        default=None,
+        description="Sort the records by the declared fields in either ascending (default) or descending (:desc) order",
+    )
+
+
+# ---------------------------------------------------------------------------
+# IPv4 base class
+# ---------------------------------------------------------------------------
+
+
+class _EpManageIpv4PrefixListsBase(PrefixListNameMixin, FabricNameMixin, NDEndpointBaseModel):
+    """
+    Base class for IPv4 prefix list endpoints.
+
+    Path: ``/api/v1/manage/fabrics/{fabricName}/ipv4PrefixLists[/{name}]``
+
+    Subclasses set ``_require_prefix_list_name = False`` for collection-level
+    operations (list, bulk-create).
+    """
+
+    _require_prefix_list_name: ClassVar[bool] = True
+
+    endpoint_params: EndpointQueryParams = Field(default_factory=EndpointQueryParams, description="Endpoint-specific query parameters")
+
+    def set_identifiers(self, identifier: IdentifierKey = None) -> None:
+        """
+        Accept either a plain name or a composite tuple ``("ipv4", name)``
+        and assign the prefix list name.
+        """
+        if isinstance(identifier, tuple):
+            self.prefix_list_name = identifier[1]
+        else:
+            self.prefix_list_name = identifier
+
+    @property
+    def path(self) -> str:
+        if self._require_prefix_list_name and self.fabric_name is None:
+            raise ValueError(f"{type(self).__name__}.path: fabric_name must be set before accessing path.")
+        if self._require_prefix_list_name and self.prefix_list_name is None:
+            raise ValueError(f"{type(self).__name__}.path: prefix_list_name must be set before accessing path.")
+        segments = ["fabrics"]
+        if self.fabric_name is not None:
+            segments.append(self.fabric_name)
+        segments.append("ipv4PrefixLists")
+        if self.prefix_list_name is not None:
+            segments.append(self.prefix_list_name)
+        base = BasePath.path(*segments)
+        qs = self.endpoint_params.to_query_string()
+        return f"{base}?{qs}" if qs else base
+
+
+class EpManageIpv4PrefixListsGet(_EpManageIpv4PrefixListsBase):
+    """
+    GET /api/v1/manage/fabrics/{fabricName}/ipv4PrefixLists/{ipv4PrefixListName}
+
+    Retrieve a specific IPv4 prefix list by name.
+    """
+
+    class_name: Literal["EpManageIpv4PrefixListsGet"] = Field(default="EpManageIpv4PrefixListsGet", description="Class name for backward compatibility")
+    endpoint_params: PrefixListsEndpointParams = Field(default_factory=PrefixListsEndpointParams)
+
+    @property
+    def verb(self) -> HttpVerbEnum:
+        return HttpVerbEnum.GET
+
+
+class EpManageIpv4PrefixListsListGet(_EpManageIpv4PrefixListsBase):
+    """
+    GET /api/v1/manage/fabrics/{fabricName}/ipv4PrefixLists
+
+    List all IPv4 prefix lists for a fabric.
+    """
+
+    _require_prefix_list_name: ClassVar[bool] = False
+
+    class_name: Literal["EpManageIpv4PrefixListsListGet"] = Field(default="EpManageIpv4PrefixListsListGet", description="Class name for backward compatibility")
+    endpoint_params: PrefixListsListEndpointParams = Field(default_factory=PrefixListsListEndpointParams)
+
+    @property
+    def verb(self) -> HttpVerbEnum:
+        return HttpVerbEnum.GET
+
+
+class EpManageIpv4PrefixListsPost(_EpManageIpv4PrefixListsBase):
+    """
+    POST /api/v1/manage/fabrics/{fabricName}/ipv4PrefixLists
+
+    Bulk-create IPv4 prefix lists.
+    Request body: ``{"ipv4PrefixLists": [...]}``.
+    """
+
+    _require_prefix_list_name: ClassVar[bool] = False
+
+    class_name: Literal["EpManageIpv4PrefixListsPost"] = Field(default="EpManageIpv4PrefixListsPost", description="Class name for backward compatibility")
+    endpoint_params: PrefixListsEndpointParams = Field(default_factory=PrefixListsEndpointParams)
+
+    @property
+    def verb(self) -> HttpVerbEnum:
+        return HttpVerbEnum.POST
+
+
+class EpManageIpv4PrefixListsPut(_EpManageIpv4PrefixListsBase):
+    """
+    PUT /api/v1/manage/fabrics/{fabricName}/ipv4PrefixLists/{ipv4PrefixListName}
+
+    Update a specific IPv4 prefix list.
+    Request body: ``ipv4PrefixListItem`` schema.
+    """
+
+    class_name: Literal["EpManageIpv4PrefixListsPut"] = Field(default="EpManageIpv4PrefixListsPut", description="Class name for backward compatibility")
+    endpoint_params: PrefixListsEndpointParams = Field(default_factory=PrefixListsEndpointParams)
+
+    @property
+    def verb(self) -> HttpVerbEnum:
+        return HttpVerbEnum.PUT
+
+
+class EpManageIpv4PrefixListsDelete(_EpManageIpv4PrefixListsBase):
+    """
+    DELETE /api/v1/manage/fabrics/{fabricName}/ipv4PrefixLists/{ipv4PrefixListName}
+
+    Delete a specific IPv4 prefix list.
+    """
+
+    class_name: Literal["EpManageIpv4PrefixListsDelete"] = Field(default="EpManageIpv4PrefixListsDelete", description="Class name for backward compatibility")
+    endpoint_params: PrefixListsEndpointParams = Field(default_factory=PrefixListsEndpointParams)
+
+    @property
+    def verb(self) -> HttpVerbEnum:
+        return HttpVerbEnum.DELETE
+
+
+class EpManageIpv4PrefixListsBulkDelete(FabricNameMixin, NDEndpointBaseModel):
+    """
+    POST /api/v1/manage/fabrics/{fabricName}/ipv4PrefixListActions/remove
+
+    Bulk-delete IPv4 prefix lists.
+    Request body: ``{"ipv4PrefixListNames": ["name1", ...]}``.
+    """
+
+    class_name: Literal["EpManageIpv4PrefixListsBulkDelete"] = Field(default="EpManageIpv4PrefixListsBulkDelete", description="Class name for backward compatibility")
+
+    @property
+    def path(self) -> str:
+        if self.fabric_name is None:
+            raise ValueError(f"{type(self).__name__}.path: fabric_name must be set before accessing path.")
+        return BasePath.path("fabrics", self.fabric_name, "ipv4PrefixListActions", "remove")
+
+    @property
+    def verb(self) -> HttpVerbEnum:
+        return HttpVerbEnum.POST
+
+
+# ---------------------------------------------------------------------------
+# IPv6 base class
+# ---------------------------------------------------------------------------
+
+
+class _EpManageIpv6PrefixListsBase(PrefixListNameMixin, FabricNameMixin, NDEndpointBaseModel):
+    """
+    Base class for IPv6 prefix list endpoints.
+
+    Path: ``/api/v1/manage/fabrics/{fabricName}/ipv6PrefixLists[/{name}]``
+    """
+
+    _require_prefix_list_name: ClassVar[bool] = True
+
+    endpoint_params: EndpointQueryParams = Field(default_factory=EndpointQueryParams, description="Endpoint-specific query parameters")
+
+    def set_identifiers(self, identifier: IdentifierKey = None) -> None:
+        """
+        Accept either a plain name or a composite tuple ``("ipv6", name)``
+        and assign the prefix list name.
+        """
+        if isinstance(identifier, tuple):
+            self.prefix_list_name = identifier[1]
+        else:
+            self.prefix_list_name = identifier
+
+    @property
+    def path(self) -> str:
+        if self._require_prefix_list_name and self.fabric_name is None:
+            raise ValueError(f"{type(self).__name__}.path: fabric_name must be set before accessing path.")
+        if self._require_prefix_list_name and self.prefix_list_name is None:
+            raise ValueError(f"{type(self).__name__}.path: prefix_list_name must be set before accessing path.")
+        segments = ["fabrics"]
+        if self.fabric_name is not None:
+            segments.append(self.fabric_name)
+        segments.append("ipv6PrefixLists")
+        if self.prefix_list_name is not None:
+            segments.append(self.prefix_list_name)
+        base = BasePath.path(*segments)
+        qs = self.endpoint_params.to_query_string()
+        return f"{base}?{qs}" if qs else base
+
+
+class EpManageIpv6PrefixListsGet(_EpManageIpv6PrefixListsBase):
+    """
+    GET /api/v1/manage/fabrics/{fabricName}/ipv6PrefixLists/{ipv6PrefixListName}
+
+    Retrieve a specific IPv6 prefix list by name.
+    """
+
+    class_name: Literal["EpManageIpv6PrefixListsGet"] = Field(default="EpManageIpv6PrefixListsGet", description="Class name for backward compatibility")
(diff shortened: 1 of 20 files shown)


plugins/modules/nd_manage_prefix_list.py
```diff
@@ -0,0 +1,312 @@
...
+        - When omitted, the default tenant is used.
+        - Allowed characters are C([A-Za-z0-9_-]).
+        type: str
+        aliases: [ tenantName ]
+      entries:
+        description:
+        - The list of prefix list entries.
+        - Each entry defines a permit/deny action for a specific IP prefix.
+        - Required for O(state=merged), O(state=replaced), and O(state=overridden).
+        - Optional for O(state=deleted), where identifier-only items
```
akinross, line 83: is required/optional documentation only or do we enforce this in the pydantic model?

plugins/modules/nd_manage_prefix_list.py
```diff
@@ -0,0 +1,312 @@
...
+        - Allowed characters are C([a-zA-Z0-9~_-]).
+        - Maximum length is 115 characters (63 for the default tenant).
+        type: str
+        required: true
+      description:
+        description:
+        - A human-readable description of the prefix list.
+        - Maximum length is 90 characters.
+        type: str
+      tenant_name:
```
akinross, line 66: can we have duplicate entries across different tenants, should this also be an identifier field of the prefix list?

plugins/modules/nd_manage_prefix_list.py
```diff
@@ -0,0 +1,312 @@
...
+            - The action to take when the prefix matches.
+            type: str
+            required: true
+            choices: [ permit, deny ]
+          prefix:
+            description:
+            - The IP prefix in CIDR notation.
+            - Must be an IPv4 CIDR (e.g. C(10.0.0.0/8)) when
+              O(config.ip_version=ipv4).
+            - Must be an IPv6 CIDR (e.g. C(2001:db8::/32)) when
```
akinross, line 111: are ipv6 address normalized? In terraform aci we had issues with plan mode when full notation was send but compressed notation was returned. Will this impact change detection?

plugins/modules/nd_manage_prefix_list.py
```diff
@@ -0,0 +1,312 @@
...
+            - Range 1-32 for IPv4, 1-128 for IPv6.
+            type: int
+            aliases: [ maxLength ]
+          mask:
+            description:
+            - Network mask in dotted-decimal format for IPv4
+              (e.g. C(255.255.255.0)) or explicit match mask in IPv6
+              format (e.g. C(ffff:ffff::)).
+            - Must be a valid IPv4 address when O(config.ip_version=ipv4).
+            - Must be a valid IPv6 address when O(config.ip_version=ipv6).
```
akinross, line 144: i am confused about this description, should it be a valid mask? also is it intended to expose only dotted decimal, or do we need numerical mask acceptance also?

further more why is the mask provided, could this not be deduced from the prefix itself?

ui names this Explicit match mask, so perhaps align the description more clear?

plugins/modules/nd_manage_prefix_list.py
```diff
@@ -0,0 +1,312 @@
...
+  O(config.entries.max_prefix_length) are set, the minimum must be less than
+  or equal to the maximum.
+- O(config.entries.sequence_number) values must be unique within each
+  prefix list.
+- O(config.description) is validated as ASCII-only.
+- O(config.tenant_name) (non-default tenant) prefix lists are not yet fully
+  supported for idempotency. Nexus Dashboard stores and returns these prefix
+  lists under the composite name C(<tenant_name>~<name>), which does not match
+  the configured O(config.name), so repeated runs may not be idempotent. Use
+  prefix lists in the default tenant until this is addressed.
```
akinross: why do we specify attribute level notes here and not in the description of attribute?

plugins/modules/nd_manage_prefix_list.py
```diff
@@ -0,0 +1,312 @@
...
+    type: str
+    required: true
+  config:
+    description:
+    - The list of prefix lists to configure.
+    type: list
+    elements: dict
+    required: True
+    suboptions:
+      ip_version:
```
akinross, line 41: did you consider introducing a default of ipv4, which is also ui default?

plugins/modules/nd_manage_prefix_list.py
```diff
@@ -0,0 +1,312 @@
...
+        required: false
+        suboptions:
+          sequence_number:
+            description:
+            - The sequence number of this entry (1-4294967294).
+            - Entries are evaluated in ascending sequence order.
+            type: int
+            required: true
+            aliases: [ sequenceNumber ]
+          action:
```
akinross, line 97: did you consider introducing a default of permit, which is also ui default?

akinross (changes requested):

plugins/module_utils/orchestrators/manage_prefix_list.py
```diff
@@ -0,0 +1,285 @@
...
+        try:
+            config = self._config_for_version(str(model_instance.ip_version))
+            api_endpoint = config["get"]()
+            api_endpoint.fabric_name = self.fabric_name
+            api_endpoint.set_identifiers(model_instance.get_identifier_value())
+            return self._request(path=api_endpoint.path, verb=api_endpoint.verb)
+        except Exception as e:
+            raise Exception(f"Query failed for {model_instance.get_identifier_value()}: {e}") from e
+
+    def query_all(self) -> ResponseType:
```
mikewiebe, line 368: **Issue**

Just something to consider here in terms of the performance and scale impact.

Every `NDStateMachine` run initializes by calling `query_all()`, and this orchestrator always fetches both IPv4 and IPv6 prefix list collections with a single GET per family. The OpenAPI list operations expose `offset`, `max`, `sort`, and `filter`, and the collection responses include `meta`, but `query_all()` does not set pagination parameters or follow metadata links.

**Evidence**

- [`plugins/module_utils/nd_state_machine.py` lines 68-80](https://github.com/CiscoDevNet/ansible-nd/blob/5d731ad693fa4da003e0c5551221491143108969/plugins/module_utils/nd_state_machine.py#L68-L80) calls `model_orchestrator.query_all()` during initialization before state-specific processing.
- [`plugins/module_utils/orchestrators/manage_prefix_list.py` lines 233-249](https://github.com/CiscoDevNet/ansible-nd/pull/286/files#diff-9903f5e5412135a57fc9b5ac60dfa8cfc14c91110c8bc985cfbc6da18eb86c97R233) fetches the IPv4 and IPv6 collections once and combines them.
- [`plugins/module_utils/endpoints/v1/manage/manage_prefix_lists.py` lines 71-108](https://github.com/CiscoDevNet/ansible-nd/pull/286/files#diff-7a0251691a8d92eae956f6c301de134b9e54b5d5444d8a1f499fe4aa633a4b17R71) defines endpoint params for `cluster_name`, `filter`, `max`, `offset`, and `sort`.
- [`plugins/module_utils/orchestrators/manage_prefix_list.py` lines 242-245](https://github.com/CiscoDevNet/ansible-nd/pull/286/files#diff-9903f5e5412135a57fc9b5ac60dfa8cfc14c91110c8bc985cfbc6da18eb86c97R242) instantiates the list endpoint without setting any filter, limit, or offset.
- OpenAPI evidence: `GET /fabrics/{fabricName}/ipv4PrefixLists` and `GET /fabrics/{fabricName}/ipv6PrefixLists` both expose `offset`, `max`, `sort`, and `filter`; `ipv4PrefixListCollection` and `ipv6PrefixListCollection` both include `meta`.


**Impact**

For large fabrics, even a one-item `merged`, `replaced`, or `deleted` task performs two unscoped collection reads before doing useful work. If the controller applies a default page size, the module can build an incomplete `before` collection, miss existing resources, fail to delete resources beyond the first page, or make `overridden` incomplete.

**Suggested fix**

Verify the API pagination default. If the endpoint is paginated, iterate `max`/`offset` or follow `meta.links.next` until all rows are collected. Separately, consider a scoped lookup path for non-`overridden` states that queries only the proposed `(ip_version, name)` identities, while keeping full scans for `overridden` and future broad `gathered` behavior.

plugins/modules/nd_manage_prefix_list.py
```diff
@@ -0,0 +1,312 @@
...
+      as defined in the configuration.
+      Prefix lists on ND that are not specified in the configuration are left unchanged.
+    - Use O(state=replaced) to replace the prefix lists specified in the configuration.
+    - Use O(state=overridden) to enforce the configuration as the single source of truth.
+      All prefix lists (both IPv4 and IPv6) on ND not present in the configuration
+      will be deleted. Use with caution.
+    - Use O(state=deleted) to remove the prefix lists specified in the configuration.
+    type: str
+    default: merged
+    choices: [ merged, replaced, overridden, deleted ]
```
mikewiebe, line 161: I can't remember but do we have an issue filed to update NDStateMachine to include gathered state?  If so, we can ignore this comment.

**Update:** I created an issue for this https://github.com/CiscoDevNet/ansible-nd/issues/340

**Issue**

New modules should generally use the standard state set, including `gathered`, unless there is a clear resource-specific reason. This module does not document, accept, implement, or test `gathered`, even though the OpenAPI spec exposes list/get/summary read operations for the resource family.

**Impact**

Without `gathered`, users do not get the standard read-only discovery path for prefix lists, and this module becomes an argspec outlier compared with the newer state vocabulary. Because the generic `NDStateMachine` does not yet implement `gathered`, this is an architecture gap rather than a simple docs-only omission.

**Suggested fix**

Either add `gathered` support for this module, including docs, return data, and tests, or document a clear reason it is intentionally deferred. If the generic state machine is extended, keep no-mutation semantics explicit and ensure write controls are rejected for `gathered`.

plugins/module_utils/endpoints/v1/manage/manage_prefix_lists.py
```diff
@@ -0,0 +1,390 @@
...
+
+
+class PrefixListsEndpointParams(EndpointQueryParams):
+    """
+    Query parameters shared by single-item prefix list endpoints.
+
+    - cluster_name: Name of the target Nexus Dashboard cluster (multi-cluster deployments)
+    """
+
+    cluster_name: Optional[str] = Field(
```
mikewiebe: **Issue**

The OpenAPI operations for prefix lists all accept optional `clusterName`, and the endpoint classes model it as `cluster_name`, but the module argspec does not expose it and the orchestrator never passes it into endpoint params. If this module is expected to work in multi-cluster deployments, the API capability is incomplete.

**Evidence**

- [`plugins/module_utils/endpoints/v1/manage/manage_prefix_lists.py` lines 57-68](https://github.com/CiscoDevNet/ansible-nd/pull/286/files#diff-7a0251691a8d92eae956f6c301de134b9e54b5d5444d8a1f499fe4aa633a4b17R57) defines `cluster_name` for single-item prefix list endpoints.
- [`plugins/module_utils/endpoints/v1/manage/manage_prefix_lists.py` lines 71-108](https://github.com/CiscoDevNet/ansible-nd/pull/286/files#diff-7a0251691a8d92eae956f6c301de134b9e54b5d5444d8a1f499fe4aa633a4b17R71) defines `cluster_name` for list endpoints.
- [`plugins/module_utils/models/manage_prefix_list/manage_prefix_list.py` lines 291-360](https://github.com/CiscoDevNet/ansible-nd/pull/286/files#diff-fadeaab1a72f95cb37e89d073ba5ddd905a06fc6540ed00b18fb799d297cf5d1R291) exposes only `fabric_name`, `config`, and `state` at the module argspec level.
- OpenAPI evidence: the list, get, create, update, single delete, and bulk delete prefix-list operations all accept optional query parameter `clusterName`.

**Impact**

This may be fine for local-cluster-only use, but the module currently gives users no way to use an API capability that the endpoint layer already partially models. In a multi-cluster deployment, users may be unable to target the intended cluster or may assume the module supports only a subset of the API without documentation saying so.

**Suggested fix**

Either expose a documented `cluster_name` option and thread it from module params into endpoint params, or document that the module intentionally does not support `clusterName`/multi-cluster targeting. Add a focused endpoint/orchestrator test if the option is exposed.

mikewiebe (changes requested):

plugins/module_utils/orchestrators/manage_prefix_list.py
```diff
@@ -0,0 +1,285 @@
...
+            status = str(item.get("status") or "").lower()
+            if status not in _FAILURE_STATUSES:
+                continue
+            name = item.get("name") or "?"
+            message = item.get("message") or "unknown error"
+            failures.append(f"{name}: {message}")
+        if failures:
+            raise Exception(f"prefix list {operation} reported per-item failures: {'; '.join(failures)}")
+
+    def _bulk_create_for_version(self, version: str, items: list[PrefixListModel]) -> ResponseType:
```
mikewiebe, line 306: **Issue**

The custom prefix-list orchestrator uses `_request()` directly for bulk create, bulk delete, and update, but does not pass `operation_type`. The shared helper defaults to `OperationType.QUERY`, so mutating calls are registered in `Results` as query operations instead of create/update/delete operations.

**Evidence**

- [`plugins/module_utils/orchestrators/base.py` lines 63-78](https://github.com/CiscoDevNet/ansible-nd/blob/5d731ad693fa4da003e0c5551221491143108969/plugins/module_utils/orchestrators/base.py#L63-L78) stores `results.action`, `results.operation_type`, and verbosity from the supplied operation type.
- [`plugins/module_utils/orchestrators/base.py` lines 80-108](https://github.com/CiscoDevNet/ansible-nd/blob/5d731ad693fa4da003e0c5551221491143108969/plugins/module_utils/orchestrators/base.py#L80-L108) defaults `_request(..., operation_type=OperationType.QUERY)`.
- [`plugins/module_utils/orchestrators/base.py` lines 122-141](https://github.com/CiscoDevNet/ansible-nd/blob/5d731ad693fa4da003e0c5551221491143108969/plugins/module_utils/orchestrators/base.py#L122-L141) shows the expected generic CRUD pattern: create, update, and delete pass `OperationType.CREATE`, `OperationType.UPDATE`, and `OperationType.DELETE`.
- [`plugins/module_utils/orchestrators/manage_prefix_list.py` lines 177-185](https://github.com/CiscoDevNet/ansible-nd/pull/286/files#diff-9903f5e5412135a57fc9b5ac60dfa8cfc14c91110c8bc985cfbc6da18eb86c97R177) calls `_request()` for bulk create without an operation type.
- [`plugins/module_utils/orchestrators/manage_prefix_list.py` lines 187-195](https://github.com/CiscoDevNet/ansible-nd/pull/286/files#diff-9903f5e5412135a57fc9b5ac60dfa8cfc14c91110c8bc985cfbc6da18eb86c97R187) calls `_request()` for bulk delete without an operation type.
- [`plugins/module_utils/orchestrators/manage_prefix_list.py` lines 204-211](https://github.com/CiscoDevNet/ansible-nd/pull/286/files#diff-9903f5e5412135a57fc9b5ac60dfa8cfc14c91110c8bc985cfbc6da18eb86c97R204) calls `_request()` for update without an operation type.

**Impact**

The controller mutation still runs, but the module's API call history and current action metadata report writes as reads. That makes troubleshooting misleading, hides write calls behind query verbosity, and leaves create/update/delete observability untested for this custom orchestrator path.

**Suggested fix**

Import `OperationType` into `manage_prefix_list.py` and pass `OperationType.CREATE` for `_bulk_create_for_version()`, `OperationType.DELETE` for `_bulk_delete_for_version()`, and `OperationType.UPDATE` for `update()`. Add focused tests that assert create, update, and delete paths register the expected operation type/action.

plugins/modules/nd_manage_prefix_list.py
gmicol: Fixed by enforcing non-empty entries for merged, replaced, and overridden; deleted still allows identifier-only config. Added unit coverage.

plugins/modules/nd_manage_prefix_list.py
gmicol: I made tenant part of the prefix-list identity, so same-name prefix lists in different tenants are handled independently.

plugins/modules/nd_manage_prefix_list.py
gmicol: Normalized prefix and mask values through `ipaddress`, including IPv6 formatting.

plugins/modules/nd_manage_prefix_list.py
gmicol: Updated the mask documentation to clarify that it is an explicit match mask, with IPv4 dotted-decimal and IPv6 address formats

plugins/modules/nd_manage_prefix_list.py
gmicol: You are right, I moved field-level validation notes out of individual attribute docs and kept them in module notes

plugins/modules/nd_manage_prefix_list.py
gmicol: I kept `ip_version` explicit rather than defaulting to IPv4 because it is part of the object identity. Updated docs to make that explicit

plugins/module_utils/orchestrators/manage_prefix_list.py
gmicol: Fixed state handling so merged, replaced, and deleted query only targeted objects, while overridden performs a full paginated scan

plugins/module_utils/endpoints/v1/manage/manage_prefix_lists.py
gmicol: Added `cluster_name` support and pass it through as `clusterName`

plugins/module_utils/orchestrators/manage_prefix_list.py
gmicol: Fixed action tracking to use the correct create, update, and delete operation types

plugins/module_utils/orchestrators/manage_prefix_list.py
mikewiebe: Resolved because the current head now uses targeted lookups for merged/replaced/deleted and paginated max/offset collection scans where a full scan is needed.

plugins/module_utils/endpoints/v1/manage/manage_prefix_lists.py
mikewiebe: Resolved because the current head now exposes `cluster_name` and threads it into endpoint params as `clusterName`.

plugins/module_utils/orchestrators/manage_prefix_list.py
mikewiebe: Resolved because the current head now passes explicit create, delete, and update `OperationType` values on the custom write paths.

plugins/modules/nd_manage_prefix_list.py
mikewiebe: Closing this thread because the shared gathered-state follow-up is tracked in issue #340. I added this PR comment link to that issue here: https://github.com/CiscoDevNet/ansible-nd/issues/340#issuecomment-4850159948

tests/integration/targets/nd_manage_prefix_list/tasks/replaced.yaml
```diff
@@ -0,0 +1,110 @@
...
+      - cm_replaced_ipv4 is changed
+      - nm_replaced_ipv4 is changed
+
+- name: "REPLACED: Verify replaced values -- only 1 entry, no description"
+  vars:
+    ipv4_after: "{{ nm_replaced_ipv4.after | selectattr('name', 'equalto', 'PL-IPV4-BORDERS') | first }}"
+  ansible.builtin.assert:
+    that:
+      - ipv4_after.entries | length == 1
+      - ipv4_after.description is none or ipv4_after.description == ""
```
mikewiebe: **Issue:** The replaced-state integration test fails when `description` is omitted from the returned object.

I reproduced this live by creating an IPv4 prefix list with a description, replacing it with a minimal config that omits `description`, and then running the same assertion used here:

```yaml
- ipv4_after.description is none or ipv4_after.description == ""
```

Observed failure:

```text
'dict object' has no attribute 'description'
```

The module result omits the `description` key when the replacement leaves it blank/absent. That seems like acceptable module behavior, but the test assertion is too strict because it assumes the key always exists.

Minimum repro:

```yaml
- name: Create prefix list with description
  cisco.nd.nd_manage_prefix_list:
    config:
      - ip_version: ipv4
        name: PR286-REPRO-PL-DESC
        description: seed description
        entries:
          - sequence_number: 10
            action: permit
            prefix: 10.0.0.0/8
    state: merged

- name: Replace prefix list without description
  cisco.nd.nd_manage_prefix_list:
    config:
      - ip_version: ipv4
        name: PR286-REPRO-PL-DESC
        entries:
          - sequence_number: 10
            action: permit
            prefix: 10.0.0.0/8
    state: replaced
  register: replaced_result
```

Suggested fix: make the assertion tolerate an omitted key, for example:

```yaml
- ipv4_after.description | default("") == ""
```

plugins/module_utils/orchestrators/manage_prefix_list.py
```diff
@@ -0,0 +1,404 @@
...
+            raw = self._request(path=api_endpoint.path, verb=api_endpoint.verb, not_found_ok=True)
+            if not raw:
+                break
+            page = raw.get(config["list_key"], []) or []
+            for item in page:
+                item["ipVersion"] = version
+                results.append(item)
+            if not self._has_next_page(raw, len(page), offset):
+                break
+            offset += _QUERY_PAGE_SIZE
```
allenrobel: **High — pagination can silently skip records.**

The offset advances by the page *size requested*, not the number of rows *received*:

```python
offset += _QUERY_PAGE_SIZE
```

but `_has_next_page` (lines 248–253) explicitly supports short non-final pages (`offset + page_count < total`, `remaining > 0`). If the controller ever returns fewer than 100 rows with more data remaining (server-side page cap, transient truncation), rows at indices `len(page)`..99 of the window are never fetched. In `overridden` state the incomplete collection is the deletion baseline, so the skipped prefix lists silently escape override deletion; `before`/`after` output is wrong for every state.

Notably, `test_manage_prefix_list_00136` models exactly this shape (a 1-row page with `"meta": {"counts": {"remaining": 1}}`) and only passes because the file-based mock sender ignores the offset param — against a real server the second GET at `offset=100` would miss the second item.

Suggested fix: `offset += len(page)` — correct for full and short pages alike.

plugins/module_utils/endpoints/v1/manage/manage_prefix_lists.py
```diff
@@ -0,0 +1,396 @@
...
+    endpoint_params: EndpointQueryParams = Field(default_factory=EndpointQueryParams, description="Endpoint-specific query parameters")
+
+    def set_identifiers(self, identifier: IdentifierKey = None) -> None:
+        """
+        Accept either a plain name, ``(ip_version, name)``, or
+        ``(ip_version, tenant_name, name)`` and assign the API prefix list name.
+        """
+        if isinstance(identifier, tuple):
+            if len(identifier) >= 3 and identifier[1]:
+                self.prefix_list_name = f"{identifier[1]}~{identifier[2]}"
```
allenrobel, line 112: **High — tenant-qualified names can overflow the 115-char endpoint cap and crash mid-run.**

`set_identifiers` assigns `f"{identifier[1]}~{identifier[2]}"` into `prefix_list_name`, whose `PrefixListNameMixin` constraint is `max_length=115` — and `NDEndpointBaseModel` sets `validate_assignment=True`, so the constraint fires on assignment. The model independently allows `name` up to 115 and `tenant_name` up to 63, with no combined-length check on `api_name`, so a config the model validates as legal (e.g. tenant 8 chars + name 110 chars = 119 qualified) raises a raw Pydantic `ValidationError` (`String should have at most 115 characters`).

Because `_query_one_existing` also goes through `set_identifiers`, this fires during the scoped preflight on every `merged`/`replaced`/`deleted` run — including for objects the module itself created successfully via bulk POST (which sends bare `name` + `tenantName`), so the module can create an object it can then never query, update, or delete.

Suggested fix: validate the combined qualified-name length up front on `PrefixListModel` (clean, early error naming both fields), and size the endpoint field to whatever the API actually accepts for the qualified form.

allenrobel (commented):
First-pass review. Overall this is a strong Gen-3 implementation — dual-family routing through a single `_VERSION_CONFIG` table, bulk create/delete with per-item 207 inspection, explicit pagination, and tenant-aware identity are all nice work, and the test coverage is extensive.

I left 10 inline comments, severity-tagged (3 **High**, 3 **Medium**, 2 **Medium-Low/Low**, 2 **Convention**). I checked them against the existing threads from @akinross and @mikewiebe to avoid re-raising anything already discussed.

A few **minor / take-or-leave** items not worth their own threads:

- `query_one()` (orchestrator line 348) returns the raw GET body without injecting `ipVersion`, unlike `query_all()` and `_query_one_existing()`. Nothing calls it today, but the first consumer that round-trips it through `PrefixListModel.from_response()` (e.g. future `gathered` support, #340) gets a `ValidationError` since `ip_version` is required. One-line fix while you're in there.
- `_endpoint_classes_for_version` (line 150) is documented as a "backwards-compatible accessor", but the file is new in this PR so there is nothing to be compatible with — suggest deleting the alias and pointing the unit tests at `_config_for_version`.
- `_raise_on_207_failures` is now the third hand-rolled 207 parser (after `maintenance_mode` and `fabric_update_group`, with subtle behavioral differences between the three). Same for the `fabric_name` property / `_configure_endpoint` pattern. Fine for this PR — flagging as a shared-helper refactor candidate for `NDBaseOrchestrator`.
- `PrefixListsListEndpointParams` hand-declares `filter`/`max`/`offset`/`sort` instead of composing the existing `LuceneQueryParams` (as `manage_switches` / `manage_fabrics_vpc_pairs` do). The inherited default `to_query_string()` does not URL-encode values, so if `filter` is ever populated (see the scoped-query comment) a filter containing spaces or special characters produces a broken query string.
- Question: `normalize_prefix` uses `ip_network(value, strict=False)`, so `10.1.1.5/24` is silently rewritten to `10.1.1.0/24` and a bare `10.1.1.5` becomes a `/32` host route. In a routing-policy object that silent rewrite can change intent — worth either rejecting host bits or documenting the normalization in the `prefix` option docs.

plugins/module_utils/models/manage_prefix_list/manage_prefix_list.py
```diff
@@ -0,0 +1,424 @@
...
+    )
+
+    name: str = Field(
+        alias="name",
+        min_length=1,
+        max_length=115,
+        description="Name of the prefix list (pattern: ^[a-zA-Z0-9~_-]+$).",
+    )
+
+    description: AsciiDescription = Field(
```
allenrobel, line 200: **High — `replaced` cannot clear an optional field that is omitted from config.**

`NDBaseModel.to_diff_dict()` always dumps with `exclude_none=True`. For `replaced`, the state machine calls `get_diff_config(proposed, exclude_unset=False)`, but the omitted `description` is `None` and is therefore still dropped from the proposed side of the comparison. With identical entries, `issubset(proposed, existing)` then returns `no_diff` → no PUT is sent, `changed=false`, and the stale description persists on ND — contradicting the documented `replaced` contract ("replace the prefix lists specified in the configuration").

Repro: create a prefix list with `description: seed`, then run `state: replaced` with the same entries and no `description` → module reports unchanged; ND keeps `seed`.

This is the mechanism underneath the behavior @mikewiebe observed live in the `replaced.yaml` thread (his comment addresses the test assertion; this comment is about the module never sending the clearing PUT when entries are unchanged). The clean fix is in the shared diff layer (replaced-mode comparison should treat proposed-None as "must be absent/empty" rather than "ignore"), which may be out of scope for this PR — if so, please document the limitation in the module docs so users know `replaced` cannot unset `description`.

plugins/module_utils/orchestrators/manage_prefix_list.py
```diff
@@ -0,0 +1,404 @@
...
+        result = self._request(path=api_endpoint.path, verb=api_endpoint.verb, not_found_ok=True)
+        if not result:
+            return None
+        result["ipVersion"] = version
+        return result
+
+    def _query_proposed_existing(self) -> list[dict[str, Any]]:
+        """Fetch only the existing prefix lists named by module ``config``."""
+        results: list[dict[str, Any]] = []
+        for version, tenant_name, name in self._proposed_identifiers():
```
allenrobel: **Medium — the scoped preflight is an N+1 request pattern with no size cutoff.**

`_query_proposed_existing` issues one GET per unique config identifier, and `_should_use_scoped_query` has no threshold — so a `merged` run creating 50 new prefix lists performs 50 sequential (mostly-404) GETs before the 1–2 bulk POSTs. That is strictly more requests than the unscoped paginated scan it bypasses (~1–2 GETs per address family for fabrics under `_QUERY_PAGE_SIZE` per family) as soon as config exceeds a couple of items.

Two cheaper options:

1. `PrefixListsListEndpointParams` already models a Lucene `filter` param that is never set — one filtered list GET per address family present in config would replace the whole loop (worth confirming the API supports name-based filters here).
2. Failing that, fall back to the unscoped scan whenever `len(identifiers)` exceeds a small threshold, bounding init cost at `min(scoped, unscoped)`.

Related micro-issue in the same path: `_proposed_identifiers` builds a full `PrefixListModel.from_config(item)` — running the entire validation chain including per-entry `ipaddress` parsing — just to read `(ip_version, tenant_name, name)`, and `NDStateMachine` immediately re-validates the same items for `self.proposed`. Reading the three keys straight from the dict (aliases are already normalized by the argspec) avoids validating every config item twice per run.

plugins/modules/nd_manage_prefix_list.py
```diff
@@ -0,0 +1,311 @@
...
+            - This value is sent to the API as a separate explicit match mask and is not deduced from O(config.entries.prefix).
+            - Must be a valid IPv4 address when O(config.ip_version=ipv4).
+            - Must be a valid IPv6 address when O(config.ip_version=ipv6).
+            type: str
+  state:
+    description:
+    - The desired state of the prefix list resources on Cisco Nexus Dashboard.
+    - Use O(state=merged) to create new prefix lists and update existing ones
+      as defined in the configuration.
+      Prefix lists on ND that are not specified in the configuration are left unchanged.
```
allenrobel, line 151: **Medium — `merged` replaces the `entries` list wholesale; the docs read as additive.**

`NDBaseModel.merge()` only recurses into nested `NDBaseModel` fields; `entries` is a plain list, so it is replaced via `setattr`. Given existing entries `[10, 20, 30]`, a `merged` task supplying only a new entry `[40]` produces a PUT whose body contains only `[40]` — deleting the other three entries under a state whose docs say unspecified config is "left unchanged". That is true at the prefix-list level but false at the entry level, and for this resource `entries` is effectively the whole object.

This is the framework-wide Gen-3 list semantic rather than something this PR introduced, but since `entries` is mandatory-and-authoritative here, please state explicitly in the `merged` description (and the `entries` option docs) that the entry list is always replaced in full — users must restate every entry they want to keep. Entry-level merging keyed on `sequence_number` would be the nicer long-term behavior if the team wants it.

plugins/module_utils/orchestrators/manage_prefix_list.py
```diff
@@ -0,0 +1,404 @@
...
+
+    def _query_one_existing(self, version: str, tenant_name: str | None, name: str) -> dict[str, Any] | None:
+        """Fetch one existing prefix list, returning ``None`` when it is absent."""
+        config = self._config_for_version(version)
+        api_endpoint = self._configure_endpoint(config["get"]())
+        api_endpoint.set_identifiers((version, tenant_name, name))
+        result = self._request(path=api_endpoint.path, verb=api_endpoint.verb, not_found_ok=True)
+        if not result:
+            return None
+        result["ipVersion"] = version
```
allenrobel, line 234: **Medium — tenant-scoped wire paths are unverified and have no fixture coverage.**

Two assumptions about the wire that decide whether tenant-scoped idempotency works at all:

1. `_query_one_existing` injects only `ipVersion` and relies on the item GET echoing `tenantName` so `normalize_tenant_scoped_name` can strip the `tenant~` prefix from the echoed name. If the GET omits `tenantName`, the existing item deserializes with identifier `("ipv4", None, "TENANT1~PL-1")` while proposed is `("ipv4", "TENANT1", "PL-1")` — the lookup misses, so `merged` re-creates an existing list (surfacing as a 207 per-item failure) and `deleted` never deletes. Since the orchestrator already knows the tenant it queried with, injecting `result["tenantName"] = tenant_name` alongside `ipVersion` removes the dependency on the echo entirely.
2. `update()` PUTs to the tenant-qualified path (`.../ipv4PrefixLists/TENANT1~PL-1`) while `to_payload()` sends the bare `name` plus `tenantName` in the body — whether ND accepts that body shape on PUT (vs. requiring the qualified name to match the path) is unverified.

`test_manage_prefix_list.json` contains no tenant-scoped response shapes, so neither path is unit-tested. Could you confirm both against the lab/OpenAPI and add tenant-scoped fixtures (item GET echo shape + update round-trip)?

plugins/module_utils/models/types.py
```diff
@@ -17,6 +17,10 @@
   UTF-8 input. Catching this client-side gives users a clear error instead of a confusing server fault.
 - `IPv4CIDR` — `str | None` validated as an IPv4 interface address in CIDR notation (e.g. `10.1.1.1/32`).
 - `IPv6CIDR` — `str | None` validated as an IPv6 interface address in CIDR notation (e.g. `2001:db8::1/128`).
+- `IPv4NetworkCIDR` — `str | None` validated as an IPv4 network prefix in CIDR notation (e.g. `10.1.0.0/16`).
```
allenrobel: **Medium-Low — these four documented types don't exist.**

This docstring addition is the only change to the shared `types.py`, and it advertises `IPv4NetworkCIDR`, `IPv6NetworkCIDR`, `IPv4AddressType`, and `IPv6AddressType` under "Available types" — but none of them are defined in this file (or anywhere in the repo). The corresponding validation actually lives as inline `normalize_prefix`/`normalize_mask` validators inside `PrefixListEntryModel`. The next feature that follows these docs (`from ...models.types import IPv4NetworkCIDR`) gets an `ImportError`.

Either implement the four Annotated types here (mirroring the existing `IPv4CIDR` pattern) and use them in `PrefixListEntryModel` — the nicer outcome, since route-maps/community-lists will want the same validation — or drop these doc lines.

plugins/module_utils/models/manage_prefix_list/manage_prefix_list.py
```diff
@@ -0,0 +1,424 @@
...
+
+    ip_version: IpVersionEnum = Field(
+        alias="ipVersion",
+        description="IP version of this prefix list: 'ipv4' or 'ipv6'. " "This is an Ansible-only field and is not sent to the API.",
+    )
+
+    name: str = Field(
+        alias="name",
+        min_length=1,
+        max_length=115,
```
allenrobel, line 196: **Low — name length limit isn't conditioned on tenant as the docs describe.**

The module docs say "Maximum length is 115 characters (63 for the default tenant)", but the model enforces `max_length=115` unconditionally. A default-tenant name of 64–115 characters passes client-side validation and only fails server-side, surfacing as a late 207 per-item error instead of the early, clear `ValueError` the model layer exists to provide. A small cross-field check (limit 63 when `tenant_name` is unset) closes the gap — and pairs naturally with the combined `tenant~name` length check suggested on the endpoint `set_identifiers` comment.

plugins/modules/nd_manage_prefix_list.py
```diff
@@ -0,0 +1,311 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2026, Gaspard Micol (@gmicol) <gmicol@cisco.com>
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
```
allenrobel: **Convention — Python 2 boilerplate in new files.**

`from __future__ import absolute_import, division, print_function` and `__metaclass__ = type` (line 9) are no longer needed — only the `annotations` future import is worth keeping where used. The same boilerplate appears in the three new unit-test files (`tests/unit/module_utils/models/test_manage_prefix_list.py:16-18`, `orchestrators/test_manage_prefix_list.py:21-23`, `endpoints/test_endpoints_api_v1_manage_prefix_lists.py:15-17`) — suggest dropping it across the PR so the new files don't need a later sweep.

plugins/module_utils/models/manage_prefix_list/manage_prefix_list.py
```diff
@@ -0,0 +1,424 @@
...
+# {"name": "PL-IPV4-1", "entries": [{"sequenceNumber": 10, "action": "permit",
+#   "prefix": "10.0.0.0/8"}, ...]}
+```
+"""
+
+from __future__ import annotations
+
+import ipaddress
+import re
+from typing import Any, ClassVar, Dict, List, Literal, Optional, Set
```
allenrobel: **Convention — legacy typing forms in new code.**

New code should use the modern PEP 585/604 forms: `X | None` instead of `Optional[X]`, and `list[X]`/`dict[K, V]`/`set[X]` instead of `List`/`Dict`/`Set`. This file imports and uses `Optional`, `Dict`, `List`, and `Set` throughout (~15 `Optional[...]` fields, `ClassVar[Optional[List[str]]]`, `Set[str]`, `List[Dict[str, Any]]`) even though it already has `from __future__ import annotations`; `endpoints/v1/manage/manage_prefix_lists.py` does the same for its six `Optional[...]` query params. The PR's own `mixins.py` addition already uses the modern `str | None` form — suggest converting the other new files to match.

gmicol:
> First-pass review. Overall this is a strong Gen-3 implementation — dual-family routing through a single `_VERSION_CONFIG` table, bulk create/delete with per-item 207 inspection, explicit pagination, and tenant-aware identity are all nice work, and the test coverage is extensive.
> 
> I left 10 inline comments, severity-tagged (3 **High**, 3 **Medium**, 2 **Medium-Low/Low**, 2 **Convention**). I checked them against the existing threads from @akinross and @mikewiebe to avoid re-raising anything already discussed.
> 
> A few **minor / take-or-leave** items not worth their own threads:
> 
> * `query_one()` (orchestrator line 348) returns the raw GET body without injecting `ipVersion`, unlike `query_all()` and `_query_one_existing()`. Nothing calls it today, but the first consumer that round-trips it through `PrefixListModel.from_response()` (e.g. future `gathered` support, #340) gets a `ValidationError` since `ip_version` is required. One-line fix while you're in there.
> * `_endpoint_classes_for_version` (line 150) is documented as a "backwards-compatible accessor", but the file is new in this PR so there is nothing to be compatible with — suggest deleting the alias and pointing the unit tests at `_config_for_version`.
> * `_raise_on_207_failures` is now the third hand-rolled 207 parser (after `maintenance_mode` and `fabric_update_group`, with subtle behavioral differences between the three). Same for the `fabric_name` property / `_configure_endpoint` pattern. Fine for this PR — flagging as a shared-helper refactor candidate for `NDBaseOrchestrator`.
> * `PrefixListsListEndpointParams` hand-declares `filter`/`max`/`offset`/`sort` instead of composing the existing `LuceneQueryParams` (as `manage_switches` / `manage_fabrics_vpc_pairs` do). The inherited default `to_query_string()` does not URL-encode values, so if `filter` is ever populated (see the scoped-query comment) a filter containing spaces or special characters produces a broken query string.
> * Question: `normalize_prefix` uses `ip_network(value, strict=False)`, so `10.1.1.5/24` is silently rewritten to `10.1.1.0/24` and a bare `10.1.1.5` becomes a `/32` host route. In a routing-policy object that silent rewrite can change intent — worth either rejecting host bits or documenting the normalization in the `prefix` option docs.

Thanks @allenrobel , I handled the local items from this note in the latest commits:

- `query_one()` now injects `ipVersion`, and `tenantName` when the model is tenant-scoped.
- Removed `_endpoint_classes_for_version` and updated tests to use `_config_for_version`.
- Moved list filtering/pagination/sort to `LuceneQueryParams` composition so filter values are encoded consistently.
- Documented the prefix normalization behavior: host bits are masked to the containing network, and bare host addresses become host routes.

I intentionally left the shared `_raise_on_207_failures` / `fabric_name` / `_configure_endpoint` helper refactor out of this PR. That would affect multiple orchestrators (this one, `maintenance_mode`, `fabric_update_group`, etc), so it is better handled as a separate framework cleanup issue.

plugins/modules/nd_manage_prefix_list.py
```diff
@@ -0,0 +1,319 @@
...
+      - ip_version: ipv4
+        name: PL-IPV4-FINAL
+        entries:
+          - sequence_number: 10
+            action: permit
+            prefix: 10.0.0.0/8
+    state: overridden
+"""
+
+RETURN = r"""
```
mikewiebe, line 244: ### Low: The public result contract is undocumented

**Issue**

The module defines an empty `RETURN` block even though it returns a structured state-machine result.

**Evidence**

- [`plugins/modules/nd_manage_prefix_list.py` line 245](https://github.com/CiscoDevNet/ansible-nd/pull/286/files#diff-ea3b3481191da6c6e2cf7b7c49fb4136069bbac14286e88d0bd37f544bd9da62R245) defines an empty `RETURN` string.
- The wrapper exits with `nd_state_machine.output.format()`, which exposes the standard state-machine result fields to callers even though none are documented by this module.
- The established module-family example, `nd_interface_port_channel_access.py`, documents `changed`, `output_level`, `before`, `after`, `diff`, `proposed`, `logs`, and failure `msg`.

**Existing PR overlap**

No matching open issue found.

**Impact**

Users and downstream automation have no published contract for the fields returned by this module or for when those fields are present.

**Suggested fix**

Document the standard ND result keys and failure shape in `RETURN`, including `changed`, `output_level`, `before`, `after`, `diff`, `proposed`, `logs`, and failure `msg`, with their descriptions, return conditions, types, elements, and representative samples.

---

gmicol: Added, thanks Mike!

mikewiebe (approved):
LGTM

akinross (approved):

mikewiebe:
Merging now with two required approvers

PR #287, opened 2026-05-20
[ND 4.2+] Add modules for both Routing Policies Community Lists and Extended Community Lists (NDA-28)

### LINKED ISSUE #241: [ND 4.2+] Add Community Lists Module

```shell
A new module to support ND 4.2+ API Community Lists is required.
```

manage

```shell

```

```yaml
TBD
```

### Ansible Version

```shell

```

```shell

```

```shell
4.2.1
```

### Cisco NX-OS Version

```shell

```

### Cisco ACI Version

```shell

```

### LINKED ISSUE #242: [ND 4.2+] Add Extended Community Lists Module

```shell
A new module to support ND 4.2+ API Extended Community Lists is required.
```

manage

```shell

```

```yaml
TBD
```

### Ansible Version

```shell

```

```shell

```

```shell
4.2.1
```

### Cisco NX-OS Version

```shell

```

### Cisco ACI Version

```shell

```

gmicol:
## Related Issue(s)

Fixes #241
Fixes #242

## Proposed Changes

- Add `nd_manage_community_list` for ND 4.2+ Routing Policies Community Lists.
- Add `nd_manage_extended_community_list` for ND 4.2+ Routing Policies Extended Community Lists.
- Implement Manage API endpoints, models, enums, and orchestrators using the current Orchestrators framework and `NDStateMachine`.
- Support `merged`, `replaced`, `overridden`, and `deleted` states.
- Add module documentation, examples, return data, and check mode support.
- Add unit coverage for endpoints, models, and orchestrators.
- Add integration test targets for both modules.

## Test Notes

- Ran focused unit tests for the new endpoint, model, and orchestrator coverage: `30 passed`.

Developed against Cisco Nexus Dashboard 4.2.1 / ND 4.2+.

* [ ] analyze
* [ ] infa
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

### DIFF
--- a/plugins/module_utils/endpoints/v1/manage/manage_community_lists.py
+++ b/plugins/module_utils/endpoints/v1/manage/manage_community_lists.py
@@ -27,8 +27,6 @@ in the ND Manage API.
 
 from __future__ import annotations
 
-__metaclass__ = type
-
 from typing import ClassVar, Literal
 from urllib.parse import quote
 
--- a/plugins/module_utils/endpoints/v1/manage/manage_community_lists.py
+++ b/plugins/module_utils/endpoints/v1/manage/manage_community_lists.py
@@ -32,24 +32,17 @@ from urllib.parse import quote
 
 from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import ConfigDict, Field
 from ansible_collections.cisco.nd.plugins.module_utils.endpoints.base import NDEndpointBaseModel
-from ansible_collections.cisco.nd.plugins.module_utils.endpoints.mixins import CommunityListNameMixin, FabricNameMixin
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.mixins import ClusterNameMixin, CommunityListNameMixin, FabricNameMixin
 from ansible_collections.cisco.nd.plugins.module_utils.endpoints.query_params import CompositeQueryParams, EndpointQueryParams, LuceneQueryParams
 from ansible_collections.cisco.nd.plugins.module_utils.endpoints.v1.manage.base_path import BasePath
 from ansible_collections.cisco.nd.plugins.module_utils.enums import HttpVerbEnum
 from ansible_collections.cisco.nd.plugins.module_utils.types import IdentifierKey
 
 
-class CommunityListsEndpointParams(EndpointQueryParams):
+class CommunityListsEndpointParams(ClusterNameMixin, EndpointQueryParams):
     """Query parameters common to community list endpoints."""
 
     model_config = ConfigDict(extra="forbid")
-
-    cluster_name: str | None = Field(
-        default=None,
-        min_length=1,
-        description="Name of the target Nexus Dashboard cluster in a multi-cluster deployment",
-    )
-
 
 class _EpManageCommunityListsBase(FabricNameMixin, CommunityListNameMixin, NDEndpointBaseModel):
     """
--- a/plugins/module_utils/endpoints/v1/manage/manage_community_lists.py
+++ b/plugins/module_utils/endpoints/v1/manage/manage_community_lists.py
@@ -44,6 +44,7 @@ class CommunityListsEndpointParams(ClusterNameMixin, EndpointQueryParams):
 
     model_config = ConfigDict(extra="forbid")
 
+
 class _EpManageCommunityListsBase(FabricNameMixin, CommunityListNameMixin, NDEndpointBaseModel):
     """
     Base class for ND Manage Community Lists endpoints.
--- a/plugins/module_utils/endpoints/v1/manage/manage_community_lists.py
+++ b/plugins/module_utils/endpoints/v1/manage/manage_community_lists.py
@@ -1,0 +1,254 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2026, Gaspard Micol (@gmicol) <gmicol@cisco.com>
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+"""
+ND Manage Community Lists endpoint models.
+
+This module contains endpoint definitions for community list operations
+in the ND Manage API.
+
+## Endpoints
+
+- `EpManageCommunityListsGet` - Get a specific community list by name
+  (GET /api/v1/manage/fabrics/{fabricName}/communityLists/{communityListName})
+- `EpManageCommunityListsListGet` - List all community lists for a fabric
+  (GET /api/v1/manage/fabrics/{fabricName}/communityLists)
+- `EpManageCommunityListsPost` - Bulk create community lists
+  (POST /api/v1/manage/fabrics/{fabricName}/communityLists)
+- `EpManageCommunityListsPut` - Update a specific community list
+  (PUT /api/v1/manage/fabrics/{fabricName}/communityLists/{communityListName})
+- `EpManageCommunityListsDelete` - Delete a specific community list
+  (DELETE /api/v1/manage/fabrics/{fabricName}/communityLists/{communityListName})
+- `EpManageCommunityListsBulkDelete` - Bulk delete community lists via action endpoint
+  (POST /api/v1/manage/fabrics/{fabricName}/communityListActions/remove)
+"""
+
+from __future__ import annotations
+
+__metaclass__ = type
+
+from typing import ClassVar, Literal
+from urllib.parse import quote
+
+from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import Field
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.base import NDEndpointBaseModel
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.mixins import CommunityListNameMixin, FabricNameMixin, FromClusterMixin
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.query_params import CompositeQueryParams, EndpointQueryParams, LuceneQueryParams
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.v1.manage.base_path import BasePath
+from ansible_collections.cisco.nd.plugins.module_utils.enums import HttpVerbEnum
+from ansible_collections.cisco.nd.plugins.module_utils.types import IdentifierKey
+
+
+class CommunityListsEndpointParams(FromClusterMixin, EndpointQueryParams):
+    """Query parameters common to community list endpoints."""
+
+
+class _EpManageCommunityListsBase(FabricNameMixin, CommunityListNameMixin, NDEndpointBaseModel):
+    """
+    Base class for ND Manage Community Lists endpoints.
+
+    Provides common functionality for all HTTP methods on the
+    /api/v1/manage/fabrics/{fabricName}/communityLists endpoint.
+
+    Subclasses may override:
+    - ``_require_community_list_name``: set to ``False`` for collection-level
+      endpoints (list, bulk create) that do not include the list name in the path.
+    """
+
+    _require_community_list_name: ClassVar[bool] = True
+
+    endpoint_params: CommunityListsEndpointParams = Field(
+        default_factory=CommunityListsEndpointParams,
+        description="Endpoint-specific query parameters",
+    )
+
+    def set_identifiers(self, identifier: IdentifierKey = None):
+        """Assign the community list name from the model identifier."""
+        self.community_list_name = identifier
+
+    @property
+    def path(self) -> str:
+        """
+        Build the endpoint path for community list operations.
+
+        Raises:
+            ValueError: if ``fabric_name`` is not set.
+            ValueError: if ``community_list_name`` is required but not set.
+        """
+        if self.fabric_name is None:
+            raise ValueError(f"{type(self).__name__}.path: fabric_name must be set before accessing path.")
+        if self._require_community_list_name and self.community_list_name is None:
+            raise ValueError(f"{type(self).__name__}.path: community_list_name must be set before accessing path.")
+        segments = ["fabrics", quote(self.fabric_name, safe=""), "communityLists"]
+        if self._require_community_list_name and self.community_list_name is not None:
+            segments.append(quote(self.community_list_name, safe=""))
+        base_path = BasePath.path(*segments)
+        query_params = CompositeQueryParams().add(self.endpoint_params)
+        lucene_params = getattr(self, "lucene_params", None)
+        if lucene_params is not None:
+            query_params.add(lucene_params)
+        query_string = query_params.to_query_string()
+        if query_string:
+            return f"{base_path}?{query_string}"
+        return base_path
+
+
+class EpManageCommunityListsGet(_EpManageCommunityListsBase):
+    """
+    ND Manage Community Lists GET Endpoint.
+
+    Retrieves details for a specific community list by name.
+
+    Path: GET /api/v1/manage/fabrics/{fabricName}/communityLists/{communityListName}
+    """
+
+    class_name: Literal["EpManageCommunityListsGet"] = Field(
+        default="EpManageCommunityListsGet",
+        frozen=True,
+        description="Class name for backward compatibility",
+    )
+
+    @property
+    def verb(self) -> HttpVerbEnum:
+        """Return the HTTP verb for this endpoint."""
+        return HttpVerbEnum.GET
+
+
+class EpManageCommunityListsListGet(_EpManageCommunityListsBase):
+    """
+    ND Manage Community Lists List GET Endpoint.
+
+    Retrieves all community lists for a given fabric.
+
+    Path: GET /api/v1/manage/fabrics/{fabricName}/communityLists
+    """
+
+    _require_community_list_name: ClassVar[bool] = False
+
+    class_name: Literal["EpManageCommunityListsListGet"] = Field(
+        default="EpManageCommunityListsListGet",
+        frozen=True,
+        description="Class name for backward compatibility",
+    )
+    lucene_params: LuceneQueryParams = Field(default_factory=LuceneQueryParams, description="Lucene-style query parameters for filtering")
+
+    @property
+    def verb(self) -> HttpVerbEnum:
+        """Return the HTTP verb for this endpoint."""
+        return HttpVerbEnum.GET
+
+
+class EpManageCommunityListsPost(_EpManageCommunityListsBase):
+    """
+    ND Manage Community Lists POST Endpoint.
+
+    Bulk-creates one or more community lists.
+    Request body: ``{"communityLists": [...]}``
+
+    Path: POST /api/v1/manage/fabrics/{fabricName}/communityLists
+    """
+
+    _require_community_list_name: ClassVar[bool] = False
+
+    class_name: Literal["EpManageCommunityListsPost"] = Field(
+        default="EpManageCommunityListsPost",
+        frozen=True,
+        description="Class name for backward compatibility",
+    )
+
+    @property
+    def verb(self) -> HttpVerbEnum:
+        """Return the HTTP verb for this endpoint."""
+        return HttpVerbEnum.POST
+
+
+class EpManageCommunityListsPut(_EpManageCommunityListsBase):
+    """
+    ND Manage Community Lists PUT Endpoint.
+
+    Updates an existing community list.
+
+    Path: PUT /api/v1/manage/fabrics/{fabricName}/communityLists/{communityListName}
+    """
+
+    class_name: Literal["EpManageCommunityListsPut"] = Field(
+        default="EpManageCommunityListsPut",
+        frozen=True,
+        description="Class name for backward compatibility",
+    )
+
+    @property
+    def verb(self) -> HttpVerbEnum:
+        """Return the HTTP verb for this endpoint."""
+        return HttpVerbEnum.PUT
+
+
+class EpManageCommunityListsDelete(_EpManageCommunityListsBase):
+    """
+    ND Manage Community Lists DELETE Endpoint.
+
+    Deletes a specific community list by name.
+
+    Path: DELETE /api/v1/manage/fabrics/{fabricName}/communityLists/{communityListName}
+    """
+
+    class_name: Literal["EpManageCommunityListsDelete"] = Field(
+        default="EpManageCommunityListsDelete",
+        frozen=True,
+        description="Class name for backward compatibility",
+    )
+
+    @property
+    def verb(self) -> HttpVerbEnum:
+        """Return the HTTP verb for this endpoint."""
+        return HttpVerbEnum.DELETE
+
+
+class _EpManageCommunityListsBulkDeleteBase(FabricNameMixin, NDEndpointBaseModel):
+    """
+    Base class for the community list bulk-delete action endpoint.
+
+    Path: /api/v1/manage/fabrics/{fabricName}/communityListActions/remove
+    """
+
+    endpoint_params: CommunityListsEndpointParams = Field(
+        default_factory=CommunityListsEndpointParams,
+        description="Endpoint-specific query parameters",
+    )
+
+    def set_identifiers(self, identifier: IdentifierKey = None):
+        """No-op: bulk delete endpoint has no per-resource identifier in the path."""
+
+    @property
+    def path(self) -> str:
+        if self.fabric_name is None:
+            raise ValueError(f"{type(self).__name__}.path: fabric_name must be set before accessing path.")
+        base_path = BasePath.path("fabrics", quote(self.fabric_name, safe=""), "communityListActions", "remove")
+        query_string = self.endpoint_params.to_query_string()
+        if query_string:
+            return f"{base_path}?{query_string}"
+        return base_path
+
+
+class EpManageCommunityListsBulkDelete(_EpManageCommunityListsBulkDeleteBase):
+    """
+    ND Manage Community Lists Bulk Delete Endpoint.
+
+    Bulk-deletes community lists by name.
+    Request body: ``{"communityListNames": [...]}``
+
+    Path: POST /api/v1/manage/fabrics/{fabricName}/communityListActions/remove
+    """
+
+    class_name: Literal["EpManageCommunityListsBulkDelete"] = Field(
+        default="EpManageCommunityListsBulkDelete",
+        frozen=True,
+        description="Class name for backward compatibility",
+    )
+
+    @property
+    def verb(self) -> HttpVerbEnum:
+        """Return the HTTP verb for this endpoint."""
+        return HttpVerbEnum.POST
--- a/plugins/module_utils/endpoints/v1/manage/manage_community_lists.py
+++ b/plugins/module_utils/endpoints/v1/manage/manage_community_lists.py
@@ -30,17 +30,25 @@ from __future__ import annotations
 from typing import ClassVar, Literal
 from urllib.parse import quote
 
-from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import Field
+from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import ConfigDict, Field
 from ansible_collections.cisco.nd.plugins.module_utils.endpoints.base import NDEndpointBaseModel
-from ansible_collections.cisco.nd.plugins.module_utils.endpoints.mixins import CommunityListNameMixin, FabricNameMixin, FromClusterMixin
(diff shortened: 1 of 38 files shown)


plugins/module_utils/models/manage_community_list/manage_community_list.py
```diff
@@ -0,0 +1,343 @@
...
+    Model for a Nexus Dashboard community list resource.
+
+    Supports both **standard** and **expanded** types; the ``type`` field
+    discriminates between them and a ``model_validator`` enforces the
+    type-specific entry constraints.
+
+    Identifier strategy: ``single`` on ``name``.
+    """
+
+    identifiers: ClassVar[Optional[List[str]]] = ["name"]
```
mikewiebe: ### High: tenant-scoped lists can collide or miss update/delete because identity is only `name`

**Issue**

Both modules expose `tenant_name`, but their model identity and API path/delete handling still use only bare `name`.

**Evidence**

- [`plugins/module_utils/models/manage_community_list/manage_community_list.py` line 167](https://github.com/CiscoDevNet/ansible-nd/pull/287/files#diff-0ebc395fb45a8af5ad591525dc39172db503a628f34e5deeae6e8640471aaa54R167) defines community-list identity as only `["name"]`, even though the same model exposes `tenant_name`.
- [`plugins/module_utils/models/manage_extended_community_list/manage_extended_community_list.py` line 222](https://github.com/CiscoDevNet/ansible-nd/pull/287/files#diff-bd87bab5b432d7626107b5f97385766a41cc715cbe6d884131473d818c98686bR222) uses the same bare-name identity for extended community lists.
- [`plugins/module_utils/orchestrators/manage_community_list.py` line 204](https://github.com/CiscoDevNet/ansible-nd/pull/287/files#diff-eb36e44d2e2ab0a872a0eceeb3d7c6c7b076d6a8660c028466c5922242c3362eR204) sends only `m.name` in the bulk delete payload; the extended-community orchestrator does the same at [`plugins/module_utils/orchestrators/manage_extended_community_list.py` line 208](https://github.com/CiscoDevNet/ansible-nd/pull/287/files#diff-d19c782e7e1480d7b673636fcebb55e2999dda6a03a0cf5c4251c64e50262af7R208).

**Existing PR overlap**

**Impact**

A list created with bare `name` plus `tenant_name` can later be returned by ND as `tenant~name`, so idempotency, update, and delete can miss the existing object. Same-named lists in different tenants can also collide or target the wrong logical resource.

**Suggested fix**

Mirror the current prefix-list pattern from PR #286: normalize `tenant~name`, make tenant part of identity, add an `api_name`, and use that API name for get/update/delete. Add tests for same-name objects in different tenants.

---

plugins/module_utils/orchestrators/manage_community_list.py
```diff
@@ -0,0 +1,235 @@
...
+
+        POST /fabrics/{fabricName}/communityLists
+        Body: {"communityLists": [...]}
+        """
+        try:
+            for model_instance in model_instances:
+                self._validate_write_model(model_instance)
+            ep = self._configure_endpoint(self.create_bulk_endpoint())
+            payload = {"communityLists": [m.to_payload() for m in model_instances]}
+            result = self._request(path=ep.path, verb=ep.verb, data=payload)
```
mikewiebe: ### Medium: mutating API calls are recorded as query operations

**Issue**

The custom create/update/delete calls do not pass an explicit `OperationType`, so they default to `QUERY`.

**Evidence**

- [`plugins/module_utils/orchestrators/manage_community_list.py` line 174](https://github.com/CiscoDevNet/ansible-nd/pull/287/files#diff-eb36e44d2e2ab0a872a0eceeb3d7c6c7b076d6a8660c028466c5922242c3362eR174) calls `_request()` for create without `operation_type`; update and delete follow the same pattern at lines 187 and 205.
- [`plugins/module_utils/orchestrators/manage_extended_community_list.py` line 178](https://github.com/CiscoDevNet/ansible-nd/pull/287/files#diff-d19c782e7e1480d7b673636fcebb55e2999dda6a03a0cf5c4251c64e50262af7R178) repeats the same pattern for extended community create; update and delete follow at lines 191 and 209.
- The shared `_request()` default is `OperationType.QUERY`, so these mutating requests are recorded as reads unless overridden.

**Existing PR overlap**

**Impact**

Create, update, and delete calls are reported in results/API history as read/query operations, which makes troubleshooting and verbose output misleading. This same pattern was fixed in PR #286 for `nd_manage_prefix_list`, but the fix is local to that module's orchestrator; PR #287 still needs the same explicit `OperationType` handling in both community-list orchestrators.

**Suggested fix**

Pass `OperationType.CREATE`, `OperationType.UPDATE`, and `OperationType.DELETE` on the custom write paths, and assert the recorded metadata action in unit tests.

---

plugins/modules/nd_manage_community_list.py
```diff
@@ -0,0 +1,299 @@
...
+"""
+
+EXAMPLES = r"""
+- name: Create a standard community list
+  cisco.nd.nd_manage_community_list:
+    fabric_name: my-fabric
+    config:
+      - name: CL-STANDARD-PROD
+        type: standard
+        description: Production community list
```
mikewiebe: ### Medium: `nd_manage_community_list` example documents unsupported `description`

**Issue**

The community-list example includes a `description` field that the module does not document, validate, or send.

**Evidence**

- [`plugins/modules/nd_manage_community_list.py` line 175](https://github.com/CiscoDevNet/ansible-nd/pull/287/files#diff-d2bdeadcd18af0dec3b0d1a841d5f40d2a9b032f038f925c19d195538a372792R175) shows `description: Production community list` in an example.
- The model and argspec expose `name`, `type`, `tenant_name`, and `entries`; no `description` option is present for this module.

**Existing PR overlap**

**Impact**

Users copying the example can hit an unsupported parameter error, or assume the module manages a field that is silently ignored by the model path.

**Suggested fix**

Remove `description` from the example, or add real `description` support across docs, argspec, model, payload, and tests if the API supports it.

---

plugins/modules/nd_manage_community_list.py
```diff
@@ -0,0 +1,299 @@
...
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+ANSIBLE_METADATA = {"metadata_version": "1.1", "status": ["preview"], "supported_by": "community"}
+
+DOCUMENTATION = r"""
+---
+module: nd_manage_community_list
+version_added: "1.6.0"
```
mikewiebe: ### Low: docs metadata needs cleanup

**Issue**

The new modules advertise the wrong `version_added`, and their `RETURN` blocks are empty.

**Evidence**

- [`plugins/modules/nd_manage_community_list.py` line 17](https://github.com/CiscoDevNet/ansible-nd/pull/287/files#diff-d2bdeadcd18af0dec3b0d1a841d5f40d2a9b032f038f925c19d195538a372792R17) sets `version_added: "1.6.0"`.
- [`plugins/modules/nd_manage_extended_community_list.py` line 17](https://github.com/CiscoDevNet/ansible-nd/pull/287/files#diff-d0a04f25fda276c6717257f12a5954e5c2fc5cb83932d65f4e5f7cdee8c7e9d6R17) also sets `version_added: "1.6.0"`.
- Both module `RETURN` blocks are empty.

**Existing PR overlap**

**Impact**

Generated documentation claims these modules existed in an older collection release and does not tell users which keys to expect from the result.

**Suggested fix**

Update `version_added` to `2.0.0` and document the standard return structure used by these state-machine modules.

tests/integration/targets/nd_manage_community_list/tasks/merged.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - nm_merged_create_expanded.after | selectattr('name', 'equalto', community_list_expanded.name) | list | length == 1
+
+- name: "MERGED IDEMPOTENT: Re-apply standard community list"
+  cisco.nd.nd_manage_community_list: *merge_standard
+  register: nm_merged_idem_standard
+
+- name: "MERGED IDEMPOTENT: Verify no change on second run"
+  ansible.builtin.assert:
+    that:
+      - nm_merged_idem_standard is not changed
```
mikewiebe, line 49: **Issue:** `nd_manage_community_list` is not idempotent for a standard community list in `state: merged`.

I reproduced this live by creating a standard community list and then reapplying the exact same config. The first apply creates the list as expected, but the second apply still reports `changed=True`.

Observed failure:

```text
Expected second apply to be idempotent, got changed=True
```

Minimum repro:

```yaml
- name: Create standard community list
  cisco.nd.nd_manage_community_list:
    config:
      - name: PR287-REPRO-CL-IDEM
        type: standard
        entries:
          - sequence_number: 10
            action: permit
            community_numbers:
              - "100:200"
          - sequence_number: 20
            action: deny
            no_export: true
    state: merged
  register: first_apply

- name: Reapply same standard community list
  cisco.nd.nd_manage_community_list:
    config:
      - name: PR287-REPRO-CL-IDEM
        type: standard
        entries:
          - sequence_number: 10
            action: permit
            community_numbers:
              - "100:200"
          - sequence_number: 20
            action: deny
            no_export: true
    state: merged
  register: second_apply
```

Expected: the second identical `state: merged` apply reports `changed=False`.

Actual: the second identical apply reports `changed=True`.

This also caused the PR-defined integration target for `nd_manage_community_list` to fail on iBGP, eBGP, and external/inter-fabric connectivity fabric types.

plugins/module_utils/models/manage_community_list/manage_community_list.py
```diff
@@ -0,0 +1,343 @@
...
+``model_validator`` on ``CommunityListModel`` enforces the type-specific
+entry constraints.
+"""
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+import re
+from typing import ClassVar, Dict, List, Optional, Set
```
allenrobel: New files should use the modern conventions from the team's code standards: "Modern type annotations (PEP 585 / PEP 604) for all new code" (`X | None` instead of `Optional[X]`, `list[X]`/`dict[K, V]`/`set[X]` instead of `List`/`Dict`/`Set`), "`__metaclass__ = type` is no longer needed and can be removed", and "Boilerplate `from future ...` imports are no longer needed and can be removed (with the exception of annotations)".

This file uses `Optional`/`List`/`Dict`/`Set` throughout (e.g. the field declarations and `-> Dict` on `get_argument_spec`), plus the `__future__` boilerplate and `__metaclass__` above. Same in the extended twin (`manage_extended_community_list.py` lines 31-36). Some of the orchestrator/test files mix both styles (`from __future__ import absolute_import, annotations, division, print_function`) — with `annotations` imported, the PEP 604 syntax is already valid, so the legacy imports can simply be dropped.

plugins/module_utils/models/manage_community_list/manage_community_list.py
```diff
@@ -0,0 +1,343 @@
...
+        """Validate tenant name matches [A-Za-z0-9_-]+."""
+        if v is None:
+            return v
+        pattern = re.compile(r"^[A-Za-z0-9_-]+$")
+        if not pattern.match(str(v)):
+            raise ValueError(f"tenant_name '{v}' contains invalid characters. " "Allowed: [A-Za-z0-9_-].")
+        return v
+
+    @model_validator(mode="after")
+    def validate_entries_match_type(self):
```
allenrobel: This unconditional `@model_validator(mode="after")` also runs when the model is built from ND response data: `NDConfigCollection.from_api_response()` calls `from_response()` for every item `query_all` returns, and the state machine builds its "before" state that way on every run. If ND ever echoes a standard-only field (or `community_number_regex`) as non-None for the other list type, query/merged/replaced/overridden runs will raise a `ValidationError` instead of reporting or remediating the existing object.

We hit this exact defect shape with the storm-control mutex validator (fixed by "Gate storm-control mutex to config input, not ND responses", PR #360 review): the fix threads `context={"mode": "config"|"response"}` through `from_config()`/`from_response()` and skips config-only enforcement when `mode == "response"`. Worth applying the same gate here (and in the extended twin, `manage_extended_community_list.py` line 299) — the PR's `from_response()` fixtures only cover the clean case, so this wouldn't show up in the unit tests.

plugins/module_utils/models/manage_community_list/manage_community_list.py
```diff
@@ -0,0 +1,343 @@
...
+        alias="name",
+        description=(
+            "Name of the community list. Allowed characters: [a-zA-Z0-9~_-]. "
+            "Max 63 characters for the default tenant, 115 when using a "
+            "non-default tenant."
+        ),
+        min_length=1,
+        max_length=115,
+    )
+    type: Optional[CommunityListTypeEnum] = Field(
```
allenrobel: The module DOCUMENTATION describes `type` (and `entries`) as "Required for create and update operations; not required when O(state=deleted)", but nothing enforces that locally: both are plain `Optional` fields with no state-aware validator, and `get_argument_spec()` doesn't mark them required. A merged/replaced/overridden task that omits them passes local validation and only fails at ND (or worse, sends an incomplete payload).

PR #266's ACL model solved this with a `ValidationInfo`-context validator (`validate_required_for_write_states`) that reads `state` from the validation context the state machine now passes to `from_ansible_config()` — after rebasing onto develop, the context will already be available here. Same applies to the extended twin (`manage_extended_community_list.py` line 247).

allenrobel (commented):
### Code review

Reviewed both new module stacks (endpoints, models, orchestrators, tests). Three findings below as inline comments; each applies equally to the extended-community-list twin at the lines noted.

plugins/module_utils/orchestrators/manage_extended_community_list.py
```diff
@@ -0,0 +1,239 @@
...
+
+        ## Raises
+
+        None
+        """
+        api_endpoint.fabric_name = self.fabric_name
+        return api_endpoint
+
+    @staticmethod
+    def _raise_on_207_action_errors(result: Any) -> None:
```
samiib, line 120: A few of these functions exist in both orchestrators and appear to be the same? Would it be possible to encapsulate and share any of the duplicated code?

eg. _raise_on_207_action_errors and create_bulk

samiib (changes requested):

plugins/module_utils/models/manage_community_list/manage_community_list.py
gmicol: Addressed in cac2256. Both models now normalize tenant-qualified names and use a computed `api_name` as their identity. The qualified API name is used for payloads, item operations, and bulk deletes. I also added unit coverage and opt-in integration coverage for same-named lists in different tenants.

plugins/module_utils/orchestrators/manage_community_list.py
gmicol: Addressed in b86e2f6. The custom create, update, and delete requests now pass `OperationType.CREATE`, `OperationType.UPDATE`, and `OperationType.DELETE` explicitly in both orchestrators. Unit tests verify the recorded operation metadata.

plugins/modules/nd_manage_community_list.py
gmicol: Addressed in 49547a5. I removed `description` from the community-list example because it is not supported by the API contract or module argument specification.

plugins/modules/nd_manage_community_list.py
gmicol: Addressed in 49547a5. Both modules now use `version_added: "2.0.0"` and document the standard state-machine return fields with representative samples.

tests/integration/targets/nd_manage_community_list/tasks/merged.yaml
gmicol: Addressed in dfbf508. Diff normalization now excludes controller-added false defaults and empty collections from standard community-list comparisons while still detecting an explicit request to clear an active flag. Regression unit coverage and normal/check-mode integration test definitions were added. Unit tests and integration tests passed.

plugins/module_utils/models/manage_community_list/manage_community_list.py
gmicol: Addressed in 2d779d0 and 1f56c67. The new module stacks now use PEP 585/604 annotations. I removed the obsolete `__metaclass__ = type` declarations, and legacy Python 2 future imports from the production and test files.

plugins/module_utils/models/manage_community_list/manage_community_list.py
gmicol: Addressed in 5b14790. Both models now use Pydantic validation context to distinguish config input from ND response parsing. Type-specific entry validation remains strict for Ansible configuration but is skipped for response data, with regression coverage for controller-added cross-type fields.

plugins/module_utils/models/manage_community_list/manage_community_list.py
gmicol: Addressed in 8443c45. A `ValidationInfo`-based validator now requires `type` and `entries` for merged, replaced, and overridden states. Deleted state and response parsing continue to permit identifier-only models.

plugins/module_utils/orchestrators/manage_extended_community_list.py
gmicol: You are absolutely right. This is actually a broader [issue](https://github.com/CiscoDevNet/ansible-nd/issues/397) that has been raised by Allen. For now, I will keep the duplicated code and will open a new PR to address it for all Routing Policies modules (as it is also a issue with Route Map, ACL and Prefix List orchestrators) once this PR is merged and most importantly, once we agreed on a solution for consolidating per-item 207 multi-status detection if that sounds good to you

sivakasi-cisco:
Kindly ensure the module support for config_actions + resource level deploy https://github.com/CiscoDevNet/ansible-nd/issues/368

samiib (dismissed):

plugins/modules/nd_manage_community_list.py
```diff
@@ -0,0 +1,383 @@
...
+    type: str
+    default: merged
+    choices: [ merged, replaced, overridden, deleted ]
+extends_documentation_fragment:
+- cisco.nd.modules
+- cisco.nd.check_mode
+notes:
+- This module is only supported on Nexus Dashboard version 4.2.1 or higher.
+- Community lists are created and deleted in bulk via separate API endpoints.
+- Standard entries that set none of the well-known community flags and leave
```
mikewiebe: **Medium: The documented flagless standard entry is rejected by Nexus Dashboard**

The documentation says a standard entry with no `community_numbers` and no well-known flags is valid and matches all communities. The model also accepts that configuration:

```yaml
- name: Create a match-all standard entry
  cisco.nd.nd_manage_community_list:
    fabric_name: site-1
    state: merged
    config:
      - name: CL-MATCH-ALL
        type: standard
        entries:
          - sequence_number: 10
            action: permit
```

Live testing of PR #287 on internal iBGP, eBGP, and external/inter-fabric fabrics produced:

```text
Invalid entries: Entry with sequence number 10 must have either
community numbers or at least one well-known community flag enabled
```

This input therefore passes local validation and check mode but fails only when the controller receives the write. The current integration coverage does not test this documented case.

Could you either:

- require at least one `community_numbers` value or supported well-known flag and correct the documentation; or
- identify the Nexus Dashboard version that supports this match-all shape and add live integration coverage proving that contract?

Closed issue [#374](https://github.com/CiscoDevNet/ansible-nd/issues/374) contains the original reproduction and was closed explicitly so this would be handled as a PR review comment.

plugins/module_utils/models/manage_extended_community_list/manage_extended_community_list.py
```diff
@@ -0,0 +1,469 @@
...
+)
+
+# ---------------------------------------------------------------------------
+# Compiled validation regexes (from OpenAPI spec patterns)
+# ---------------------------------------------------------------------------
+
+# MAC: eeee.eeee.eeee  OR  ee:ee:ee:ee:ee:ee  OR  ee-ee-ee-ee-ee-ee
+_ROUTER_MAC_RE = re.compile(r"^(?:[a-fA-F0-9]{4}\.[a-fA-F0-9]{4}\.[a-fA-F0-9]{4}" + r"|([a-fA-F0-9]{2}[:\-]){5}[a-fA-F0-9]{2})$")
+
+# Route target: ASN2:NN  |  ASN4:NN  |  IPv4:NN  (single value, no commas)
```
mikewiebe: **Low/non-blocking: Route-target validation rejects a schema-valid controller value**

The OpenAPI pattern for each `routeTargetCollection` item permits multiple comma-separated route targets, but `_ROUTE_TARGET_RE` accepts only one value.

For example, assume Nexus Dashboard already contains and returns this extended community list:

```json
{
  "name": "RT-IMPORT",
  "type": "standard",
  "entries": [
    {
      "sequenceNumber": 10,
      "action": "permit",
      "routeTargetCollection": [
        "65000:100,65000:200"
      ]
    }
  ]
}
```

A user then attempts to delete it:

```yaml
- name: Delete the existing extended community list
  cisco.nd.nd_manage_extended_community_list:
    fabric_name: site-1
    state: deleted
    config:
      - name: RT-IMPORT
```

Before sending the delete, the state machine queries and parses existing controller state. The field validator tests the complete string against the single-value regex and raises:

```text
route_target_collection entry '65000:100,65000:200'
must be in ASN2:NN, ASN4:NN, or IPv4:NN format
```

The delete request is never sent. The same response would prevent merged, replaced, or overridden reconciliation from starting, even though the user did not supply the comma-packed value.

Could the model either accept the OpenAPI item grammar or normalize the response to `["65000:100", "65000:200"]`? Please cover both playbook input and `from_response()` behavior.

I have not observed a supported ND version actually return this packed form, so this is a defensive schema-compatibility finding unless controller behavior confirms it.

plugins/modules/nd_manage_community_list.py
gmicol: Addressed in 023f6e7. Standard community-list entries supplied through Ansible config must now contain at least one community number or have at least one supported well-known community flag enabled. Empty community-number lists and flags set only to false are rejected locally, including in check mode.

ND response parsing remains permissive so existing controller objects can still be queried or deleted. I also corrected the documentation, added unit coverage for valid and invalid selectors, and added a check-mode integration-test definition.

plugins/module_utils/models/manage_extended_community_list/manage_extended_community_list.py
gmicol: Addressed in 83c2c89. The extended-community model now accepts the comma-packed route-target grammar from OpenAPI for both playbook config and ND responses. Packed items are normalized to a canonical flat list before serialization and reconciliation.

This also makes a packed controller response idempotent with equivalent unpacked playbook config. Malformed tokens, empty tokens, trailing commas, and schema-invalid whitespace remain rejected. Documentation and unit coverage for config parsing, response parsing, payload serialization, invalid input, and diff behavior were added.

samiib (dismissed):

plugins/module_utils/models/manage_community_list/manage_community_list.py
```diff
@@ -0,0 +1,423 @@
...
+        return self.name
+
+    def to_payload(self, **kwargs) -> dict[str, Any]:
+        """Export an API payload with a fully qualified tenant-scoped name."""
+        data = super().to_payload(**kwargs)
+        if self.tenant_name:
+            data["name"] = self.api_name
+        return data
+
+    def to_diff_dict(self, **kwargs) -> dict[str, Any]:
```
allenrobel, line 247: This `to_diff_dict()` override (and its twin in `manage_extended_community_list.py` line 298) works around an ND wire-shape discrepancy: ND echoes `false` well-known-flag defaults (and empty collections, in the extended model) for fields the user never set. Per the team's workaround-marker convention, each such site needs its own `TODO(4.2.1) <slug>` marker so the workaround can be audited and removed later — same pattern as the route-map model's next-hop boolean stripping in `manage_route_map.py`. Until we've found a way to share the bug-tracker data that backs the slugs, please use `TBD` as the slug (as the route-map PR did), e.g.:

```python
# TODO(4.2.1) TBD
# ND echoes false/empty defaults for unset standard-entry fields on GET; strip them so diffs stay idempotent.
```

One marker here and one on the extended twin.

allenrobel (commented):
### Code review (follow-up)

Re-reviewed the commits pushed since my last review. All ten review-response fixes check out as claimed — nice work, especially the tenant-scoped `api_name` identity and the diff-normalization/idempotency mechanics, which I traced end to end. Three additive findings below as inline comments.

plugins/module_utils/models/manage_extended_community_list/manage_extended_community_list.py
```diff
@@ -0,0 +1,476 @@
...
+
+        for i, entry in enumerate(self.entries):
+            if self.type == ExtendedCommunityListTypeEnum.EXPANDED:
+                if not entry.community_number_regex:
+                    raise ValueError(f"entries[{i}].community_number_regex is required " "for type='expanded'.")
+                for field in standard_only_fields:
+                    if getattr(entry, field) is not None:
+                        raise ValueError(f"entries[{i}].{field} must not be set for " "type='expanded'. Use community_number_regex instead.")
+            else:
+                # standard
```
allenrobel, line 397: Follow-up to "Reject flagless standard community entries": that commit added an at-least-one-selector rule to `CommunityListModel` (a standard entry must set at least one community number or well-known flag), but the standard branch here never got the equivalent. All five selector collections default to `None`, so a standard entry with only `sequence_number` + `action` passes local validation and is sent to ND — the same failure mode Mike lab-verified for community lists. The module docs also still say entries "may set any combination of" the collections, with no at-least-one requirement.

Suggest mirroring the community-list fix: for config input, require at least one of the five collections to be non-empty, update the DOCUMENTATION/notes, and add the matching unit coverage. (Caveat: I have not lab-verified that ND rejects the extended variant the way it rejects flagless standard community entries — worth a quick check while you're in there.)

plugins/module_utils/endpoints/v1/manage/manage_community_lists.py
```diff
@@ -0,0 +1,260 @@
...
+from ansible_collections.cisco.nd.plugins.module_utils.enums import HttpVerbEnum
+from ansible_collections.cisco.nd.plugins.module_utils.types import IdentifierKey
+
+
+class CommunityListsEndpointParams(EndpointQueryParams):
+    """Query parameters common to community list endpoints."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    cluster_name: str | None = Field(
```
allenrobel: `cluster_name` is hand-declared here (and in `manage_extended_community_lists.py` line 47) with a definition identical to the existing `ClusterNameMixin` in `endpoints/mixins.py`, whose stated purpose is adding common fields to endpoint models without duplication. The other endpoint-params classes in the collection compose the mixin for this exact field, and `model_config = ConfigDict(extra="forbid")` layers fine on top of it elsewhere. Suggest `class CommunityListsEndpointParams(ClusterNameMixin, EndpointQueryParams)` (and the twin) and dropping the local field.

plugins/module_utils/models/manage_community_list/manage_community_list.py
gmicol: Addressed in 0475ce5. I added a separate `# TODO(4.2.1) TBD` marker at both the community-list and extended-community-list to_diff_dict() workarounds. Each marker documents that ND adds false or empty defaults to GET responses and that these values are temporarily stripped to preserve idempotency.

plugins/module_utils/models/manage_extended_community_list/manage_extended_community_list.py
gmicol: Addressed in c0a3c93. Standard extended-community entries now require at least one non-empty selector collection during configuration validation. Response parsing remains permissive so sparse or ND-defaulted GET responses are still accepted. The module documentation was updated, and unit coverage now includes missing selectors, all-empty selectors, each of the five valid selector collections, and sparse response data.

plugins/module_utils/endpoints/v1/manage/manage_community_lists.py
gmicol: Addressed in 228cdee. Both CommunityListsEndpointParams and ExtendedCommunityListsEndpointParams now inherit from the shared ClusterNameMixin; the duplicate cluster_name fields were removed. extra="forbid" and existing clusterName query serialization remain unchanged, with unit tests covering the shared inheritance contract.

mikewiebe (approved):
LGTM.  Non-blocking request to remove mutation check in a future PR that removes the check from all route-control modules

allenrobel (approved):
Re-approving.

PR #288, opened 2026-05-20
[ND 4.2+] Add module for Routing Policies Route Maps (NDA-29)

### LINKED ISSUE #238: [ND 4.2+] Add Route-Map Module

```shell
A new module to support ND 4.2+ API Route-Maps is required.
```

manage

```shell

```

```yaml
TBD
```

### Ansible Version

```shell

```

```shell

```

```shell
4.2.1
```

### Cisco NX-OS Version

```shell

```

### Cisco ACI Version

```shell

```

gmicol:
## Related Issue(s)

- Resolves CiscoDevNet/ansible-nd#238

## Proposed Changes

- Add initial `nd_manage_route_map` support for Cisco Nexus Dashboard 4.2+ Route Maps.
- Add route-map endpoint models for:
  - list/get route maps
  - bulk create route maps
  - replace route map
  - delete route map
  - bulk remove route maps
- Add Pydantic models and enums for route maps, entries, and supported match/set rule entries.
- Add a fabric-scoped route-map orchestrator using the current Orchestrators framework and `NDStateMachine`.
- Support `merged`, `replaced`, `overridden`, and `deleted` states.
- Support tenant-specific route maps through full route-map names and optional `tenant_name`.
- Add module documentation, examples, and route-map integration target files.
- Add unit coverage for endpoint, model, and orchestrator behavior.

## Test Notes

- Focused unit tests:
  - `python -m pytest tests/unit/module_utils/endpoints/test_manage_route_maps.py tests/unit/module_utils/models/test_manage_route_map.py tests/unit/module_utils/orchestrators/test_manage_route_map.py -q`
  - Result: `29 passed`

- Developed against Cisco Nexus Dashboard 4.2.1 / ND 4.2+

* [ ] analyze
* [ ] infa
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

### DIFF
--- a/plugins/module_utils/endpoints/v1/manage/manage_route_maps.py
+++ b/plugins/module_utils/endpoints/v1/manage/manage_route_maps.py
@@ -1,0 +1,479 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2026, Gaspard Micol (@gmicol) <gmicol@cisco.com>
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+"""
+ND Manage Route Maps endpoint models.
+
+This module contains endpoint definitions for route-map-related operations
+in the ND Manage API.
+
+## Endpoints
+
+- `EpManageRouteMapsGet` - Get a specific route map by name
+  (GET /api/v1/manage/fabrics/{fabricName}/routeMaps/{routeMapName})
+- `EpManageRouteMapsListGet` - List all route maps for a fabric
+  (GET /api/v1/manage/fabrics/{fabricName}/routeMaps)
+- `EpManageRouteMapsPost` - Bulk-create route maps for a fabric
+  (POST /api/v1/manage/fabrics/{fabricName}/routeMaps)
+- `EpManageRouteMapsPut` - Update a specific route map
+  (PUT /api/v1/manage/fabrics/{fabricName}/routeMaps/{routeMapName})
+- `EpManageRouteMapsDelete` - Delete a specific route map
+  (DELETE /api/v1/manage/fabrics/{fabricName}/routeMaps/{routeMapName})
+- `EpManageRouteMapsBulkDelete` - Bulk-delete route maps by name
+  (POST /api/v1/manage/fabrics/{fabricName}/routeMapActions/remove)
+"""
+
+from __future__ import annotations
+
+__metaclass__ = type
+
+from typing import ClassVar, Literal, Optional
+from urllib.parse import quote
+
+from ansible_collections.cisco.nd.plugins.module_utils.enums import HttpVerbEnum
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.v1.manage.base_path import BasePath
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.mixins import FabricNameMixin, RouteMapNameMixin
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.base import NDEndpointBaseModel
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.query_params import EndpointQueryParams
+from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import Field
+from ansible_collections.cisco.nd.plugins.module_utils.types import IdentifierKey
+
+
+class RouteMapsEndpointParams(EndpointQueryParams):
+    """
+    # Summary
+
+    Query parameters shared by single-item route map endpoints.
+
+    ## Parameters
+
+    - cluster_name: Name of the target Nexus Dashboard cluster (multi-cluster deployments)
+
+    ## Usage
+
+    ```python
+    params = RouteMapsEndpointParams(cluster_name="cluster1")
+    query_string = params.to_query_string()
+    # Returns: "clusterName=cluster1"
+    ```
+    """
+
+    cluster_name: Optional[str] = Field(
+        default=None,
+        min_length=1,
+        description="Name of the target Nexus Dashboard cluster to execute this API, in a multi-cluster deployment",
+    )
+
+
+class RouteMapsListEndpointParams(EndpointQueryParams):
+    """
+    # Summary
+
+    Query parameters for the GET /fabrics/{fabricName}/routeMaps list endpoint.
+
+    ## Parameters
+
+    - cluster_name: Name of the target Nexus Dashboard cluster (multi-cluster deployments)
+    - filter: Lucene-format filter string
+    - max: Maximum number of records to return
+    - offset: Number of records to skip for pagination
+    - sort: Sort field with optional ``:desc`` suffix
+
+    ## Usage
+
+    ```python
+    params = RouteMapsListEndpointParams(max=10, offset=0)
+    query_string = params.to_query_string()
+    # Returns: "max=10&offset=0"
+    ```
+    """
+
+    cluster_name: Optional[str] = Field(
+        default=None,
+        min_length=1,
+        description="Name of the target Nexus Dashboard cluster to execute this API, in a multi-cluster deployment",
+    )
+
+    filter: Optional[str] = Field(
+        default=None,
+        description="Lucene format filter - Filter the response based on this filter field",
+    )
+
+    max: Optional[int] = Field(
+        default=None,
+        ge=1,
+        description="Number of records to return",
+    )
+
+    offset: Optional[int] = Field(
+        default=None,
+        ge=0,
+        description="Number of records to skip for pagination",
+    )
+
+    sort: Optional[str] = Field(
+        default=None,
+        description="Sort the records by the declared fields in either ascending (default) or descending (:desc) order",
+    )
+
+
+class _EpManageRouteMapsBase(RouteMapNameMixin, FabricNameMixin, NDEndpointBaseModel):
+    """
+    Base class for ND Manage Route Maps endpoints.
+
+    Provides common functionality for all HTTP methods on the
+    /api/v1/manage/fabrics/{fabricName}/routeMaps endpoint.
+
+    Subclasses may override:
+    - ``_require_fabric_name``: set to ``False`` if fabric name is not required.
+    - ``_require_route_map_name``: set to ``False`` for collection-level endpoints
+      (list, bulk-create) that do not include a route map name in the path.
+    """
+
+    _require_fabric_name: ClassVar[bool] = True
+    _require_route_map_name: ClassVar[bool] = True
+
+    endpoint_params: EndpointQueryParams = Field(default_factory=EndpointQueryParams, description="Endpoint-specific query parameters")
+
+    def set_identifiers(self, identifier: IdentifierKey = None) -> None:
+        """Set the route_map_name from the identifier value."""
+        self.route_map_name = identifier
+
+    @property
+    def path(self) -> str:
+        """
+        # Summary
+
+        Build the endpoint path for route maps.
+
+        ## Returns
+
+        - Complete endpoint path string including fabric name, optional route map name,
+          and query string.
+
+        ## Raises
+
+        - ``ValueError`` if ``fabric_name`` is required but not set.
+        - ``ValueError`` if ``route_map_name`` is required but not set.
+        """
+        if self._require_fabric_name and self.fabric_name is None:
+            raise ValueError(f"{type(self).__name__}.path: fabric_name must be set before accessing path.")
+        if self._require_route_map_name and self.route_map_name is None:
+            raise ValueError(f"{type(self).__name__}.path: route_map_name must be set before accessing path.")
+
+        segments = ["fabrics"]
+        if self.fabric_name is not None:
+            segments.append(quote(self.fabric_name, safe=""))
+        segments.append("routeMaps")
+        if self.route_map_name is not None:
+            segments.append(quote(self.route_map_name, safe=""))
+
+        base_path = BasePath.path(*segments)
+        query_string = self.endpoint_params.to_query_string()
+        if query_string:
+            return f"{base_path}?{query_string}"
+        return base_path
+
+
+class EpManageRouteMapsGet(_EpManageRouteMapsBase):
+    """
+    # Summary
+
+    ND Manage Route Maps GET Endpoint
+
+    ## Description
+
+    Endpoint to retrieve a specific route map by name from a fabric.
+    Both ``fabric_name`` and ``route_map_name`` are required path parameters.
+
+    ## Path
+
+    - ``/api/v1/manage/fabrics/{fabricName}/routeMaps/{routeMapName}``
+    - ``/api/v1/manage/fabrics/{fabricName}/routeMaps/{routeMapName}?clusterName=cluster1``
+
+    ## Verb
+
+    - GET
+
+    ## Raises
+
+    - ``ValueError`` if ``fabric_name`` or ``route_map_name`` is not set when accessing ``path``
+
+    ## Usage
+
+    ```python
+    ep = EpManageRouteMapsGet()
+    ep.fabric_name = "my-fabric"
+    ep.route_map_name = "my-route-map"
+    path = ep.path
+    verb = ep.verb
+    ```
+    """
+
+    class_name: Literal["EpManageRouteMapsGet"] = Field(default="EpManageRouteMapsGet", description="Class name for backward compatibility")
+
+    endpoint_params: RouteMapsEndpointParams = Field(default_factory=RouteMapsEndpointParams, description="Endpoint-specific query parameters")
+
+    @property
+    def verb(self) -> HttpVerbEnum:
+        """Return the HTTP verb for this endpoint."""
+        return HttpVerbEnum.GET
+
+
+class EpManageRouteMapsListGet(_EpManageRouteMapsBase):
+    """
+    # Summary
+
+    ND Manage Route Maps List GET Endpoint
+
+    ## Description
+
+    Endpoint to list all route maps for a given fabric.
+    Supports optional query parameters for filtering, pagination, and sorting.
+
+    ## Path
+
+    - ``/api/v1/manage/fabrics/{fabricName}/routeMaps``
+    - ``/api/v1/manage/fabrics/{fabricName}/routeMaps?max=10&offset=0``
+
+    ## Verb
+
+    - GET
+
+    ## Raises
+
+    - ``ValueError`` if ``fabric_name`` is not set when accessing ``path``
+
+    ## Usage
+
+    ```python
+    ep = EpManageRouteMapsListGet()
+    ep.fabric_name = "my-fabric"
+    path = ep.path
+    verb = ep.verb
+    ```
+    """
+
+    _require_route_map_name: ClassVar[bool] = False
+
+    class_name: Literal["EpManageRouteMapsListGet"] = Field(default="EpManageRouteMapsListGet", description="Class name for backward compatibility")
+
+    endpoint_params: RouteMapsListEndpointParams = Field(default_factory=RouteMapsListEndpointParams, description="Endpoint-specific query parameters")
+
+    @property
+    def verb(self) -> HttpVerbEnum:
+        """Return the HTTP verb for this endpoint."""
+        return HttpVerbEnum.GET
+
+
+class EpManageRouteMapsPost(_EpManageRouteMapsBase):
+    """
+    # Summary
+
+    ND Manage Route Maps POST Endpoint
+
+    ## Description
+
+    Endpoint to bulk-create route maps for a given fabric.
+    The request body must conform to the ``CreateRouteMapsRequest`` schema:
+    ``{"routeMaps": [{"name": ..., "entries": [...]}, ...]}``.
+
+    ## Path
+
+    - ``/api/v1/manage/fabrics/{fabricName}/routeMaps``
+
+    ## Verb
+
+    - POST
+
+    ## Request Body
+
+    ``CreateRouteMapsRequest`` schema -- ``{"routeMaps": [RouteMap, ...]}``.
+
+    ## Raises
+
+    - ``ValueError`` if ``fabric_name`` is not set when accessing ``path``
+
+    ## Usage
+
+    ```python
+    ep = EpManageRouteMapsPost()
+    ep.fabric_name = "my-fabric"
+    rest_send.path = ep.path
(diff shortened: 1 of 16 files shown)


plugins/module_utils/models/manage_route_map/manage_route_map.py
```diff
@@ -0,0 +1,503 @@
...
+from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import Field
+from ansible_collections.cisco.nd.plugins.module_utils.models.base import NDBaseModel
+from ansible_collections.cisco.nd.plugins.module_utils.models.nested import NDNestedModel
+from ansible_collections.cisco.nd.plugins.module_utils.models.manage_route_map.enums import ActionEnum, RuleTypeEnum
+
+# Rule type choices list (used in argument_spec)
+RULE_TYPE_CHOICES = [e.value for e in RuleTypeEnum]
+
+
+class RouteMapRuleEntryModel(NDNestedModel):
```
mikewiebe, line 148: ### High: route-map `rule_type` variants do not enforce required fields locally

**Issue**

The module docs say several fields are required depending on `rule_type`, but `RouteMapRuleEntryModel` makes the variant-specific fields optional and has no discriminator validator.

**Evidence**

- [`plugins/module_utils/models/manage_route_map/manage_route_map.py` line 56](https://github.com/CiscoDevNet/ansible-nd/pull/288/files#diff-2e1742ebdbe056f9d6b775600f7f4beac0a23a9863495e424c174af66c719330R56) defines `rule_type`, while variant-specific fields such as `prefix_list_names`, `community_list_names`, `tags`, `community_numbers`, `extended_community_list_name`, and `value` are optional.
- [`plugins/modules/nd_manage_route_map.py` line 372](https://github.com/CiscoDevNet/ansible-nd/pull/288/files#diff-8b50878e2fdc2f98e697da6dfabfb8c9492bfe0f807cf0e292282e5860d7864bR372) only checks that route maps have `entries`; it does not validate each `rule_entries` item against its `rule_type`.
- A direct model construction test accepted and serialized configs containing only `rule_type: matchIpv4PrefixList` and only `rule_type: setLocalPreference`, without their documented required fields.

**Existing PR overlap**

**Impact**

Invalid route-map configs can pass module/check-mode validation and fail only when the controller rejects them. That makes dry-run behavior less trustworthy and pushes schema errors to ND instead of catching them at the module boundary.

**Suggested fix**

Add a model validator mapping each `rule_type` to its required fields, allowed fields, and range checks, then add negative unit tests for every rule family.

---

plugins/module_utils/models/manage_route_map/manage_route_map.py
```diff
@@ -0,0 +1,503 @@
...
+
+    - ``last_update_timestamp`` is a read-only field returned by the API.
+      It is excluded from payload output and diff comparisons.
+    - ``fabric_name`` is managed at the orchestrator level and is NOT part of
+      this model; path construction is handled by the endpoint classes.
+    """
+
+    # --- Identifier Configuration ---
+
+    identifiers: ClassVar[Optional[List[str]]] = ["name"]
```
mikewiebe: ### Medium: `tenant_name` is exposed but identity is still bare-name based

**Issue**

`tenant_name` is part of the route-map payload model, but resource identity and delete/update lookup still use only bare `name`.

**Evidence**

- [`plugins/module_utils/models/manage_route_map/manage_route_map.py` line 308](https://github.com/CiscoDevNet/ansible-nd/pull/288/files#diff-2e1742ebdbe056f9d6b775600f7f4beac0a23a9863495e424c174af66c719330R308) identifies route maps by only `["name"]`.
- [`plugins/module_utils/models/manage_route_map/manage_route_map.py` line 332](https://github.com/CiscoDevNet/ansible-nd/pull/288/files#diff-2e1742ebdbe056f9d6b775600f7f4beac0a23a9863495e424c174af66c719330R332) exposes `tenant_name`.
- [`plugins/module_utils/orchestrators/manage_route_map.py` line 193](https://github.com/CiscoDevNet/ansible-nd/pull/288/files#diff-6f7e930799bd4ef606a6f45ad84c59e6873d3ee81ab76c83d1e42c2f60ac4067R193) sends `item.name` in the bulk delete payload.

**Existing PR overlap**

**Impact**

A user can create a tenant-scoped route map with bare `name` plus `tenant_name`, but later miss it on get/update/delete if ND keys or returns the object as `tenant~name`.

**Suggested fix**

Either implement the PR #286 tenant-aware normalization and `api_name` pattern, or reject/document that `config.name` must already be fully qualified whenever `tenant_name` is set.

---

plugins/modules/nd_manage_route_map.py
```diff
@@ -0,0 +1,427 @@
...
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+ANSIBLE_METADATA = {"metadata_version": "1.1", "status": ["preview"], "supported_by": "community"}
+
+DOCUMENTATION = r"""
+---
+module: nd_manage_route_map
+version_added: "1.6.0"
```
mikewiebe: ### Low: docs metadata needs cleanup

**Issue**

The new module advertises the wrong `version_added`, and its `RETURN` block is empty.

**Evidence**

- [`plugins/modules/nd_manage_route_map.py` line 14](https://github.com/CiscoDevNet/ansible-nd/pull/288/files#diff-8b50878e2fdc2f98e697da6dfabfb8c9492bfe0f807cf0e292282e5860d7864bR14) sets `version_added: "1.6.0"`.
- The module `RETURN` block is empty.

**Existing PR overlap**

**Impact**

Generated documentation claims this module existed in an older collection release and does not tell users which keys to expect from the result.

**Suggested fix**

Update `version_added` to `2.0.0` and document the standard return structure used by these state-machine modules.

plugins/module_utils/models/manage_route_map/manage_route_map.py
```diff
@@ -0,0 +1,503 @@
...
+
+    value: Optional[int] = Field(
+        default=None,
+        alias="value",
+        description="Local preference value (0-4294967295).",
+    )
+
+    # --- setIpv4NextHop / setIpv6NextHop ---
+
+    next_hop_ip_collection: Optional[List[str]] = Field(
```
mikewiebe: **Issue:** `nd_manage_route_map` is not idempotent for next-hop rule entries.

I reproduced this live with a minimal route map containing a single `setIpv4NextHop` rule. The first apply creates the route map, but reapplying the exact same config reports `changed=True`.

Minimum repro:

```yaml
- name: Create next-hop route map
  cisco.nd.nd_manage_route_map:
    config:
      - name: PR288-REPRO-RM-IDEM
        entries:
          - sequence_number: 10
            action: permit
            rule_entries:
              - rule_type: setIpv4NextHop
                next_hop_ip_collection:
                  - 192.0.2.10
    state: merged
  register: first_apply

- name: Reapply same next-hop route map
  cisco.nd.nd_manage_route_map:
    config:
      - name: PR288-REPRO-RM-IDEM
        entries:
          - sequence_number: 10
            action: permit
            rule_entries:
              - rule_type: setIpv4NextHop
                next_hop_ip_collection:
                  - 192.0.2.10
    state: merged
  register: second_apply
```

Observed failure:

```text
Expected second apply to be idempotent, got changed=True
```

I also ran a focused probe across route-map rule shapes. These next-hop variants all reported changed on reapply:

- `setIpv4NextHop` with `next_hop_ip_collection`
- `setIpv4NextHop` with `use_peer_address`
- `setIpv4NextHop` with `unchanged`
- `setIpv6NextHop` with `next_hop_ip_collection`
- `setIpv6NextHop` with `use_peer_address`
- `setIpv6NextHop` with `unchanged`

The same probe showed that `matchExtendedCommunity`, `setExtendedCommunityList`, and `setCommunity` with flags were idempotent, so this looks specific to next-hop rule normalization/diffing.

Expected: the second identical `state: merged` apply should report `changed=False`.

Actual: the second identical apply reports `changed=True`.

plugins/module_utils/models/manage_route_map/manage_route_map.py
```diff
@@ -0,0 +1,503 @@
...
+                                   ``loadShare``, ``enforceOrder``,
+                                   ``verifyAvailability``, ``usePeerAddress``,
+                                   ``redistributeUnchanged``, ``unchanged``,
+                                   ``trackId``
+    - ``setIpv6NextHop``         - same optional fields as ``setIpv4NextHop``
+    """
+
+    # --- Discriminator (required for every rule entry) ---
+
+    rule_type: str = Field(alias="ruleType", description="Rule type discriminator.")
```
mikewiebe, line 185: **Issue:** Route-map rule entries need discriminator-aware validation.

I reproduced this live with `matchIpv4Acl` in check mode. The rule is missing its required companion field, `access_control_list_name`, but the module reports `changed=True` instead of failing local validation.

Minimum repro:

```yaml
- name: Missing matchIpv4Acl access_control_list_name should fail locally
  cisco.nd.nd_manage_route_map:
    config:
      - name: PR288-REPRO-RM-MISSING-ACL
        entries:
          - sequence_number: 10
            action: permit
            rule_entries:
              - rule_type: matchIpv4Acl
    state: merged
  check_mode: true
  register: missing_acl_result
```

Observed result:

```text
Expected local validation failure for matchIpv4Acl without access_control_list_name, got failed=False changed=True.
```

This is probably broader than `matchIpv4Acl`. `rule_type` is effectively a discriminator, and each discriminator value has its own required/allowed companion fields. Examples:

| Rule type | Required companion field(s) |
|---|---|
| `matchIpv4Acl`, `matchIpv6Acl` | `access_control_list_name` |
| `matchIpv4PrefixList`, `matchIpv6PrefixList` | `prefix_list_names` |
| `matchCommunity` | `community_list_names` |
| `matchExtendedCommunity` | `extended_community_list_names` |
| `matchTag` | `tags` |
| `setExtendedCommunityList` | `extended_community_list_name` |
| `setLocalPreference` | `value` |
| `setIpv4NextHop`, `setIpv6NextHop` | one valid next-hop mode, with mutually exclusive combinations |

Worst likely outcome if we do not support this validation: users can submit syntactically valid Ansible that is semantically invalid for the selected `rule_type`. Check mode may say a change would be made even though the real run can fail later against ND, wrong companion fields may be silently ignored, and idempotency can become unstable across route-map rule permutations.

I do not think this necessarily creates a silent destructive config risk because ND appears to reject many invalid payloads. The practical risk is unreliable automation: check mode lies, invalid configs get through local validation, and users get delayed API errors instead of clear Ansible validation messages.

Suggested direction: add rule-type-specific validation for required fields, allowed fields, and mutually exclusive field combinations, then add a validation matrix covering the route-map rule permutations.

This may be worth discussing in the next team meeting because the validation matrix could be fairly large, especially for next-hop rule combinations.

gmicol: I implemented a solution in commit [[ignore] Validate route-map rule permutations](https://github.com/CiscoDevNet/ansible-nd/pull/288/commits/19360ff80757ea79ab3bf9b666f077908b253cef), let me know what you think

mikewiebe: Thanks @gmicol sorry, just saw this.  Will take a look

mikewiebe: Thanks, this looks like it addresses the original discriminator-validation gap and looks like a good overall approach.

My understanding is that the bundled `routeMapSchemas.yaml`  OpenAPI schema is the source of truth for the required and allowed rule fields. Can you confirm that `_REQUIRED_RULE_FIELDS` and `_ALLOWED_RULE_FIELDS` are intended to mirror those schemas?

If so, we should acknowledge that this manually maintained ruleset must be reviewed and refreshed whenever the bundled OpenAPI schema changes. A reference to the source schema in the code, or an automated parity test, could help prevent the two definitions from drifting.

I could not find the following additional constraints expressed in the OpenAPI schema:

- Exactly one next-hop mode must be selected.
- `track_id` requires `next_hop_ip_collection`.
- Every property not listed for a rule type must be rejected. The schemas do not appear to set `additionalProperties: false`.

How were these rules confirmed? For example, were they derived from observed ND controller validation, the ND UI, another product specification, or guidance from the API owners?

I would like to distinguish the rules directly derived from OpenAPI from stricter controller-specific constraints so we can document and maintain each appropriately.

mikewiebe: I assume this same approach will be followed for your other PRs?

mikewiebe (changes requested):

plugins/module_utils/orchestrators/manage_route_map.py
```diff
@@ -0,0 +1,197 @@
...
+        """Validate that the target fabric can be mutated before writes."""
+        if model_instances:
+            self.fabric_context.validate_for_mutation()
+
+    @staticmethod
+    def _raise_on_bulk_errors(result: ResponseType, action: str) -> None:
+        """Raise when a 207 bulk response contains failed per-item results."""
+        if not isinstance(result, dict):
+            return
+        failures = [item for item in result.get("results", []) if isinstance(item, dict) and item.get("status") not in (None, "", "success")]
```
allenrobel: This treats any per-item `status` not in `(None, "", "success")` as a failure — an allowlist. The sibling ACL orchestrator (`orchestrators/manage_acl.py`, merged in PR #266, same Routing Policies API family and same 207 Multi-Status body shape) deliberately uses the opposite polarity:

```python
# Per-item ``status`` values in a 207 Multi-Status body that count as a failure. Anything else --
# ``success``, missing, empty, or future progress tokens -- is tolerated so informational rows do
# not surface as spurious errors. Mirrors the prefix-list orchestrator's denylist approach.
_FAILURE_STATUSES = frozenset({"failed", "failure", "error"})
```

With the allowlist, any benign informational/in-progress status ND emits in a bulk response raises `RuntimeError` and fails the module run. Suggest switching to the denylist approach to match the siblings.

plugins/module_utils/models/manage_route_map/manage_route_map.py
```diff
@@ -0,0 +1,674 @@
...
+        "use_peer_address",
+        "verify_availability",
+    },
+}
+
+_NEXT_HOP_RULE_TYPES = {
+    RuleTypeEnum.SET_IPV4_NEXT_HOP.value,
+    RuleTypeEnum.SET_IPV6_NEXT_HOP.value,
+}
+_NEXT_HOP_FALSE_DEFAULT_FIELDS = (
```
allenrobel, line 137: `_NEXT_HOP_FALSE_DEFAULT_FIELDS` and the stripping in `to_diff_dict()` (lines 503–513) work around an ND API discrepancy: ND echoes these next-hop booleans back as `false` on GET even when they were never sent, which is what caused the next-hop idempotency failures reported earlier on this PR. CLAUDE.md's Workaround Markers section applies: "Any code that works around an ND API discrepancy (wire shape vs. schema, wrong HTTP status, silent field drops, etc.) MUST be flagged with a `TODO(X.Y.Z)` marker so the workaround can be audited and removed when the ND release cadence permits."

Since the bug-tracker vault isn't shared yet, please add `TODO(4.2.1) TBD` (with a line or two describing what ND echoes vs. what we send) at both sites — we'll sweep the `TBD` slugs later and match them to bug-tracker documents.

plugins/module_utils/orchestrators/manage_route_map.py
```diff
@@ -0,0 +1,197 @@
...
+        """
+        List all route maps for the configured fabric.
+
+        The API response is wrapped under the ``"routeMaps"`` key; this method
+        extracts and returns the list directly.
+        """
+        try:
+            api_endpoint = self._configure_endpoint(self.query_all_endpoint())
+            result = self._request(path=api_endpoint.path, verb=api_endpoint.verb, not_found_ok=True)
+            return result.get("routeMaps", []) or []
```
allenrobel: `query_all()` issues a single GET and never sets `max`/`offset`, but this list endpoint paginates (`RouteMapsListEndpointParams` exposes the same `max`/`offset` contract as the ACL list endpoint). With more route maps than the controller's default page size, only page 1 is returned: the state machine misses existing route maps, breaking idempotency and — worst case — leaving route maps undeleted under `overridden`.

`orchestrators/manage_acl.py` (PR #266) handles this with a paged walk — `query_all_page_size`/`query_all_max_pages` ClassVars, looping on `max`/`offset` until a short/empty page, with a seen-name dedup guard. Its docstring: "The list endpoint paginates, so this walks the collection with `max`/`offset` until a page arrives short... or empty." Suggest mirroring that here.

plugins/module_utils/endpoints/v1/manage/manage_route_maps.py
```diff
@@ -0,0 +1,477 @@
...
+    ```
+    """
+
+    cluster_name: str | None = Field(
+        default=None,
+        min_length=1,
+        description="Name of the target Nexus Dashboard cluster to execute this API, in a multi-cluster deployment",
+    )
+
+    filter: str | None = Field(
```
allenrobel: `filter`, `max`, `offset`, and `sort` are re-declared as bare fields here instead of composing the existing `LuceneQueryParams` (or `CompositeQueryParams`) from `endpoints/query_params.py`. That loses three things the shared class provides: the `asc`/`desc` sort-direction validator, the `max` upper bound (`le=10000` — this declaration only has `ge=1`), and the Lucene-safe URL encoding that preserves `:` in `filter` (`field:value` syntax). Same point raised on PR #197; `manage_fabrics_policies.py` and `manage_fabrics_vpc_pairs.py` show the composed pattern. (`manage_acl.py` shares this issue — worth a follow-up there too.)

allenrobel (commented):
### Code review

Four findings, each as an inline comment: 207 bulk-error status polarity, a missing `TODO(X.Y.Z)` workaround marker, unpaginated `query_all()`, and re-declared list query params. Overall the module is in good shape — CRUD wiring, identifier composition, and the next-hop diff normalization are solid and well tested.

plugins/module_utils/models/manage_route_map/manage_route_map.py
gmicol: Thanks Mike, that distinction makes sense.

Yes, `_REQUIRED_RULE_FIELDS` and most of `_ALLOWED_RULE_FIELDS` are intended to mirror the bundled route-map OpenAPI schemas: required fields come from each rule schema’s `required` list, and allowed fields come from each schema’s `properties`.

The extra-property rejection is a local guardrail around our flat Ansible model. OpenAPI models each rule type as a separate schema, but our model exposes all possible rule fields on one class. Without the allowed-field check, users could mix fields from unrelated rule types and we would accept semantically invalid configs locally. I agree this should be documented as an Ansible-side semantic check, not as `additionalProperties: false` from OpenAPI.

For next-hop, I verified the stricter rules against the ND UI. They are not fully expressed in OpenAPI, so they should be documented separately from the schema-derived matrix.

OpenAPI gives us the available fields, numeric ranges, and IPv4/IPv6 address formats. The UI/controller adds these semantic constraints for `setIpv4NextHop` / `setIpv6NextHop`:

- At least one next-hop IP option must be provided.
- `verify_availability` requires `next_hop_ip_collection`.
- `track_id` is only available when `next_hop_ip_collection` is set and `verify_availability` is enabled.
- `use_peer_address` can be used by itself, or with `next_hop_ip_collection`, `verify_availability`, and `track_id` when the above rule is satisfied.
- `use_peer_address` cannot be mixed with `drop_on_fail`, `load_share`, `enforce_order`, `redistribute_unchanged`, or `unchanged`.
- `use_peer_address`, `redistribute_unchanged`, and `unchanged` cannot be mixed with `drop_on_fail`, `load_share`, or `enforce_order`.
- Otherwise, if none of those rules are violated, the UI allows the remaining combinations.

The UI errors I used to confirm this were:

- `At least one next hop IP option must be provided`
- `Cannot mix 'use peer address' with other ip next-hop configurations`
- `Cannot mix 'use peer address', 'redistribute unchanged', or 'unchanged' with any of 'drop on fail', 'enable load sharing' or 'force order' ip next-hop configurations`

So I adjusted the current validator: the “exactly one next-hop mode” rule is too strict based on UI behavior. I updated it to match the UI/controller semantics, and I added code comments or tests that clearly separate OpenAPI-derived validation from UI/controller-derived semantic validation.

For similar flat discriminator models in other PRs, I’ll follow the same split: schema-derived validation matrix where possible, plus clearly documented local semantic guardrails only when they are verified against controller/UI behavior or API-owner guidance.

plugins/module_utils/orchestrators/manage_route_map.py
gmicol: Addressed. I switched route-map 207 handling to the same denylist approach as ACL, where only failed/failure/error statuses are treated as per-item failures. I also added coverage confirming informational statuses such as accepted, empty status, and omitted status do not fail create_bulk.

plugins/module_utils/models/manage_route_map/manage_route_map.py
gmicol: Addressed. Added `TODO(4.2.1) TBD` markers at both workaround sites: the next-hop false-default field list and the to_diff_dict normalization loop. The comments document that ND echoes unsent next-hop booleans as false on GET.

plugins/module_utils/orchestrators/manage_route_map.py
gmicol: Addressed. query_all now walks the route-map list endpoint with max/offset, preserves clusterName, deduplicates rows by route-map name, and has a max-page bound. Added unit coverage for multi-page responses and duplicate suppression.

I also opened an issue to [add reusable, safe pagination for ND collection queries in the Orchestrators framework](https://github.com/CiscoDevNet/ansible-nd/issues/407) to consolidate on pagination across all orchestrators.

plugins/module_utils/endpoints/v1/manage/manage_route_maps.py
gmicol: This was addressed in the branch by composing `RouteMapsListEndpointParams` with `LuceneQueryParams` through `CompositeQueryParams`. I added extra tests for the shared Lucene max upper bound and sort-direction validator, in addition to existing filter encoding coverage. But I do agree this is worth a follow-up

allenrobel (approved):
LGTM

mikewiebe (approved):
LGTM

PR #291, opened 2026-05-21
nd_fabric_prepare_update -  ND 4.2 software staging

### LINKED ISSUE #448: [ND 4.2+] Add Fabric Software Prepare/Stage Module (nd_fabric_prepare_update)
## Summary

Add `nd_fabric_prepare_update` to support the Nexus Dashboard 4.2 Fabric Software Management Prepare stage (`softwareUpdatePlan/actions/stage`).

## Related PR

- #291

## Scope

- Stage and validate the configured image for one or more update groups.
- Reject mixed-role update groups before staging.
- Remain idempotent when update groups are already staged.
- Support optional completion polling with `wait: true`.
- Add the required endpoint models, orchestrator logic, documentation, unit tests, and integration tests.

## Acceptance Criteria

- Mixed-role groups are rejected before an update is started.
- Already-staged groups report no change.
- `wait: true` polls safely to completion and tolerates transient transport errors.
- Sanity, unit, and live ND 4.2.1 integration tests pass.

## Out of Scope

- Image upload, software installation, and activation.

allenrobel:
## Related Issue(s)

N/A — new module.

## Proposed Changes

- New module **`nd_fabric_prepare_update`** — the Fabric Software Management "Prepare" step on Nexus Dashboard 4.2, i.e. the `softwareUpdatePlan/actions/stage` action: stages and validates the configured image on the switches of one or more update groups.
- One-shot action module: pre-flight check rejects mixed-role update groups (ND will not prepare them), idempotent against already-staged groups, and with `wait: true` polls to completion.
- `FabricPrepareUpdateOrchestrator` — drives the stage action and a poll loop that tolerates transient transport errors during the long wait.
- New endpoint models `EpFabricSoftwareUpdatePlanStage` and `EpFabricSoftwareUpdatePlanSummary`; Pydantic models for the `softwareUpdatePlan/summary` response.
- 33 unit tests (orchestrator + models) and an integration test target.

## Test Notes

- **Unit:** `python -m pytest tests/unit/module_utils/models/test_fabric_prepare_update.py tests/unit/module_utils/orchestrators/test_fabric_prepare_update.py` — 33 pass.
- **Sanity:** `ansible-test sanity --docker` passes for all new/changed files.
- **Integration:** `ansible-test integration nd_fabric_prepare_update` — green against a live ND 4.2.1 lab (fabric SITE1), covering mixed-role pre-flight rejection, check mode, a real stage + wait, and idempotency.

4.2.1

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

### DIFF
--- a/plugins/module_utils/endpoints/v1/manage/software_update_plan_summary.py
+++ b/plugins/module_utils/endpoints/v1/manage/software_update_plan_summary.py
@@ -18,8 +18,27 @@ from urllib.parse import quote
 from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import Field
 from ansible_collections.cisco.nd.plugins.module_utils.endpoints.base import NDEndpointBaseModel
 from ansible_collections.cisco.nd.plugins.module_utils.endpoints.mixins import FabricNameMixin
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.query_params import EndpointQueryParams
 from ansible_collections.cisco.nd.plugins.module_utils.endpoints.v1.manage.base_path import BasePath
 from ansible_collections.cisco.nd.plugins.module_utils.enums import HttpVerbEnum
+
+
+class SoftwareUpdatePlanSummaryEndpointParams(EndpointQueryParams):
+    """
+    # Summary
+
+    Endpoint-specific query parameters for the software update plan summary endpoint.
+
+    `update_group_name` is rendered as the API's `updateGroupName` query parameter (snake_case ->
+    camelCase is automatic via `EndpointQueryParams`). When set, Nexus Dashboard scopes the summary
+    to that single update group instead of returning the fabric-wide plan.
+
+    ## Raises
+
+    None
+    """
+
+    update_group_name: str | None = Field(default=None, min_length=1, description="Scope the summary to a single update group")
 
 
 class EpFabricSoftwareUpdatePlanSummary(FabricNameMixin, NDEndpointBaseModel):
@@ -31,6 +50,9 @@ class EpFabricSoftwareUpdatePlanSummary(FabricNameMixin, NDEndpointBaseModel):
     Returns the fabric-wide software update plan: every update group with its per-switch stage /
     validate / install status. Used to drive the `nd_fabric_prepare_update` pre-flight role check
     and to poll for staging completion.
+
+    Set `endpoint_params.update_group_name` to scope the summary to a single update group via the
+    API's optional `updateGroupName` query parameter; leave it unset for the fabric-wide summary.
 
     - Path: `/api/v1/manage/fabrics/{fabric_name}/softwareUpdatePlan/summary`
     - Verb: GET
@@ -45,13 +67,17 @@ class EpFabricSoftwareUpdatePlanSummary(FabricNameMixin, NDEndpointBaseModel):
     class_name: Literal["EpFabricSoftwareUpdatePlanSummary"] = Field(
         default="EpFabricSoftwareUpdatePlanSummary", frozen=True, description="Class name for backward compatibility"
     )
+    endpoint_params: SoftwareUpdatePlanSummaryEndpointParams = Field(
+        default_factory=SoftwareUpdatePlanSummaryEndpointParams, description="Endpoint-specific query parameters"
+    )
 
     @property
     def path(self) -> str:
         """
         # Summary
 
-        Build the software update plan summary endpoint path. `fabric_name` is percent-encoded with `safe=""`.
+        Build the software update plan summary endpoint path. `fabric_name` is percent-encoded with `safe=""`. When
+        `endpoint_params.update_group_name` is set, the `updateGroupName` query string is appended to scope the summary.
 
         ## Raises
 
@@ -61,7 +87,11 @@ class EpFabricSoftwareUpdatePlanSummary(FabricNameMixin, NDEndpointBaseModel):
         """
         if self.fabric_name is None:
             raise ValueError(f"{type(self).__name__}.path: fabric_name must be set before accessing path.")
-        return BasePath.path("fabrics", quote(self.fabric_name, safe=""), "softwareUpdatePlan", "summary")
+        base_path = BasePath.path("fabrics", quote(self.fabric_name, safe=""), "softwareUpdatePlan", "summary")
+        query_string = self.endpoint_params.to_query_string()
+        if query_string:
+            return f"{base_path}?{query_string}"
+        return base_path
 
     @property
     def verb(self) -> HttpVerbEnum:
--- a/plugins/module_utils/endpoints/v1/manage/software_update_plan_summary.py
+++ b/plugins/module_utils/endpoints/v1/manage/software_update_plan_summary.py
@@ -1,0 +1,77 @@
+# Copyright: (c) 2026, Allen Robel (@allenrobel) <arobel@cisco.com>
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+"""
+ND Manage Fabric Software Management software update plan summary endpoint model.
+
+## Endpoints
+
+- `EpFabricSoftwareUpdatePlanSummary` - Software update plan summary for a fabric
+  (GET /api/v1/manage/fabrics/{fabric_name}/softwareUpdatePlan/summary)
+"""
+
+from __future__ import annotations
+
+from typing import Literal
+from urllib.parse import quote
+
+from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import Field
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.base import NDEndpointBaseModel
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.mixins import FabricNameMixin
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.v1.manage.base_path import BasePath
+from ansible_collections.cisco.nd.plugins.module_utils.enums import HttpVerbEnum
+
+
+class EpFabricSoftwareUpdatePlanSummary(FabricNameMixin, NDEndpointBaseModel):
+    """
+    # Summary
+
+    Software update plan summary for a fabric.
+
+    Returns the fabric-wide software update plan: every update group with its per-switch stage /
+    validate / install status. Used to drive the `nd_fabric_prepare_update` pre-flight role check
+    and to poll for staging completion.
+
+    - Path: `/api/v1/manage/fabrics/{fabric_name}/softwareUpdatePlan/summary`
+    - Verb: GET
+
+    ## Raises
+
+    ### ValueError
+
+    - Via `path` property if `fabric_name` is not set.
+    """
+
+    class_name: Literal["EpFabricSoftwareUpdatePlanSummary"] = Field(
+        default="EpFabricSoftwareUpdatePlanSummary", frozen=True, description="Class name for backward compatibility"
+    )
+
+    @property
+    def path(self) -> str:
+        """
+        # Summary
+
+        Build the software update plan summary endpoint path. `fabric_name` is percent-encoded with `safe=""`.
+
+        ## Raises
+
+        ### ValueError
+
+        - If `fabric_name` is not set before accessing `path`.
+        """
+        if self.fabric_name is None:
+            raise ValueError(f"{type(self).__name__}.path: fabric_name must be set before accessing path.")
+        return BasePath.path("fabrics", quote(self.fabric_name, safe=""), "softwareUpdatePlan", "summary")
+
+    @property
+    def verb(self) -> HttpVerbEnum:
+        """
+        # Summary
+
+        Return `HttpVerbEnum.GET`.
+
+        ## Raises
+
+        None
+        """
+        return HttpVerbEnum.GET
--- a/plugins/module_utils/models/fabric_prepare_update/software_update_plan_summary.py
+++ b/plugins/module_utils/models/fabric_prepare_update/software_update_plan_summary.py
@@ -12,8 +12,16 @@ they parse the wire shape and are never serialized back to ND.
 
 from __future__ import annotations
 
+from typing import Literal, TypeAlias
+
 from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import Field
 from ansible_collections.cisco.nd.plugins.module_utils.models.nested import NDNestedModel
+
+# Known per-switch stage / validate statuses on ND 4.2.1. The `str` fallback in `ImageStatus` lets
+# an unrecognized status from a newer ND release parse rather than raising during a poll; type
+# checkers collapse the union to `str`, so the Literal is documentation of the known vocabulary.
+KnownImageStatus: TypeAlias = Literal["none", "inProgress", "success", "failed", "skipped"]
+ImageStatus: TypeAlias = KnownImageStatus | str
 
 
 class UpdateGroupWarningModel(NDNestedModel):
@@ -39,9 +47,10 @@ class SwitchStageStatusModel(NDNestedModel):
     Per-switch stage / validate status within an update group, as reported by the
     `softwareUpdatePlan/summary` endpoint.
 
-    `image_staged_status` and `image_validated_status` are kept as free-form strings (not a
-    `Literal`) so an unrecognized status from a newer ND release parses rather than raising during
-    a poll. Known values: `none`, `inProgress`, `success`, `failed`, `skipped`.
+    `image_staged_status` and `image_validated_status` are typed `ImageStatus`
+    (`KnownImageStatus | str`): the `Literal` half documents the known vocabulary while the `str`
+    fallback lets an unrecognized status from a newer ND release parse rather than raising during
+    a poll.
 
     ## Raises
 
@@ -54,8 +63,8 @@ class SwitchStageStatusModel(NDNestedModel):
     switch_management_ip: str | None = Field(default=None, alias="switchManagementIP")
     switch_version: str | None = Field(default=None, alias="switchVersion")
     selected_version: str | None = Field(default=None, alias="selectedVersion")
-    image_staged_status: str | None = Field(default=None, alias="imageStagedStatus")
-    image_validated_status: str | None = Field(default=None, alias="imageValidatedStatus")
+    image_staged_status: ImageStatus | None = Field(default=None, alias="imageStagedStatus")
+    image_validated_status: ImageStatus | None = Field(default=None, alias="imageValidatedStatus")
     switch_stage_validate_percentage: int | None = Field(default=None, alias="switchStageValidatePercentage")
 
 
@@ -69,6 +78,11 @@ class UpdateGroupStatusModel(NDNestedModel):
     The `warnings` and `update_group_switches` lists default to empty: a key absent from the
     response body parses the same as an explicitly empty list, since every consumer treats the two
     identically.
+
+    `update_group_status` is kept as a free-form string (not a `Literal`) for the same reason as
+    the per-switch statuses in `SwitchStageStatusModel`: an unrecognized status from a newer ND
+    release must parse rather than raise during a poll. Its vocabulary is not lab-confirmed, so no
+    known-values alias is provided.
 
     ## Raises
 
--- a/plugins/module_utils/models/fabric_prepare_update/software_update_plan_summary.py
+++ b/plugins/module_utils/models/fabric_prepare_update/software_update_plan_summary.py
@@ -66,6 +66,10 @@ class UpdateGroupStatusModel(NDNestedModel):
     A single update group's status within the software update plan summary, including its member
     switches and any advisory warnings ND raises for the group.
 
+    The `warnings` and `update_group_switches` lists default to empty: a key absent from the
+    response body parses the same as an explicitly empty list, since every consumer treats the two
+    identically.
+
     ## Raises
 
     None
@@ -76,8 +80,8 @@ class UpdateGroupStatusModel(NDNestedModel):
     update_type: str | None = Field(default=None, alias="updateType")
     stage_validate_percentage: int | None = Field(default=None, alias="stageValidatePercentage")
     switch_count: int | None = Field(default=None, alias="switchCount")
-    warnings: list[UpdateGroupWarningModel] | None = Field(default=None, alias="warnings")
-    update_group_switches: list[SwitchStageStatusModel] | None = Field(default=None, alias="updateGroupSwitches")
+    warnings: list[UpdateGroupWarningModel] = Field(default_factory=list, alias="warnings")
+    update_group_switches: list[SwitchStageStatusModel] = Field(default_factory=list, alias="updateGroupSwitches")
 
 
 class SoftwareUpdatePlanSummaryModel(NDNestedModel):
@@ -86,10 +90,12 @@ class SoftwareUpdatePlanSummaryModel(NDNestedModel):
 
     Top-level parse of the `softwareUpdatePlan/summary` response. Only the `updateGroups` list is
     modeled; the `softwareUpdateSummary` and `tableHeaders` blocks are ignored (`extra="ignore"`).
+    `update_groups` defaults to empty, so a response missing the key parses the same as an
+    explicitly empty plan.
 
     ## Raises
 
     None
     """
 
-    update_groups: list[UpdateGroupStatusModel] | None = Field(default=None, alias="updateGroups")
+    update_groups: list[UpdateGroupStatusModel] = Field(default_factory=list, alias="updateGroups")
--- a/plugins/module_utils/models/fabric_prepare_update/software_update_plan_summary.py
+++ b/plugins/module_utils/models/fabric_prepare_update/software_update_plan_summary.py
@@ -1,0 +1,95 @@
+# Copyright: (c) 2026, Allen Robel (@allenrobel) <arobel@cisco.com>
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+"""
+Pydantic models for parsing the ND `softwareUpdatePlan/summary` response.
+
+The summary endpoint returns the fabric-wide software update plan: every update group with its
+per-switch stage / validate / install status. `nd_fabric_prepare_update` uses these models for the
+pre-flight switch-role check and to poll for staging completion. The models are response-only -
+they parse the wire shape and are never serialized back to ND.
+"""
+
+from __future__ import annotations
+
+from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import Field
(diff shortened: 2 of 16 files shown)


plugins/modules/nd_fabric_prepare_update.py
```diff
@@ -0,0 +1,219 @@
...
+
+    after = orchestrator.status_snapshot(update_groups, summary=final_summary)
+    return orchestrator.results, {"changed": True, "before": before, "after": after}
+
+
+def main():
+    argument_spec = nd_argument_spec()
+    argument_spec.update(
+        fabric_name=dict(type="str", required=True),
+        update_groups=dict(type="list", elements="str", required=True),
```
akinross, line 215: should this specify min length of 1? i do not see it enforced in the pydantic model, so could also be handled there.

plugins/module_utils/orchestrators/fabric_prepare_update.py
```diff
@@ -0,0 +1,444 @@
...
+
+        Returns False for an empty snapshot (nothing to confirm prepared).
+
+        ## Raises
+
+        None
+        """
+        switches = [sw for group in snapshot for sw in group.get("switches", [])]
+        if not switches:
+            return False
```
akinross: should this be True? where there are no switches does not that qualify as a success or failure?

plugins/module_utils/endpoints/v1/manage/software_update_plan_summary.py
```diff
@@ -0,0 +1,77 @@
...
+
+        ## Raises
+
+        ### ValueError
+
+        - If `fabric_name` is not set before accessing `path`.
+        """
+        if self.fabric_name is None:
+            raise ValueError(f"{type(self).__name__}.path: fabric_name must be set before accessing path.")
+        return BasePath.path("fabrics", quote(self.fabric_name, safe=""), "softwareUpdatePlan", "summary")
```
mikewiebe: ### Medium: Polling always reads the full software update plan instead of using the API's update-group filter

**Issue**

The summary endpoint and orchestrator always fetch the fabric-wide `softwareUpdatePlan/summary` response, then filter requested update groups locally. The OpenAPI contract exposes an `updateGroupName` query parameter on that same endpoint, but the endpoint class does not model it and the poll loop cannot use it.

**Evidence**

- [`plugins/module_utils/endpoints/v1/manage/software_update_plan_summary.py` line 64](https://github.com/CiscoDevNet/ansible-nd/pull/291/files#diff-286a96e50e30b5e0fbb68a75a4b284c05325bba69a2fca3b97aa121278440d8eR64)
  builds `/softwareUpdatePlan/summary` without any query-string support.
- [`plugins/module_utils/orchestrators/fabric_prepare_update.py` lines 193-197](https://github.com/CiscoDevNet/ansible-nd/pull/291/files#diff-6869ad3b5dd6d652dc2e9d5091ffa23343421a574d4e1c32e39839da75411abdR193)
  calls that unscoped summary endpoint for every preflight, idempotency, and wait poll.
- [`plugins/module_utils/orchestrators/fabric_prepare_update.py` lines 399-402](https://github.com/CiscoDevNet/ansible-nd/pull/291/files#diff-6869ad3b5dd6d652dc2e9d5091ffa23343421a574d4e1c32e39839da75411abdR399)
  repeats the full summary fetch inside the wait loop and filters target groups in memory.
- OpenAPI evidence: `GET /fabrics/{fabricName}/softwareUpdatePlan/summary` includes optional query parameter `updateGroupName`.

**Impact**

On large fabrics, a prepare task for one update group can repeatedly transfer and parse the entire fabric update plan every `wait_interval` seconds for up to `wait_timeout`. That is avoidable controller load and task latency during the longest-running path in this module.

**Suggested fix**

Model the summary endpoint's `updateGroupName` query parameter with the shared query-params pattern, and use it when the request can be scoped safely. If the API only accepts one group at a time, use scoped reads for single-group prepares and document or test why multi-group prepares intentionally fall back to the fabric-wide summary.

---

plugins/modules/nd_fabric_prepare_update.py
```diff
@@ -0,0 +1,219 @@
...
+            update_groups,
+            timeout=module.params["wait_timeout"],
+            interval=module.params["wait_interval"],
+        )
+
+    after = orchestrator.status_snapshot(update_groups, summary=final_summary)
+    return orchestrator.results, {"changed": True, "before": before, "after": after}
+
+
+def main():
```
mikewiebe, line 211: ### Low: Module-wrapper behavior has no focused unit coverage

**Issue**

The PR adds strong model and orchestrator unit tests, but there is no `tests/unit/modules/test_nd_fabric_prepare_update.py` coverage for the module wrapper that owns the argspec, dependency check, check-mode handoff, and parameter validation surface.

**Evidence**

- [`plugins/modules/nd_fabric_prepare_update.py` lines 194-206](https://github.com/CiscoDevNet/ansible-nd/pull/291/files#diff-bf975127beb2e5715317b385312125224e9f0924145cf7193b491e94c60e7b86R194)
  hand-builds the module argspec and calls `require_pydantic(module)`.
- [`tests/unit/module_utils/orchestrators/test_fabric_prepare_update.py` lines 415-445](https://github.com/CiscoDevNet/ansible-nd/pull/291/files#diff-ab16e9d0b551791ffd22fcb5451eee7dfeca48ba51e8623e81d1c1147c4b31c1R415)
  covers the orchestrator stage payload, but not Ansible argument validation or module-wrapper behavior.
- [`tests/integration/targets/nd_fabric_prepare_update/tasks/prepare.yaml` lines 25-74](https://github.com/CiscoDevNet/ansible-nd/pull/291/files#diff-a9a51f81334fc9abe60ea252e47269d183c80f53f59a85fd98f746ceadc240c8R25)
  covers end-to-end check mode, stage, and idempotency paths, but not cheap edge cases like empty lists or invalid intervals.
- Changed-file evidence: PR #291 adds no `tests/unit/modules/test_nd_fabric_prepare_update.py` file.

**Existing PR overlap**

**Impact**

The current test suite validates the deeper orchestration behavior well, but the user-facing Ansible contract can still drift. The empty `update_groups` and unconstrained `wait_interval` issues are both wrapper-surface defects that focused unit tests would catch quickly without requiring a live Nexus Dashboard fabric.

**Suggested fix**

Add a small module-wrapper unit test file that exercises argspec constraints, `require_pydantic(module)`, check-mode return behavior, and invalid parameter cases. Keep the existing orchestrator/model tests for REST and parsing behavior; this is complementary coverage, not a replacement.

plugins/module_utils/orchestrators/fabric_prepare_update.py
```diff
@@ -0,0 +1,431 @@
...
+
+        ### RuntimeError
+
+        - If a requested update group is missing from the summary.
+        - If any switch reports a staging or validation failure.
+        - If the summary poll fails more than `_MAX_CONSECUTIVE_POLL_FAILURES` times in a row.
+        - If staging does not complete within `timeout` seconds.
+        """
+        deadline = time.monotonic() + max(timeout, 0)
+        consecutive_failures = 0
```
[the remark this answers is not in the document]
allenrobel: Fixed in aba01553 — `wait_for_completion` now clamps `interval = max(interval, 0)` at the top (mirroring the existing `max(timeout, 0)`), so a negative `wait_interval` becomes a no-sleep retry bounded by the deadline instead of raising from `time.sleep`.

plugins/modules/nd_fabric_prepare_update.py
```diff
@@ -0,0 +1,207 @@
...
+- name: Prepare multiple update groups with a longer timeout
+  cisco.nd.nd_fabric_prepare_update:
+    fabric_name: SITE1
+    update_groups:
+      - SITE1_N9K_leaf
+      - SITE1_N9K_spine
+    wait: true
+    wait_timeout: 3600
+    wait_interval: 60
+    state: merged
```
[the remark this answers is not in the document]
allenrobel: Fixed in aba01553 — lowered the multi-group example to `wait_interval: 20` so it no longer contradicts the "<30s `persistent_command_timeout`" guidance in the docs.

plugins/modules/nd_fabric_prepare_update.py
```diff
@@ -0,0 +1,207 @@
...
+    )
+
+    module = AnsibleModule(argument_spec=argument_spec, supports_check_mode=True)
+    require_pydantic(module)
+
+    try:
+        result = _run_prepare(module)
+        module.exit_json(**result)
+    except Exception as e:
+        module.fail_json(msg=f"Module execution failed: {str(e)}")
```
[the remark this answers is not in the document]
allenrobel: This is already handled in the current code: `main()` calls `module.fail_json(msg=..., **output.format())`, so structured output is included on failure. The thread is outdated against the present `main()`.

plugins/modules/nd_fabric_prepare_update.py
allenrobel: Fixed in aba01553 — added a `_validate_update_groups` guard in the module wrapper that fails when `update_groups` is empty (Ansible's `required=True` only guarantees the key is present, not that the list is non-empty). Kept it at the module surface rather than the model so the failure is a clean Ansible error; covered by a new module-wrapper unit test.

plugins/module_utils/orchestrators/fabric_prepare_update.py
allenrobel: Changed in aba01553 to return True. A switch-less snapshot is now treated as vacuously prepared (`changed=False`), which matches `wait_for_completion` treating a switch-less group as already satisfied. Previously a switch-less group reported `changed=True` and fired a no-op stage POST. Parametrized `snapshot_fully_prepared` tests updated to expect True for the empty / no-switches cases.

plugins/module_utils/endpoints/v1/manage/software_update_plan_summary.py
allenrobel: Fixed in aba01553. Modeled the `updateGroupName` query parameter on `EpFabricSoftwareUpdatePlanSummary` via `EndpointQueryParams` (`SoftwareUpdatePlanSummaryEndpointParams`). `get_summary(update_group_name=...)` now scopes the read, and the single-group prepare path — the common case, including every wait poll — uses it (`scope_for()` picks the single name). Multi-group prepares intentionally keep the one fabric-wide GET and filter in memory: the param accepts a single group at a time, so looping per-group would be N requests per poll, worse than one fabric-wide fetch and a no-N+1 violation. Added endpoint unit tests for the scoped vs. unscoped path.

plugins/modules/nd_fabric_prepare_update.py
allenrobel: Added in aba01553 — `tests/unit/modules/test_nd_fabric_prepare_update.py` covers the argspec surface (`_validate_update_groups` empty-list guard) and the `_run_prepare` decision surface: idempotent short-circuit (no stage), check-mode handoff (changed without staging), and the wait / no-wait branches, with the orchestrator methods monkeypatched so no controller I/O runs. The existing orchestrator/model tests are kept as complementary coverage.

plugins/modules/nd_fabric_prepare_update.py
allenrobel: Sorry about the single commit addressing all comments.  I've instructed Claude that we want single focused diff per comment in the future.

plugins/module_utils/orchestrators/fabric_prepare_update.py
allenrobel: Sorry about the single commit addressing all comments.  I've instructed Claude that we want single focused diff per comment in the future.

plugins/module_utils/endpoints/v1/manage/software_update_plan_summary.py
allenrobel: Sorry about the single commit addressing all comments.  I've instructed Claude that we want single focused diff per comment in the future.

plugins/modules/nd_fabric_prepare_update.py
allenrobel: Sorry about the single commit addressing all comments.  I've instructed Claude that we want single focused diff per comment in the future.

plugins/modules/nd_fabric_prepare_update.py
allenrobel: Resolving after 1 week.

plugins/module_utils/orchestrators/fabric_prepare_update.py
allenrobel: Resolving after 1 week.

plugins/module_utils/endpoints/v1/manage/software_update_plan_summary.py
allenrobel: Resolving after 1 week.

plugins/modules/nd_fabric_prepare_update.py
allenrobel: Resolving after 1 week.

akinross (approved):

gmicol (approved):

plugins/module_utils/models/fabric_prepare_update/software_update_plan_summary.py
```diff
@@ -0,0 +1,95 @@
...
+    A single update group's status within the software update plan summary, including its member
+    switches and any advisory warnings ND raises for the group.
+
+    ## Raises
+
+    None
+    """
+
+    update_group_name: str | None = Field(default=None, alias="updateGroupName")
+    update_group_status: str | None = Field(default=None, alias="updateGroupStatus")
```
akinross, line 93: does the same free-form [comment](https://github.com/CiscoDevNet/ansible-nd/pull/291/changes#diff-f55ea7281469d4e7afb25d6699792a4dca41fa3b42abe8d52989e655f550d915R42-R44) about string apply to update_group_status?

plugins/module_utils/models/fabric_prepare_update/software_update_plan_summary.py
```diff
@@ -0,0 +1,95 @@
...
+class SwitchStageStatusModel(NDNestedModel):
+    """
+    # Summary
+
+    Per-switch stage / validate status within an update group, as reported by the
+    `softwareUpdatePlan/summary` endpoint.
+
+    `image_staged_status` and `image_validated_status` are kept as free-form strings (not a
+    `Literal`) so an unrecognized status from a newer ND release parses rather than raising during
+    a poll. Known values: `none`, `inProgress`, `success`, `failed`, `skipped`.
```
akinross: Did you consider a Literal and TypeAlias combination for this case, in a way that known values for the status are Literals and the states can contain that TypeAlias or a string?

```
from typing import Literal, TypeAlias

KnownImageStatus: TypeAlias = Literal[
    "none", "inProgress", "success", "failed", "skipped",
]
ImageStatus: TypeAlias = KnownImageStatus | str
```

At runtime that is effectively str, but it makes the known vocabulary visible to readers and can generate a schema with both the enum branch and the general-string fallback.

plugins/module_utils/orchestrators/fabric_prepare_update.py
```diff
@@ -0,0 +1,471 @@
...
+        interval = max(interval, 0)
+        # Scope the (repeated) poll to a single update group when possible; multi-group prepares
+        # fetch the fabric-wide summary once per poll and filter in memory.
+        scope = self.scope_for(update_group_names)
+        consecutive_failures = 0
+        while True:
+            try:
+                summary = self.get_summary(update_group_name=scope)
+                groups = self._resolve_groups(summary, update_group_names)
+                switches = [sw for group in groups for sw in (group.update_group_switches or [])]
```
akinross: I assume we need the `or []` because group.update_group_switches can be None according to `update_group_switches: list[SwitchStageStatusModel] | None = Field(default=None, alias="updateGroupSwitches")`. 

I am confused on when the case would be None, is this when the field is missing from response (since the model is only used for response as documented)? Should we simplify the model to not write None and default to an empty list? Another option would be to error when updateGroupSwitches is not part of the response, assuming that would lead to None?

plugins/module_utils/orchestrators/fabric_prepare_update.py
```diff
@@ -0,0 +1,471 @@
...
+                    raise RuntimeError(
+                        f"Timed out after {timeout}s waiting for staging of update group(s) {update_group_names} "
+                        f"in fabric '{self.fabric_name}' to complete; last poll error: {e}"
+                    ) from e
+                if consecutive_failures > _MAX_CONSECUTIVE_POLL_FAILURES:
+                    raise RuntimeError(
+                        f"Polling staging status for update group(s) {update_group_names} in fabric "
+                        f"'{self.fabric_name}' failed {consecutive_failures} times in a row: {e}"
+                    ) from e
+                time.sleep(interval)
```
akinross: When the remaining timeout is shorter than wait_interval, the loop still sleeps for the full interval before checking the deadline again. For example, wait_timeout=65 and wait_interval=30 can time out around 90 seconds. I think we should wait for a maximum of the remaining time until deadline, so in the example wait 30, 30 and 5 instead of 30, 30, 30. 

same applies to the timeout below

plugins/module_utils/orchestrators/fabric_prepare_update.py
```diff
@@ -0,0 +1,471 @@
...
+        # Scope the (repeated) poll to a single update group when possible; multi-group prepares
+        # fetch the fabric-wide summary once per poll and filter in memory.
+        scope = self.scope_for(update_group_names)
+        consecutive_failures = 0
+        while True:
+            try:
+                summary = self.get_summary(update_group_name=scope)
+                groups = self._resolve_groups(summary, update_group_names)
+                switches = [sw for group in groups for sw in (group.update_group_switches or [])]
+            except Exception as e:  # pylint: disable=broad-except
```
akinross, line 431: Do we want to retry pydantic validation errors that potentially could be raised in this try/except block? I assume the chance we hit those are small, but they should not be retried in my opinion.

plugins/modules/nd_fabric_prepare_update.py
```diff
@@ -0,0 +1,237 @@
...
+    - The list of update group names to prepare.
+    - Each named update group must already exist in O(fabric_name). Create update groups with M(cisco.nd.nd_fabric_update_group).
+    type: list
+    elements: str
+    required: true
+  wait:
+    description:
+    - Whether to wait for staging to complete on every switch before returning.
+    - When V(true), the module polls Nexus Dashboard until every switch has staged and validated, or until O(wait_timeout) is reached.
+    - When V(false), the module returns as soon as Nexus Dashboard accepts the stage action; staging continues on Nexus Dashboard.
```
akinross: I think we need to reword this because the module still performs the status snapshot for after in results, so it does not return immediately.

plugins/modules/nd_fabric_prepare_update.py
```diff
@@ -0,0 +1,237 @@
...
+  configured image, the module reports no change and does not stage again.
+author:
+- Allen Robel (@allenrobel)
+options:
+  fabric_name:
+    description:
+    - The name of the fabric that contains the update groups to prepare.
+    type: str
+    required: true
+  update_groups:
```
akinross, line 37: should we document the validation rule that at least on entry is required?

plugins/modules/nd_fabric_prepare_update.py
```diff
@@ -0,0 +1,237 @@
...
+- Preparing is a long-running, asynchronous operation. By default the module waits for staging to complete on
+  every switch before returning.
+- If every switch in every requested update group is already staged and validated for the update group's
+  configured image, the module reports no change and does not stage again.
+author:
+- Allen Robel (@allenrobel)
+options:
+  fabric_name:
+    description:
+    - The name of the fabric that contains the update groups to prepare.
```
akinross, line 33: should we add a similar line as https://github.com/CiscoDevNet/ansible-nd/pull/291/changes#diff-bf975127beb2e5715317b385312125224e9f0924145cf7193b491e94c60e7b86R39 for fabric name?

plugins/modules/nd_fabric_prepare_update.py
```diff
@@ -0,0 +1,237 @@
...
+    """
+    sender = Sender()
+    sender.ansible_module = module
+    rest_send_params = dict(module.params)
+    rest_send_params["check_mode"] = module.check_mode
+    rest_send = RestSend(rest_send_params)
+    rest_send.sender = sender
+    rest_send.response_handler = ResponseHandler()
+
+    orchestrator = FabricPrepareUpdateOrchestrator(rest_send=rest_send, results=Results())
```
akinross, line 177: not wrong but more stylistic inconsistency question as to why some parts are assigned to variable first but FabricPrepareUpdateOrchestrator is constructed directly with the objects and functions.

```suggestion
    rest_send = RestSend({**module.params, "check_mode": module.check_mode})
    rest_send.sender = Sender(ansible_module=module)
    rest_send.response_handler = ResponseHandler()

    orchestrator = FabricPrepareUpdateOrchestrator(rest_send=rest_send, results=Results())
```

plugins/modules/nd_fabric_prepare_update.py
```diff
@@ -0,0 +1,237 @@
...
+    choices: [ merged ]
+extends_documentation_fragment:
+- cisco.nd.modules
+- cisco.nd.check_mode
+notes:
+- This module is only supported on Nexus Dashboard 4.2.1 or higher.
+- The target image, whether the upgrade is disruptive or non-disruptive, maintenance mode, and the report checks
+  are properties of the update group itself, not of this module. Configure them with M(cisco.nd.nd_fabric_update_group)
+  before preparing. This module only selects which update groups to prepare.
+- In check mode, when staging is required, the module reports a change but does not stage, because the stage
```
akinross: I would slight reword the description in check mode

```suggestion
- In check mode, if staging is required, the module reports changed=true but does not invoke the stage action because it cannot be previewed. Consequently, the returned after status is unchanged from before.
```

akinross (changes requested):

plugins/module_utils/models/fabric_prepare_update/software_update_plan_summary.py
allenrobel: Yes — same rationale: an unrecognized group-level status from a newer ND release should parse rather than raise mid-poll. Now documented on `UpdateGroupStatusModel` in "Type per-switch image statuses as ImageStatus alias; document free-form update_group_status".

plugins/module_utils/models/fabric_prepare_update/software_update_plan_summary.py
allenrobel: Adopted for the per-switch statuses in "Type per-switch image statuses as ImageStatus alias; document free-form update_group_status": `KnownImageStatus` Literal + `ImageStatus: TypeAlias = KnownImageStatus | str`, with the docstring pointing at the alias instead of listing values in prose. `update_group_status` stays a bare `str` — its vocabulary isn't lab-confirmed, so there's nothing reliable to enumerate yet.

plugins/module_utils/orchestrators/fabric_prepare_update.py
allenrobel: The `None` case was exactly "key absent from the response body". Since every consumer coalesced `None` to `[]`, the distinction carried no information — the three list fields now default to `[]` via `default_factory` and the `or []` sites are gone ("Default summary-model list fields to empty lists instead of None"). I didn't take the error-on-absence option: ND 4.2.x omits keys in places, so absence isn't a trustworthy failure signal for a poll loop.

plugins/module_utils/orchestrators/fabric_prepare_update.py
allenrobel: Fixed in "Cap wait_for_completion poll sleeps at the remaining deadline" — both sleep sites now sleep `min(interval, remaining-time-to-deadline)`, so the loop takes one final poll at the deadline. Your 65/30 example polls at 0/30/60/65 and fails at ~65s instead of ~90s.

plugins/module_utils/orchestrators/fabric_prepare_update.py
allenrobel: Left as-is deliberately — open to changing if you disagree. A ValidationError from a real contract break recurs on every poll, so the retry budget aborts it within `_MAX_CONSECUTIVE_POLL_FAILURES` (3) polls with the original error chained; the cost is at most two extra intervals. The transient case this loop defends against (a partial body during failover) can trip validation just as easily as an unresolved group, and retrying is exactly what we want there. In practice ValidationError is nearly unreachable here: all fields are optional with extra ignored, so only a wire-type change (e.g. an int arriving as an object) can raise.

plugins/modules/nd_fabric_prepare_update.py
allenrobel: Reworded in "Docs: clarify wait=false return, check-mode wording, required update_groups, fabric preexistence": the module does not wait for staging to complete; it records one post-stage status snapshot (returned as the task's after status) and returns while staging continues on Nexus Dashboard.

plugins/modules/nd_fabric_prepare_update.py
allenrobel: Added in "Docs: clarify wait=false return, check-mode wording, required update_groups, fabric preexistence": "At least one update group name is required; an empty list fails validation."

plugins/modules/nd_fabric_prepare_update.py
allenrobel: Added in the same docs commit: "The fabric must already exist on Nexus Dashboard." I stopped short of a create-module pointer like the update_groups line has, since fabric creation spans several modules depending on fabric type.

plugins/modules/nd_fabric_prepare_update.py
allenrobel: Kept the property-assignment form for consistency: this block matches the wiring in nd_fabric_update_group (this feature family's sibling), nd_state_machine, and nd_manage_policy_group verbatim, and the same idiom is the collection's unit-test setup pattern. Your version is valid — `Sender` does accept `ansible_module` as a constructor kwarg — but converting one module would create the cross-module inconsistency this comment is about. A collection-wide sweep to constructor-style wiring would be a reasonable follow-up outside this PR.

plugins/modules/nd_fabric_prepare_update.py
allenrobel: Took your wording (with doc markup) in "Docs: clarify wait=false return, check-mode wording, required update_groups, fabric preexistence": "In check mode, if staging is required, the module reports C(changed=true) but does not invoke the stage action, because it cannot be previewed. Consequently, the returned after status is unchanged from before."

akinross (approved):

mikewiebe (approved):

allenrobel:
Merging on approvals from @mikewiebe and @akinross

PR #294, opened 2026-05-27
Add shared module utility support

### LINKED ISSUE #450: [ND 4.x] Add shared module utility support for nested resources
## Summary

Add the shared state-machine and utility behavior needed for nested ND 4.x resources such as VRF-Lite attachments and links.

## Related PR

- #294

## Scope

- Allow `state: merged` comparisons to ignore additional controller fields that were not supplied in the playbook.
- Support correct comparison of nested list resources without false changes.
- Track delete operations so follow-up config-save and deploy actions know which resources changed.
- Ensure check mode and skipped or failed operations do not produce false deploy triggers.
- Add focused unit coverage for the shared comparison and state-machine behavior.

## Acceptance Criteria

- Nested resources remain idempotent when controller responses include additional fields.
- List matching produces deterministic and correct merged-state results.
- Only operations actually sent to Nexus Dashboard are tracked for follow-up actions.
- Check mode does not trigger save or deploy operations.
- Relevant unit tests and sanity validation pass.

## Dependency

This shared framework work is required by the VRF-Lite implementation in PR #281.

sivakasi-cisco:
## Related Issue(s)

Fixes #450

**Problem Description**

This PR separates out a few shared framework changes that are needed before modules like `nd_manage_vrf_lite` can cleanly follow the generic ND 4.x module architecture - https://github.com/CiscoDevNet/ansible-nd/pull/281.

Some ND 4.x resources are simple: one playbook item maps directly to one controller object. In those cases, the existing generic state machine can compare “what the user wants” with “what exists on the controller” and decide create, update, delete, or no-op.

But some resources are nested. VRF Lite is a good example: the user gives VRF-level config, but the actual work happens under smaller child objects like VRF attachments and VRF Lite links. Without small shared framework improvements, each nested module has to write its own custom comparison logic.

This PR adds the common support needed for that style of module.

**Why These Common Changes Are Needed**

This PR updates the common state machine and utility logic used by the VRF Lite workflow.

The main change is to make `merged` state behave correctly when the controller already has extra data that the user did not provide in the playbook. In that case, the module should not report a change just because existing data has additional fields.

This PR also tracks deleted items in the state machine, so follow-up actions like config save and deploy can know which VRFs were changed by a delete operation.


plugins/module_utils/nd_state_machine.py
```diff
@@ -195,6 +200,15 @@ def _manage_override_deletions(self) -> None:
         items_to_delete = [existing_item for identifier in diff_identifiers if (existing_item := self.existing.get(identifier)) is not None]
         self._delete_items(items_to_delete)
 
+    def _manage_replace_deletions(self) -> None:
```
sivakasi-cisco: Kindly comment on "replace" state if it is okay

plugins/module_utils/common/data.py
```diff
@@ -0,0 +1,76 @@
+# -*- coding: utf-8 -*-
```
akinross: what is the difference between data.py and the utils.py, they both seem helper functions is there a reason to add this into a separate file/location?

plugins/module_utils/common/data.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def loads_maybe_json(value: Any) -> Any:
+    """Parse JSON or Python-literal strings while leaving parsed values unchanged."""
+    if isinstance(value, (dict, list)):
+        return value
+    if value is None:
+        return None
+
+    text = str(value).strip()
```
akinross: should we wrap this into a try/except to be sure the str() works?

plugins/module_utils/common/data.py
akinross: please add unit tests for all the helper functions

plugins/module_utils/common/data.py
```diff
@@ -0,0 +1,76 @@
...
+    if not text:
+        return None
```
akinross: why return None for an empty text (assuming here text is always a string) and not an empty string to preserve it's type?

plugins/module_utils/common/data.py
```diff
@@ -0,0 +1,76 @@
...
+        return source
+
+    params = getattr(source, "params", None)
+    if isinstance(params, dict):
+        return params
+
+    return {}
+
+
+def loads_maybe_json(value: Any) -> Any:
```
akinross: Why name the function loads_maybe_json when it could be json or python literal string? I think we should make this more naming more specific, the function more generic or the function more specific. 

Personally i would make the function more generic and extensible by calling it parse_value, where the function takes a value and a list of parsers like json.loads and ast.literal_eval, this way if you want to extend parsing capabilities in the future you should be able to leverage the function.

plugins/module_utils/common/data.py
```diff
@@ -0,0 +1,76 @@
...
+        return None
+
+    text = str(value).strip()
+    if not text:
+        return None
+
+    try:
+        return json.loads(text)
+    except Exception:
+        try:
```
akinross: nested try/except are typically a code smell, think we should use a loop alternative

plugins/module_utils/common/data.py
```diff
@@ -0,0 +1,76 @@
...
+            return ast.literal_eval(text)
+        except Exception:
+            return None
+
+
+def coerce_dict_list(data: Any, list_keys: tuple[str, ...] = ("DATA", "data", "items")) -> list[dict[str, Any]]:
```
akinross: should this function also support deeper levels of nesting?
can you document the restrictions of this function more clearly?

plugins/module_utils/common/data.py
```diff
@@ -0,0 +1,76 @@
...
+    if isinstance(data, dict):
+        for key in list_keys:
+            value = data.get(key)
+            if isinstance(value, list):
+                return [item for item in value if isinstance(item, dict)]
+
+    return []
+
+
+def copy_dict_items(items: Any) -> list[dict[str, Any]]:
```
akinross: we should a stricter type hinting on items, probably something like `Optional[Iterable[Any]]` or `Iterable[Mapping[str, Any] | BaseModel]`

plugins/module_utils/nd_state_machine.py
```diff
@@ -69,14 +67,18 @@ def __init__(self, module: AnsibleModule, model_orchestrator: Union[Type[NDBaseO
         # Initialize collections
         try:
             response_data = self.model_orchestrator.query_all()
+            config_data = self.module.params.get("config", [])
```
akinross: dict.get(key, default) only returns default when the key is missing. If config is explicitly present with value None, you get None back — not []. is this intended else change to below

```suggestion
            config_data = self.module.params.get("config") or []
```

akinross (changes requested):

plugins/module_utils/common/data.py
sivakasi-cisco: Sure

plugins/module_utils/common/data.py
sivakasi-cisco: In data.py, the helpers are about normalizing controller-returned data and serialized values. In utils.py currently has more generic dict/diff helpers like 'issubset' and key removal. 

As there is no technical blocker, if preferred, they can be combined.

sivakasi-cisco: Added

plugins/module_utils/common/data.py
sivakasi-cisco: Agreed

plugins/module_utils/common/data.py
sivakasi-cisco: For this use case, an empty string means “nothing useful to parse” from the controller, so 'None' is the safer default for the current callers. I made that behaviour explicit through the 'default' argument, so a future caller can preserve "" by passing default="".

plugins/module_utils/common/data.py
sivakasi-cisco: I refactored this into a parser loop so each parser gets a chance in order, and the function returns the default only if all parsers fail.

plugins/module_utils/common/data.py
sivakasi-cisco: Added

plugins/module_utils/common/data.py
sivakasi-cisco: I have updated the type hint to Optional[Iterable[Any]] and changed the dict check to 'Mapping' so dict-like objects are handled more accurately without coupling this helper directly to a specific BaseModel type.

plugins/module_utils/nd_state_machine.py
sivakasi-cisco: Applied

sivakasi-cisco (commented):
Addressed your comments. 

For the comment on data.py and utils.py files, I can combine them if required.

You can consider the latest changes for review and subsequent approval.

plugins/module_utils/nd_state_machine.py
```diff
@@ -35,25 +34,24 @@ def __init__(self, module: AnsibleModule, model_orchestrator: Union[Type[NDBaseO
         sender = Sender()
         sender.ansible_module = self.module
 
-        rest_send_params = dict(self.module.params)
-        rest_send_params["check_mode"] = self.module.check_mode
-        self.rest_send = RestSend(rest_send_params)
```
samiib, line 39: Was the intention of this change to strip out all the RestSend params down to just `check_most` and `state`?
I can see in other parts of the code that `rest_send.params.get("fabric_name")` is used.
If all the params get removed then fabric_name will always be None. Modules like nd_interface_loopback could be broken after this change. The unit tests won't be able to pick this up due to unit isolation. Has this PR been integration tested yet?

samiib (changes requested):

plugins/module_utils/nd_state_machine.py
sivakasi-cisco: Good catch. I did not intend to strip the module params from RestSend, instead tried to preserve the runtime check_mode/state values. I missed that interface orchestrators still use rest_send.params as shared module context, including fabric_name and config. It is required.

I’ll restore passing a copy of module.params with check_mode and interface so this path is covered.

The integration tests for VRFLite is checked against, though it had not flagged here as VRFLite mostly reads module.params directly

sivakasi-cisco (commented):
Resolved the comments

akinross (dismissed):

gmicol (dismissed):
LGTM

allenrobel:
### Code review

Found 4 issues:

1. `_delete_items` adds items to `self.sent` unconditionally. `_execute_operation` returns `None` without calling the API in check_mode, and orchestrators may skip items during delete (see #307 for the same unconditional pattern with `existing.delete_many`), yet all `items` are still added to `sent`. This contradicts the guard used for creates/updates in `_manage_create_update_state` ("Mark as sent only after successful API operations", L182-L185). Since the PR's stated purpose for `sent` is to drive follow-up config save/deploy, check_mode runs and orchestrator-skipped items would produce false deploy triggers.

https://github.com/CiscoDevNet/ansible-nd/blob/28f6cdcb17559ee49d3d1122655819e79c0b8c2d/plugins/module_utils/nd_state_machine.py#L217-L221

2. The new `issubset(allow_superset=True)` path has no unit tests. It relaxes the bidirectional list-element match that was deliberately added as a bug fix in #209, and the final diff contains no test files (the tests added earlier in this PR were removed along with `common/data.py`). An earlier review comment on this PR asked for unit tests for all the helper functions; that still applies to `issubset` and to `get_diff(exclude_unset=True)` with list-valued fields.

https://github.com/CiscoDevNet/ansible-nd/blob/28f6cdcb17559ee49d3d1122655819e79c0b8c2d/plugins/module_utils/utils.py#L32-L43

3. Greedy first-match list matching can report false diffs with `allow_superset=True`. Each subset item consumes the first matching candidate (`del remaining[index]`), so a less-specific proposed item can consume a candidate needed by a more-specific one. Example: proposed `[{"a": 1}, {"a": 1, "b": 2}]` vs existing `[{"a": 1, "b": 2}, {"a": 1, "b": 3}]` — `{"a": 1}` consumes `{"a": 1, "b": 2}`, then `{"a": 1, "b": 2}` fails against `{"a": 1, "b": 3}`, reporting a diff although a valid pairing exists. One-directional matching makes this ambiguity much more likely than the bidirectional default. Result is spurious `changed=True` (broken idempotency) in merged state for lists of similar dicts — exactly the shape of VRF attachments.

https://github.com/CiscoDevNet/ansible-nd/blob/28f6cdcb17559ee49d3d1122655819e79c0b8c2d/plugins/module_utils/utils.py#L52-L64

4. The new docstring claim "When False (default) both directions are required, which is equivalent to equality for lists of dicts" is inaccurate: the dict branch skips `None`-valued keys (L68-L69), so `{"a": 1, "b": None}` and `{"a": 1}` match bidirectionally despite not being equal. Harmless for current callers (`to_diff_dict` uses `exclude_none=True`) but misleading for future direct callers of `issubset`.

https://github.com/CiscoDevNet/ansible-nd/blob/28f6cdcb17559ee49d3d1122655819e79c0b8c2d/plugins/module_utils/utils.py#L38-L43

plugins/module_utils/models/base.py
```diff
@@ -204,11 +204,14 @@ def get_diff(self, other: "NDBaseModel", exclude_unset: bool = False) -> bool:
             exclude_unset: When True, only compare fields explicitly set in
                 ``other`` (via Pydantic's ``exclude_unset``). This prevents
                 default values from triggering false diffs during merge
-                operations.
+                operations.  List elements are matched one-directionally so
```
samiib:
```suggestion
                operations. List elements are matched one-directionally so
```

plugins/module_utils/models/base.py
```diff
@@ -204,11 +204,14 @@ def get_diff(self, other: "NDBaseModel", exclude_unset: bool = False) -> bool:
...
+                that an existing item with extra fields (e.g. ``deploy``) does
+                not trigger a spurious diff when the proposed item omits those
+                fields.
         """
         self_data = self.to_diff_dict()
         other_data = other.to_diff_dict(exclude_unset=exclude_unset)
-        return issubset(other_data, self_data)
+        return issubset(other_data, self_data, allow_superset=exclude_unset)
```
samiib: Are `exclude_unset` and `allow_superset` required to be coupled together? Would it be better to separate these concerns instead? Perhaps just having `allow_superset` as an argument to `get_diff`? This would allow callers to opt into each behaviour separately.

sivakasi-cisco:
> > ### Code review
> > Found 4 issues:
> > 
> > 1. `_delete_items` adds items to `self.sent` unconditionally. `_execute_operation` returns `None` without calling the API in check_mode, and orchestrators may skip items during delete (see [NDStateMachine._delete_items removes orchestrator-skipped items from existing, corrupting 'after' state #307](https://github.com/CiscoDevNet/ansible-nd/issues/307) for the same unconditional pattern with `existing.delete_many`), yet all `items` are still added to `sent`. This contradicts the guard used for creates/updates in `_manage_create_update_state` ("Mark as sent only after successful API operations", L182-L185). Since the PR's stated purpose for `sent` is to drive follow-up config save/deploy, check_mode runs and orchestrator-skipped items would produce false deploy triggers.
> > 
> > https://github.com/CiscoDevNet/ansible-nd/blob/28f6cdcb17559ee49d3d1122655819e79c0b8c2d/plugins/module_utils/nd_state_machine.py#L217-L221
> > 
**Reply:** Fixed — only items actually pushed are marked as sent now, so check-mode and skipped/failed items no longer trigger a deploy.
> > 
> > 2. The new `issubset(allow_superset=True)` path has no unit tests. It relaxes the bidirectional list-element match that was deliberately added as a bug fix in [Ansible ND 4.X Fabric Modules for iBGP, eBGP and External Fabric Types #209](https://github.com/CiscoDevNet/ansible-nd/pull/209), and the final diff contains no test files (the tests added earlier in this PR were removed along with `common/data.py`). An earlier review comment on this PR asked for unit tests for all the helper functions; that still applies to `issubset` and to `get_diff(exclude_unset=True)` with list-valued fields.
> > 
> > https://github.com/CiscoDevNet/ansible-nd/blob/28f6cdcb17559ee49d3d1122655819e79c0b8c2d/plugins/module_utils/utils.py#L32-L43
> > 
> > 
**Reply:** Added tests/unit/module_utils/test_utils.py covering 'issubset' and 'get_diff'


> > 3. Greedy first-match list matching can report false diffs with `allow_superset=True`. Each subset item consumes the first matching candidate (`del remaining[index]`), so a less-specific proposed item can consume a candidate needed by a more-specific one. Example: proposed `[{"a": 1}, {"a": 1, "b": 2}]` vs existing `[{"a": 1, "b": 2}, {"a": 1, "b": 3}]` — `{"a": 1}` consumes `{"a": 1, "b": 2}`, then `{"a": 1, "b": 2}` fails against `{"a": 1, "b": 3}`, reporting a diff although a valid pairing exists. One-directional matching makes this ambiguity much more likely than the bidirectional default. Result is spurious `changed=True` (broken idempotency) in merged state for lists of similar dicts — exactly the shape of VRF attachments.
> > 
> > https://github.com/CiscoDevNet/ansible-nd/blob/28f6cdcb17559ee49d3d1122655819e79c0b8c2d/plugins/module_utils/utils.py#L52-L64
> > 
> > **Reply:**: Sure. Swapped the greedy first-match for proper bipartite matching so as to find a valid pairing when it exists. Added a regression test as well.

> > 4. The new docstring claim "When False (default) both directions are required, which is equivalent to equality for lists of dicts" is inaccurate: the dict branch skips `None`-valued keys (L68-L69), so `{"a": 1, "b": None}` and `{"a": 1}` match bidirectionally despite not being equal. Harmless for current callers (`to_diff_dict` uses `exclude_none=True`) but misleading for future direct callers of `issubset`.
> > 
> > https://github.com/CiscoDevNet/ansible-nd/blob/28f6cdcb17559ee49d3d1122655819e79c0b8c2d/plugins/module_utils/utils.py#L38-L43
> > 

**Reply:**: As the dict comparison ignores 'None' keys it was not strict ==. Corrected the docstring now.

> > 🤖 Generated with [Claude Code](https://claude.ai/code)
> > 
> > * If this code review was useful, please react with 👍. Otherwise, react with 👎.
>

plugins/module_utils/models/base.py
sivakasi-cisco: Agreed, separated them.

sivakasi-cisco (commented):
Thanks for all the comments. Addressed them all and rebased the branch to develop.

This is focussed towards the major general requirements and I have not handled the bulk delete case. I think it would require more work that could be handled seperately in the issue - https://github.com/CiscoDevNet/ansible-nd/issues/307

allenrobel (dismissed):
### Re-review: all 4 issues from my [earlier review](https://github.com/CiscoDevNet/ansible-nd/pull/294#issuecomment-4674314545) are addressed

Verified against head `75189dc6`.

**1. `_delete_items` marked items `sent` unconditionally → false deploy triggers in check_mode / on skipped items.** Fixed in `nd_state_machine.py`. `_delete_items` now builds a `deleted` list from items whose `_execute_operation` did not fail, `delete_many(...)` receives only those `deleted` items (also closes the #307-style after-state corruption), and the `sent` update is now guarded by `if not self.check_mode and deleted:` — matching the create/update guard. Check-mode runs no longer populate `sent`.

**2. `allow_superset=True` path had no unit tests.** Added `tests/unit/module_utils/test_utils.py` (+305) covering `issubset` with `allow_superset=True` (including candidate non-reuse) and `get_diff(exclude_unset=True)` over list-valued fields.

**3. Greedy first-match could report false diffs under `allow_superset=True`.** Replaced the greedy `del remaining[index]` logic with maximum bipartite matching (`_has_perfect_matching`, Kuhn's augmenting-path). My exact counterexample is now a passing regression test (`test_issubset_one_directional_no_greedy_false_diff`).

**4. Docstring "equivalent to equality" was inaccurate (None-keys skipped).** Docstring corrected to "equivalent to equality *after* `None`-valued keys are dropped from both sides ... not strict `==` equality," and documented by `test_issubset_dicts_none_keys_not_strict_equality`.

Thanks for the thorough follow-up — LGTM.

plugins/module_utils/nd_state_machine.py
allenrobel: Reviewed the `replaced`-state handling at head `75189dc6`. It looks good. 👍

Note this thread is now outdated: the `_manage_replace_deletions()` method this comment was anchored to has been removed, and `replaced` is folded into the unified `_manage_create_update_state()` (the only state-specific deletion path now is `_manage_override_deletions()` for `overridden`). That consolidation is the right call.

Why the current behavior is correct `replaced` semantics:

- `exclude_unset` is False for `replaced` (only `merged` sets it True), so the full proposed model (defaults included) is compared. A field whose model default differs from the controller value is detected and reset on update — proper per-object full replace.
- `allow_superset` is False for `replaced`, so list elements are matched bidirectionally. An existing parent carrying an extra child (e.g. a VRF attachment) is a length mismatch → `changed` → update fires. Nested drift is detected.
- `replaced` does not delete unmentioned resources — that's correctly reserved to `overridden`.

One thing to verify downstream (not in this shared-infra PR): the generic state machine only *detects* nested-child drift and calls `orchestrator.update(proposed)`. Pruning stale child objects (attachments/lite-links on the controller but absent from the replaced parent) is delegated entirely to the orchestrator's `update`. So for VRF Lite (#281), `replaced` is only as correct as that orchestrator's reconciliation — worth confirming `update` removes stale children there.

Resolving this thread since the method it referenced no longer exists.

samiib (dismissed):

tests/unit/module_utils/test_utils.py
```diff
@@ -0,0 +1,304 @@
...
+        ([{"a": 1}], [{"a": 1, "b": 2}], True),
+        # Multiple elements, each a subset of a distinct candidate -> True
+        (
+            [{"a": 1}, {"c": 3}],
+            [{"a": 1, "b": 2}, {"c": 3, "d": 4}],
+            True,
+        ),
+        # Subset element with extra key not in candidate -> still False
+        ([{"a": 1, "z": 9}], [{"a": 1, "b": 2}], False),
+        # Length must still match
```
mikewiebe: **Issue**

The new `allow_superset=True` path allows extra keys inside a matched list element, but it still requires the two lists to have the same length. That does not fully match the PR's stated merged-state goal of ignoring extra existing nested data that the user did not provide.

**Evidence**

- [`plugins/module_utils/utils.py` lines 91-94](https://github.com/CiscoDevNet/ansible-nd/pull/294/files#diff-f3e3720de5c754cfeb714aa1ed0f9d0a1a043f8c0005944d58565decefe1817R91) returns `False` immediately when list lengths differ, even when `allow_superset=True`.
- [`plugins/module_utils/nd_state_machine.py` lines 143-149](https://github.com/CiscoDevNet/ansible-nd/pull/294/files#diff-a9d5c9c64ffd6ae9a3be4b6ff08263cff8995bc962ca9b9da693231f8cba6866R143) enables `allow_superset` for `state: merged`.
- [`plugins/module_utils/nd_config_collection.py` lines 161-163](https://github.com/CiscoDevNet/ansible-nd/pull/294/files#diff-a17f9b688b15df6baa2b44e937f5158ed93276e0496d5154dd37bc3d6fe90149R161) documents `allow_superset` as allowing an existing item to carry extra list elements or extra dict keys.
- [`tests/unit/module_utils/test_utils.py` lines 152-153](https://github.com/CiscoDevNet/ansible-nd/pull/294/files#diff-99d54281a02e6a2ca6df0e406350cd15d6c449b7661676241959a5086b8ea9d4R152) locks in the current behavior by asserting that list length must still match under `allow_superset=True`.
- Local reproduction artifact: `shared_module_utils/pr_294_allow_superset_repro.txt` shows `same_length_extra_keys: True` but `existing_has_extra_list_entry: False`.

**Impact**

For nested resources such as VRF Lite, a `merged` task that names only one child while the controller already has additional children can still report `changed`. Depending on the orchestrator's `update()` behavior, that can create noisy non-idempotency or accidentally push a parent payload that omits controller-side children the user intended to leave alone.

**Suggested fix**

Decide the intended contract and make code, docs, and tests agree. If `merged` should tolerate extra existing child entries, change the list branch so `allow_superset=True` permits `len(subset) <= len(superset)` and still requires a perfect one-to-one match for every proposed item. Add a regression test where the proposed list is shorter than the existing list but every proposed child matches an existing child.

plugins/module_utils/nd_state_machine.py
```diff
@@ -105,17 +105,27 @@ def _execute_operation(
         *args: Any,
         error_msg_prefix: str = "Operation failed",
         **kwargs: Any,
-    ) -> ResponseType | None:
-        """Execute an API operation with standardized error handling."""
+    ) -> bool:
```
mikewiebe, line 162: **Issue**

This PR changes core `NDStateMachine` behavior for operation success tracking, `sent` population, check mode, ignored errors, and delete after-state updates, but the added tests only cover `issubset()` and `NDBaseModel.get_diff()`.

**Evidence**

- [`plugins/module_utils/nd_state_machine.py` lines 108-128](https://github.com/CiscoDevNet/ansible-nd/pull/294/files#diff-a9d5c9c64ffd6ae9a3be4b6ff08263cff8995bc962ca9b9da693231f8cba6866R108) changes `_execute_operation()` to return a Boolean success signal.
- [`plugins/module_utils/nd_state_machine.py` lines 181-203](https://github.com/CiscoDevNet/ansible-nd/pull/294/files#diff-a9d5c9c64ffd6ae9a3be4b6ff08263cff8995bc962ca9b9da693231f8cba6866R181) changes when creates and updates are added to `sent`.
- [`plugins/module_utils/nd_state_machine.py` lines 223-249](https://github.com/CiscoDevNet/ansible-nd/pull/294/files#diff-a9d5c9c64ffd6ae9a3be4b6ff08263cff8995bc962ca9b9da693231f8cba6866R223) changes delete handling so only successfully deleted items are removed from `existing` and added to `sent`.
- [`tests/unit/module_utils/test_utils.py` lines 7-12](https://github.com/CiscoDevNet/ansible-nd/pull/294/files#diff-99d54281a02e6a2ca6df0e406350cd15d6c449b7661676241959a5086b8ea9d4R7) describes the added unit coverage as focused on `utils.py`, `issubset`, and `NDBaseModel.get_diff`.
- A repository search of the extracted PR head found no direct unit test file for `NDStateMachine`, `_execute_operation()`, `_delete_items()`, or the changed `sent`/check-mode behavior.

**Impact**

These are shared framework paths used by every generic model-backed module. A regression here can create false deploy triggers, wrong `changed` output, incorrect `after` state, or missed deletes across multiple modules, and isolated `issubset()` tests will not catch those state-machine wiring failures.

**Suggested fix**

Add focused `NDStateMachine` unit tests with a tiny fake model and fake orchestrator. Cover at least check-mode create/update/delete, ignored delete errors, bulk-delete failure behavior, and the `config=None` normalization case. Keep the existing `test_utils.py` coverage for the pure diff helper.

tests/unit/module_utils/test_utils.py
sivakasi-cisco: Yes, fixed now.

For allow_superset=True with merged state, issubset, with the changeset,  allows the proposed list to be shorter than the controller list. Every proposed child must still match a distinct existing child, but extra existing children are left alone and no longer cause a false changed result. Have updated the older tests as well.

plugins/module_utils/nd_state_machine.py
sivakasi-cisco: Fixed. I added focused NDStateMachine unit tests using a fake model and fake orchestrator.

The tests now cover check-mode create/update/delete behaviour, ignored delete errors, bulk delete failure handling, sent tracking and config=None normalisation.

mikewiebe:
@sivakasi-cisco Please fix the CI tests above that are failing

sivakasi-cisco (commented):
Addressed the review comments

plugins/module_utils/nd_config_collection.py
```diff
@@ -234,11 +237,15 @@ def to_payload_list(self, **kwargs) -> List[Dict[str, Any]]:
         return [item.to_payload(**kwargs) for item in self._items]
 
     @staticmethod
-    def from_ansible_config(data: List[Dict], model_class: type[NDBaseModel], **kwargs) -> "NDConfigCollection":
+    def from_ansible_config(data: Optional[List[Dict]], model_class: type[NDBaseModel], **kwargs) -> "NDConfigCollection":
```
mikewiebe: ### High: `config: null` can turn `state: overridden` into delete-all

**Issue**

The new `from_ansible_config()` normalization deliberately maps an omitted or null `config` to an empty collection. For modules whose `config` is optional, `state: overridden` then treats that accidental empty collection as the complete desired state and deletes every existing resource.

**Evidence**

- [`plugins/module_utils/nd_config_collection.py` lines 240-248](https://github.com/CiscoDevNet/ansible-nd/pull/294/files#diff-a17f9b688b15df6baa2b44e937f5158ed93276e0496d5154dd37bc3d6fe90149R240) explicitly makes `None` and `[]` indistinguishable.
- [`plugins/module_utils/nd_state_machine.py` lines 214-220](https://github.com/CiscoDevNet/ansible-nd/pull/294/files#diff-a9d5c9c64ffd6ae9a3be4b6ff08263cff8995bc962ca9b9da693231f8cba6866R214) computes overridden deletions as every identifier present in `before` but absent from that empty proposed collection.
- [`plugins/module_utils/models/manage_fabric/manage_fabric_base.py` line 167](https://github.com/CiscoDevNet/ansible-nd/blob/aea71148c827dce423394d67c0c759a302f70fdd/plugins/module_utils/models/manage_fabric/manage_fabric_base.py#L167) defines `config` as optional for fabric modules.
- A focused state-machine reproduction with two existing resources and `config=None`, `state=overridden` deleted both resources and left zero in `after`.
- The new test at [`tests/unit/module_utils/test_nd_state_machine_operations.py` line 452](https://github.com/CiscoDevNet/ansible-nd/pull/294/files#diff-a6236cc68a3916a1b6a0451db5eba6dee738577089a523330cfb052a6132597dR452) checks only that null produces an empty collection; it does not run the state-dependent overridden workflow.

**Existing PR overlap**

No matching existing PR comment found. The resolved comment at https://github.com/CiscoDevNet/ansible-nd/pull/294#discussion_r3311396121 requested null normalization, but it did not identify that collapsing `None` into `[]` makes an omitted value destructive under `state: overridden`.

No matching open issue found.

**Impact**

A playbook with an omitted variable or `config: null` can remove every fabric or other state-machine-managed object instead of failing validation. This is materially different from an explicit `config: []`, where delete-all may be intentional.

**Suggested fix**

Preserve the distinction between absent/null and an explicitly empty list. Reject absent or null `config` for `merged`, `replaced`, and `overridden` before any mutation; retain explicit `[]` only where delete-all is a documented contract. Add state-machine tests for omitted, null, and explicit-empty config under every state, including an overridden test with non-empty existing data.

---

plugins/module_utils/nd_config_collection.py
```diff
@@ -169,7 +172,7 @@ def get_diff_config(self, new_item: NDBaseModel, exclude_unset: bool = False) ->
         if existing is None:
             return "new"
 
-        is_subset = existing.get_diff(new_item, exclude_unset=exclude_unset)
+        is_subset = existing.get_diff(new_item, exclude_unset=exclude_unset, allow_superset=allow_superset)
```
mikewiebe, line 185: ### High: the new `get_diff` keyword breaks `nd_maintenance_mode`

**Issue**

@sivakasi-cisco Please check with Allen on this one since he wrote the `nd_maintenance_mode` module

`NDConfigCollection.get_diff_config()` now unconditionally passes `allow_superset` to model overrides, but the existing `MaintenanceModeModel.get_diff()` override still implements the old signature. Any maintenance-mode reconciliation that finds the singleton existing snapshot fails before it can decide whether a change is required.

**Evidence**

- [`plugins/module_utils/nd_config_collection.py` line 175](https://github.com/CiscoDevNet/ansible-nd/pull/294/files#diff-a17f9b688b15df6baa2b44e937f5158ed93276e0496d5154dd37bc3d6fe90149R175) calls `existing.get_diff(..., allow_superset=allow_superset)` for every model.
- [`plugins/module_utils/models/maintenance_mode/maintenance_mode.py` lines 121-143](https://github.com/CiscoDevNet/ansible-nd/blob/aea71148c827dce423394d67c0c759a302f70fdd/plugins/module_utils/models/maintenance_mode/maintenance_mode.py#L121) overrides `get_diff(self, other, exclude_unset=False)` without the new keyword.
- A direct production-model collection reproduction fails with `TypeError: MaintenanceModeModel.get_diff() got an unexpected keyword argument 'allow_superset'`.
- The full merged unit suite passes because maintenance-mode model tests and state-machine tests exercise the components separately; no test composes `NDStateMachine`/`NDConfigCollection` with `MaintenanceModeModel`.

**Existing PR overlap**

No matching existing PR comment found. The resolved state-machine test-coverage thread at https://github.com/CiscoDevNet/ansible-nd/pull/294#discussion_r3445198107 requested direct state-machine tests, but the added tests use a fake base model and do not cover production `get_diff` overrides.

No matching open issue found.

**Impact**

Merging the PR breaks the existing `nd_maintenance_mode` module for normal `merged`, `replaced`, and `overridden` workflows, including idempotent requests.

**Suggested fix**

Update `MaintenanceModeModel.get_diff()` to accept `allow_superset` and either ignore it explicitly or apply the intended custom semantics. 

---

sivakasi-cisco: MaintenanceModeModel.get_diff() now accepts the 'allow_superset' for compatibility with NDConfigCollection. It intentionally ignores it because maintenance mode uses its own per-switch comparison.

plugins/module_utils/nd_config_collection.py
sivakasi-cisco: For the config: null issue, NDStateMachine now rejects missing or null config for merged, replaced and overridden before any mutation.

For the Explicit config: [] is still allowed, so intentional empty config behaviour is preserved.

plugins/module_utils/nd_config_collection.py
mikewiebe, line 185: **High: PR #294 changes a shared method signature that ready PRs #286 and #312 override**

This line changes the call from:

```python
existing.get_diff(new_item, exclude_unset=exclude_unset)
```

to:

```python
existing.get_diff(
    new_item,
    exclude_unset=exclude_unset,
    allow_superset=allow_superset,
)
```

The important detail is that Python calls the `get_diff()` implementation on the concrete model, not necessarily `NDBaseModel.get_diff()`.

Both open module PRs define their own implementation using the previous signature:

```python
# PR #286: PrefixListModel
def get_diff(self, other, exclude_unset=False)

# PR #312: NDLinkModel
def get_diff(self, other, exclude_unset=False)
```

### Runtime failure

The runtime failure is a Python `TypeError` during local diff calculation. For [PR #286](https://github.com/CiscoDevNet/ansible-nd/pull/286), the combined production-model call was reproduced failing with:

```text
TypeError: PrefixListModel.get_diff() got an unexpected keyword argument 'allow_superset'
```

For [PR #312](https://github.com/CiscoDevNet/ansible-nd/pull/312), retaining its current override after resolving the textual merge conflicts produces the equivalent failure:

```text
TypeError: NDLinkModel.get_diff() got an unexpected keyword argument 'allow_superset'
```

The failure sequence is:

1. The state machine asks `NDConfigCollection` to compare a proposed item with an existing item.
2. This line passes `allow_superset` to the concrete item model.
3. Python dispatches the call to `PrefixListModel.get_diff()` or `NDLinkModel.get_diff()`.
4. The override does not accept that keyword, so Python raises `TypeError`.

This happens before comparison completes and before any requested controller operation is sent. A completely new resource can avoid this path because there is no existing item to compare, but updating or checking idempotency for an existing resource can abort the task.

### Where the fixes belong

This requires coordinated changes rather than a fix in only one PR:

- **PR #294 owns the shared API contract.** It should establish that `get_diff()` overrides accept `allow_superset`, document or otherwise make that subclass contract explicit, and add shared regression coverage for invoking an override through `NDConfigCollection`.
- **PR #286 owns the `PrefixListModel` implementation.** It must accept the new keyword while retaining strict, field-aware comparison for authoritative prefix-list `entries`.
- **PR #312 owns the `NDLinkModel` implementation.** It must accept the new keyword while retaining the special preprovision-link comparison behavior.

Do not hide the incompatibility by catching `TypeError` or inspecting method signatures at runtime. That would conceal invalid overrides and could silently bypass the intended comparison policy.

### Recommended integration order

1. Finalize the shared `get_diff()` contract and tests in #294.
2. Merge #294 into `develop` once its behavior against currently merged models is correct.
3. Rebase #286 and #312 onto that updated `develop` branch.
4. Update their model-specific `get_diff()` signatures and preserve their existing comparison semantics.
5. Run composition tests using the actual #286 and #312 production model classes before either module PR is merged.

PR #286 currently merges cleanly with #294 but fails at runtime. PR #312 first requires textual conflict resolution and will then fail at runtime unless its override is updated during that rebase.

mikewiebe:
**High: PRs #281 and #360 require semantic conflict reconciliation after #294**

“Conflict reconciliation” means manually combining both PRs’ behaviors after #294 merges. Selecting Git’s entire “ours” or “theirs” version would break one side.

Currently, [PR #281](https://github.com/CiscoDevNet/ansible-nd/pull/281) already conflicts with `develop`. [PR #360](https://github.com/CiscoDevNet/ansible-nd/pull/360) merges with current `develop`, but conflicts after applying #294.

### PR #360

Both PRs modify `NDBaseModel.get_diff()`:

- #294 adds the `allow_superset` argument for merged-list comparison.
- #360 adds `merge_would_change()` to detect storm-control percentage/PPS transitions that have merge side effects.

The combined implementation needs both:

```python
def get_diff(
    self,
    other,
    exclude_unset=False,
    allow_superset=False,
):
    self_data = self.to_diff_dict()
    other_data = other.to_diff_dict(exclude_unset=exclude_unset)

    is_subset = issubset(
        other_data,
        self_data,
        allow_superset=allow_superset,
    )

    if is_subset and exclude_unset and self.merge_would_change(other):
        return False

    return is_subset
```

Keep #360’s `merge_would_change()` implementation too.

Taking only #294 would lose storm-control transition detection. Taking only #360 would remove `allow_superset`, causing the same keyword-argument compatibility failure discussed for #286 and #312.

The combined tests should prove that:

- Existing percentage can transition to PPS.
- Existing PPS can transition to percentage.
- A controller response containing both values is remediated.
- #294’s merged-list behavior still works.

### PR #281

PR #281 conflicts in three shared files and also has semantic overlaps that Git may auto-merge incorrectly.

#### `models/base.py`

- Keep #294’s recursive removal of implicit `None` values in `from_config()`.
- Do not restore #281’s direct validation of the uncleaned Ansible dictionary.

#### `nd_config_collection.py`

- The direct conflict is mostly documentation.
- Keep the contract that `None` can be normalized to an empty collection only after the caller has validated whether omitted config is legal for the requested state.

#### `nd_state_machine.py`

- Keep #294’s raw `config is None` validation for write states.
- Keep state-aware Pydantic validation and preflight checks.
- Keep #294’s rule that `sent` contains only operations actually sent to ND and remains empty in check mode.
- Preserve #281’s `track_deletes_in_sent` behavior as a VRF Lite opt-in, with the generic default remaining `False`.

Two additional #281 behaviors need deliberate adjustment:

- #281 converts omitted config to `[]` before creating the state machine. It must validate raw user input first so omitted config is not confused with an intentional empty configuration.
- #281 calculates `_changed_vrfs` from `sent`. In check mode, that collection is intentionally empty under #294. Planned VRFs should instead be derived from the before/after preview while still preventing save or deploy operations.

Its keyed VRF Lite attachment merge behavior must also remain intact.

### Recommended integration order

1. Finalize and merge #294.
2. Rebase #360 and manually combine both `get_diff()` behaviors.
3. Rebase #281, dropping duplicated shared-framework changes where #294 now supplies them.
4. Resolve the three #281 shared files according to the contracts above.
5. Run each module’s tests together with #294’s shared state-machine and diff tests.

plugins/module_utils/nd_config_collection.py
sivakasi-cisco: Thanks Mike. I have added

- A Subclass contract note on NDBaseModel.get_diff  — any override must accept both exclude_unset and allow_superset. An override can ignore semantically but it still has to accepts the keyword.
- A note at the get_diff_config
- Added a UT test_nd_config_collection.py for NDConfigCollection

sivakasi-cisco:
Have tested the nd_vrf_lite module with the latest changes in https://github.com/CiscoDevNet/ansible-nd/pull/294.

plugins/module_utils/nd_config_collection.py
mikewiebe: ### High: The production prefix-list model is still incompatible

**Issue**

Thanks for adding the subclass-contract documentation and the `NDConfigCollection` unit test. Those changes describe the expected interface, but they do not fix the production `PrefixListModel` implementation at the current PR head.

**Evidence**

- `NDConfigCollection.get_diff_config()` now unconditionally calls the concrete model with `allow_superset=...`.
- `PrefixListModel.get_diff()` still has the signature `get_diff(self, other, exclude_unset=False)` and therefore does not accept `allow_superset`.
- Python dispatches this call to `PrefixListModel.get_diff()`, not to the documented base-class implementation. Comparing an existing prefix list therefore still raises:

```text
TypeError: PrefixListModel.get_diff() got an unexpected keyword argument `allow_superset`
```

- The added unit test uses a fake model to demonstrate the contract, but it does not invoke the real `PrefixListModel`, so this production incompatibility is not detected.

**Existing PR overlap**

This is a follow-up to the issue originally reported in this thread. The thread had been marked resolved, but the concrete production override remains unchanged and the failure is still reproducible at the current head.

No matching open issue was found. Issue #386 concerns deep-merge behavior rather than this method-signature incompatibility.

**Impact**

Updating or checking idempotency for an existing prefix list can terminate with `TypeError` before comparison completes or any controller operation is sent.

**Suggested fix**

Update `PrefixListModel.get_diff()` and every other production override to accept the complete shared signature, forwarding or intentionally ignoring `allow_superset` as appropriate. Add a regression test that reconciles the real production model through `NDConfigCollection`, rather than testing only a fake override.

plugins/module_utils/nd_config_collection.py
mikewiebe: ### High: Production wrappers still bypass the null-config guard

**Issue**

Thanks for adding the `NDStateMachine` validation. That guard rejects a raw `None`, but production callers still convert omitted or null configuration to `[]` before constructing the state machine. The guard therefore never sees `None` in those workflows.

**Evidence**

- `nd_manage_networks.py` and `network_workflow_coordinator.py` normalize configuration with logic equivalent to `module_args.get("config") or []`.
- `nd_manage_vpc_pair.py` similarly builds runtime configuration from `(module_config.config or [])`.
- Both expressions convert `None` to `[]`, so the state-machine validation cannot distinguish omitted configuration from an explicitly empty desired configuration.
- The direct state-machine tests validate raw `None`, but no composed production-wrapper test covers this conversion.
- Practical example: assume Nexus Dashboard currently contains the networks `production` and `storage`. An automation task runs with:

```yaml
- name: Reconcile all networks
  cisco.nd.nd_manage_networks:
    state: overridden
    config: "{{ desired_networks | default(omit) }}"
```

If `desired_networks` is undefined because of an inventory or pipeline error, Ansible omits `config`. The production wrapper converts the omitted value to `[]`. The state machine then sees the same input as an intentional `config: []` request and can calculate deletion of both `production` and `storage`, instead of rejecting the task.

**Existing PR overlap**

This is a follow-up to the original finding in this thread. The state-machine-local fix addresses only the value presented directly to `NDStateMachine`; the production wrapper boundary remains unsafe at the current head.

No matching open issue was found.

**Impact**

A missing variable or incomplete generated task using `state: overridden` can delete every managed resource in scope instead of failing safely.

**Suggested fix**

Preserve `None` through the production modules and coordinators until shared write-state validation runs. Normalize to `[]` only after confirming that the user explicitly supplied an empty list. Add composed tests through each affected production wrapper, including `state: overridden`, check mode, non-empty existing resources, omitted config, null config, and explicit `config: []`.

sivakasi-cisco: Hi Mike,

This is an important case on the overridden that required addressing.
1. I have removed "default=[]" in nd_manage_networks config arg spec so that the null value would be "None" instead of "[]" - before validation.
2. The "validate_config_presence" would run on the raw config in network and vpc_pair module before any [] normalisation so tht "None" is rejected for merge, replaced and overridden states.
3. For delete all under overrridden, "config: []" works when mentioned explicitly
4. Added tests through 3 modules -vrf_lite, network and vpc_pair to cover omitted, null, config: [] witj overridden, check-mode, non-empty existing resource. This is to make sure that they would assert and get failed before the controller call. So, there will be no query or delete.

plugins/module_utils/nd_config_collection.py
sivakasi-cisco: I understand the untouched problem here. 

- The shared method signature was updated with allow_superset and made the collection to pass it. The base class was made to accept it. But the "PrefixListModel" from [PR286](https://github.com/CiscoDevNet/ansible-nd/pull/286) still had old shape, which had no allow_superset param in it.
- I had a docstring in my earlier commit which was a rule mention so the real crash would have been hidden as the UT was also sending no real prefix list
- Hence TypeError was reproducible 

Now, 

- The "PrefixListModel" uses the full signature (self, other, exclude_unset=False, allow_superset=False) and forwards allow_superset to super().get_diff(...) with strict comparison for entries. So the prefix list semantics have nothing changed but abides the contract.
- I have replaced the fake model coverage with a real PrefixListModel through NDConfigCollection.get_diff_config(..., allow_superset=True), covering the diff, no-diff and new cases. So, without the fix, this test reproduces the Typeerror.

sivakasi-cisco:
Hi Mike,

This is really helpful with the integration order.

For PR294 - the current one - base merged to 'develop' branch.

Now the combined version of base.py is below with allow_superset and merge_would_change.

self_data = self.to_diff_dict()
other_data = other.to_diff_dict(exclude_unset=exclude_unset)
is_subset = issubset(other_data, self_data, allow_superset=allow_superset)
if is_subset and exclude_unset and self.merge_would_change(other):
    return False
return is_subset

Both [#294](https://github.com/CiscoDevNet/ansible-nd/pull/294) and [#360](https://github.com/CiscoDevNet/ansible-nd/pull/360) touch NDBaseModel.get_diff()

#294 added allow_superset for list comparison
#360 added merge_would_change to catch storm-control percentag

plugins/module_utils/nd_state_machine.py
```diff
@@ -204,21 +235,28 @@ def _manage_create_update_state(self) -> None:
...
-        successfully_sent = items_to_update + items_to_create
-        if successfully_sent:
+                    if self._execute_operation(self.model_orchestrator.create, item, error_msg_prefix=f"Failed to create {item.get_identifier_value()}"):
+                        successfully_sent.append(item)
+
+        # Mark as sent only for items actually pushed to the controller. In
+        # check mode no API call is made, so nothing is marked as sent (avoids
+        # false deploy triggers); the previewed 'after' state is still reflected
+        # in self.existing (mutated above), which is what drives 'changed'.
+        if not self.check_mode and successfully_sent:
```
allenrobel: Gating `self.sent.add_many(...)` on `not self.check_mode` (here and in the delete path below) reverses a deliberate earlier change ("Remove self.send from check_mode guards in NDStateMachine", PR #225), which kept `sent` populated during dry runs so downstream consumers could see what a real run would send.

Five fabric modules (`nd_manage_fabric_external`, `nd_manage_fabric_{ai_,}ebgp_vxlan`, `nd_manage_fabric_{ai_,}ibgp_vxlan`) gate their config-save/deploy call solely on `len(nd_state_machine.sent) > 0`, with no check-mode guard of their own — they rely on `execute_config_actions()` being check-mode-safe via `_request()` simulation (which it is, per the `ConfigActionsMixin` docstring). With this change, a `--check` run that has real pending changes leaves `sent` empty, so those modules silently skip the config-save/deploy preview.

The "false deploy triggers" concern in the comment above doesn't apply in check mode, since `execute_config_actions()` already simulates rather than sends. Follow-up to the earlier (resolved) test-coverage thread at line 162: the new tests lock in this behavior, but the behavior itself looks like a regression for the fabric modules. Suggest keeping `sent` populated in check mode (per-item success gating for real runs is a clear improvement and can stay), or alternatively updating the five fabric modules to preview config actions under check mode some other way.

plugins/module_utils/nd_config_collection.py
```diff
@@ -240,11 +252,16 @@ def to_payload_list(self, **kwargs) -> List[Dict[str, Any]]:
         return [item.to_payload(**kwargs) for item in self._items]
 
     @staticmethod
-    def from_ansible_config(data: List[Dict], model_class: type[NDBaseModel], **kwargs) -> "NDConfigCollection":
+    def from_ansible_config(data: Optional[List[Dict]], model_class: type[NDBaseModel], **kwargs) -> "NDConfigCollection":
```
allenrobel: This touched signature keeps the legacy annotation `Optional[List[Dict]]`. CLAUDE.md: "Modern type annotations (PEP 585 / PEP 604) for all new code, and when touching existing annotations: `X | None` instead of `Optional[X]`" — should be `list[dict] | None`.

plugins/module_utils/utils.py
```diff
@@ -6,7 +6,7 @@
 
 import logging
 from copy import deepcopy
-from typing import Any
+from typing import Any, Dict, List, Set
```
allenrobel: New code in this file introduces legacy typing imports and usages: `Dict[int, int]` (line 53), `Set[int]` (line 62 area / `_try_assign`), `List[List[int]]` (line 116). CLAUDE.md: "`list[X]`, `dict[K, V]`, `set[X]`, `tuple[X, ...]` instead of `List`, `Dict`, `Set`, `Tuple`" — please use the PEP 585 builtins and drop `Dict`/`List`/`Set` from this import (keep `Any`).

allenrobel (commented):
### Code review

Fresh pass over the current head (my June review was dismissed by subsequent pushes). One behavioral concern with check-mode `sent` tracking and two type-annotation notes, each as an inline comment.

plugins/module_utils/nd_state_machine.py
sivakasi-cisco: This was an unintended re-introduction of the check-mode guard that PR #225 removed. Reverted it: on both the create/update and delete paths no longer gates on - not self.check_mode,  so  that 'sent' stays populated during --checkand the five fabric modules' config-save/deploy preview works again.

plugins/module_utils/nd_config_collection.py
sivakasi-cisco: updated the touched signature to "list[dict] | None" as per the PEP 585/604 convention

plugins/module_utils/utils.py
sivakasi-cisco: Switched the new code to the modern built-in types

allenrobel (approved):

mikewiebe:
@sivakasi-cisco  This PR is out-of-date with the base branch now so changes from `develop` need to be merged into this branch

PR #334, opened 2026-06-16
nd_interface_vpc_base

allenrobel:
# Add shared vPC interface infrastructure (VpcInterfaceBaseOrchestrator)

## Related Issue(s)

Unblocks the vPC interface modules (PRs #279 `nd_interface_vpc_access`, #280 `nd_interface_vpc_trunk_host`) after PR #278 (`nd_vpc_pair`) was closed in favor of #197 (`nd_manage_vpc_pair`).

Code-review findings deferred (not fixed in this PR) with their tracking issues:

- **#348** — `VpcInterfaceBaseOrchestrator` CRUD (`create`/`update`/`query_one`/`create_bulk`) and the new `_switches_to_query` / `_managed_vpc_interfaces` helpers duplicate `PortChannelBaseOrchestrator`; lift into the shared base. Maintenance, not a bug.
- **#349** — `query_all` issues one interface-list GET per switch under `state: overridden` (ND 4.2.1 has no fabric-wide interface-list endpoint). This PR already narrows non-overridden states to config switches via `_switches_to_query()`, so the residual fan-out is overridden-only. (This is the reviewer thread on the fabric-wide scan.)
- **#354** — `delete()` queues a deploy even when the per-interface DELETE was a 404 no-op (a 204 success and a 404 both return a falsy body, so the deploy cannot be cheaply gated without surfacing the return code / lab-verifying the deploy is a no-op for an absent interface).

## Proposed Changes

PR #278 (`nd_vpc_pair`) carried the shared vPC *interface* infrastructure that `nd_interface_vpc_access` and `nd_interface_vpc_trunk_host` depend on. With #278 closed, that infrastructure is re-homed here on a dedicated base branch so the two interface modules can reach `develop` independently:

- **`orchestrators/vpc_interface_base.py`** — `VpcInterfaceBaseOrchestrator`, the shared CRUD base for vPC host-interface modules (`accessVpcHost`, `trunkVpcHost`). Renamed from the original `VpcBaseOrchestrator` to be clearly interface-scoped (parallels `base_interface.py` / `NDBaseInterfaceOrchestrator`), distinct from `develop`'s pair-domain `orchestrators/manage_vpc_pair.py`.
- **`endpoints/v1/manage/manage_interfaces.py`** — adds `EpManageInterfacesDelete` (per-interface DELETE) used for vPC interface removal.
- Peer-serial resolution is **rewired to `develop`'s existing `EpVpcPairGet`** (`manage_fabrics_switches_vpc_pair.py`, from #197) rather than carrying a duplicate vpcPair endpoint — the only genuine overlap with #197 (whose `VpcPairOrchestrator` manages a different resource and cannot serve as a base for interface modules).

## Test Notes

- Unit tests pass on this branch and downstream: `python -m pytest tests/unit/module_utils/orchestrators/ tests/unit/module_utils/endpoints/` (457 passing on this base; 470/483 on the stacked interface branches).
- The downstream interface modules' integration targets (`nd_interface_vpc_access`, `nd_interface_vpc_trunk_host`) pass end-to-end against ND 4.2 / NX-OS 10.6(2) on the SITE1 lab, using `VpcInterfaceBaseOrchestrator` for peer resolution and CRUD.

4.2.1

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

### DIFF
--- a/plugins/module_utils/orchestrators/vpc_interface_base.py
+++ b/plugins/module_utils/orchestrators/vpc_interface_base.py
@@ -362,6 +362,6 @@ class VpcInterfaceBaseOrchestrator(NDBaseInterfaceOrchestrator[ModelType]):
                     existing = interfaces_by_name.get(name)
                     if existing is None or switch_id < existing[0]:
                         interfaces_by_name[name] = (switch_id, iface)
-            return [entry for _, entry in interfaces_by_name.values()]
+            return [entry[1] for entry in interfaces_by_name.values()]
         except Exception as e:
             raise RuntimeError(f"Query all failed: {e}") from e
--- a/plugins/module_utils/orchestrators/vpc_interface_base.py
+++ b/plugins/module_utils/orchestrators/vpc_interface_base.py
@@ -45,6 +45,7 @@ from ansible_collections.cisco.nd.plugins.module_utils.endpoints.v1.manage.manag
     EpManageInterfacesPut,
 )
 from ansible_collections.cisco.nd.plugins.module_utils.models.base import NDBaseModel
+from ansible_collections.cisco.nd.plugins.module_utils.orchestrators.base import requires_bulk_support
 from ansible_collections.cisco.nd.plugins.module_utils.orchestrators.base_interface import NDBaseInterfaceOrchestrator
 from ansible_collections.cisco.nd.plugins.module_utils.orchestrators.types import ResponseType
 
@@ -63,9 +64,9 @@ class VpcInterfaceBaseOrchestrator(NDBaseInterfaceOrchestrator[ModelType]):
     Each create/update reads the `vpcPair` record for the primary switch to obtain the peer serial, which is
     then injected as `peerSwitchId` in the payload. Lookups are cached per orchestrator instance.
 
-    Mutation methods (`create`, `update`) queue deploys for bulk execution. `delete` queues vPC interfaces for
-    bulk removal via `interfaceActions/remove` and bulk deploy. Call `remove_pending` and `deploy_pending` after
-    all mutations are complete.
+    Mutation methods (`create`, `update`) queue deploys for bulk execution. `delete` issues an immediate
+    per-interface `DELETE /interfaces/{name}` (the bulk `interfaceActions/remove` endpoint rejects vPC interfaces;
+    see the `TODO(4.2.1)` below) and queues a deploy. Call `deploy_pending` after all mutations are complete.
 
     ## Raises
 
@@ -82,7 +83,8 @@ class VpcInterfaceBaseOrchestrator(NDBaseInterfaceOrchestrator[ModelType]):
     - Via `query_all` if the query API request fails.
     """
 
-    # TODO(4.2.1) The bulk `/api/v1/manage/fabrics/{fabric}/interfaceActions/remove` endpoint returns
+    # TODO(4.2.1) vpc-interface-bulk-delete-silent-fail
+    # The bulk `/api/v1/manage/fabrics/{fabric}/interfaceActions/remove` endpoint returns
     # `{"status":"Failed","message":"Invalid Interface"}` inside a 207 for vPC interfaces (lab-verified). The
     # per-interface `DELETE /interfaces/{name}` endpoint works (returns 204). We therefore disable bulk delete and
     # use the per-interface DELETE via `delete_endpoint`.
@@ -163,16 +165,22 @@ class VpcInterfaceBaseOrchestrator(NDBaseInterfaceOrchestrator[ModelType]):
         """
         # Summary
 
-        Inject `peerSwitchId` into the nested `configData.networkOS.policy` block of an interface payload. The model
-        already nests `peer_switch_id` under `policy`, so this just sets the value if not already present.
+        Inject `peerSwitchId` into the nested `configData.networkOS.policy` block of an interface payload. Every vPC
+        interface requires a peer serial, so a payload without a `policy` block is structurally invalid and is rejected
+        here rather than silently sent to ND as a one-sided vPC.
 
         ## Raises
 
-        None
+        ### RuntimeError
+
+        - If the payload has no `configData.networkOS.policy` block to inject `peerSwitchId` into.
         """
-        policy = payload.get("configData", {}).get("networkOS", {}).get("policy")
-        if policy is not None:
-            policy["peerSwitchId"] = peer_serial
+        policy = (payload.get("configData") or {}).get("networkOS", {}).get("policy")
+        if policy is None:
+            raise RuntimeError(
+                f"vPC interface payload is missing the 'configData.networkOS.policy' block required to inject 'peerSwitchId'; received: {payload!r}"
+            )
+        policy["peerSwitchId"] = peer_serial
         return payload
 
     def create(self, model_instance: ModelType, **kwargs) -> ResponseType:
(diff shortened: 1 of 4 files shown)


plugins/module_utils/orchestrators/vpc_interface_base.py
```diff
@@ -0,0 +1,367 @@
...
+
+        ## Raises
+
+        ### NotImplementedError
+
+        - Always, if not overridden by a subclass.
+        """
+        raise NotImplementedError("Subclasses must implement _managed_policy_types()")
+
+    def _resolve_peer_switch_id(self, switch_ip: str, primary_serial: str) -> str:
```
mikewiebe, line 131: ### High: Add unit coverage for the shared vPC interface base

**Issue**

This PR adds the shared vPC interface orchestration layer and a new per-interface DELETE endpoint, but it does not add any unit tests for either change.

**Evidence**

- [`plugins/module_utils/orchestrators/vpc_interface_base.py` lines 129-160](https://github.com/CiscoDevNet/ansible-nd/pull/334/files#diff-e8ad326e0cde4b80ec8422d231810a48322151232983290010bfb0ee7957b61cR129) resolves and caches the vPC peer switch serial from the `vpcPair` endpoint.
- [`plugins/module_utils/orchestrators/vpc_interface_base.py` lines 236-258](https://github.com/CiscoDevNet/ansible-nd/pull/334/files#diff-e8ad326e0cde4b80ec8422d231810a48322151232983290010bfb0ee7957b61cR236) implements vPC delete through the per-interface DELETE path.
- [`plugins/module_utils/endpoints/v1/manage/manage_interfaces.py` lines 282-323](https://github.com/CiscoDevNet/ansible-nd/pull/334/files#diff-d39008b70fe4bc837f50eb937a31cd7e9f302f0036f341326676391cdf46316eR282) adds `EpManageInterfacesDelete`.
- The fetched PR file list contains only two changed Python files and no `tests/unit/...` files.

**Existing PR overlap**

**Impact**

The vPC access and vPC trunk modules will depend on this base for peer lookup, payload injection, delete behavior, and deduped query results. A regression here would affect every vPC interface module and could produce wrong deploy targets, missed peer IDs, or untested delete behavior on large fabrics.

**Suggested fix**

Add focused unit tests for `_resolve_peer_switch_id` success, missing-pair, and missing-`peerSwitchId` cases; create/update payload injection; per-interface DELETE path construction; delete queuing; `query_all` deduplication; and path encoding for fabric, switch, and interface identifiers.

---

plugins/module_utils/orchestrators/vpc_interface_base.py
```diff
@@ -0,0 +1,367 @@
...
+        """
+        try:
+            switch_id = self._resolve_switch_id(model_instance.switch_ip)
+            api_endpoint = self._configure_endpoint(self.query_one_endpoint(), switch_sn=switch_id)
+            api_endpoint.set_identifiers(model_instance.interface_name)
+            return self._request(path=api_endpoint.path, verb=api_endpoint.verb)
+        except Exception as e:
+            raise RuntimeError(f"Query failed for {model_instance.get_identifier_value()}: {e}") from e
+
+    def query_all(self, model_instance: ModelType | None = None, **kwargs) -> ResponseType:
```
mikewiebe, line 411: ### Medium: Fabric-wide vPC query scans every switch and filters locally

**Issue**

`query_all()` walks every switch in the fabric, fetches the full interface list for each switch, then filters down to vPC interfaces and managed policy types in Python.

@allenrobel I have been reviewing PRs for any potential performance bottlenecks so this may or may not be an issue.  Let's discuss if this is even viable or desired.

**Evidence**

- [`plugins/module_utils/orchestrators/vpc_interface_base.py` lines 318-365](https://github.com/CiscoDevNet/ansible-nd/pull/334/files#diff-e8ad326e0cde4b80ec8422d231810a48322151232983290010bfb0ee7957b61cR318) validates the fabric, iterates `fabric_context.switch_map`, calls the list endpoint for every switch, filters by `interfaceType == "vpc"`, filters by policy type, and deduplicates by interface name.

**Existing PR overlap**

**Impact**

On large fabrics, this becomes one controller request per switch plus a full interface payload per switch. That can make `overridden`, future `gathered`, or any full-state comparison slow and increases controller load even when the target set is small.

**Suggested fix**

If the Manage API supports server-side filtering for interface type or policy type, use it. If it does not, add tests that lock in the expected request count and dedupe behavior, and consider pruning obvious non-vPC candidates before issuing per-switch calls where fabric context exposes that information.

plugins/module_utils/orchestrators/vpc_interface_base.py
allenrobel: Added in 012bbc99 — `tests/unit/module_utils/orchestrators/test_vpc_interface_base.py` (24 tests, 100% line coverage of `vpc_interface_base.py`), driving the abstract base through a minimal `_StubVpcOrchestrator` / `_StubVpcInterfaceModel` (the concrete access/trunk models live on the branches stacked above, so the base is tested directly here).

Covers each case you called out:

- `_resolve_peer_switch_id`: success, per-instance cache (second call issues no second `vpcPair` GET), missing-pair (404 → "not in a vPC pair"), and missing-`peerSwitchId`.
- create/update payload injection: `switchId` + `peerSwitchId` set, `switchIp` excluded, endpoint path/verb asserted.
- per-interface DELETE: path construction (`.../interfaces/{name}`, verb DELETE), deploy queued, no bulk remove queued.
- `create_bulk`: per-switch grouping with a single peer lookup per primary switch (cache).
- `query_all`: dedup of the per-peer copies; path encoding is exercised via the fabric/switch/interface path assertions.

plugins/module_utils/orchestrators/vpc_interface_base.py
allenrobel: Good to discuss. I checked the Manage API (nd-openapi, 4.2.1): there is no fabric-wide endpoint that returns full per-interface `configData`. `interfacesOverview` / `interfacesSummary` return only aggregate counts by type/status, and the only full-config source is the per-switch `.../switches/{switchId}/interfaces` list. So one request per switch is inherent to the API for the data we need for idempotency — server-side filtering (the `filter` / `interfaceType` query params) would narrow each payload but not reduce the request count.

Given that, I took the fallback you suggested: `test_vpc_interface_base_00900` now locks the request count to `2 + (one GET per switch)` (summary + switches-list + per-switch scan) plus the per-peer dedup, so a regression to N+1 fails the suite. Added in 012bbc99.

If you'd prefer we cap/scope the scan differently (e.g. prune via `fabric_context` before issuing per-switch calls, or only scan vPC-capable switches once `capableSwitches` lands), happy to follow up — leaving this thread for you to weigh in.

allenrobel: Addressed in the PR #334 review pass. `query_all` now adopts the sibling's `_switches_to_query()`: for every state except `overridden` it scans only the switches named in the user config, so a merged/replaced/query run no longer fans out one GET per fabric switch. The remaining fabric-wide scan is `overridden`-only, and as confirmed here ND 4.2.1 has no fabric-wide interface-list endpoint (the only full-config source is per-switch `.../switches/{switchId}/interfaces`), so that residual is inherent to the API — tracked in #349. The per-peer dedup now also prefers the config-supplied peer IP so idempotency holds regardless of which peer is named. Resolving; reopen if you'd prefer a different scoping for the overridden case.

mikewiebe (approved):
LGTM

gmicol (approved):
LGTM

Issue #293, opened 2026-05-26
Define a common Ansible action-plugin based testing strategy for ND 4.x modules


sivakasi-cisco:
Problem statement: In ND 4.x, each module integration tests have their own playbook structure, setup and cleanup pattern, assertion style, and execution flow. The structure is not standardised.

The goal here is to define one testing strategy that all ND 4.x modules can follow. 

Action plugins can help us avoid repeating the same testing code in every module test and agree to a common pattern. It can run and help with common tasks like:

- Loading shared test variables
- Preparing module inputs
- Running the module in a standard way
- Checking idempotency
- Validating the expected changed/unchanged results
- Saving logs/debugs
- Running cleanups
- handling common failures

Document a common playbook directory that can be created later, for example:

playbooks/nd4x_module_tests/
  inventory/
    hosts.yaml
    group_vars/
  common/
    setup.yaml
    cleanup.yaml
    assert_result.yaml
    run_module_test.yaml
  module_tests/
    fabrics/
    vpc_pair/
    vrf/
    vrf_lite/
    policy/
    policy_group/
    l3out_links/
    resource_manager/
    acl/
    interfaces/
  README.md
Each module folder should contain simple playbooks for the states that the module supports.

Example:

module_tests/vrf_lite/
  merged.yaml
  replaced.yaml
  overridden.yaml
  deleted.yaml
  gathered.yaml

mtarking:
Please address the versioning check raised [here](https://github.com/CiscoDevNet/ansible-nd/pull/260#discussion_r3307830358)

sivakasi-cisco:
https://github.com/CiscoDevNet/ansible-nd/pull/337

astawast-cisco:
Planning details for Issue:

ND 4.x Standardized Testing and Automation Plan

Task 1 — Review existing tests and define the standard strategy - (May 27 - June 5) —— Completed
Task 2 — Implement the shared ND 4.x action plugin - (June 6 - June 12) —— Completed
Task 3 — Migrate the first module: nd_interface_loopback - (June 13 - June 20) —— Completed
Task 4 — Harden the shared action plugin - (June 21 - June 27) —— Completed
Task 5 — Add action-plugin unit tests  - (June 28 - July 3) —— Completed
Task 6 — Migrate the second module: nd_interface_ethernet_access - (July 4 - July 13) —— Completed
Task 7 — Add safety, backward compatibility and opt-in execution - (July 14 - July 20) —— Completed
[...]

PR #337, opened 2026-06-17
Demo For Common Ansible Action-Plugin Based Testing Strategy for ND 4…

astawast-cisco:
## Related Issue(s)

Part of #293

# ND 4.x Common Test Harness – Module Migration

## Overview

This PR completes the migration of the following ND 4.x integration test modules to the proposed common Ansible action-plugin based test harness:

* `nd_interface_loopback`
* `nd_interface_ethernet_access`
* `nd_interface_vpc_access`

The goal is to centralize common integration-test execution and validation logic while keeping module-specific assertions within the individual test playbooks.

No module functionality is changed. This PR focuses on standardizing and strengthening the integration-test framework and coverage.

### 1. Shared Action Plugin

Implemented the reusable action plugin:

`plugins/action/nd4x_module_test.py`

The plugin standardizes the common test workflow, including:

* Check-mode execution
* Normal module execution
* Idempotency verification
* Expected changed/unchanged validation
* Expected failure validation
* Debug and result collection

This removes duplicated execution logic across individual module test scenarios.

### 2. Proof of Concept – `nd_interface_loopback`

The shared action plugin was initially demonstrated using `nd_interface_loopback`.

Demo scenarios have been implemented for:

* Merged
* Replaced
* Overridden
* Deleted

This validates that the same framework can support different module states without modifying the core testing logic.

### 3. Lightweight Sanity Test Flow

Added a smaller sanity workflow for faster validation during pull requests.

The flow includes:

* Environment setup
* Module execution
* Idempotency verification
* Validation
* Cleanup

This provides a faster alternative to running the complete integration suite while maintaining confidence in the core functionality.

### 4. Standardized Expected-Result Contract

Introduced a simplified expected-result schema:

yaml
expected:
  check_mode_changed: true
  apply_changed: true
  idempotency: true
  failed: false

This provides a consistent and readable contract across module tests.

### 5. ND API Query and Validation

Added optional ND REST API validation to the shared action plugin.

The framework supports:

* Executing ND REST API queries through `cisco.nd.nd_rest`
* Validating responses using JSONPath expressions
* Comparing actual API values against expected values
* Validating expected HTTP status codes
* Supporting `404` validation for deleted resources
* Reporting API validation failures as part of the test result

This allows the tests to validate both the module result and the actual state in the ND controller.

Example validations include:

* Confirming an interface was created
* Confirming interface attributes were updated correctly
* Confirming a deleted resource is no longer present

### 6. Separation of Generic and Module-Specific Validation

The framework separates common test behavior from feature-specific assertions.

The shared action plugin handles:

* Module execution
* Idempotency checks
* Expected-result validation
* ND API validation

Individual module playbooks continue to handle module-specific validations such as:

* Interface configuration
* Descriptions
* IP addresses
* Resource-specific assertions
* Negative cases

This keeps the framework reusable across different ND modules.

### 7. Cleanup Strategy

Test cleanup is implemented using Ansible `always` blocks to ensure resources are removed regardless of test success or failure.

This prevents leftover configuration from impacting subsequent test runs.

### 8. Environment and Integration Improvements

Resolved several integration issues encountered during the migration, including:

* Generalizing `ansible_python_interpreter`
* Correcting `ansible_network_os` configuration
* Resolving dependency issues, including Pydantic
* Resolving `ansible.netcommon` and `ansible.utils` visibility issues
* Validating required fabric and switch variables
* Adding switch ID support for ND API endpoint validation
* Successfully executing the POC test flow

## Test Coverage

The migrated integration scenarios now use the shared harness to consistently validate:

* Predictive check-mode behavior
* Controller-state preservation during check mode
* Real configuration application
* Idempotency
* Post-apply REST API validation
* Module-specific assertions
* Negative test cases
* Test resource setup and cleanup

## Parity Updates

The corresponding `PARITY.md` files have been updated to:

* Map the original integration scenarios to their harness-based replacements
* Document the migrated test coverage
* Capture validation details
* Record execution results

## Current Status

The POC demonstrates that the common action-plugin based framework can successfully centralize ND 4.x integration-test behavior while keeping module-specific validations independent and reusable.

The three modules listed above have been migrated and validated using the new framework.

## Running the ND 4.x Demo

From the collection root:

bash
cd /Users/<user>/ansible_collections/cisco/nd

Install/check the required collections:

bash
ansible-galaxy collection install ansible.netcommon ansible.utils

Create a local inventory:

bash
cp tests/integration/inventory.networking tests/integration/inventory.LOCAL.networking

Update the following values in the local inventory:

text
ansible_host=
ansible_user=
ansible_password=
nd_test_fabric_name=
nd_test_switch_ip=
nd_test_switch_id=<switch id/serial>

Run the complete ND 4.x demo sanity flow:

bash
ANSIBLE_CONFIG=ansible.cfg \
ANSIBLE_COLLECTIONS_PATH=/Users/:/Users/<user>/.ansible/collections:/usr/share/ansible/collections \
python -m ansible_test._util.target.cli.ansible_test_cli_stub \
network-integration nd_interface_loopback \
--inventory tests/integration/inventory.LOCAL.networking \
--tags nd4x_demo \
-vvv

To run an individual state, use one of:

bash
--tags nd4x_demo_merged
--tags nd4x_demo_replaced
--tags nd4x_demo_overridden
--tags nd4x_demo_deleted

### Expected Result

The final `PLAY RECAP` should report:

text
failed=0

## Summary

This PR introduces and applies the common ND 4.x test-harness approach across three interface modules. It reduces duplicated test logic, adds consistent check-mode/idempotency/API validation, preserves module-specific assertions, and provides updated parity documentation for the migrated scenarios.

**No module functionality is changed.**


tests/integration/inventory.networking
```diff
@@ -1,31 +1,19 @@
 [nd]
-nd ansible_host=<TODO>
+nd ansible_host=10.122.84.112
 
 [nd:vars]
 ansible_connection=ansible.netcommon.httpapi
-ansible_python_interpreter=/usr/bin/python3.9
+ansible_python_interpreter=/Users/astawast/ABC/ansible_collections/cisco/nd/ansible-nd/.venv-run/bin/python
```
sivakasi-cisco: Generalise the path

tests/integration/playbooks/nd4x_module_tests/module_tests/nd_interface_loopback/merged.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - name: "MERGED CREATE: Create loopback100 using standard runner"
+      cisco.nd.nd4x_module_test:
+        module: cisco.nd.nd_interface_loopback
+        state: merged
+        common_args:
+          output_level: "{{ nd_info.output_level }}"
+          fabric_name: "{{ test_fabric_name }}"
+        config:
+          - "{{ loopback_100 }}"
+        expected:
```
sivakasi-cisco, line 15: Can we change this to 
(For merged/overridden/replaced/delete)
expected:
          check_mode_changed: true
          apply_changed: true
          idempotency: true
          failed: false

(For gathered)
expected:
  changed: false 
  failed: false

Applicable in other expected keys as well

tests/integration/inventory.networking
```diff
@@ -1,31 +1,20 @@
 [nd]
```
sivakasi-cisco, line 1: This file can be made generic when adding for review as it contains live credentials

tests/integration/targets/nd_interface_loopback/tasks/main.yaml
```diff
@@ -30,28 +30,93 @@
...
+        # - nd4x_demo_cleanup
 
-    - name: Run nd_interface_loopback overridden state tests
-      ansible.builtin.include_tasks: overridden.yaml
+    
+    - name: Run nd_interface_loopback replaced state tests using ND4.x demo harness
+      ansible.builtin.include_tasks: nd4x_demo_replaced.yaml
+      tags:
+         - nd4x_demo
+         - nd4x_demo_replaced
```
sivakasi-cisco: Hope other files would also be added like nd4x_demo_replaced.yaml, nd4x_demo_overridden.yaml and nd4x_demo_deleted.yaml, similar to nd4x_demo_merged.yaml

tests/integration/ND4X_MODULE_TEST_STRATEGY_PROPOSAL.md
```diff
@@ -0,0 +1,479 @@
+# ND 4.x Module Integration Test Strategy Proposal
```
sivakasi-cisco: Can be removed from review

tests/integration/ND4X_MODULE_TEST_STRATEGY_PROPOSAL.md
```diff
@@ -0,0 +1,479 @@
...
+      cisco.nd.nd4x_module_test:
+        module: cisco.nd.nd_vrf_lite
+        state: merged
+        common_args:
+          output_level: "{{ nd4x_output_level }}"
+          fabric_name: "{{ nd4x_fabric_name }}"
+        config:
+          - "{{ vrf_lite_100 }}"
+        expected:
+          check_mode_changed: true
```
sivakasi-cisco: Can we rename it as run_check_mode instead of check_mode_changed?

plugins/action/nd4x_module_test.py
```diff
@@ -0,0 +1,410 @@
...
+import json
+import time
+from jsonpath_ng import parse
+
+from ansible.errors import AnsibleActionFail
+from ansible.plugins.action import ActionBase
+
+
+
+class ActionModule(ActionBase):
```
sivakasi-cisco, line 23: Can you add a doc stub here?

plugins/action/nd4x_module_test.py
```diff
@@ -0,0 +1,410 @@
...
+# Copyright: (c) 2026, Cisco Systems, Inc.
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+import json
+import time
+from jsonpath_ng import parse
```
sivakasi-cisco: Can you refer plugins/module_utils/ndi.py to guard this with try/except?

tests/integration/ND4X_MODULE_TEST_STRATEGY_PROPOSAL.md
astawast-cisco: Removed ND4X_MODULE_TEST_STRATEGY_PROPOSAL.md from the PR as requested.

tests/integration/inventory.networking
astawast-cisco: Addressed.

tests/integration/playbooks/nd4x_module_tests/module_tests/nd_interface_loopback/merged.yaml
astawast-cisco: Addressed.

tests/integration/inventory.networking
astawast-cisco: Addressed.

tests/integration/targets/nd_interface_loopback/tasks/main.yaml
astawast-cisco: Addressed.

tests/integration/ND4X_MODULE_TEST_STRATEGY_PROPOSAL.md
astawast-cisco: run_check_mode and check_mode_changed have different meanings.

check_mode_changed currently represents the expected changed result from the check-mode execution, while run_check_mode represents whether check mode should execute.

plugins/action/nd4x_module_test.py
astawast-cisco: Added a matching documentation-only module stub at plugins/modules/nd4x_module_test.py, including the action plugin options, examples, and return values.

plugins/action/nd4x_module_test.py
astawast-cisco: Addressed.

tests/integration/targets/nd_interface_loopback/tasks/nd4x_demo_replaced.yaml
```diff
@@ -0,0 +1,82 @@
...
+- name: Run standardized nd_interface_loopback replaced tests
+  block:
+    - name: "REPLACED SETUP: Create loopback100 original config"
+      cisco.nd.nd_interface_loopback:
+        output_level: "{{ nd_info.output_level }}"
+        fabric_name: "{{ test_fabric_name }}"
+        config:
+          - "{{ loopback_100 }}"
+        state: merged
+        tags:
```
sivakasi-cisco: These tags are inside the module arguments, so nd_interface_loopback will treat tags as an unsupported parameter. Can we move them to the task level? The same issue is present in the cleanup task.

tests/integration/targets/nd_interface_loopback/tasks/main.yaml
```diff
@@ -30,28 +30,93 @@
...
 
-    - name: Run nd_interface_loopback merged state tests
-      ansible.builtin.include_tasks: merged.yaml
+     #- name: Run nd_interface_loopback merged state tests
+      # ansible.builtin.include_tasks: merged.yaml
 
-    - name: Run nd_interface_loopback replaced state tests
-      ansible.builtin.include_tasks: replaced.yaml
+    - name: Run nd_interface_loopback merged state tests using ND4.x demo harness
+      ansible.builtin.include_tasks: nd4x_demo_merged.yaml
```
sivakasi-cisco: Are we intentionally replacing the existing integration tests with the demo tests? Can we keep the original tests and make the demo opt-in using the never tag, similar to the Ethernet access tests?

tests/integration/targets/nd_interface_loopback/tasks/nd4x_demo_overridden.yaml
```diff
@@ -0,0 +1,76 @@
...
+      cisco.nd.nd4x_module_test:
+        module: cisco.nd.nd_interface_loopback
+        state: overridden
+        common_args:
+          output_level: "{{ nd_info.output_level }}"
+          fabric_name: "{{ test_fabric_name }}"
+        config:
+          - "{{ loopback_100_updated }}"
+        expected:
+          check_mode_changed: true
```
sivakasi-cisco: Can we add opt-in and preflight check similar to the ethernet overridden test here? This could delete existing loopbacks that are not included in the config.

sivakasi-cisco:
Can we add unit tests for the action plugin?

tests/integration/inventory.networking
```diff
@@ -1,31 +1,20 @@
...
-nd ansible_host=<TODO>
+nd ansible_host=10.122.84.112
```
sivakasi-cisco: This still contains a lab IP address. Can we change it back to a generic placeholder?

astawast-cisco: Updated `tests/integration/inventory.networking` to replace the lab-specific IP address with the reserved documentation placeholder `192.0.2.1`. Thanks for catching this.

tests/integration/targets/nd_interface_loopback/tasks/nd4x_demo_overridden.yaml
astawast-cisco: Added an explicit opt-in and read-only check-mode preflight to the loopback overridden test. It now refuses to apply the override if managed loopbacks exist on another switch or outside the reserved loopback100–loopback109 range. The destructive test remains disabled by default.

tests/integration/targets/nd_interface_loopback/tasks/main.yaml
astawast-cisco: Yes—the current loopback main.yaml replaces the original tests with demo tests. Have updated the loopback test entry point to retain the existing integration suite as the default execution path. The ND 4.x demo tests are now opt-in using the never tag, consistent with the Ethernet access tests. Thanks for catching this.

tests/integration/targets/nd_interface_loopback/tasks/nd4x_demo_replaced.yaml
astawast-cisco: Moved the tags out of the module arguments and variables to the Ansible task level, including the cleanup task.

mikewiebe:
**Medium: predictive check mode does not verify that controller state stayed unchanged**

The harness currently trusts the target module’s check-mode implementation.

During [predictive check mode](https://github.com/CiscoDevNet/ansible-nd/pull/337/files#diff-205d27ebb635125c160485a2b5d7f68e182c11176efa25eae6511069a447a90bR331), it:

1. Runs the target module with check mode enabled.
2. Checks only the returned `changed` and `failed` values.
3. Proceeds to the real apply when those expectations pass.

It does not:

- Capture controller state before check mode.
- Query the controller immediately afterward.
- Compare before and after controller state.
- Detect an unintended mutation elsewhere in the fabric.

The target module normally performs GET requests to calculate its predicted changes, but that does not prove it avoided a write.

Also, the configured [`nd_queries`](https://github.com/CiscoDevNet/ansible-nd/pull/337/files#diff-205d27ebb635125c160485a2b5d7f68e182c11176efa25eae6511069a447a90bR394) do not help here—they run only after the real apply and idempotency phases.

To verify check-mode safety, the sequence should be:

```text
GET relevant controller state
        ↓
run target module in check mode only
        ↓
GET the same controller state again
        ↓
normalize and compare before == after
        ↓
only then run the real apply
```

For an interface test, the snapshot should cover all relevant managed interfaces on the test switch—not just the requested interface—so an unintended mutation to another interface is also detected.

The cleanest enhancement would be for the harness to support something like:

```yaml
check_mode_queries:
  - name: Snapshot managed interfaces
    path: >-
      /api/v1/manage/fabrics/{{ fabric_name
      }}/switches/{{ switch_id }}/interfaces
```

The harness would run that query before and after predictive check mode and require equivalent normalized controller state before proceeding.

The current phase verifies the module’s reported prediction; it does **not** verify that predictive execution was mutation-free.

mikewiebe:
**Medium: add a contributor guide for writing integration tests with this framework**

The PR explains the framework’s purpose and demonstrates it with loopback and Ethernet-access playbooks, but it does not provide a clear authoring guide for a contributor migrating another module.

The [`nd4x_module_test` documentation stub](https://github.com/CiscoDevNet/ansible-nd/pull/337/files#diff-e5a4de54d6a1ab6fc66ddfc77d92482c14b53bf06cf6e007d7423fe2aa064e55R11) describes the top-level options, while the target-local demos show selected implementations. A contributor must still infer the framework’s structure, responsibilities, safety requirements, and migration process from the source.

Could we add an authoritative guide such as:

```text
tests/integration/playbooks/nd4x_module_tests/README.md
```

At minimum, it should document:

1. **Responsibility split**
   - What the action plugin handles.
   - What remains in the module-specific playbook.
   - That setup, feature-specific assertions, and cleanup are not performed automatically by the harness.

2. **Complete input contract**
   - `module`, `state`, `config`, `common_args`, and `module_args`.
   - The nested `expected` structure.
   - The nested `nd_queries` and JSONPath expectation structure.
   - Defaults, supported values, and phase-skipping behavior.

3. **Canonical directory and file structure**
   - Where a migrated module’s state files should live.
   - Whether the shared hierarchy or target-local hierarchy is authoritative.
   - A copyable template for adding another module.

4. **Scenario patterns**
   - Simple create/delete tests.
   - Multi-step create → update → delete workflows.
   - Multi-resource, fan-out, and split-configuration tests.
   - Passing module-specific options such as `deploy: false`.

5. **Safety and lifecycle**
   - Pre-test cleanup and `block`/`always` cleanup.
   - Destructive-test opt-in and preflight requirements.
   - ND-version gating.
   - Resolving current switch IDs from management IPs instead of maintaining two independent identities.
   - Querying controller state before and after predictive check mode to prove that no mutation occurred.

6. **Assertions and negative tests**
   - Checking phase results and module-specific returned fields.
   - REST validation and normalized controller-state comparisons.
   - Verifying the intended failure reason rather than accepting any `failed: true` result.

7. **Tags and execution**
   - Safe aggregate, state-specific, fine-grained, and destructive tags.
   - Exact `ansible-test` commands for each supported execution mode.

8. **Migration acceptance checklist**
   - Inventory every original test scenario.
   - Map each scenario to its harness replacement.
   - Confirm parity before removing or retiring the original suite.

Without this guide, the PR demonstrates the framework but does not yet make it straightforward for another contributor to use it consistently or confirm that a module’s original integration coverage has been preserved and it will help our team review it's capabilities.

mikewiebe:
**Medium: demonstrate full test parity for all four migrated modules**

Before treating this framework as complete, could we migrate every existing integration-test scenario for:

- `nd_interface_loopback`
- `nd_manage_switches`
- `nd_manage_vrf`

The current demos prove that the framework supports selected workflows, but they do not yet demonstrate that it can preserve the modules’ complete existing test coverage. Exercising all four full suites would also expose framework gaps that simplified examples may not reveal.

For each module, please:

1. Inventory every scenario in the original integration suite.
2. Map each original scenario to its framework replacement.
3. Preserve the original module-result and controller-state assertions.
4. Run the original and replacement suites against the same environment.
5. Document any scenario that cannot be represented by the framework.
6. Retain the original tests until complete parity has been demonstrated.

The migrated coverage should include, where applicable:

- Every supported state.
- Create → update → delete workflows.
- Multiple resources and split or fan-out configurations.
- Module-specific options such as `deploy: false`.
- Predictive check mode with controller queries before and after execution.
- Idempotency.
- Negative tests that verify the intended failure reason.
- ND-version and destructive-test safety gates.
- Guaranteed cleanup, including after failures.
- Controller-state validation rather than only checking returned `changed` and `failed` values.

Setup, feature-specific assertions, and cleanup can remain in the module-specific playbooks where appropriate; the framework does not need to internalize every responsibility. However, every original behavior should have an identifiable replacement.

If an existing scenario cannot be represented cleanly, that should be treated as a framework gap to address rather than dropping or weakening the test. The original suites should only be retired after the parity mapping and replacement runs are complete.

nikhilsrikrishna:
I was trying to use this framework for the Interface Groups integration tests and noticed one possible improvement around idempotency. The second execution currently verifies `changed: false`, but that does not prove it avoided redundant mutation or deployment API calls.

Could you please evaluate supporting an optional expectation such as:

```yaml
expected:
  idempotency:
    changed: false
    failed: false
    non_get_call_count: 0
```

The harness can calculate this from the idempotency result fields `api_verbs` and `api_paths`: pair the lists, exclude `GET`, and count the remaining calls. If the expected count does not match, the assertion should report the non-GET verbs and their corresponding paths.

This would detect redundant `POST`, `PUT`, `PATCH`, `DELETE`, deployment, configuration-save, or other action calls even when the module reports `changed: false`.

nikhilsrikrishna:
I was trying to use this framework for the Interface Groups integration tests and noticed another area that could improve negative-test validation. Currently, phase expectations validate only `changed` and `failed`. A test expecting `failed: true` could therefore pass even if the module failed for an unrelated reason.

Could you please evaluate supporting a per-phase message assertion, for example:

```yaml
expected:
  apply:
    changed: false
    failed: true
    msg_contains: "does not exist"
```

This could validate the `msg` returned by the target module, whether it originated from local validation or an ND API failure, and avoid separate assertion tasks in every module test suite.

mikewiebe:
### Medium: the harness is not yet at 100% `develop`-branch coverage for all three target modules

**Issue**

To determine whether the harness conversion provides 100% coverage, I compared
**all existing integration-test coverage on the `develop` branch** for each of
the following three modules against its harness-based replacement in this PR:

- `nd_interface_loopback`
- `nd_interface_vpc_access`

This was a full module-by-module comparison, not a sample of representative
tests. It included every setup path, state workflow, input variant, assertion,
idempotency check, negative path, optional execution mode, and cleanup path in
the three `develop` integration targets. The comparison found missing,
weakened, and unverified coverage in the PR, so none of the three modules has
yet demonstrated 100% parity with its complete `develop` suite.

**Evidence**

- [`nd_interface_loopback/PARITY.md` lines 132-140](https://github.com/CiscoDevNet/ansible-nd/pull/337/files#diff-798fd4c0c2bf8efbd583eb62dc39c99ee964729d0cd0eed3ec4c04785c7a25a6R132)
  says no scenarios are unmapped and all original scenarios have replacements,
  but the exact source comparison found the gaps below.
- The comparison used exact commits:
  [`develop` `937380de`](https://github.com/CiscoDevNet/ansible-nd/commit/937380de68e252dcec3dc5c5a6aa4d61169efc69)
  and [PR head `a0e0aaee`](https://github.com/CiscoDevNet/ansible-nd/commit/a0e0aaee2a519f4351181361263b6921923eb516).

These percentages estimate **scenario parity**, not Python line or branch
coverage. Each distinct `develop` behavior is scored as full (`1.0`), partial
or weakened (`0.5`), or missing or broken (`0`). Harness-only additions,
cutover status, CI status, and live-evidence quality are excluded from the
percentage. The ranges reflect reasonable differences in how closely related
behaviors are grouped.

| Module compared in full against `develop` | Estimated effective parity | Reasonable range | Score | 100% parity demonstrated? |
| --- | ---: | ---: | --- | --- |
| `nd_interface_loopback` | **~85%** | 84-90% | 17 full-equivalent behaviors out of 20 | No |
| `nd_interface_ethernet_access` | **~95%** | 95-98% | 26 full + 3 partial out of 29 | No |
| `nd_interface_vpc_access` | **~55%** | 52-58% | 12 full + 7 partial + 9 missing out of 28 | No |

The weighted estimate across the three complete `develop` comparisons is
approximately **78%** (`60` full-equivalent behavior points out of `77`).
Ethernet access has a harness counterpart for every named scenario, but its
effective parity is below 100% because the replacement workflow and assertion
fidelity are weaker than the complete sequential `develop` behavior.

The detailed gaps found while comparing every scenario are:

| Classification | Module / area | Coverage gap |
| --- | --- | --- |
| Missing | `nd_interface_loopback` CIDR regression | Current `develop` [creates `loopback104` from `10.100.104.1/32`, verifies the normalized bare address, reapplies it idempotently, and cleans it up](https://github.com/CiscoDevNet/ansible-nd/blob/937380de68e252dcec3dc5c5a6aa4d61169efc69/tests/integration/targets/nd_interface_loopback/tasks/merged.yaml#L146-L192). No harness scenario covers this workflow. |
| Broken replacement | `nd_interface_loopback` no-deploy | Current `develop` [uses `config_actions.deploy: false`](https://github.com/CiscoDevNet/ansible-nd/blob/937380de68e252dcec3dc5c5a6aa4d61169efc69/tests/integration/targets/nd_interface_loopback/tasks/merged.yaml#L111-L131), but the harness [passes unsupported top-level `deploy: false`](https://github.com/CiscoDevNet/ansible-nd/pull/337/files#diff-719aa0d2227fc2f5e8bedd7a23568d04fceb31ea9803c280a34c72aeffeb3351R254). The [action plugin forwards `module_args` unchanged](https://github.com/CiscoDevNet/ansible-nd/blob/a0e0aaee2a519f4351181361263b6921923eb516/plugins/action/nd4x_module_test.py#L368-L374), while the [module reads only `config_actions.deploy`](https://github.com/CiscoDevNet/ansible-nd/blob/a0e0aaee2a519f4351181361263b6921923eb516/plugins/modules/nd_interface_loopback.py#L295-L322), so the current replacement cannot execute the intended workflow. |
| Partial / weakened | `nd_interface_ethernet_access` replaced workflow | The sequential `develop` suite [creates Ethernet1/45-47 and Ethernet1/49-50](https://github.com/CiscoDevNet/ansible-nd/blob/937380de68e252dcec3dc5c5a6aa4d61169efc69/tests/integration/targets/nd_interface_ethernet_access/tasks/merged.yaml#L217-L302) before entering `replaced`, but the harness [reconstructs only Ethernet1/41-44](https://github.com/CiscoDevNet/ansible-nd/pull/337/files#diff-eb2bcdbe2de6d29068db5f1c4af0b2847dda9fd3d037eecd73b119ae0df3b474R4). This removes coverage for replacement while unrelated managed access interfaces coexist. |
| Missing | `nd_interface_vpc_access` pre-existing-pair mode | Current `develop` [supports `nd_test_manage_vpc_pairs=false`](https://github.com/CiscoDevNet/ansible-nd/blob/937380de68e252dcec3dc5c5a6aa4d61169efc69/tests/integration/targets/nd_interface_vpc_access/vars/main.yaml#L43-L46), which leaves an existing pair untouched during setup and teardown. Every mutable harness state instead invokes [`nd4x_demo_pair_setup.yaml`](https://github.com/CiscoDevNet/ansible-nd/pull/337/files#diff-84e8b678ecb9cd3b9c285f2fa3be89d43dd1ee4fcbadc18563969b5db284e9c8R2), requires pair-reset opt-in, and deletes and recreates the pair. |
| Missing | `nd_interface_vpc_access` multi-pair workflow | Current `develop` [tests the second-pair precondition, the same vPC name and ID on two different pairs, same-pair duplicate rejection, idempotency, pair-scoped `overridden`, and cleanup](https://github.com/CiscoDevNet/ansible-nd/blob/937380de68e252dcec3dc5c5a6aa4d61169efc69/tests/integration/targets/nd_interface_vpc_access/tasks/multi_pair.yaml#L16-L154). The PR [explicitly acknowledges that this workflow is not migrated](https://github.com/CiscoDevNet/ansible-nd/pull/337/files#diff-3d5893b7972144408d130054791cf32c019c46094be54716e322a3fe7f70b5f9R148). |
| Partial / weakened | `nd_interface_vpc_access` workflow and fixture fidelity | The original suite runs `merged` through `deleted` sequentially, while the harness reconstructs isolated pre-states and its own parity document [lists configured fields those pre-state assertions do not verify](https://github.com/CiscoDevNet/ansible-nd/pull/337/files#diff-3d5893b7972144408d130054791cf32c019c46094be54716e322a3fe7f70b5f9R137). Shared fixtures also change the original virtual peer link with explicit pair details to a physical peer link without those details, and change the updated member from Ethernet1/6 to Ethernet1/8, so the original and replacement runs do not exercise equivalent fixture semantics. |
| Incomplete conversion | Cutover status | The [legacy state sequence remains the default and the harness remains behind `never` tags](https://github.com/CiscoDevNet/ansible-nd/pull/337/files#diff-5d45216f0503901fd71ecb038dcdb1d2057d5131ba1423bba3903ed6a0bd26c5R74). Consequently, none of the three targets has actually been cut over to the harness. Keeping the originals enabled is the correct safeguard until parity is complete. |
| Unverified | Reproducible proof | The retained live records do not demonstrate exact-head versus exact-`develop` parity: [loopback runs predate the current CIDR and `config_actions` changes](https://github.com/CiscoDevNet/ansible-nd/pull/337/files#diff-798fd4c0c2bf8efbd583eb62dc39c99ee964729d0cd0eed3ec4c04785c7a25a6R111); [Ethernet](https://github.com/CiscoDevNet/ansible-nd/pull/337/files#diff-49197e6272d695ac43ce6f11356ca0141e26f2ab92e550f433ddf879a8d4be3bR160) and [vPC](https://github.com/CiscoDevNet/ansible-nd/pull/337/files#diff-3d5893b7972144408d130054791cf32c019c46094be54716e322a3fe7f70b5f9R95) records used older baselines plus uncommitted harness content; the referenced JUnit files are absent from the PR; and the current `Integration Tests` check is skipped. No interface harness scenario requests `config_actions.deploy: true`, so these records prove staged controller intent rather than switch deployment. |

**Existing PR overlap**

Related to [the earlier complete-parity request from `@mikewiebe`](https://github.com/CiscoDevNet/ansible-nd/pull/337#issuecomment-5136030850).
The vPC parity document acknowledges the multi-pair gap, but no existing PR
comment consolidates the exact current-`develop` comparison or the remaining
loopback, Ethernet, vPC, cutover, and evidence gaps above.

Related open issue [#293](https://github.com/CiscoDevNet/ansible-nd/issues/293)
defines the common harness strategy and ND-version-gating goal, but it does not
track these concrete three-module parity and cutover gaps.

**Impact**

The original suites still protect default coverage today. However, the result
of the full comparison is currently **zero of three modules at demonstrated
100% `develop` parity**: approximately 85% for loopback, 95% for Ethernet
access, and 55% for vPC access, with a weighted estimate of approximately 78%
across all three. Accepting the current parity claims or retiring those suites
would lose regression coverage for CIDR normalization, no-deploy behavior,
Ethernet replacement with coexisting resources, pre-existing vPC pairs, and
multi-pair vPC identity and override behavior. The existing execution records
also cannot prove that the exact code proposed by this PR passed equivalent
testing.

**Suggested fix**

Close the identified gaps for all three modules, then repeat the complete
module-by-module comparison against the current `develop` branch:

- [x] Add the loopback CIDR normalization and idempotency workflow.
- [x] Change the loopback no-deploy harness arguments to
      `config_actions: {deploy: false}`.
- [x] Reconstruct the complete Ethernet post-merged state before `replaced`.
- [ ] Add a vPC path that preserves a pre-existing pair, or explicitly obtain
      agreement to retire that mode.
- [x] Migrate the complete vPC multi-pair workflow.
- [ ] Preserve equivalent vPC fixture semantics and fully assert reconstructed
      pre-states.
- [ ] Run the exact PR head and exact current `develop` suites on the same
      testbed, retain reviewable artifacts, and document whether deployment to
      switches is in scope.
- [ ] Keep the legacy suites enabled until the comparison is complete; then
      make the harness the default and perform the actual cutover.

astawast-cisco:
The original harness review comments have been resolved. Loopback CIDR normalization/idempotency coverage and the nested config_actions.deploy: false workflow are now implemented. Ethernet-access and vPC harness parity coverage has also been completed. The remaining loopback gaps are the newer IOS-XE, IPv6 CIDR, and explicit discriminator scenarios introduced later in develop; these are intentionally not claimed as complete in this PR.

astawast-cisco:
Identified harness fixes have been implemented: the no-deploy workflow now uses module_args.config_actions.deploy: false, and the CIDR regression now covers loopback204 with IPv4 and IPv6 normalization, idempotency, cleanup, and API validation. Remaining validation is the optional IOS-XE harness coverage.

astawast-cisco:
This change adds harness coverage for module nd_interface_ethernet_trunk_host ,the original merged, replaced, overridden, and deleted workflows, including:

- Check-mode validation with complete fabric snapshots
- Apply and idempotency phase assertions
- Controller-state validation through nd_queries
- Exact module-result assertions
- Independent prerequisite reconstruction for each mutable workflow
- Shared ND-version and switch-identifier discovery
- Reserved-interface and destructive-scope safety checks
- Cleanup re-query probe coverage
- Storm-control regression scenarios
- Exact negative validation coverage
- Separate destructive tags: nd4x_demo_overridden and nd4x_demo_destructive

tests/integration/inventory.networking
```diff
@@ -1,31 +1,20 @@
 [nd]
-nd ansible_host=<TODO>
+nd ansible_host=192.0.2.1
 
 [nd:vars]
```
allenrobel, line 4: This rewrite drops the whole `[nd:vars]` scaffolding that the other integration targets rely on: `insights_group`, `site_name`, `site_host`, `site_username`, `site_password`, `dns_server`, `app_network`, `service_network`, `ntp_server`, `serial_number`, `management_ip_address`, `deployment_password`, `management_ip`, `management_gateway`, `external_management_service_ip`, `external_data_service_ip`, `data_ip`, `data_gateway`, `service_package_host`.

Twenty targets that this PR does not touch template those keys directly with no fallback defaults (`nd_setup`, `nd_site`, `nd_pcv`, all `nd_compliance_*`, `nd_service`, `nd_service_instance`, `nd_snapshot`, `nd_flow_rules`, ...). After this change a `network-integration` run of any of them fails with an undefined-variable error instead of a soft placeholder.

Suggest keeping the existing `<TODO>` placeholders and adding the new `nd_test_fabric_name` / `nd_test_switch_ip` / `nd_test_switch_id` keys alongside them. (Replacing the lab IP in `service_package_host` with a placeholder is a good change on its own.)

tests/integration/playbooks/nd4x_module_tests/module_tests/nd_interface_loopback/merged.yaml
```diff
@@ -0,0 +1,49 @@
...
+          check_mode_changed: true
+          apply_changed: true
+          idempotency: true
+          failed: false
+        idempotency_retries: 6
+        idempotency_delay: 10
```
allenrobel, line 21: This example cannot run as written against the plugin in this PR:

- `expected` uses the flat `check_mode_changed` / `apply_changed` / `idempotency` / `failed` keys. `_normalize_expected()` calls `_reject_unknown_keys()` against `ALLOWED_EXPECTED_PHASES = {"check_mode", "apply", "idempotency"}`, so this raises `AnsibleActionFail` on the first key. The task files under `targets/` use the correct nested shape (`expected: {check_mode: {changed: true, failed: false}, ...}`).
- `idempotency_retries: 6` and `idempotency_delay: 10` are hard-rejected in `run()` ("must be 1" / "must be 0").

Since this is the one example a new contributor will copy, it should match the plugin. Separately, the README in this same directory says the `playbooks/` hierarchy "is not a second source of runnable integration scenarios", yet this file is a full runnable scenario (register, assert, `always` cleanup, tags) that duplicates `targets/nd_interface_loopback/tasks/nd4x_demo_merged.yaml`. Consider either trimming it to a non-runnable snippet or removing it in favor of the target-local file.

tests/integration/playbooks/nd4x_module_tests/README.md
```diff
@@ -0,0 +1,1879 @@
...
+common_args: {}
+```
+
+Example:
+
+```yaml
+common_args:
+  fabric_name: "{{ test_fabric_name }}"
+  output_level: "{{ nd_info.output_level }}"
+  timeout: 300
```
allenrobel, line 249: `timeout: 300` inside `common_args` is passed straight through to the target module (the plugin merges `common_args` into the top-level module args), and no module in the collection accepts a top-level `timeout` option. The only `timeout` that exists is `nd_manage_vpc_pair`'s `verify.timeout` suboption; the others are named `wait_timeout` / `job_wait_timeout`. So anyone copying this example gets an argument-validation failure. PR #454 hit exactly this and PR #486 fixed it at the source.

The task files in this PR already do the right thing with a block-scoped `vars: ansible_command_timeout: 300`; the README example (here and again at line 947) should show that instead.

plugins/action/nd4x_module_test.py
```diff
@@ -0,0 +1,1082 @@
...
+        "state",
+        "config",
+        "common_args",
+        "module_args",
+        "expected",
+        "check_mode",
+        "check_mode_queries",
+        "idempotency",
+        "idempotency_retries",
+        "idempotency_delay",
```
allenrobel, line 40: `idempotency_retries` and `idempotency_delay` are accepted here, so `_reject_unknown_keys()` will not flag them, but they appear nowhere in the `DOCUMENTATION` of `plugins/modules/nd4x_module_test.py`, and `run()` hard-fails any value other than 1 / 0. A user who guesses `idempotency_retries: 3` gets a cryptic failure with no doc entry explaining why. Either document both (with the restriction) or drop them from `ALLOWED_ARGUMENTS` until retries are actually supported.

allenrobel (commented):
### Code review

Nine inline comments. The two that matter most: the rewritten `inventory.networking` drops the `[nd:vars]` placeholders that twenty untouched integration targets template directly, and the one shipped example under `playbooks/` uses an `expected` shape and retry knobs the action plugin itself rejects. The rest are doc/convention items.

plugins/modules/nd4x_module_test.py
```diff
@@ -0,0 +1,187 @@
...
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+DOCUMENTATION = r"""
+---
+module: nd4x_module_test
+short_description: Run standardized integration tests for ND 4.x modules
+version_added: "1.6.0"
```
allenrobel, line 15: `version_added: "1.6.0"`: the latest tag is v1.5.0 and the unreleased cohort on develop declares `version_added: "2.0.0"` (31 modules). PR #366 settled on marking new modules with the release they first ship in, so this should be `"2.0.0"`.

tests/integration/targets/nd_manage_fabric/tasks/main.yaml
```diff
@@ -5,6 +5,9 @@
 - name: Run nd_manage_fabric eBGP tests
   ansible.builtin.include_tasks: fabric_ebgp.yaml
 
+#- name: Run nd_manage_fabric eBGP tests
+ # ansible.builtin.import_tasks: fabric_ebgp.yaml
```
allenrobel, line 9: Leftover commented-out duplicate of the eBGP include just above it, in a target this PR otherwise does not touch. Please drop it.

plugins/action/nd4x_module_test.py
```diff
@@ -0,0 +1,1082 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2026, Cisco Systems, Inc.
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
```
allenrobel, line 8: Python 2 boilerplate in a new file. CLAUDE.md: "`__metaclass__ = type` is no longer needed and can be removed" and "Boilerplate `from future ...` imports are no longer needed and can be removed (with the exception of annotations)". The collection floor is Python 3.10.

plugins/modules/nd4x_module_test.py
```diff
@@ -0,0 +1,187 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2026, Cisco Systems, Inc.
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
```
allenrobel, line 9: Same as the action plugin: drop `from __future__ import absolute_import, division, print_function` and `__metaclass__ = type`. If you add `X | None` annotations later, `from __future__ import annotations` is the one future import that stays.

tests/integration/targets/nd_interface_ethernet_trunk_host/tasks/main.yaml
```diff
@@ -16,7 +16,8 @@
 # TODO: Modify the above (and below) if the team decides to use tests/integration/integration_config.yml instead.
 #
 # supports_vlan_mapping (bool, default: false)
-#   Enables the VLAN-mapping / dot1q-tunnel test block in merged.yaml. Skipped by
+#   Enables the VLAN-mapping / dot1q-tunnel test workflow in
+#   nd4x_demo_vlan_mapping.yaml. Skipped by
```
allenrobel, line 20: This comment now points at `nd4x_demo_vlan_mapping.yaml`, but that file is `never`-tagged and only runs with `--tags nd4x_demo...`. The block that `supports_vlan_mapping=true` actually enables in a normal `network-integration` run is still the one in `merged.yaml` (unchanged by this PR). Suggest mentioning both, or restoring the original wording.

requirements.txt
```diff
@@ -2,3 +2,4 @@ requests_toolbelt
 jsonpath-ng
 lxml
 pydantic==2.12.5
+eval_type_backport; python_version < "3.10"
```
allenrobel, line 5: `eval_type_backport; python_version < "3.10"` can never install: the collection requires Python >= 3.10 (`pyproject.toml` and CLAUDE.md), so the marker is always false. If nothing in this PR needs it on 3.10+, drop the line here and in `tests/integration/network-integration.requirements.txt`.

tests/integration/inventory.networking
astawast-cisco: Thanks, good catch. I restored the complete existing [nd:vars] scaffold, including the variables used by the unrelated integration targets. I also replaced the lab-specific service_package_host value with a <TODO> placeholder and added nd_test_fabric_name, nd_test_switch_ip, and nd_test_switch_id for the ND4X harness. Real testbed values remain only in my private inventory.LOCAL.networking file and are not committed. The fix has been pushed.

Issue #335, opened 2026-06-16
nd_manage_vpc_pair: physical vPC peer-link creation fails (deploys via fabric actions/configSave; switchActions/deploy works)


allenrobel:
## Summary

`nd_manage_vpc_pair` (added in #197) cannot create a vPC pair with a **physical peer-link**. The intent `PUT .../vpcPair` succeeds (HTTP 204), but the deploy step — fabric-level `POST .../actions/configSave` — returns **HTTP 500** and the pair never forms:

```
[<sn>/<host>] - Unexpected error generating vPC configuration. Last Status [Peer Link Interface]
Switch [<host>/<sn>]: Member port Ethernet1/2 is a member of VPC_PAIR, remove Ethernet1/2 from VPC_PAIR first.
```

A **virtual** peer-link (`use_virtual_peer_link: true`) works fine through the same module. The failure is specific to **physical** peer-links (i.e. configs that set `switch_member_interfaces` / `peer_switch_member_interfaces` / `switch_po_id` under `vpc_pair_details`).

## Root cause

It's the **deploy mechanism**, not the intent. The config-generation succeeds via the **switch-scoped** deploy and fails via the **fabric-scoped** config-save — same intent, same switches, only the endpoint differs (lab-verified via the raw ND API, independent of any module):

| Path | Endpoint(s) | Physical peer-link |
| --- | --- | --- |
| Switch-level | `POST .../switchActions/deploy {"switchIds":[le1,le2]}` | ✅ works — `switchActions/preview` shows the generated CLI; `deploy` returns "Deployed Successfully"; pair forms (`vpcConfigured: true` on both, roles elected) |
| Fabric-level | `POST .../actions/configSave` then `.../actions/deploy` | ❌ HTTP 500 "Member port … is a member of VPC_PAIR" |

`nd_manage_vpc_pair` deploys via the **fabric-level `actions/configSave` + `actions/deploy`** path (`plugins/module_utils/manage_vpc_pair/deploy.py` → `EpFabricConfigSavePost` / `EpFabricDeployPost`). The peer-link member interface is a clean `trunkHost` when unpaired, so this is ND's config generator self-conflicting on the peer-link member through the fabric path — not a pre-existing claim on the interface.

(There is no switch-level config-save endpoint; `switchActions/deploy` itself generates+pushes. So the fabric `configSave` is the only path `nd_manage_vpc_pair` currently uses, and it's the broken one for physical peer-links.)

## Why it wasn't caught

[...]

PR #338, opened 2026-06-18
Fix vPC pair physical interface deployment

sivakasi-cisco:
## Related Issue(s)

- CiscoDevNet/ansible-nd#335

## Problem Statement

The nd_manage_vpc_pair was failing when creating a vPC pair with a physical peer-link.
The vpc_pair module was able to tell ND what vPC pair we wanted, but the final deploy step failed before the configuration could actually be pushed to the switches. This only happened for physical peer-links with member interfaces. Virtual peer-links were working fine.

## Root cause

The module was using the fabric-level deploy flow: `actions/configSave` + `actions/deploy`

That path fails in ND for physical vPC peer-link creation because ND tries to generate the configuration fabric-wide and incorrectly reports that the peer-link member port is already part of the vPC pair.

The same configuration works when deployed directly at the switch level using `switchActions/deploy` So the problem was the deploy API path being used.

## Proposed Changes

Updated `nd_manage_vpc_pair` so the default switch-scoped deployment uses `switchActions/deploy` with the affected switch serial numbers.

This means the module now asks ND to deploy only to the switches involved in the vPC pair, which avoids the failing fabric-level config-save path and allows physical peer-link vPC pair creation to complete successfully.

The fabric-level deploy path is still kept for global deployment mode or fallback cases.

## Test Notes

Updated the vPC pair integration task assertions for merge, replace, override, and delete flows to verify that switch-level deployment uses `/switchActions/deploy` instead of the fabric-level `/actions/configSave` path.


allenrobel (commented):
### Code review

Reviewed the switch-scoped deploy refactor for correctness and CLAUDE.md compliance. Three findings inline: a missing workaround marker (CLAUDE.md-mandated), an unchecked 207 multi-status response that can mask per-switch deploy failures, and a now-stale `custom_vpc_deploy` docstring. No issues found in the integration-test or unit-test changes.

plugins/module_utils/manage_vpc_pair/deploy.py
```diff
@@ -282,6 +329,118 @@ def custom_vpc_deploy(nrm: Any, fabric_name: str, result: dict[str, Any]) -> dic
...
+        return deployment_info
+
+    # Initialize RestSend via NDModuleV2
+    nd_v2 = NDModuleV2(nrm.module)
+    results = Results()
+
+    # Switch-scoped deploy (config_actions.type == "switch", the default) targets
+    # only the affected switches via switchActions/deploy, which generates and
+    # pushes config in a single call. This avoids the fabric-level actions/configSave
+    # path that returns HTTP 500 for vPC pairs with a physical peer-link (issue #335).
```
allenrobel: This switch-scoped deploy path works around a documented ND API discrepancy — fabric-level `actions/configSave` returns HTTP 500 for vPC pairs with a physical peer-link. Per CLAUDE.md's *Workaround Markers* section: "Any code that works around an ND API discrepancy (wire shape vs. schema, wrong HTTP status, silent field drops, etc.) MUST be flagged with a `TODO(X.Y.Z) <slug>` marker so the workaround can be audited and removed when the ND release cadence permits." This site (and the `_switch_level_deploy` docstring at line 193) cites `issue #335` in prose but carries no marker. A matching bug-tracker vault note already exists — please add `# TODO(4.2.1) vpc-physical-peer-link-configsave-fails` above the switch-vs-fabric dispatch.

plugins/module_utils/manage_vpc_pair/deploy.py
```diff
@@ -102,104 +109,144 @@ def _is_non_fatal_config_save_error(error: NDModuleError) -> bool:
...
-    # Initialize RestSend via NDModuleV2
-    nd_v2 = NDModuleV2(nrm.module)
+    Raises:
+        VpcPairResourceError: If the deploy call fails.
+    """
+    deploy_path = _build_switch_deploy_path(fabric_name)
+    payload = {"switchIds": switch_ids}
+
+    try:
+        nd_v2.request(deploy_path, HttpVerbEnum.POST, payload)
```
allenrobel: `_switch_level_deploy` discards the response of `switchActions/deploy` and unconditionally registers `success=True, changed=True` unless an `NDModuleError` is raised. But per the ND OpenAPI spec this endpoint returns HTTP 207 (Multi-Status) with a `DATA.switchIds[]` array carrying a per-switch `status: success|failed` — the spec's own example shows one switch succeeding and its peer failing in the same 207. `NdV1Strategy.is_success()` treats 207 as success as long as there is no top-level `ERROR`/`DATA.error` key; it never inspects `switchIds[].status`, so `nd_v2.request()` does not raise on a partial failure. Since `_collect_affected_switch_ids()` typically sends both peers of a pair in one call, a per-switch deploy failure (one box fails, its peer succeeds) is silently reported as full success/changed, defeating idempotency. Consider capturing the return value and failing if any `switchIds[].status == "failed"`.

plugins/module_utils/manage_vpc_pair/deploy.py
```diff
@@ -282,6 +329,118 @@ def custom_vpc_deploy(nrm: Any, fabric_name: str, result: dict[str, Any]) -> dic
             final_msg = final_result.pop("msg", "Fabric deployment failed")
             _raise_vpc_error(msg=final_msg, **final_result)
 
+
+def custom_vpc_deploy(nrm: Any, fabric_name: str, result: dict[str, Any]) -> dict[str, Any]:
+    """
+    Custom save/deploy action handler for vPC fabric changes using RestSend.
+
+    - Smart action decision (_needs_deployment)
+    - Optional Step 1: Save fabric configuration
```
allenrobel, line 144: This docstring still describes only the old fabric-only two-step flow (`Optional Step 1: Save fabric configuration` / `Optional Step 2: Deploy fabric with forceShowRun=true`), but after this PR the default path for `config_actions.type == "switch"` is the single-call switch-scoped `switchActions/deploy` (`_switch_level_deploy`); the two-step save/deploy is now only the `type == "global"` / no-serials fallback. Relatedly, `Raises: NDModuleError` (line 353) is stale — the caller-visible exception is `VpcPairResourceError` (raised via `_raise_vpc_error`). Worth updating the docstring to describe the switch-scoped-vs-fabric branching.

sivakasi-cisco:
Hi Allen,

Thanks for the review. I rechecked the physical vPC pair behavior against ND/UI.

The earlier assumption was that physical peer-link vPC creation needed a different deploy path. After checking the actual UI/API behavior, that is not the real issue.

The issue is not that physical peer-link vPC pairing needs a switch-level deploy workaround. The issue is that for VXLAN iBGP/eBGP fabrics, ND/UI only supports the simple pair payload with `switchId`, `peerSwitchId`, and `useVirtualPeerLink=false`. It does not accept `vpcPairDetails` for that scenario. It accepts only the below.
{
  "vpcAction": "pair",
  "switchId": "<switch1Serial>",
  "peerSwitchId": "<switch2Serial>",
  "useVirtualPeerLink": false
}

The UI itself does not send or accept `vpcPairDetails` for this scenario. If we send extra details like domain ID, keepalive IPs, peer-link port-channel, member interfaces, VLANs, etc., ND rejects it / can hit the HTTP 500 path.

I removed the switch deploy workaround from the PR. The PR now only blocks unsupported `vpc_pair_details` on VXLAN iBGP/eBGP fabrics and keeps those details allowed for supported fabric types.

So the earlier comments on `_switch_level_deploy`, HTTP 207 handling, and the deploy docstring should be obsolete for the current diff. Please take another look at the updated PR.

So the original switch-level deploy workaround is no longer needed. I removed that change from the PR. The PR now focuses only on validating the input earlier:

- block `vpc_pair_details` for VXLAN iBGP/eBGP fabrics
- keep allowing `vpc_pair_details` for fabrics where those details are supported
- add integration coverage for both cases

Kindly re-review the updated diff when you get a chance and also share your comments on this

plugins/module_utils/manage_vpc_pair/deploy.py
sivakasi-cisco: Added a consolidated reply

plugins/module_utils/manage_vpc_pair/deploy.py
sivakasi-cisco: Added a consolidated reply

plugins/module_utils/manage_vpc_pair/deploy.py
sivakasi-cisco: Added a consolidated reply

allenrobel:
> Kindly re-review the updated diff when you get a chance and also share your comments on this

Have re-reviewed based on your analysis of the root cause and this looks good.  Thank you @sivakasi-cisco!

allenrobel (approved):
LGTM

plugins/module_utils/manage_vpc_pair/actions.py
```diff
@@ -39,6 +41,118 @@
...
+    """
+    cached = module.params.get("_fabric_type_token")
+    if isinstance(cached, str) and cached:
+        return cached
+
+    details_path = f"/api/v1/manage/fabrics/{quote(fabric_name, safe='')}"
+    try:
+        details = nd_v2.request(details_path, HttpVerbEnum.GET)
+    except Exception as exc:
+        module.warn(f"Unable to determine fabric type for '{fabric_name}' while validating vpc_pair_details: " f"{str(exc).splitlines()[0]}")
```
akinross: when splitlines() returns [] it will result in IndexError raise

plugins/module_utils/manage_vpc_pair/actions.py
```diff
@@ -39,6 +41,118 @@
...
+        for key in ("fabricType", "fabricTechnology", "type"):
+            value = properties.get(key)
+            if isinstance(value, str):
+                candidates.append(value)
+
+    for candidate in candidates:
+        token = _normalize_fabric_type_token(candidate)
+        if token:
+            module.params["_fabric_type_token"] = token
+            return token
```
akinross: could we add some explanation on how/why this is the order is set for candidates while determining the type? could you get into a wrong order and return the wrong type?

plugins/module_utils/manage_vpc_pair/actions.py
```diff
@@ -39,6 +41,118 @@
...
+    if not isinstance(value, str):
+        return ""
+    return re.sub(r"[^a-z0-9]", "", value.strip().lower())
+
+
+def _resolve_fabric_type_token(nd_v2: NDModuleV2, fabric_name: str, module: Any) -> str:
+    """
+    Resolve and cache the normalized fabric-type token for current run.
+
+    Best-effort lookup from fabric details endpoint. Returns empty string when
```
akinross: are there cases where it is not possible to determine the fabric_type? I am a bit confused why this function is best-effort? can this lead to downstream error when the type is ""?

plugins/module_utils/manage_vpc_pair/actions.py
```diff
@@ -39,6 +41,118 @@
...
+    for candidate in candidates:
+        token = _normalize_fabric_type_token(candidate)
+        if token:
+            module.params["_fabric_type_token"] = token
+            return token
+
+    return ""
+
+
+def _get_proposed_vpc_pair_details(proposed_config: Any) -> Any:
```
akinross: should we make the return type more explicit? because inputting and returning any seems like a very unpredictable behaviour

plugins/module_utils/manage_vpc_pair/actions.py
```diff
@@ -39,6 +41,118 @@
...
+    """Return proposed vpc_pair_details (snake_case or API key), if present."""
+    if not isinstance(proposed_config, dict):
+        return None
+
+    details = proposed_config.get("vpc_pair_details")
+    if details is None:
```
akinross, line 143: why only this check on None and not on empty values like {}, or []? because the caller function later does `if not proposed_details:`, should this also be None check?

akinross (changes requested):

plugins/module_utils/manage_vpc_pair/actions.py
sivakasi-cisco: Replaced splitlines()[0] style warning extraction with a safe helper to avoid  indexing errors using safe first-line helper with fallback text when exception message is empty

plugins/module_utils/manage_vpc_pair/actions.py
sivakasi-cisco: I have clarified the behavior contract in docstring. This path is intentionally fail-open. If fabric type cannot be determined due transient lookup/API shape, we return empty and skip the block check to avoid false negatives. The block is only enforced when we can confidently classify the fabric as VXLAN iBGP/eBGP.

plugins/module_utils/manage_vpc_pair/actions.py
sivakasi-cisco: I have updated the helper docstring to document the lookup order. At first, we check the top-level fabric fields, then management.type, and finally properties fields.

This ensures nested metadata does not accidentally override canonical top-level fabric type values.

plugins/module_utils/manage_vpc_pair/actions.py
sivakasi-cisco: I have cleaned this up so empty or invalid values are treated as None in the helper. The caller now follows that same rule.

sivakasi-cisco (commented):
Addressed the comments

plugins/module_utils/manage_vpc_pair/actions.py
```diff
@@ -39,6 +41,136 @@
...
+
+        The lookup is fail-open: when type cannot be determined, return empty string
+        and skip the VXLAN iBGP/eBGP vpc_pair_details block to avoid false
+        negatives caused by transient fabric-details lookup failures.
+
+        Candidate precedence is deterministic and favors top-level fabric type
+        fields over nested properties:
+            1. details.fabricType / fabricTechnology / type / category
+            2. details.management.type
+            3. details.properties.fabricType / fabricTechnology / type
```
akinross: I do not understand why this is needed because fabric endpoint always exposes management.type. Tested it for 14 fabric types

```
## Summary

| Fabric name | `management.type` |
|---|---|
| `akinross_vxlan_ibgp` | `vxlanIbgp` |
| `akinross_vxlan_ebgp` | `vxlanEbgp` |
| `akinross_vxlan_campus` | `vxlanCampus` |
| `akinross_vxlan_external` | `vxlanExternal` |
| `akinross_classic_lan_enhanced` | `classicLanEnhanced` |
| `akinross_classic_lan` | `classicLan` |
| `akinross_aiml_vxlan_ibgp` | `aimlVxlanIbgp` |
| `akinross_aiml_vxlan_ebgp` | `aimlVxlanEbgp` |
| `akinross_aiml_routed` | `aimlRouted` |
| `akinross_external_connectivity` | `externalConnectivity` |
| `akinross_routed` | `routed` |
| `akinross_ipfm_enhanced` | `ipfmEnhanced` |
| `akinross_ipfm` | `ipfm` |
| `akinross_data_broker` | `dataBroker` |

**Review by akinross — CHANGES_REQUESTED** [review]

File: plugins/module_utils/manage_vpc_pair/actions.py (hunk 9 above)
Comment (sivakasi-cisco) [reply plugins/module_utils/manage_vpc_pair/actions.py]: We need to block the other unsupported fabric types here. We need to allow only ibgp and ebgp.

For that, we require the fabricType to be extracted.

At some cases, We do not see the ' fabricType'. As mentioned above, in case of switch endpoint - GET /api/v1/manage/fabrics/<fabricName>/switches

So, I fallback to management.type to fetch that information instead of 'fabricType'

Comment (sivakasi-cisco) [reply plugins/module_utils/manage_vpc_pair/actions.py]: Hi Akkini,

I have addressed the comments we discussed about. Kindly review.

File: plugins/module_utils/manage_vpc_pair/actions.py [hunk 10]
```diff
@@ -34,11 +35,114 @@
...
+    Resolve and cache the canonical fabric type for the current run.
+
+    Nexus Dashboard returns the fabric type in ``management.type``. The lookup
+    is fail-open: when the type cannot be determined, return an empty string
+    and skip the VXLAN iBGP/eBGP vpc_pair_details block to avoid false
+    negatives caused by transient lookup failures.
+
+    Candidate precedence is deterministic and based on observed response shapes:
+      1. /api/v1/manage/fabrics/<name> -> management.type
+      2. /api/v1/manage/fabrics/<name>/switches -> fabricType
```
akinross: when is /switches used? 

I see either module.params.get("_fabric_type"), assume this is some cached value? Why do we store this in module_params should we not have a dedicated cache somewhere if we want to leverage caching values?

plugins/module_utils/manage_vpc_pair/actions.py
```diff
@@ -34,11 +35,114 @@
...
+    """
+    cached = module.params.get("_fabric_type")
+    if isinstance(cached, str) and cached:
+        return cached
```
akinross: not sure I understand this isinstance check since it is only ever written as a string? this also goes back to separate caching mechanism. If we create a dedicated caching mechanism ( perhaps a class ) then these can contain methods for retrieving and setting cached items.

plugins/module_utils/manage_vpc_pair/actions.py
```diff
@@ -34,11 +35,114 @@
...
+
+    details_path = f"/api/v1/manage/fabrics/{quote(fabric_name, safe='')}"
```
akinross: why don't we use the endpoint for this from the collection because we now define path in multiple locations?

plugins/module_utils/manage_vpc_pair/actions.py
```diff
@@ -34,11 +35,114 @@
...
+    try:
+        details = nd_v2.request(details_path, HttpVerbEnum.GET)
```
akinross, line 71: i am confused on the nd_v2 usage, cant we leverage restsend for this?

plugins/module_utils/manage_vpc_pair/actions.py
```diff
@@ -34,11 +35,114 @@
...
+    except Exception as exc:
```
akinross: we should specify more clear exceptions because this is very broad exception handling

plugins/module_utils/manage_vpc_pair/actions.py
```diff
@@ -34,11 +35,114 @@
...
+}
+
+
+def _first_line(value: Any) -> str:
+    """Return first non-empty line from value converted to string."""
+    text = str(value).strip()
+    if not text:
+        return "unknown error"
+    lines = text.splitlines()
+    return lines[0] if lines else "unknown error"
```
akinross: why accept any value? if None would be passed in this would return "None"?

why return "unknown error" and not raise an actual error?

I think the function name is now very generic for the thing you are trying to resolve. So think the function should either be more generic to match the current function name or the function name should be restricted.

plugins/module_utils/manage_vpc_pair/actions.py
```diff
@@ -34,11 +35,114 @@
...
+
+
+def _resolve_fabric_type(nd_v2: NDModuleV2, fabric_name: str, module: Any) -> str:
+    """
+    Resolve and cache the canonical fabric type for the current run.
+
+    Nexus Dashboard returns the fabric type in ``management.type``. The lookup
+    is fail-open: when the type cannot be determined, return an empty string
```
akinross: I think we should never be able to not not determine a fabric type, this should always be the case and else raise an error. The caller can then handle the error accordingly.

plugins/module_utils/manage_vpc_pair/actions.py
```diff
@@ -34,11 +35,114 @@
...
+
+    details = proposed_config.get("vpc_pair_details")
+    if details is None:
+        details = proposed_config.get(VpcFieldNames.VPC_PAIR_DETAILS)
+    if not isinstance(details, dict) or not details:
+        return None
+    return details
+
+
+def _validate_vpc_pair_details_fabric_support(
```
akinross: what is the reason for doing this validation? does the api not error on wrong configuration? 

I am not sure we want to force our selves into these type of validations because whenever vpc is allowed on new fabric types we would need to adjust this logic.

akinross (changes requested):

plugins/module_utils/manage_vpc_pair/actions.py
sivakasi-cisco: Yes, for /switches cache class is optional.

plugins/module_utils/manage_vpc_pair/actions.py
sivakasi-cisco: The API does reject this configuration, but for vxlanIbgp and vxlanEbgp it can return an unclear HTTP 500 instead of explaining that vpcPairDetails is unsupported. The UI also sends only switchId, peerSwitchId, and useVirtualPeerLink for these fabric types.  
This validation fails early with a clear Ansible error and avoids sending a known-invalid request. The check only blocks the two currently unsupported fabric types, so new fabric types remain allowed by default. If these two types support vpcPairDetails in the future, we will update the blocked list.

plugins/module_utils/manage_vpc_pair/actions.py
sivakasi-cisco: As NDModuleV2 does not replace or bypass RestSend, but a wrapper around it, when we call nd_v2.request(path, GET), it internally sets the path and verb on RestSend and calls rest_send.commit(). It also converts failed API responses into NDModuleError.

Also, as manage_vpc_pair implementation consistently uses NDModuleV2.request(), I had used it here.

allenrobel: @sivakasi-cisco My understanding is that nd_v2.py is deprecated and no new modules should use it.  We should add a comment in nd_v2.py to this effect, but I remember this being the concensus during a meeting with Matt and Mike.  It you're going to use it, at least check with them, OK?

mikewiebe: vPC pair is one of the older style modules so I would have expected it for this module that has already been reviewed/appproved and merged.  This PR is just an update/fix to the existing module.  Agree that no new modules should use it but this one can continue until it gets refactored at a later date but don't want to risk that now

plugins/module_utils/manage_vpc_pair/actions.py
sivakasi-cisco: Now /switches is not used. Corrected the documentation

plugins/module_utils/manage_vpc_pair/actions.py
sivakasi-cisco: Thank you for the comments @akinross  @allenrobel . Thank you for adding detailed clarity @mikewiebe.

plugins/module_utils/manage_vpc_pair/actions.py
sivakasi-cisco: Broad exception handling added with NDModuleError, TypeError, ValueError and runtimeError

plugins/module_utils/manage_vpc_pair/actions.py
sivakasi-cisco: Addressed. I renamed the helper from _first_line to _first_exception_line and restricted its argument to exception. In the latest code, if an invalid value is passed, say "None", it now raises TypeError instead of returning "None".

plugins/module_utils/manage_vpc_pair/actions.py
sivakasi-cisco: Addressed. The fabric-type lookup is now updated not be fail-open and does not return an empty string.

When non-empty vpc_pair_details requires validation, the resolver must obtain a non-empty management.type; otherwise, it raises a contextual VpcPairResourceError.

Controller request failures, malformed responses, and missing or empty management.type are handled explicitly. Only successfully resolved fabric types are cached.

plugins/module_utils/manage_vpc_pair/actions.py
sivakasi-cisco: In the original implementation, _fabric_type had been inside module.params.

I removed that parameter-based caching. Fabric type is now explicit per-run state on the vPC resource as self.fabric_type: str | None. Once the validation path gets resolved, it assigns it only after a successful validation and subsequently reuses it

plugins/module_utils/manage_vpc_pair/actions.py
sivakasi-cisco: Addressed. Previously, actions.py manually constructed the path which now uses details_path = VpcPairEndpoints.fabric_details(fabric_name) which uses the collection’s existing EpManageFabricsGet endpoint model.

akinross (approved):

mikewiebe:
### High: `config_actions.type: switch` still uses fabric-scoped deployment

**Issue**

The supported VXLAN workflow is now clear: iBGP/eBGP fabrics correctly reject `vpc_pair_details`, and with `use_virtual_peer_link: false` ND automatically discovers the physical interfaces between the pair. That does not make `config_actions.type` advisory, however. The current implementation sends both `type: switch` and `type: global` through the fabric-level `actions/configSave` and `actions/deploy` flow instead of selecting `switchActions/deploy` for switch scope.

**Evidence**

- [`deploy.py` lines 204-257](https://github.com/CiscoDevNet/ansible-nd/blob/b664ba54475222a1750bde734825e59d28284f98/plugins/module_utils/manage_vpc_pair/deploy.py#L204-L257) always calls `FabricUtils.save_config()` and `FabricUtils.deploy_config()` after reading `config_actions.type`; there is no branch to `switchActions/deploy`.
- [The current PR description](https://github.com/CiscoDevNet/ansible-nd/pull/338) still says the default switch-scoped path uses `switchActions/deploy` with the affected switch serials, but `deploy.py` is not changed in the current eight-file diff because commit `c9aacd9` reverted that implementation.
- [The author's scope-update comment](https://github.com/CiscoDevNet/ansible-nd/pull/338#issuecomment-4936284626) confirms that switch deployment was removed in favor of VXLAN input validation. Live testing confirms that validation and ND-managed physical-interface discovery are correct, while captured action paths still show fabric-level save/deploy for `type: switch`.

**Existing PR overlap**

Related to the author's scope-update comment above, which explains why the earlier switch-deploy workaround was removed. It does not address the separate public contract that `config_actions.type: switch` selects switch-level deployment, and it conflicts with the PR description that still claims this dispatch is implemented.

Related open issue [#368](https://github.com/CiscoDevNet/ansible-nd/issues/368) tracks the collection-wide `config_actions` contract. The PR-specific delta is actionable here because this PR advertises switch-scoped deployment as its fix while the current head does not implement it.

**Impact**

Users cannot select the documented deployment scope: `switch` and `global` execute the same fabric operation. This makes the module contract misleading and leaves switch-targeted runs with fabric-wide save/deploy behavior even though ND correctly handles the physical VXLAN pair itself through interface auto-discovery.

**Suggested fix**

Keep the new VXLAN validation and auto-discovery workflow unchanged. Route `config_actions.type: switch` to `POST .../switchActions/deploy` with the affected peer switch IDs, retain `actions/configSave` plus `actions/deploy` for `type: global`, and add unit/integration assertions for both endpoint and payload selections. Update the PR description if switch-scoped dispatch is intentionally deferred to #368.

tests/integration/targets/nd_vpc_pair/tasks/nd_vpc_pair_merge.yaml
```diff
@@ -1816,10 +1816,47 @@
...
+        peer2_switch_id: "{{ test_switch2 }}"
+        use_virtual_peer_link: false
+        vpc_pair_details:
+          type: default
+          domain_id: 1
+          keep_alive_vrf: management
+          switch_keep_alive_local_ip: "192.0.2.11"
+          peer_switch_keep_alive_local_ip: "192.0.2.12"
+  register: tc23b_result
+  failed_when: false
```
mikewiebe: ### High: The new VXLAN integration regression case cannot pass

**Issue**

The new negative regression test has two independent control-flow defects. On iBGP/eBGP VXLAN fabrics, the suite first runs TC1 with the newly prohibited `vpc_pair_details`, so execution stops before TC23b is reached. If TC23b is reached separately, `failed_when: false` overwrites the expected module failure before the next task asserts that the registered result has `failed == true`.

**Evidence**

- [`nd_vpc_pair_merge.yaml` line 1844](https://github.com/CiscoDevNet/ansible-nd/pull/338/files#diff-a2041261b6916a3ca38168d34a2a2a23c787c3693a9a18c40c06f84d512c5b5bR1844) sets `failed_when: false`, while the immediately following TC23b assertion requires `tc23b_result.failed == true` at line 1852. Ansible applies `failed_when` to the registered task result, so a correctly rejected request is recorded with `failed: false`; an incorrectly accepted request is also `failed: false`.
- [`nd_vpc_pair_merge.yaml` lines 20-32](https://github.com/CiscoDevNet/ansible-nd/blob/b664ba54475222a1750bde734825e59d28284f98/tests/integration/targets/nd_vpc_pair/tasks/nd_vpc_pair_merge.yaml#L20-L32) build TC1's full configuration with a non-empty `vpc_pair_details` block.
- [`nd_vpc_pair_merge.yaml` lines 89-100](https://github.com/CiscoDevNet/ansible-nd/blob/b664ba54475222a1750bde734825e59d28284f98/tests/integration/targets/nd_vpc_pair/tasks/nd_vpc_pair_merge.yaml#L89-L100) execute that full configuration unconditionally. On `vxlanIbgp` or `vxlanEbgp`, the PR's validation correctly rejects TC1 and stops the play before the new TC23b case at line 1823. Live testing reproduced that early rejection on both VXLAN fabric types.

**Existing PR overlap**

No matching open issue found.

**Impact**

The integration target cannot verify the behavior introduced by this PR on either supported VXLAN topology: the normal run stops before the regression case, and the regression case cannot pass even when reached. Because the CI Integration Tests job is currently skipped, the passing unit and sanity jobs do not expose this failure.

**Suggested fix**

Guard TC1 and the other detail-bearing positive cases so they run only on fabric types that support `vpc_pair_details`; use the ID-only configuration for iBGP/eBGP VXLAN. In TC23b, replace `failed_when: false` with `ignore_errors: true` or a `block`/`rescue` pattern that allows execution to continue while preserving `tc23b_result.failed == true`, then retain the assertions on the failure message.

tests/integration/targets/nd_vpc_pair/tasks/base_tasks.yaml
```diff
@@ -25,6 +25,17 @@
...
+    test_fabric_type_normalized: "{{ (fabric_type | default('LANClassic') | string | lower) }}"
+    vpc_pair_details_disallowed_fabric: >-
+      {{
+        (fabric_type | default('LANClassic') | string | lower)
+        in ['vxlanfabric', 'vxlanibgp', 'vxlanebgp']
+      }}
+    vpc_pair_details_expected_supported: >-
+      {{
+        (fabric_type | default('LANClassic') | string | lower)
+        in ['lanclassic', 'external', 'isn']
```
mikewiebe: ### Medium: Canonical `externalConnectivity` input skips extended External detail tests

**Issue**

The new positive-test selector lowercases `fabric_type` but recognizes only `external`, while ND and the collection use the canonical value `externalConnectivity`. With that canonical input, the guarded TC7b preservation checks and TC23 extended-detail checks are skipped. Basic TC7/TC8 detail tests are unguarded and still run, so the gap is specifically the preservation and extended payload coverage rather than all External positive coverage.

**Evidence**

- [`base_tasks.yaml` line 37](https://github.com/CiscoDevNet/ansible-nd/pull/338/files#diff-1e2e046f781e5211e8a26a2b5e8cacdd943d923f0e1ebffeee26a0f9528c5bcdR37) evaluates the lowercased input against `['lanclassic', 'external', 'isn']`. Canonical `externalConnectivity` becomes `externalconnectivity`, so `vpc_pair_details_expected_supported` is false.
- [`FabricTypeEnum.EXTERNAL_CONNECTIVITY`](https://github.com/CiscoDevNet/ansible-nd/blob/b664ba54475222a1750bde734825e59d28284f98/plugins/module_utils/models/manage_fabric/enums.py#L38) and the External fabric module both define the management type as `externalConnectivity`.
- [`nd_vpc_pair_merge.yaml` lines 611-674](https://github.com/CiscoDevNet/ansible-nd/blob/b664ba54475222a1750bde734825e59d28284f98/tests/integration/targets/nd_vpc_pair/tasks/nd_vpc_pair_merge.yaml#L611-L674) guard TC7b's omitted-detail preservation workflow with this false value, and [lines 1763-1821](https://github.com/CiscoDevNet/ansible-nd/blob/b664ba54475222a1750bde734825e59d28284f98/tests/integration/targets/nd_vpc_pair/tasks/nd_vpc_pair_merge.yaml#L1763-L1821) apply the same guard to TC23's extended payload, direct API query, validation, and assertions.
- TC7 and TC8 remain unguarded, so they still provide basic default/custom detail coverage. If an external inventory deliberately passes the friendly alias `External`, it normalizes to `external` and the guarded tests run; however, that inventory contract is not defined in the repository.

**Existing PR overlap**

Related to [`@akinross`'s fabric-type discussion](https://github.com/CiscoDevNet/ansible-nd/pull/338#discussion_r3585151707), which records `externalConnectivity` as the value returned in `management.type`. That thread concerns runtime type resolution and does not identify the integration selector or skipped tests; this finding applies its evidence to the test matrix.

No matching open issue found.

**Impact**

An External integration run using the controller/collection's canonical type can complete without proving that omitted details are preserved or that the extended physical-detail payload is accepted and returned correctly. The skipped tasks can therefore leave regressions in the PR's supported External path undetected.

**Suggested fix**

Normalize documented aliases centrally or include `externalconnectivity` alongside `external` in the supported set, then add an assertion that both accepted inputs enable TC7b and TC23. Prefer using the collection's fabric-type enum values in integration inventory so the tests exercise the same identifiers as runtime code.

plugins/module_utils/manage_vpc_pair/actions.py
```diff
@@ -34,11 +34,137 @@
...
+    proposed_details = _get_proposed_vpc_pair_details(nrm.proposed_config)
+    if proposed_details is None:
+        return
+
+    fabric_type = nrm.fabric_type
+    if fabric_type is None:
+        fabric_type = _resolve_fabric_type(nd_v2, fabric_name)
+        nrm.fabric_type = fabric_type
+
+    if fabric_type in _BLOCKED_FABRIC_TYPES_FOR_VPC_PAIR_DETAILS:
```
mikewiebe, line 186: ### Medium: The new fabric-specific input restriction is absent from the public contract

**Issue**

The PR adds a fabric-dependent validation rule to the module's public input contract: any non-empty `vpc_pair_details` is rejected for `vxlanIbgp` and `vxlanEbgp`, while the option remains valid on supported fabrics such as External Connectivity. `DOCUMENTATION` still describes the option only as optional template data, and the examples do not identify their fabric type or demonstrate either side of this contract.

**Evidence**

- [`actions.py` lines 156-161](https://github.com/CiscoDevNet/ansible-nd/pull/338/files#diff-a304622ca452c2276503bf6c2b8ef634d8f2033419faf4bb30403894024e1457R156) introduce the runtime rejection for iBGP/eBGP VXLAN and direct users to submit only peer IDs plus `use_virtual_peer_link`.
- [`nd_manage_vpc_pair.py` lines 115-118](https://github.com/CiscoDevNet/ansible-nd/blob/b664ba54475222a1750bde734825e59d28284f98/plugins/modules/nd_manage_vpc_pair.py#L115-L118) describe `vpc_pair_details` only as "Optional vPC pair template details" without listing blocked or supported fabric types.
- [`nd_manage_vpc_pair.py` lines 131-205](https://github.com/CiscoDevNet/ansible-nd/blob/b664ba54475222a1750bde734825e59d28284f98/plugins/modules/nd_manage_vpc_pair.py#L131-L205) provide fabric-neutral examples and no example containing `vpc_pair_details`. Users therefore cannot tell whether the examples target VXLAN, External Connectivity, or another fabric type.
- Live iBGP/eBGP testing confirmed the supported physical workflow: omit `vpc_pair_details`, set `use_virtual_peer_link: false`, and ND automatically discovers the directly connected peer interfaces. Live External testing confirmed that explicit physical details remain accepted and deployable there.

**Existing PR overlap**

Related to [the author's scope-update comment](https://github.com/CiscoDevNet/ansible-nd/pull/338#issuecomment-4936284626) and [the validation-rationale thread](https://github.com/CiscoDevNet/ansible-nd/pull/338#discussion_r3593038331), which explain the controller/UI behavior to reviewers. Neither updates the module's user-facing `DOCUMENTATION` or `EXAMPLES`, so this finding is specifically about the public Ansible contract.

No matching open issue found.

**Impact**

Users following the module documentation can supply a documented option that works on one fabric type but fails local validation on another, with no way to predict that result before execution. Declarative automation also lacks a documented rule for choosing between the ID-only VXLAN payload and explicit External physical details.

**Suggested fix**

Update the `vpc_pair_details` description or module notes to name `vxlanIbgp` and `vxlanEbgp` as unsupported and identify the supported fabric types. Add one iBGP/eBGP physical-pair example that omits details and uses `use_virtual_peer_link: false`, plus one External Connectivity example with explicit `vpc_pair_details`. State that ND discovers the direct peer interfaces for the VXLAN form, and keep the examples aligned with the validation tests.

plugins/module_utils/manage_vpc_pair/actions.py
```diff
@@ -34,11 +34,137 @@
...
+    fabric_name: str,
+) -> None:
+    """
+    Block vpc_pair_details on fabrics where it is not supported.
+
+    Currently blocked for VXLAN iBGP/eBGP fabrics. Allowed for other fabric
+    types (for example External/ISN-style deployments) where the controller
+    supports the additional settings.
+    """
+    proposed_details = _get_proposed_vpc_pair_details(nrm.proposed_config)
```
mikewiebe: **Medium: Fabric-support validation examines reconciled state instead of raw user intent**

The validator reads `nrm.proposed_config`, but it runs only inside `custom_vpc_create()` and `custom_vpc_update()`. By that point, the state machine has either returned early for `no_diff` or replaced the raw playbook intent with a model reconciled against controller state.

**Case 1: Explicitly invalid input bypasses validation when idempotent**

Assume an iBGP/eBGP VXLAN pair already has `vpcPairDetails: {domainId: 10}` in the controller response. If the user explicitly supplies the same unsupported `vpc_pair_details`, `resources.py:351-359` classifies the proposed values as a subset of the existing values. The `no_diff` branch at `resources.py:361-367` records `no_change` and returns before either action handler invokes this validator.

The same prohibited field is therefore rejected when it changes state but accepted when it already matches state.

**Case 2: Omitted input is mistaken for explicit input after reconciliation**

Assume the same controller state, but the user correctly omits `vpc_pair_details` and changes only `use_virtual_peer_link`. For `state: merged`, `resources.py:369-378` merges the existing model with the proposed item. `VpcPairModel.merge()` preserves existing fields omitted by the user, and `to_payload()` serializes the inherited `vpcPairDetails` into `nrm.proposed_config`.

This validator then sees the reconciled field and can reject the request even though the user never supplied it. If validation were merely removed or moved without addressing payload construction, the inherited field could instead be sent back to ND on a blocked fabric type.

An exact-head state-machine reproduction confirmed both control paths: identical explicit details produced a `no_change` log without invoking the action handler, while a details-free merged update produced a final payload containing inherited `vpcPairDetails`. This is modeled state-machine evidence; it does not assert that every live VXLAN controller response contains that field.

**Suggested fix**

Validate each normalized `proposed_item` before diffing and before the `no_diff` early return. Use `model_fields_set` or `model_dump(exclude_unset=True)` so validation considers only fields explicitly supplied by the user, and resolve/cache the fabric type for this preflight validation.

Separately, remove inherited `vpcPairDetails` when constructing create/update payloads for blocked iBGP/eBGP VXLAN fabrics. External fabrics should continue accepting and preserving these details. The generic `merge()` behavior should remain unchanged because preserving omitted existing values is appropriate for merged state and supported External workflows.

Please add tests proving that:

1. Explicit unsupported details are rejected even when they match existing state.
2. A merged update omitting details succeeds when controller state contains them.
3. That update does not send `vpcPairDetails`.
4. External fabrics continue accepting and preserving details.
5. The same rules apply in check mode.

sivakasi-cisco:
> ### High: `config_actions.type: switch` still uses fabric-scoped deployment
> **Issue**
> 
> The supported VXLAN workflow is now clear: iBGP/eBGP fabrics correctly reject `vpc_pair_details`, and with `use_virtual_peer_link: false` ND automatically discovers the physical interfaces between the pair. That does not make `config_actions.type` advisory, however. The current implementation sends both `type: switch` and `type: global` through the fabric-level `actions/configSave` and `actions/deploy` flow instead of selecting `switchActions/deploy` for switch scope.
> 
> **Evidence**
> 
> * [`deploy.py` lines 204-257](https://github.com/CiscoDevNet/ansible-nd/blob/b664ba54475222a1750bde734825e59d28284f98/plugins/module_utils/manage_vpc_pair/deploy.py#L204-L257) always calls `FabricUtils.save_config()` and `FabricUtils.deploy_config()` after reading `config_actions.type`; there is no branch to `switchActions/deploy`.
> * [The current PR description](https://github.com/CiscoDevNet/ansible-nd/pull/338) still says the default switch-scoped path uses `switchActions/deploy` with the affected switch serials, but `deploy.py` is not changed in the current eight-file diff because commit `c9aacd9` reverted that implementation.
> * [The author's scope-update comment](https://github.com/CiscoDevNet/ansible-nd/pull/338#issuecomment-4936284626) confirms that switch deployment was removed in favor of VXLAN input validation. Live testing confirms that validation and ND-managed physical-interface discovery are correct, while captured action paths still show fabric-level save/deploy for `type: switch`.
> 
> **Existing PR overlap**
> 
> Related to the author's scope-update comment above, which explains why the earlier switch-deploy workaround was removed. It does not address the separate public contract that `config_actions.type: switch` selects switch-level deployment, and it conflicts with the PR description that still claims this dispatch is implemented.
> 
> **Existing open issue overlap**
> 
> Related open issue [#368](https://github.com/CiscoDevNet/ansible-nd/issues/368) tracks the collection-wide `config_actions` contract. The PR-specific delta is actionable here because this PR advertises switch-scoped deployment as its fix while the current head does not implement it.
> 
> **Impact**
> 
> Users cannot select the documented deployment scope: `switch` and `global` execute the same fabric operation. This makes the module contract misleading and leaves switch-targeted runs with fabric-wide save/deploy behavior even though ND correctly handles the physical VXLAN pair itself through interface auto-discovery.
> 
> **Suggested fix**
> 
> Keep the new VXLAN validation and auto-discovery workflow unchanged. Route `config_actions.type: switch` to `POST .../switchActions/deploy` with the affected peer switch IDs, retain `actions/configSave` plus `actions/deploy` for `type: global`, and add unit/integration assertions for both endpoint and payload selections. Update the PR description if switch-scoped dispatch is intentionally deferred to #368.

Changed failed_when: false → ignore_errors: true so the failure is actually captured; gated the fabric-detail test cases TC1, 7, 8, 12, 13 and  23b so they only run on fabrics that support them.

For example, on an eBGP VXLAN fabric, when it was senidng the whole vpc_pair_details which is rejected by module and gets aborted. Now, it is skipped there and asserts with the rejection message.

plugins/module_utils/manage_vpc_pair/actions.py
sivakasi-cisco: Added a preflight reject which runs before diffing. So that the idempotent and the check_mode cases are covered too. This would strip the inherited details on merge for blocked fabrics.

sivakasi-cisco:
> ### High: `config_actions.type: switch` still uses fabric-scoped deployment
> **Issue**
> 
> The supported VXLAN workflow is now clear: iBGP/eBGP fabrics correctly reject `vpc_pair_details`, and with `use_virtual_peer_link: false` ND automatically discovers the physical interfaces between the pair. That does not make `config_actions.type` advisory, however. The current implementation sends both `type: switch` and `type: global` through the fabric-level `actions/configSave` and `actions/deploy` flow instead of selecting `switchActions/deploy` for switch scope.
> 
> **Evidence**
> 
> * [`deploy.py` lines 204-257](https://github.com/CiscoDevNet/ansible-nd/blob/b664ba54475222a1750bde734825e59d28284f98/plugins/module_utils/manage_vpc_pair/deploy.py#L204-L257) always calls `FabricUtils.save_config()` and `FabricUtils.deploy_config()` after reading `config_actions.type`; there is no branch to `switchActions/deploy`.
> * [The current PR description](https://github.com/CiscoDevNet/ansible-nd/pull/338) still says the default switch-scoped path uses `switchActions/deploy` with the affected switch serials, but `deploy.py` is not changed in the current eight-file diff because commit `c9aacd9` reverted that implementation.
> * [The author's scope-update comment](https://github.com/CiscoDevNet/ansible-nd/pull/338#issuecomment-4936284626) confirms that switch deployment was removed in favor of VXLAN input validation. Live testing confirms that validation and ND-managed physical-interface discovery are correct, while captured action paths still show fabric-level save/deploy for `type: switch`.
> 
> **Existing PR overlap**
> 
> Related to the author's scope-update comment above, which explains why the earlier switch-deploy workaround was removed. It does not address the separate public contract that `config_actions.type: switch` selects switch-level deployment, and it conflicts with the PR description that still claims this dispatch is implemented.
> 
> **Existing open issue overlap**
> 
> Related open issue [#368](https://github.com/CiscoDevNet/ansible-nd/issues/368) tracks the collection-wide `config_actions` contract. The PR-specific delta is actionable here because this PR advertises switch-scoped deployment as its fix while the current head does not implement it.
> 
> **Impact**
> 
> Users cannot select the documented deployment scope: `switch` and `global` execute the same fabric operation. This makes the module contract misleading and leaves switch-targeted runs with fabric-wide save/deploy behavior even though ND correctly handles the physical VXLAN pair itself through interface auto-discovery.
> 
> **Suggested fix**
> 
> Keep the new VXLAN validation and auto-discovery workflow unchanged. Route `config_actions.type: switch` to `POST .../switchActions/deploy` with the affected peer switch IDs, retain `actions/configSave` plus `actions/deploy` for `type: global`, and add unit/integration assertions for both endpoint and payload selections. Update the PR description if switch-scoped dispatch is intentionally deferred to #368.

Fixed them up to use respective deploy method.  Now, we look up the fabric and only deploy the switches that are actually out of sync via ../switchActions/deploy for type:switch. The global still does the whole-fabric deploy. Save stays on configSave for both. Added unit tests for the endpoint + payload of each scope. Not deferred to #368 anymore

tests/integration/targets/nd_vpc_pair/tasks/base_tasks.yaml
sivakasi-cisco: I had missed the actual type name. Added 'externalConnectivity' to the supported list so External fabrics would be recognized properly,  instead of being treated as unsupported.

tests/integration/targets/nd_vpc_pair/tasks/nd_vpc_pair_merge.yaml
sivakasi-cisco: This is a valid concern, which is left out. TC23b now uses ignore_errors: true instead of failed_when: false so the module failure is preserved but the assert actually checks 'failed == true'.

I have gated TC1 and the related cases to only run on fabrics that support vpc_pair_details, with a separate ID-only TC1 for iBGP/eBGP VXLAN so the play doesn't abort before TC23b. This would run clean on both now.

plugins/module_utils/manage_vpc_pair/actions.py
sivakasi-cisco: Yes. This fabric-specific restriction needs to be part of the documentation.

- Updated the vpc_pair_details documentation to state that it is unsupported on vxlanIbgp and vxlanEbgp.
- Documented that these VXLAN fabrics should use the peer switch IDs with use_virtual_peer_link: false, allowing Nexus Dashboard to discover the physical peer-link interfaces.
- Identified LAN Classic, External Connectivity, and ISN as fabrics that accept explicit physical peer-link details.
- Added an iBGP/eBGP VXLAN physical-pair example without vpc_pair_details.
- Added an External Connectivity example using explicit vpc_pair_details.
- Updated the integration coverage to gate detail-bearing cases by fabric support and added a negative test confirming that vpc_pair_details is rejected on iBGP/eBGP VXLAN.
- The full integration suite was also validated successfully on a live vxlanIbgp fabric.

sivakasi-cisco:
Verified the changes with IT on the latest changes

plugins/module_utils/manage_vpc_pair/resources.py
```diff
@@ -341,6 +346,12 @@ def _manage_create_update_state(self, state: str, unwanted_keys: list[Any]) -> N
             try:
```
mikewiebe, line 346: Not blocking for this PR

**Issue**

The module accepts `config` as optional, then normalizes both omission and explicit
null to `[]`. Under `state: overridden`, that empty desired collection deletes every
existing vPC pair. The runtime therefore cannot distinguish a malformed task from the
documented, intentionally destructive `config: []` request.

**Evidence**

- [`resources.py` lines 346-357](https://github.com/CiscoDevNet/ansible-nd/pull/338/files#diff-a90aab5e2b82da84f8a6e19039c24f49ed7c17e87337e6249586bcab3204f27dR346)
  adds state-aware preflight inside the proposed-item loop. Once null config has
  become `[]`, that loop performs no validation before override reconciliation.
- `vpc_pair_model.py` lines 418-420 and 466-475 make `config` optional in both
  Pydantic and Ansible contracts.
- [`nd_manage_vpc_pair.py` lines 549-552](https://github.com/CiscoDevNet/ansible-nd/pull/338/files#diff-c35e6892ac4fa124c686ae2b66ddddfcbeb9e0472f102b3618e5acb158d52ef6R549)
  maps `None` to an empty runtime list.
- [`resources.py` lines 321-324 and 442-467](https://github.com/CiscoDevNet/ansible-nd/pull/338/files#diff-a90aab5e2b82da84f8a6e19039c24f49ed7c17e87337e6249586bcab3204f27dR321)
  performs override deletions for every existing pair absent from that list.
- Exact current-head probes confirmed that both omitted and explicit-null config
  validate and normalize to `[]`; explicit `config: []` is separately tested as
  intentional purge-all behavior.
- A practical failure is a site-scoped task that allows its generated inventory
  variable to be absent:

  ```yaml
  - name: Enforce the site's vPC pairs
    cisco.nd.nd_manage_vpc_pair:
      fabric_name: site-1
      state: overridden
      config: "{{ site_vpc_pairs | default(none) }}"
  ```

  If a missing inventory file, typo, or failed data-generation step leaves
  `site_vpc_pairs` undefined, Ansible passes `config: null`. The wrapper converts
  that to `[]`, and `overridden` interprets the empty list as an authoritative
  request to delete every existing vPC pair instead of rejecting unavailable
  desired-state data.

No matching open issue found.

**Impact**

A missing variable or YAML null can delete every vPC pair in the fabric when the task
uses `overridden`, including in automation that never intentionally supplied an empty
set.

**Suggested fix**

Reject omitted/null config for `merged`, `replaced`, and `overridden` before converting
the playbook model to runtime data. Preserve explicit `config: []` only where the
destructive contract is intentional and documented.

---

mikewiebe (approved):
LGTM, one additional non-blocking comment left

Issue #348, opened 2026-06-25
Lift duplicated interface orchestrator CRUD + policy-type serializer into shared bases


allenrobel:
## Summary

Deferred finding from the code review of the per-policy interface stacks. Several per-feature interface
orchestrators (and one policy-model serializer) re-implement logic byte-for-byte instead of inheriting it
from the shared `NDBaseInterfaceOrchestrator` / `NDNestedModel`. Maintenance cleanup, no functional bug.

## Duplication sites

**Orchestrator CRUD** — `PortChannelBaseOrchestrator` (`plugins/module_utils/orchestrators/port_channel_base.py`)
re-implements `create` / `update` / `delete` / `create_bulk` / `delete_bulk` / `query_one` essentially
character-for-character with `LoopbackInterfaceOrchestrator`
(`plugins/module_utils/orchestrators/loopback_interface.py`). The only port-channel-specific logic is the
`query_all` policy-type filter. Both already extend `NDBaseInterfaceOrchestrator`, which could own the shared
write/queue path.

**Orchestrator + model + helper** — `SubinterfaceUnmanagedInterfaceOrchestrator` / model duplicate the managed
sibling (`SubinterfaceManagedInterface*`) structure; the unmanaged variant differs only by the `policyType`
discriminator and the reduced policy body.

**Policy-type serializer** — `_strip_policy_type_in_config` (the wrap-mode `model_serializer` that pops
`policy_type` / `policyType` when `mode == "config"`) is near-verbatim across the per-policy policy models
(e.g. `port_channel_access_interface.py`, `ethernet_access_interface.py`). It should live in a shared
`NDNestedModel` mixin so the config-mode detection and the `AssertionError` invariant have a single source of truth.

## Why it matters

Any change to the shared interface write path (switchId injection, deploy/remove queuing, error-message format)
or to the config-mode serializer must be applied in every copy and will silently drift in the one that gets
missed. This is the exact class of drift that #336 calls out for `_request` / `_register_api_call`; this issue is
the interface-CRUD / policy-model analogue.

## Affected PRs

- #269 `nd_interface_port_channel_access` — `PortChannelBaseOrchestrator` CRUD dup + serializer dup
- #270 `nd_interface_port_channel_trunk_host` — same `PortChannelBaseOrchestrator` CRUD dup + ported `_strip_policy_type_in_config` serializer dup
- #297 `nd_interface_subinterface_unmanaged` — orchestrator / model / helper dup vs the managed sibling
[...]

PR #360, opened 2026-06-27
Enforce storm-control percentage/pps mutual exclusivity (#351)

### LINKED ISSUE #351: Storm-control percentage and pps levels are not mutually exclusive in interface policy models
## Summary

Deferred finding from the code review of #269 `nd_interface_port_channel_access`. For each storm-control traffic
class the model exposes both a percentage level and a packets-per-second level as independently-settable fields,
with no mutual-exclusion validator. Needs ND-behavior verification before deciding on a fix.

## Detail

`PortChannelAccessPolicyModel` (`plugins/module_utils/models/interfaces/port_channel_access_interface.py`) carries,
per traffic class:

- `storm_control_broadcast_level` (float %) and `storm_control_broadcast_level_pps` (int pps)
- `storm_control_multicast_level` / `storm_control_multicast_level_pps`
- `storm_control_unicast_level` / `storm_control_unicast_level_pps`

A playbook can set both the `%` and the `pps` variant for the same class; both are serialized into the policy
payload.

## Why it matters

Percentage and pps are expected to be mutually exclusive for a given storm-control class. If both are sent, ND
likely either rejects the create/update with an API error, or silently echoes back only one — producing a
perpetual diff (`changed: true` every run) when the dropped field is compared against the user's config.

## Affected PRs

- #269 `nd_interface_port_channel_access`

(The same paired `%` / `pps` fields will appear on the trunk-host port-channel and ethernet policy models, so the
resolution should be shared.)

## Proposed approach

1. Verify on an ND 4.2.1 lab what ND does when both `%` and `pps` are sent for the same class (reject? drop one?
   which one?). Record the finding in the bug-tracker vault if it is a wire deviation.
2. Add a `model_validator(mode="after")` rejecting both-set-for-the-same-class with a clear message, and document
   the mutual exclusivity in the module docs.

allenrobel:
## Related Issue(s)

Closes #351

## Proposed Changes

Storm-control percentage (`storm_control_<class>_level`) and pps (`storm_control_<class>_level_pps`) levels are mutually exclusive per traffic class on NX-OS, but nothing enforced this. Lab verification (ND **4.3.1.75** and **4.2.1.10**, see #351) showed ND accepts both, echoes both back on GET — so there is **no perpetual diff** — but at config generation it silently applies the percentage and **drops the pps**, losing the user's intent with no error.

- Add `StormControlMutexMixin` (`plugins/module_utils/models/interfaces/storm_control.py`): a shared `model_validator(mode="after")` that rejects setting both the percentage and pps for the same class, naming the offending class(es).
- Reparent all five interface policy models to the mixin: `ethernet_access`, `port_channel_access`, `port_channel_trunk_host`, `vpc_access`, `vpc_trunk_host`.
- Document the per-class mutual exclusivity in the four modules that lacked the note (`ethernet_access` already had it).
- Add parametrized unit tests across all five models.

**No `TODO(X.Y.Z)` workaround marker** — the wire round-trips faithfully, so there is no wire deviation to suppress; this is a client-side fail-fast UX guard. Behavior recorded in the bug-tracker vault (`storm-control-percent-pps-not-mutually-exclusive`).

## Test Notes

- New `tests/unit/module_utils/models/test_storm_control_mutex.py` — 45 cases (5 models × the conflict/single/none/cross-class/multi-class matrix), all passing via `ndpytest`.
- Full model + orchestrator suite green (960 passed) — reparenting changes no existing behavior.
- `validate-modules` sanity clean on the four doc-changed modules.
- `black`, `isort`, `pylint`, `mypy` run on changed files (remaining pylint `R0801` duplicate-code is pre-existing sibling-argspec duplication tracked by #348; mypy `import-not-found` is the baseline collection-path environment behavior, identical on untouched siblings).

4.2.1.10 (CCO) and 4.3.1.75 (development image)

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

### DIFF
--- a/plugins/module_utils/models/interfaces/storm_control.py
+++ b/plugins/module_utils/models/interfaces/storm_control.py
@@ -49,7 +49,10 @@ class StormControlMutexMixin(NDNestedModel):  # pylint: disable=too-few-public-m
 
     ## Raises
 
-    None
+    ### ValueError
+
+    - If both the percentage and pps level are set for the same storm-control class (raised by
+      `_reject_storm_control_level_and_pps` during model validation; surfaces as a Pydantic `ValidationError` when constructing an inheriting model).
     """
 
     @model_validator(mode="after")
--- a/plugins/module_utils/models/interfaces/storm_control.py
+++ b/plugins/module_utils/models/interfaces/storm_control.py
@@ -18,11 +18,17 @@ intent is silently lost with no error and no diff to warn them.
 Because the wire round-trips faithfully there is no wire deviation to work around (no `TODO(X.Y.Z)` marker); this is
 a client-side fail-fast guard so the user learns at validation time rather than discovering the dropped pps on the
 device. `StormControlMutexMixin` centralizes the check so all five interface policy models share one implementation.
+
+The guard is scoped to user/proposed input only. Since ND legitimately echoes both values on GET, enforcing the
+rule while parsing an ND response would make `query`/`gathered`/`diff` raise before the module could report or
+remediate a device already in that state. The check therefore fires on `from_config()` (and any direct/non-response
+construction) but is skipped when the validation context carries `mode="response"` (set by `from_response()`), so
+the write path fails fast while the read path stays permissive.
 """
 
 from __future__ import annotations
 
-from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import model_validator
+from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import ValidationInfo, model_validator
 from ansible_collections.cisco.nd.plugins.module_utils.models.nested import NDNestedModel
 
 # (traffic class label, percentage attribute, pps attribute) for each storm-control class. The attribute names are
@@ -47,28 +53,39 @@ class StormControlMutexMixin(NDNestedModel):  # pylint: disable=too-few-public-m
     fields. The check reads the field values by attribute name, so a model missing one of the pairs is handled
     gracefully (a `None` from `getattr` simply cannot conflict).
 
+    The check is skipped when the validation context carries `mode="response"` (set by `NDBaseModel.from_response`),
+    so parsing an ND echo that carries both values never raises; it fires on `from_config` and any other
+    (non-response) construction so user/proposed input still fails fast. See the module docstring for the rationale.
+
     ## Raises
 
     ### ValueError
 
-    - If both the percentage and pps level are set for the same storm-control class (raised by
-      `_reject_storm_control_level_and_pps` during model validation; surfaces as a Pydantic `ValidationError` when constructing an inheriting model).
+    - If both the percentage and pps level are set for the same storm-control class in a non-response context (raised
+      by `_reject_storm_control_level_and_pps` during model validation; surfaces as a Pydantic `ValidationError` when constructing an inheriting model).
     """
 
     @model_validator(mode="after")
-    def _reject_storm_control_level_and_pps(self) -> "StormControlMutexMixin":
+    def _reject_storm_control_level_and_pps(self, info: ValidationInfo) -> "StormControlMutexMixin":
         """
         # Summary
 
         Reject any storm-control traffic class that has both its percentage level and its pps level set, naming the
         offending class(es) in the error so the user knows exactly which one to drop.
 
+        The check is skipped when the validation context carries `mode="response"` — ND legitimately echoes both
+        values on GET (issue #351), so `from_response()` parsing must stay permissive or `query`/`gathered`/`diff`
+        would raise before the module could report or remediate an existing device already in that state. Every other
+        path (`from_config`, direct construction, any non-response context) enforces the rule so writes fail fast.
+
         ## Raises
 
         ### ValueError
 
-        - If both the percentage and pps level are set for the same storm-control class (broadcast, multicast, or unicast).
+        - If both the percentage and pps level are set for the same storm-control class (broadcast, multicast, or unicast) outside a response context.
         """
+        if info.context and info.context.get("mode") == "response":
+            return self
         conflicts = [
             traffic_class
             for traffic_class, pct_attr, pps_attr in _STORM_CONTROL_CLASSES
--- a/plugins/module_utils/models/interfaces/storm_control.py
+++ b/plugins/module_utils/models/interfaces/storm_control.py
@@ -1,0 +1,81 @@
+# Copyright: (c) 2026, Allen Robel (@allenrobel)
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+"""
+Shared storm-control mutual-exclusion validator for interface policy models.
+
+Every physical/port-channel/vPC host interface policy exposes, per traffic class (broadcast, multicast, unicast),
+both a percentage level (`storm_control_*_level`) and a packets-per-second level (`storm_control_*_level_pps`).
+These are mutually exclusive on NX-OS: `storm-control <class> level <pct>` and `storm-control <class> level pps <n>`
+are the same single setting, so only one may apply per class.
+
+ND does not enforce this at the API layer. Lab-verified on ND 4.3.1.75 and 4.2.1.10 (issue #351): a create/update
+that sets both for one class is accepted (HTTP 204/207) and ND echoes both values back on GET (so there is no
+perpetual diff), but at config generation ND silently applies the percentage and drops the pps. The user's pps
+intent is silently lost with no error and no diff to warn them.
+
+Because the wire round-trips faithfully there is no wire deviation to work around (no `TODO(X.Y.Z)` marker); this is
+a client-side fail-fast guard so the user learns at validation time rather than discovering the dropped pps on the
+device. `StormControlMutexMixin` centralizes the check so all five interface policy models share one implementation.
+"""
+
+from __future__ import annotations
+
+from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import model_validator
+from ansible_collections.cisco.nd.plugins.module_utils.models.nested import NDNestedModel
+
+# (traffic class label, percentage attribute, pps attribute) for each storm-control class. The attribute names are
+# identical across all five interface policy models, so the mixin reads them positionally via getattr.
+_STORM_CONTROL_CLASSES: tuple[tuple[str, str, str], ...] = (
+    ("broadcast", "storm_control_broadcast_level", "storm_control_broadcast_level_pps"),
+    ("multicast", "storm_control_multicast_level", "storm_control_multicast_level_pps"),
+    ("unicast", "storm_control_unicast_level", "storm_control_unicast_level_pps"),
+)
+
+
+class StormControlMutexMixin(NDNestedModel):  # pylint: disable=too-few-public-methods
+    """
+    # Summary
+
+    Mixin for interface policy models that carry the paired `storm_control_<class>_level` (percentage) and
+    `storm_control_<class>_level_pps` (packets-per-second) fields. Rejects a config that sets both for the same
+    traffic class, since the two are mutually exclusive on NX-OS and ND silently keeps the percentage while dropping
+    the pps at config generation (issue #351, lab-verified on ND 4.3.1.75 and 4.2.1.10).
+
+    Add it to a policy model's bases (replacing the direct `NDNestedModel` base); the model keeps defining its own
+    fields. The check reads the field values by attribute name, so a model missing one of the pairs is handled
+    gracefully (a `None` from `getattr` simply cannot conflict).
+
+    ## Raises
+
+    None
+    """
+
+    @model_validator(mode="after")
+    def _reject_storm_control_level_and_pps(self) -> "StormControlMutexMixin":
+        """
+        # Summary
+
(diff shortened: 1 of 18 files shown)


plugins/module_utils/models/interfaces/storm_control.py
```diff
@@ -0,0 +1,81 @@
...
+    traffic class, since the two are mutually exclusive on NX-OS and ND silently keeps the percentage while dropping
+    the pps at config generation (issue #351, lab-verified on ND 4.3.1.75 and 4.2.1.10).
+
+    Add it to a policy model's bases (replacing the direct `NDNestedModel` base); the model keeps defining its own
+    fields. The check reads the field values by attribute name, so a model missing one of the pairs is handled
+    gracefully (a `None` from `getattr` simply cannot conflict).
+
+    ## Raises
+
+    None
```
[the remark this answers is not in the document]
allenrobel: Fixed in "Document StormControlMutexMixin ValueError in class docstring (PR #360 review)" — the class-level `## Raises` now documents the `ValueError` (surfacing as a Pydantic `ValidationError` on construction of inheriting models).

akinross (approved):

plugins/module_utils/models/interfaces/storm_control.py
```diff
@@ -0,0 +1,84 @@
...
+    ## Raises
+
+    ### ValueError
+
+    - If both the percentage and pps level are set for the same storm-control class (raised by
+      `_reject_storm_control_level_and_pps` during model validation; surfaces as a Pydantic `ValidationError` when constructing an inheriting model).
+    """
+
+    @model_validator(mode="after")
+    def _reject_storm_control_level_and_pps(self) -> "StormControlMutexMixin":
```
gmicol: issue: Avoid rejecting existing ND storm-control echoes during response parsing - High

This validator runs during both user config parsing and `from_response()`. The PR notes that ND can already accept and echo both percentage and pps values on GET, so existing objects in that state will fail `query_all()`/gathered/diff initialization before the module can report or remediate them.

suggested fix:
Keep the write-side fail-fast behavior, but gate it to user config/proposed state or move it to the module/orchestrator input boundary. Add coverage showing `from_config()` rejects the conflict while `from_response()` accepts the ND echo.

gmicol (changes requested):

plugins/module_utils/models/interfaces/storm_control.py
allenrobel: Good catch — confirmed and fixed in "Gate storm-control mutex to config input, not ND responses (PR #360 review)".

You're right that the `model_validator(mode="after")` fired on every construction, including `from_response()`. Since ND accepts and echoes both the percentage and pps on GET (the behavior this PR itself documents), a device already in that state would have made `query`/`gathered`/`diff` raise before the module could report or remediate it.

Rather than move the check to the orchestrator boundary, I kept it centralized in the mixin and gated it on validation context, reusing the existing `context={"mode": ...}` idiom already on `to_payload`/`to_config`:

- `from_config()` sets `mode="config"`, `from_response()` sets `mode="response"` (in `NDBaseModel`).
- The mixin skips enforcement when `mode == "response"` and enforces on every other path — `from_config`, direct construction, any non-response context — so the write path still fails fast while the read path stays permissive. Pydantic propagates context to nested validation, so a top-level `from_response()` reaches the nested policy model where the mixin lives.

Added the coverage you asked for, parametrized across all five models × three classes: `from_config()` rejects the conflict; `from_response()` accepts the ND echo and asserts both values round-trip onto the model. Full suite green (1464 passed).

gmicol (approved):
LGTM

tests/unit/module_utils/models/test_storm_control_mutex.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+# Every interface policy model that carries the paired storm-control fields.
+POLICY_MODELS = [
+    EthernetAccessPolicyModel,
+    PortChannelAccessPolicyModel,
+    PortChannelTrunkHostPolicyModel,
+    AccessVpcHostPolicyModel,
+    TrunkVpcHostPolicyModel,
+]
```
mikewiebe, line 69: ### High: Physical Ethernet `trunkHost` responses remain unparseable

**Issue**

This matrix omits `EthernetTrunkHostPolicyModel`, so the response-aware mutex is applied to only five of the six user-facing policy families. Physical Ethernet `trunkHost` still inherits directly from `NDNestedModel` and retains its older unconditional validator.

The required behavior differs by input source:

- Proposed user configuration containing both percentage and PPS should be rejected because NX-OS can apply only one setting.
- Existing ND state containing both must remain readable so the module can construct `before` and report, delete, or remediate it.

Issue [#351](https://github.com/CiscoDevNet/ansible-nd/issues/351) demonstrated this exact state with a physical `trunkHost`: ND 4.2.1.10 and 4.3.1.75 accepted and returned both values, while generated CLI used only percentage.

```yaml
switchIp: 192.0.2.10
interfaceName: Ethernet1/1
configData:
  networkOS:
    policy:
      policyType: trunkHost
      stormControlBroadcastLevel: 42.5
      stormControlBroadcastLevelPps: 12345
```

Even a valid task containing no storm-control fields fails against that existing record:

```yaml
- name: Delete the physical trunk-host interface configuration
  cisco.nd.nd_interface_ethernet_trunk_host:
    fabric_name: FABRIC1
    state: deleted
    config:
      - switch_ip: 192.0.2.10
        interface_name: Ethernet1/1
```

The state machine queries existing interfaces first. While constructing its `before` collection, `EthernetTrunkHostInterfaceModel.from_response()` reaches the nested policy's unconditional mutex and raises `ValidationError`, so deletion never runs. The same initialization failure blocks `merged`, `replaced`, and `overridden`.

I reproduced this on the exact PR head with the ND-shaped response above. The error is:

```text
storm_control_broadcast_level and storm_control_broadcast_level_pps are mutually exclusive; set only one.
```

**Impact**

The PR still cannot read or remediate the precise physical `trunkHost` state used to validate #351, despite closing that issue.

**Suggested fix**

Reparent `EthernetTrunkHostPolicyModel` to `StormControlMutexMixin`, remove or make its duplicate validator response-aware, and include it as the sixth model in this matrix. Add a top-level `EthernetTrunkHostInterfaceModel.from_response()` regression using this nested ND payload. Config input with both values should continue to fail; only controller-response parsing should be permissive.

tests/unit/module_utils/models/test_storm_control_mutex.py
```diff
@@ -0,0 +1,214 @@
...
+    ## Classes and Methods
+
+    - NDBaseModel.from_config()
+    - StormControlMutexMixin._reject_storm_control_level_and_pps()
+    """
+    with pytest.raises(ValidationError, match=label):
+        model_cls.from_config({pct_attr: PCT, pps_attr: PPS})
+
+
+@pytest.mark.parametrize("model_cls", POLICY_MODELS)
```
mikewiebe, line 213: ### High: `state: merged` cannot switch storm-control units or remediate dual-valued ND state

**Issue**

This response test stops after model construction, but the accepted model cannot complete the normal response-to-diff-to-merge lifecycle. Percentage and PPS are two representations of one NX-OS threshold, while the generic merge path treats them as independent fields.

**Failure mode 1: switching percentage to PPS raises**

Assume ND currently returns a valid percentage-only policy:

```yaml
stormControlBroadcastLevel: 50.0
```

The user submits valid PPS-only intent:

```yaml
- name: Switch broadcast storm control from percentage to PPS
    fabric_name: FABRIC1
    state: merged
    config:
      - switch_ip: 192.0.2.10
        interface_name: Ethernet1/1
        config_data:
          network_os:
            policy:
              storm_control_broadcast_level_pps: 1000
```

The existing response and proposed config are each valid, and diffing reports a change. `NDBaseModel.merge()` preserves the omitted percentage and assigns PPS through `setattr()`. Assignment validation then sees both values without the original response context and raises before any update is sent.

**Failure mode 2: matching PPS intent returns `no_diff`**

Assume ND returns the controller-accepted state demonstrated in [#351](https://github.com/CiscoDevNet/ansible-nd/issues/351):

```yaml
stormControlBroadcastLevel: 50.0
stormControlBroadcastLevelPps: 1000
```

The same playbook asks for PPS `1000`. The one-way subset comparison sees that PPS already matches and ignores the extra percentage field. The state machine reports `no_diff`, sends no update, and leaves ND's percentage-wins behavior in place. Requesting PPS `2000` detects a difference but then fails through the assignment-validation path above.

Generic `merged` behavior normally preserves omitted fields, but these two fields are not independent settings. Explicitly selecting one must clear the other:

| Existing | Desired | Expected result |
| --- | --- | --- |
| Percentage 50 | PPS 1000 | Remove percentage; set PPS 1000 |
| PPS 1000 | Percentage 50 | Remove PPS; set percentage 50 |
| Percentage 50 plus PPS 1000 | PPS 1000 | Remove percentage |
| Percentage 50 plus PPS 1000 | PPS 2000 | Remove percentage; update PPS to 2000 |

Proposed user configuration containing both values should continue to be rejected.

**Impact**

All five newly mixed-in interface families cannot perform a normal percentage-to-PPS or PPS-to-percentage `merged` transition. They also cannot reliably remediate the dual-valued controller state that motivated the response exemption. Check mode follows the same reconciliation path and either raises or incorrectly reports no change.

**Suggested fix**

Treat each percentage/PPS pair as one logical setting during diff and merge. When proposed config explicitly selects one variant, count an existing counterpart as a difference and clear it before validating the final model. Add parameterized top-level lifecycle tests for percent-to-PPS, PPS-to-percent, dual-response-to-single-value remediation with matching and changed values, unrelated merged updates, and check mode.

tests/unit/module_utils/models/test_storm_control_mutex.py
allenrobel: Confirmed and fixed in "Storm-control mutex: add trunkHost; treat pct/pps as one setting in merge (PR #360 review)".

`EthernetTrunkHostPolicyModel` is reparented onto `StormControlMutexMixin` and its older unconditional validator is removed, so the response-aware mutex now covers all six user-facing policy families. It's the sixth model in the test matrix (from_config rejects the conflict; from_response accepts the echo), plus a top-level `EthernetTrunkHostInterfaceModel.from_response()` regression using the exact nested ND payload from your comment — including the previously-blocked scenario where a storm-control-free `deleted` task failed during before-collection initialization.

One deliberate non-change: `InterfaceDefaultPolicyModel` stays off the mixin — it sets both variants to `0.0`/`0` by design (the internal normalize/delete template, not user-facing input), and the mixin would reject it.

Integration-verified against ND (SITE1): the full `nd_interface_ethernet_trunk_host` target passes, including a new dual-state block that seeds the exact #351 controller state via raw REST and confirms the module can read, diff, and remediate it (see "Integration: storm-control merge lifecycle + dual-state remediation (PR #360 review)").

tests/unit/module_utils/models/test_storm_control_mutex.py
allenrobel: Confirmed both failure modes and fixed in "Storm-control mutex: add trunkHost; treat pct/pps as one setting in merge (PR #360 review)". Each percentage/pps pair is now one logical setting during diff and merge:

- **Merge**: `StormControlMutexMixin.merge()` clears an existing counterpart before delegating to the generic merge whenever proposed config explicitly selects the other variant. Mutex enforcement is suspended for the duration of the in-place merge — assignment validation re-runs the validator without context on every setattr, which is also why an *unrelated* merged update against a dual-valued echo used to raise. The suspension can't mask a real conflict: proposed input was validated at construction, and the clearing rule guarantees the merge can't introduce a new dual state.
- **Diff**: a generic `NDBaseModel.merge_would_change()` hook recurses into nested models and is consulted by `get_diff(exclude_unset=True)`, so dual-echo-plus-matching-pps is classified `changed` rather than `no_diff`. All four rows of your table behave as specified.

Chasing this end-to-end surfaced a third bug your scenario was masking: the update executed on ND but the module still reported `changed: false`, because `NDOutput` derives `changed` from a one-way subset comparison that can't see field *removals* (the cleared percentage made `after` a strict subset of `before`). Fixed in "Report field removals as changed in before/after comparison (PR #360 review)".

Coverage: unit lifecycle tests parametrized across all six interface families (pct→pps, pps→pct, dual+matching, dual+changed, unrelated merge preserving the dual echo, per-class scope, idempotency guard) — check mode follows the same reconciliation path, so these cover it. Lab verification in "Integration: storm-control merge lifecycle + dual-state remediation (PR #360 review)": both unit switches prove on the wire that ND clears the omitted counterpart on PUT (post-switch idempotency holds), and the dual-state block seeds the exact #351 intent via raw REST, confirms ND echoes both, and asserts matching-pps remediation is a change that clears the percentage. Full target: ok=117 changed=40 failed=0 against ND SITE1.

One scope note: `replaced` state's subset comparison has a related blind spot for *any* field the proposed config omits — that's pre-existing replaced-state semantics on develop, beyond this PR's scope.

allenrobel: Follow-up: the scope note above is now tracked as #410 (replaced/overridden states cannot detect field removals) — general-case fix for the non-merged states, separate from this PR.

mikewiebe (approved):
LGTM

PR #362, opened 2026-06-29
Add policy-required-on-create preflight for interface modules (#350)

### LINKED ISSUE #350: Interface create accepts a policy-less config item (config_data optional, no create-time validator)
## Summary

Deferred finding from the code review of #269 `nd_interface_port_channel_access`. The model permits a create payload that carries `interface_name` / `interface_type` but no `config_data` (and therefore no policy), because `config_data` is `Optional` with no model-level validator requiring a policy on create. No functional regression today (matches the sibling interface models), but worth a deliberate decision.

## Detail

`PortChannelAccessInterfaceModel.config_data` (`plugins/module_utils/models/interfaces/port_channel_access_interface.py`) is
`PortChannelAccessConfigDataModel | None = None`, and the argspec marks only `switch_ip` / `interface_name` required. A config item supplying just `switch_ip` + `interface_name` is accepted, and `create()` emits a payload with `configData` absent/null.

For `state: deleted` an item with no `config_data` is correct and intended (only the identifier is needed), so a blanket "config_data required" rule would be wrong. The gap is specifically the **create** path under `merged` / `replaced` / `overridden`.

## Why it matters

On create, ND either rejects the policy-less POST with an opaque 4xx that surfaces as a generic
`Create failed ...` `RuntimeError`, or creates a port-channel with no member ports/policy the user did not intend. A state-aware validator would fail fast with a clear message.

## Affected PRs

- #269 `nd_interface_port_channel_access`

Note: the same `config_data: Optional` shape exists across the sibling interface models (loopback, ethernet, svi, subinterface), so any fix should consider a shared state-aware "policy required on create" check rather than a one-off.

## Proposed approach

Add a state-aware validation (in the orchestrator's create/preflight path, where `state` is available — not an unconditional model validator) that rejects a create item with no `config_data.network_os.policy`, with a message naming the offending `(switch_ip, interface_name)`. Confirm the exact ND failure mode for a policy-less create against a 4.2.1 lab before deciding between fail-fast vs. documenting.

allenrobel:
## Related Issue(s)

Closes #350

## Proposed Changes

Adds a state-aware fail-fast guard that rejects an interface **create** item carrying no policy, before any API round-trip (in `--check` mode too), naming the offending `(switch_ip, interface_name)` in module terms.

Background: interface argspecs require only `switch_ip` / `interface_name`, and `config_data` is `Optional` with no create-time validator, so a policy-less config item is accepted on create. A policy-less item is correct and intended for `state: deleted` (identifier only), so a blanket model validator would be wrong — the gap is specifically the **create** path under `merged` / `replaced` / `overridden`.

Lab verification (this work, ND 4.2.1.10, fabric SITE1 / leaf S1_LE1) confirmed ND rejects a policy-less create **per-interface** and never creates an empty interface — but the failure surfaces as `interface[0] '<name>'` inside a generic create error after a round-trip. So this is a nice-to-have fail-fast UX guard, not a correctness gap (down-scoped per the issue discussion).

Changes:

- `orchestrators/base.py` — generic no-op `preflight_create()` hook on `NDBaseOrchestrator`, mirroring the existing `preflight()` hook so non-interface orchestrators (local_user, fabric_*) are unaffected.
- `orchestrators/base_interface.py` — override that aggregates create items missing `config_data.network_os.policy` into a single `RuntimeError`, naming each offender and the fabric.
- `nd_state_machine.py` — calls `preflight_create(items_to_create)` after the create/update split and before the mutation loops. **Create-only by construction**: a `merged`/`replaced` update that legitimately omits a policy already present on the switch is never validated, and `state: deleted` does not route through the hook.

Mirrors the structure of the merged capability-preflight PR #275 (hook + override + state-machine wiring), but is purely local (no API call).

## Test Notes

Unit-only, matching PR #275 (no integration precedent for a preflight failure; ND's real behavior is already lab-verified):

- `test_base_interface.py` — 7 `preflight_create` behavior tests (offenders raise naming switch_ip/interface_name, `config_data` None vs `policy` None, multiple/mixed offenders, empty no-op, still raises in `--check`).
- `test_base_orchestrator.py` — base hook is a no-op for non-interface orchestrators.
- `test_nd_state_machine.py` — wiring: `preflight_create` runs with the create subset ahead of the check-mode gate; create-only scoping (an already-present item re-submitted without a policy is not validated); failure propagates before any mutation; not called for `deleted`.

Two integration tests should be added to each nd_interface_* module (issues will be opened).  1) Merge/Create with no config and 2) Merge/Create with empty config.
 
Verification (in the `nd-dev` container):

- Full unit suite: 1222 passed, 0 failures.
- `black` / `isort`: clean.
- `pylint`: message-id baseline identical to develop (zero new findings).
- `mypy`: changed-file errors identical to develop (only pre-existing environmental import-not-found).
- `ansible-test sanity --test pep8` on the changed files: clean.

4.2.1.10

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

### DIFF
--- a/plugins/module_utils/nd_state_machine.py
+++ b/plugins/module_utils/nd_state_machine.py
@@ -93,9 +93,19 @@ class NDStateMachine:
         Manage state according to desired configuration.
         """
         if self.state in ["merged", "replaced", "overridden"]:
+            proposed_items = list(self.proposed)
+
+            # Policy-required-on-create guard (issue #350) runs FIRST: it is local-only (self.existing is
+            # already in memory), so it fails before the API-backed capability preflight below and before
+            # _manage_create_update_state mutates self.existing, which NDOutput aliases as `after`. Create
+            # subset = proposed items not present in the existing inventory -- the same key-membership
+            # criterion get_diff_config uses to classify "new" (PR #362 review).
+            items_to_create = [item for item in proposed_items if self.existing.get(item.get_identifier_value()) is None]
+            self.model_orchestrator.preflight_create(items_to_create)
+
             # Capability preflight runs here -- before _manage_create_update_state, whose mutations are
             # skipped in check mode -- so dry-runs surface incapable switches (PR #275 / issue #273).
-            self.model_orchestrator.preflight(list(self.proposed))
+            self.model_orchestrator.preflight(proposed_items)
             self._manage_create_update_state()
 
             if self.state == "overridden":
@@ -176,10 +186,8 @@ class NDStateMachine:
                 if not self.ignore_errors:
                     raise NDStateMachineError(error_msg) from e
 
-        # Fail fast before any mutation if a create item carries no policy (issue #350). Create-only:
-        # a merged/replaced update may legitimately omit a policy that already exists on the switch, and
-        # state: deleted does not route through here, so only genuine new items are validated.
-        self.model_orchestrator.preflight_create(items_to_create)
+        # The policy-required-on-create guard (issue #350) runs in manage_state, before the capability
+        # preflight and before this method mutates self.existing (PR #362 review).
 
         # Execute updates (always individual)
         for item in items_to_update:
--- a/plugins/module_utils/nd_state_machine.py
+++ b/plugins/module_utils/nd_state_machine.py
@@ -101,11 +101,26 @@ class NDStateMachine:
             # subset = proposed items not present in the existing inventory -- the same key-membership
             # criterion get_diff_config uses to classify "new" (PR #362 review).
             items_to_create = [item for item in proposed_items if self.existing.get(item.get_identifier_value()) is None]
-            self.model_orchestrator.preflight_create(items_to_create)
 
-            # Capability preflight runs here -- before _manage_create_update_state, whose mutations are
-            # skipped in check mode -- so dry-runs surface incapable switches (PR #275 / issue #273).
-            self.model_orchestrator.preflight(proposed_items)
+            # Normalize preflight failures to NDStateMachineError (PR #362 review, gmicol). Both preflight
+            # hooks raise a bare RuntimeError (base_interface.preflight_create / the capability preflight),
+            # but nd_interface_svi and nd_interface_subinterface_managed/_unmanaged catch only
+            # NDStateMachineError at their entrypoint. Without this wrap a policy-less (or capability) preflight
+            # failure in those modules escapes as an unhandled RuntimeError, bypassing fail_json and losing the
+            # structured before/after/changed output the guard exists to provide. This wrap deliberately does
+            # NOT route through _execute_operation: both preflights must run in check mode too, and
+            # _execute_operation skips execution during a dry-run.
+            try:
+                self.model_orchestrator.preflight_create(items_to_create)
+
+                # Capability preflight runs here -- before _manage_create_update_state, whose mutations are
+                # skipped in check mode -- so dry-runs surface incapable switches (PR #275 / issue #273).
+                self.model_orchestrator.preflight(proposed_items)
+            except NDStateMachineError:
+                raise
+            except Exception as e:
+                raise NDStateMachineError(f"Preflight failed: {e}") from e
+
             self._manage_create_update_state()
 
             if self.state == "overridden":
--- a/plugins/module_utils/nd_state_machine.py
+++ b/plugins/module_utils/nd_state_machine.py
@@ -176,6 +176,11 @@ class NDStateMachine:
                 if not self.ignore_errors:
                     raise NDStateMachineError(error_msg) from e
 
+        # Fail fast before any mutation if a create item carries no policy (issue #350). Create-only:
+        # a merged/replaced update may legitimately omit a policy that already exists on the switch, and
+        # state: deleted does not route through here, so only genuine new items are validated.
+        self.model_orchestrator.preflight_create(items_to_create)
+
         # Execute updates (always individual)
         for item in items_to_update:
             self._execute_operation(self.model_orchestrator.update, item, error_msg_prefix=f"Failed to update {item.get_identifier_value()}")
--- a/plugins/module_utils/orchestrators/base.py
+++ b/plugins/module_utils/orchestrators/base.py
@@ -138,6 +138,22 @@ class NDBaseOrchestrator(BaseModel, Generic[ModelType]):
         """
         return
 
+    def preflight_create(self, model_instances: Sequence[ModelType]) -> None:
+        """
+        # Summary
+
+        Pre-mutation hook invoked by `NDStateMachine` with only the items diffed as create (`new`) — before any
+        create/update API call, so it fails fast in check mode too. Distinct from `preflight`, which receives the
+        full proposed set; this hook sees the create-only subset so subclasses can enforce create-specific
+        requirements without rejecting legitimate policy-less updates. Base implementation is a no-op; interface
+        orchestrators override to require a policy on create.
+
+        ## Raises
+
+        None
+        """
+        return
+
     # NOTE: Generic CRUD API operations for simple endpoints with single identifier (e.g. "api/v1/infra/aaa/LocalUsers/{loginID}")
     def create(self, model_instance: ModelType, **kwargs) -> ResponseType:
         try:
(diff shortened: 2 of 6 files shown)


akinross (approved):

plugins/module_utils/nd_state_machine.py
```diff
@@ -172,6 +172,11 @@ def _manage_create_update_state(self) -> None:
                 if not self.ignore_errors:
                     raise NDStateMachineError(error_msg) from e
 
+        # Fail fast before any mutation if a create item carries no policy (issue #350). Create-only:
+        # a merged/replaced update may legitimately omit a policy that already exists on the switch, and
+        # state: deleted does not route through here, so only genuine new items are validated.
+        self.model_orchestrator.preflight_create(items_to_create)
```
mikewiebe: ### Medium: policy preflight failures can report `changed=True` and an `after` state even though no mutation ran

**Issue**

When `preflight_create()` rejects a policy-less create, `NDStateMachine` has already mutated its `existing` collection during diff preparation. Since `NDOutput` holds the same collection reference as `after`, the module can return failure output showing `changed=True` and the proposed interface in `after`, even though no create/update API call ran.

**Evidence**

`NDStateMachine._manage_create_update_state()` merges or adds proposed items into `self.existing` before calling `preflight_create()`. `NDOutput.format()` then derives `changed=True` by comparing the referenced `before` and `after` collections.

Local reproduction from this PR head:

```text
RuntimeError Cannot create interface(s) without a policy
changed= True
before= []
after= [{'switch_ip': '192.168.12.151', 'interface_name': 'loopback10', 'interface_type': 'loopback'}]
calls= ['preflight', 'preflight_create']
```

The new failure test asserts no `create` or `create_bulk` call ran, but it does not assert the failed result output.

**Impact**

Automation consuming the failed task can see `changed=True` and an `after` snapshot for an interface that was never sent to ND. That is especially misleading for this new validation path, whose purpose is to fail before mutation.

**Suggested fix**

Avoid mutating `self.existing` before all pre-mutation validation passes. One approach is to compute `items_to_create` and `items_to_update` using temporary/final objects, run `preflight_create(items_to_create)`, and only then update `self.existing` / `after`. Add a unit test that asserts a policy-preflight failure returns unchanged output with no created item in `after`.

plugins/module_utils/nd_state_machine.py
```diff
@@ -172,6 +172,11 @@ def _manage_create_update_state(self) -> None:
                 if not self.ignore_errors:
                     raise NDStateMachineError(error_msg) from e
 
+        # Fail fast before any mutation if a create item carries no policy (issue #350). Create-only:
```
mikewiebe: ### Medium: policy-less create validation can be masked by API-backed interface preflight

**Issue**

The new guard is described as failing fast before API calls, but it runs after the general interface `preflight()`. For interface orchestrators, `preflight()` resolves switches and may call the `capableSwitches` endpoint before `preflight_create()` has a chance to reject the missing policy. That means a policy-less create can fail on switch resolution or capability lookup first, rather than returning the clearer `config_data.network_os.policy` error.

**Evidence**

`manage_state()` calls `self.model_orchestrator.preflight(list(self.proposed))` before `_manage_create_update_state()`. `preflight_create(items_to_create)` runs later in this block. `NDBaseInterfaceOrchestrator.preflight()` delegates to switch capability validation, which resolves switch IDs and can call the API-backed capability endpoint.

Local ordering check:

```text
RuntimeError capability/switch preflight failed before policy guard
calls= ['preflight']
```

**Impact**

The PR improves the eventual create failure, but it does not reliably surface the intended policy-specific error first. In check mode or with an invalid/unreachable switch, users can still see switch/capability preflight failures before being told the create payload is missing its required policy.

**Suggested fix**

Add an earlier, local-only validation step for policy shape before API-backed capability preflight, or split the create-only policy check so the missing-policy condition can be detected from proposed config before switch/capability validation. Keep the create-only diff-aware hook as needed to avoid rejecting legitimate policy-less updates. Add a test where `preflight()` would raise and verify the missing-policy error wins for create payloads.

plugins/module_utils/nd_state_machine.py
allenrobel: Confirmed and fixed in commit "Run policy preflight before capability preflight and before existing-collection mutation (PR #362 review)". The guard now runs in `manage_state()` before `_manage_create_update_state()` ever touches `self.existing`, so a rejected policy-less create can no longer leak a phantom item into `after` — new test `test_nd_state_machine_00160` asserts the failed output is `changed=False` with `after == before`. One note on scope: the same `after`-aliasing can still misreport when a real create/update **API call** fails mid-run — that's pre-existing NDStateMachine/NDOutput behavior unrelated to this PR's guard, and is now tracked in #389

plugins/module_utils/nd_state_machine.py
allenrobel: Confirmed and fixed in commit "Run policy preflight before capability preflight and before existing-collection mutation (PR #362 review)". The policy guard now runs before the API-backed capability preflight: the create subset is computed by key-membership against the in-memory `existing` inventory (the same criterion `get_diff_config` uses to classify `new`), so the reorder needs no additional API calls and the create-only scoping is preserved. New test `test_nd_state_machine_00170` rigs the capability preflight to fail and asserts the missing-policy error wins for a policy-less create. The `preflight_create` docstring's "before any API call" claim is also corrected (the initial inventory fetch still precedes it).

plugins/module_utils/nd_state_machine.py
allenrobel: The pre-existing API-failure aliasing mentioned above is now tracked in #389 (covers create/update over-report, delete under-report, and the ignore_errors path, across all 17 NDStateMachine-driven modules).

plugins/module_utils/nd_state_machine.py
```diff
@@ -89,9 +89,19 @@ def manage_state(self) -> None:
...
         if self.state in ["merged", "replaced", "overridden"]:
+            proposed_items = list(self.proposed)
+
+            # Policy-required-on-create guard (issue #350) runs FIRST: it is local-only (self.existing is
+            # already in memory), so it fails before the API-backed capability preflight below and before
+            # _manage_create_update_state mutates self.existing, which NDOutput aliases as `after`. Create
+            # subset = proposed items not present in the existing inventory -- the same key-membership
+            # criterion get_diff_config uses to classify "new" (PR #362 review).
+            items_to_create = [item for item in proposed_items if self.existing.get(item.get_identifier_value()) is None]
+            self.model_orchestrator.preflight_create(items_to_create)
```
gmicol: issue: Normalize preflight failures before module entrypoints - Medium

`preflight_create()` can raise `RuntimeError` here, and `manage_state()` lets it leave as-is. Some interface modules using this state machine only catch `NDStateMachineError` at the entrypoint (`nd_interface_svi`, `nd_interface_subinterface_managed`), so a policy-less create in those modules bypasses their `fail_json` path and loses the structured `before` / `after` / `changed` output this guard is meant to provide.

suggested fix:
Wrap `preflight_create()` / `preflight()` exceptions in `NDStateMachineError` inside `manage_state()` before returning to module entrypoints, or add broad exception handling to the affected modules. Add a module-level test for one affected module that asserts policy-less create is reported through `fail_json`.

gmicol (changes requested):

plugins/module_utils/nd_state_machine.py
allenrobel: Confirmed and fixed in commit "Normalize preflight failures to NDStateMachineError so all modules route through fail_json (PR #362 review)". You're right that `preflight_create()` raises a bare `RuntimeError` and `manage_state()` let it leave un-normalized, so a policy-less create in a module that catches only `NDStateMachineError` at its entrypoint escaped as an unhandled exception and bypassed `fail_json`.

I went with your first suggested fix — wrapping the preflight calls in `manage_state()` rather than adding broad handlers per module — because it centralizes the normalization and also closes a latent sibling gap: the API-backed capability `preflight()` on the next line raises a bare `RuntimeError` too, so those same modules were already exposed to it independent of this PR. Both calls are now wrapped and re-raised as `NDStateMachineError` (preserving the original as `__cause__`). The wrap deliberately does **not** route through `_execute_operation`, since that skips execution in check mode and both preflights must run during dry-runs.

One addition to your module list: alongside `nd_interface_svi` and `nd_interface_subinterface_managed`, **`nd_interface_subinterface_unmanaged`** is also affected (same entrypoint pattern) — the central wrap covers all three.

Tests: `test_nd_state_machine_00150/00160/00170` now assert the failure surfaces as `NDStateMachineError` (00150 also asserts the chained `RuntimeError` cause), and new `test_nd_state_machine_00180` covers the capability-preflight normalization leg specifically. Full unit suite green (3021 passed).

nikhilsrikrishna:
LGTM

gmicol (approved):
LGTM

mikewiebe (approved):
LGTM

Issue #384, opened 2026-07-02
Tracking: Gen-3 argspec imports via nd_argument_specs.py


allenrobel:
## Summary

Two related cleanups to finish decoupling Generation-3 modules from legacy module_utils:

1. **Gen-3 modules should import common argspec fragments from `plugins/module_utils/nd_argument_specs.py`** instead of pulling `nd_argument_spec` from the legacy Gen-1 `plugins/module_utils/nd.py`. The file already exists and houses shared fragments (`ntp_server_spec`, `ntp_keys_spec`, `network_spec`, `bgp_spec`), but `nd_argument_spec()` itself is still defined in `nd.py`, so every Gen-3 module carries an import of the Gen-1 runtime file purely for the argspec.
2. **`nd_manage_switches` should re-point its `nd_argument_spec` import** from the deprecated Gen-2 `plugins/module_utils/nd_v2.py` to `nd_argument_specs.py`. (Its full Gen-3 rewrite was originally task 2 of this issue but is deliberately out of scope here — it will be tracked separately, keeping this issue focused on argspec sharing only.)

## Definitions

- **Gen-1**: Original ND modules
- **Gen-2**: ND modules that leveraged `nd_v2.py` or some variant (basically anything not Gen-1 and not Gen-3)
- **Gen-3**: Our new modules that are fully compliant with the orchestrator and state-machine frameworks

### 0. Prerequisite

- [ ] Relocate `nd_argument_spec()` into the existing `plugins/module_utils/nd_argument_specs.py` (alongside `ntp_server_spec`, `network_spec`, etc.). Keep `nd.py` re-exporting it so Gen-1 modules are untouched (same pattern as the `issubset()` relocation, #310). — PR #387 (also adds the shared `config_actions_spec()` fragment with include-allowlist)

### 1. Re-point Gen-3 argspec imports to `nd_argument_specs.py`

- [ ] nd_fabric_update_group
- [ ] nd_interface_ethernet_access
- [ ] nd_interface_ethernet_trunk_host
- [ ] nd_interface_loopback
- [ ] nd_interface_port_channel_access
- [ ] nd_interface_port_channel_trunk_host
- [ ] nd_interface_subinterface_managed
- [ ] nd_interface_subinterface_unmanaged
- [ ] nd_interface_svi
- [ ] nd_interface_vpc_access
- [ ] nd_interface_vpc_trunk_host
- [ ] nd_local_user
- [ ] nd_maintenance_mode
[...]

Issue #310, opened 2026-06-10
issubset() relocated from nd.py to utils.py: add explicit re-export note + changelog entry


allenrobel:
Deferred finding from review of #198 (applies once that PR merges). Low severity.

## Problem

#198 moves `issubset()` from `plugins/module_utils/nd.py` to `plugins/module_utils/utils.py`. All in-tree imports are migrated, and external code doing `from ansible_collections.cisco.nd.plugins.module_utils.nd import issubset` still works — but only as a side effect of `nd.py` importing the symbol for its own use. `module_utils` is public API for a released collection, so this accidental compatibility could be broken unknowingly by a future cleanup that drops the now-"unused-looking" import.

## Suggested fix

- Add a short comment on the `issubset` import in `nd.py` marking it as an intentional backward-compat re-export.
- Add a changelog fragment noting the relocation and that the old import path remains supported.

Issue #368, opened 2026-06-30
Module Support for config_actions + resource level deploy where applicable


mikewiebe:
### New module/feature description:

We need to make sure that each module supports a `config_actions` dictionary at the global level and a resource level `deploy` key.

```
    config_actions:
        description:
        - Controls save and deploy behavior after inventory is updated.  This option is mutually exclusive with the per resource deploy key.
        type: dict
        suboptions:
            save:
                description:
                - Save/Recalculate the configuration of the fabric after resource is updated.
                type: bool
                default: true
            deploy:
                description:
                - Deploy the pending configuration after resource is updated.
                - When set to C(true), C(save) must also be C(true). (Only if save is required for the module)
                type: bool
                default: true
            type:
                description:
                - Scope of the deploy operation.
                - C(resource) deploys using resource level deploy
                - C(switch) deploys using switch level deploy
                - C(global) deploys using global fabric level deploy
                type: str
                default: switch
                choices:
                - resource
                - switch
                - global
```

Notes:

- `config_actions` and resource level deploy key are mutually exclusive.  If both `config_actions` and resource level deploy are included in the module inputs an error should be generated to indicate that the user needs to pick one and the module documentation should clearly call this out.
- `config_actions.save` does is not needed for all modules.  Only needed for modules like  fabric, switches, vpc_pair, tor_pair , few others...
- `config_actions.type` under config_actions controls deployment type for all resources managed by the module 
- If the module cannot support the contract above **because of of the way ND functions** document the reasons clearly and call this out in the PR

### Which resource area of Cisco Nexus Dashboard does this new feature belong to?

manage

### New or Affected Documentation for New Feature

```shell

```

### Potential Ansible Task Config

```yaml
- name: Create an L3 Network associated with a VRF
  cisco.nd.nd_manage_networks:
    fabric_name: fab1
    state: merged
    config_actions:  <------------------- Config Actions
      deploy: true
      type: switch
    config:
      - network_name: Network_L3
        is_l2only: false
        vrf_name: Tenant_A
        network_id: 50020
        vlan_id: 2002
        vlan_name: Network_L3_VLAN
        gateway_ipv4_address: 10.10.20.1/24
        arp_suppression: true
        routing_tag: 12345
        deploy: true  <------------------- Resource level deploy
```

Note: One of these options must be selected by users, not both.  If both `config_actions` and `deploy` key under `config:` are specified the module should generate an error indicating that one method must be selected.

### Ansible Version

[...]

PR #387, opened 2026-07-02
Relocate nd_argument_spec to nd_argument_specs.py; add config_actions_spec fragment

allenrobel:
## Related Issue(s)

Prerequisite task (task 0) of #384 — not `Fixes`, since #384 remains open as the tracking issue.

## Proposed Changes

- Relocate `nd_argument_spec()` to `plugins/module_utils/nd_argument_specs.py` (single source of truth), joining the existing shared fragments (`ntp_server_spec`, `network_spec`, ...). The relocated function adopts the typed variant that previously lived in `nd_v2.py`.
- `nd.py` and `nd_v2.py` re-export it with explicit backward-compat comments, so **no module imports change in this PR** (re-pointing Gen-3 modules is tracked per-module under #384 task 1). This intentionally does better than the `issubset()` relocation (#310), which shipped without the re-export note or changelog entry.
- Remove `nd_v2.py`'s duplicate definition in favor of the shared one.
- Add `config_actions_spec()` shared argspec fragment (full option set: `save`/`deploy`/`type`, matching the superset used by `nd_manage_switches`), with an include-allowlist via `_select_options()` so modules compose only the options they support — `config_actions_spec(include=("deploy",))` reproduces the deploy-only block currently hand-written in the eight `nd_interface_*` modules byte-for-byte. An allowlist (not an exclude list) is deliberate: options added to a shared fragment later must never silently appear in modules that composed it earlier.
- Add unit tests (`tests/unit/module_utils/test_nd_argument_specs.py`) and a changelog fragment.

## Test Notes

- `ndpytest tests/unit/` — 2169 passed (includes 6 new tests)
- `black`, `isort`: clean; `ansible-lint`: 0 failures on changed files; changelog fragment passes YAML validation
- `pylint`/`mypy`: no new findings attributable to this change (both files carry pre-existing environmental import-resolution noise; verified against the develop baseline)
- Runtime verification inside nd-dev: `nd.py` and `nd_v2.py` re-exports resolve to the same function object, and `config_actions_spec(include=("deploy",))` equals the current interface-module block
- `ansible-test sanity --test import` could not run in the offline nd-dev machine (pip bootstrap requires network); CI will exercise it

4.2.1

* [ ] analyze
* [ ] infra
* [ ] manage
* [ ] onemanage
* [x] other

## Checklist

https://claude.ai/code/session_01Deb5UWoTfAZa6kq7khAEu6

### DIFF
--- a/changelogs/fragments/relocate_nd_argument_spec.yml
+++ b/changelogs/fragments/relocate_nd_argument_spec.yml
@@ -2,5 +2,6 @@
 minor_changes:
   - Relocate ``nd_argument_spec()`` to ``plugins/module_utils/nd_argument_specs.py`` as the single source of truth;
     ``nd.py`` and ``nd_v2.py`` continue to re-export it, so existing imports keep working unchanged.
-  - Add shared ``config_actions_spec()`` argument-spec fragment (``save``/``deploy``/``type``) with an include-allowlist,
-    so modules can compose the subset of config actions they support instead of hand-writing the block.
+  - Add shared ``config_actions_spec()`` argument-spec fragment (``save``/``deploy``/``type``, where ``type`` accepts
+    ``resource``, ``switch``, or ``global``) with an include-allowlist, so modules can compose the subset of config actions
+    they support instead of hand-writing the block.
--- a/changelogs/fragments/relocate_nd_argument_spec.yml
+++ b/changelogs/fragments/relocate_nd_argument_spec.yml
@@ -1,0 +1,6 @@
+---
+minor_changes:
+  - Relocate ``nd_argument_spec()`` to ``plugins/module_utils/nd_argument_specs.py`` as the single source of truth;
+    ``nd.py`` and ``nd_v2.py`` continue to re-export it, so existing imports keep working unchanged.
+  - Add shared ``config_actions_spec()`` argument-spec fragment (``save``/``deploy``/``type``) with an include-allowlist,
+    so modules can compose the subset of config actions they support instead of hand-writing the block.
--- a/plugins/module_utils/nd_argument_specs.py
+++ b/plugins/module_utils/nd_argument_specs.py
@@ -1,8 +1,90 @@
 # Copyright: (c) 2023, Shreyas Srish (@shrsr) <ssrish@cisco.com>
 
 # GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+"""
+# nd_argument_specs.py
 
-from __future__ import absolute_import, division, print_function
+Shared argument-spec building blocks for ND modules: the common connection/authentication spec (`nd_argument_spec`) and reusable
+fragments (`config_actions_spec`, `ntp_server_spec`, ...) that modules compose into their `argument_spec`.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Iterable
+from typing import Any
+
+from ansible.module_utils.basic import env_fallback
+
+
+def nd_argument_spec() -> dict[str, Any]:
+    """
+    # Summary
+
+    Return the common connection/authentication argument spec accepted by all ND modules.
+
+    ## Raises
+
+    None
+    """
+    return {
+        "host": {"type": "str", "required": False, "aliases": ["hostname"], "fallback": (env_fallback, ["ND_HOST"])},
+        "port": {"type": "int", "required": False, "fallback": (env_fallback, ["ND_PORT"])},
+        "username": {"type": "str", "fallback": (env_fallback, ["ND_USERNAME", "ANSIBLE_NET_USERNAME"])},
+        "password": {"type": "str", "required": False, "no_log": True, "fallback": (env_fallback, ["ND_PASSWORD", "ANSIBLE_NET_PASSWORD"])},
+        "output_level": {"type": "str", "default": "normal", "choices": ["debug", "info", "normal"], "fallback": (env_fallback, ["ND_OUTPUT_LEVEL"])},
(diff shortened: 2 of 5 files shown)


.github/workflows/ansible-test.yml
```diff
@@ -115,7 +115,7 @@ jobs:
           path: .cache/collection-tarballs
 
       - name: Install the collection tarball
-        run: ansible-galaxy collection install .cache/collection-tarballs/*.tar.gz
+        run: ansible-galaxy collection install .cache/collection-tarballs/*.tar.gz --no-deps
```
mtarking: Why the no dependencies?

changelogs/fragments/relocate_nd_argument_spec.yml
```diff
@@ -0,0 +1,6 @@
+---
```
mtarking, line 1: Are we committing individual changelogs?

plugins/module_utils/nd_argument_specs.py
```diff
@@ -1,8 +1,90 @@
 # Copyright: (c) 2023, Shreyas Srish (@shrsr) <ssrish@cisco.com>
 
 # GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+"""
+# nd_argument_specs.py
```
mtarking, line 5: Why are we pulling the arg spec out instead of consolidating down to just one "nd.py" vs keeping nd.py and nd_v2.py; e.g. single NDModule object?

mtarking (changes requested):

changelogs/fragments/relocate_nd_argument_spec.yml
allenrobel: Yes — deliberately. `changelogs/config.yaml` already declares `notesdir: fragments` with `keep_fragments: false`, i.e. the standard antsibull-changelog flow: fragments are committed during development, then automatically consumed (deleted) into `changelog.yaml` by `antsibull-changelog release` at release time. Until now `changelog.yaml` has been hand-edited at release; committing fragments moves changelog authorship to PR time, where the context lives. Happy to drop the fragment if the team prefers the current release-time process.

plugins/module_utils/nd_argument_specs.py
allenrobel: They're complementary rather than competing. Gen-3 modules (orchestrator + state-machine based) don't use *any* `NDModule` — `nd_argument_spec` was the only symbol they still imported from `nd.py`, so relocating it severs Gen-3's last dependency on the Gen-1 runtime without touching Gen-1 behavior. Consolidating the two `NDModule` classes is a much bigger job (fail_json-based vs exception-based semantics) and is intentionally out of scope for #384, which now tracks the argspec sharing work only — the Gen-3 rewrite of `nd_manage_switches` (the sole `nd_v2.py` consumer) will be tracked in a separate issue, per review discussion. Once that rewrite lands, `nd_v2.py` has zero importers and can be deleted, leaving exactly one `NDModule` (legacy `nd.py`, used by Gen-1 modules only) — and new modules never import an `NDModule` at all.

akinross (approved):

.github/workflows/ansible-test.yml
allenrobel: Good catch — this wasn't part of the PR's intended scope. A `copilot-swe-agent` commit ("Fix CI: add --no-deps to ansible-galaxy collection install to avoid Galaxy connectivity failures") was pushed to the branch to work around galaxy.ansible.com flakiness when resolving the `ansible.netcommon` dependency during tarball install. I've dropped that commit from this branch and filed #388 to handle it properly: `--no-deps` is safe for the galaxy-importer/sanity/units jobs, but the integration job genuinely needs `ansible.netcommon` (`tests/integration/inventory.networking` uses `ansible_connection=ansible.netcommon.httpapi`), so a blanket `--no-deps` isn't right there.

gmicol (approved):
LGTM

AKDRG:
Hey Allen,
As per https://github.com/CiscoDevNet/ansible-nd/issues/368, we are discussing on a resource level deploy for VRF, Networks and potentially other modules. So, we need to support that as well in the specs.

mtarking (approved):

allenrobel:
Hi @AKDRG — thanks, good catch.

I've added `resource` to the `config_actions.type` choices, so the fragment now offers all three (`resource` / `switch` / `global`) per the contract in #368.

I've held off on the companion per-resource `deploy` key for now, because I don't think #368 has consensus yet on how it interacts with `config_actions`. The issue body says the two are **mutually exclusive** (error if both are supplied), while your follow-up comment says the per-resource `deploy` key is accepted **only when `config_actions.type == "resource"`** — i.e. they're used together, with `type` as the gate. Those imply different validation, so I'd rather not encode a guess into a shared fragment that every module will inherit.

Worth noting too that neither rule can be expressed in the argument spec alone: the constraint spans the top-level `config_actions` and the per-item `config` suboptions, so `mutually_exclusive` won't reach it. It'll need a shared runtime validation helper alongside the fragment — which is a bit more surface area than this PR's relocation scope, and has no consumer until `nd_manage_vrf` / `nd_manage_networks` land.

Once the behavior is settled on #368, I'll add the resource-deploy fragment plus its validator as the next PR in the #384 stack. I've noted the deferral in the `config_actions_spec()` docstring so the omission is traceable.

AKDRG:
@allenrobel Agreed. I was planning to take up the per-item config deploy and config level actions "resource deploy type" validation separately. 

This PR looks good to me now. Can be merged.

AKDRG (approved):
Looks Good!

allenrobel:
Merging with 4x approvals (3x Matt's team 1x Mike's team).  Thanks all!

Issue #340, opened 2026-06-20
Track state: gathered support for NDStateMachine/orchestrator-backed modules


mikewiebe:
### New module/feature description:

Track and implement consistent `state: gathered` support for modules that use the Gen-3 `NDStateMachine` + orchestrator architecture.

Right now, some new modules support `gathered` and others only support mutating states such as `merged`, `replaced`, `overridden`, and `deleted`. Every supported state needs documentation examples and meaningful tests, and `gathered` is a standard expectation for modules where the Manage API exposes a read/list operation.

The ansible state `gathered` should return a list of resources for the module in the format of the Ansible argspec for that module.

This issue should track a repository-wide audit and implementation plan:

- Identify every module backed by `NDStateMachine` and a resource-specific orchestrator.
- Add `gathered` to the module argspec and DOCUMENTATION when supported.
- Add at least one `gathered` example per supported module.
- Add meaningful unit and/or integration coverage for gathered behavior.
- Make sure the shared state-machine/orchestrator contract supports gathered consistently rather than requiring each module to hand-roll the same pattern.
- Document intentional exceptions where gathered does not make sense or the API cannot support it. (I can't think of any exceptions at this point)
- Need to decide as a team if `gathered` should support filtering functionality using the `config:` block settings or if it should just return a full list of resources each time

Related issue:

### Which resource area of Cisco Nexus Dashboard does this new feature belong to?

manage

### New or Affected Documentation for New Feature

- Module DOCUMENTATION for every affected module.
- Module EXAMPLES for every affected module.
- Contributor/developer guidance for Gen-3 `NDStateMachine` + orchestrator modules.
- Integration test guidance so each supported state has a corresponding workflow, including `gathered` where applicable.

### Potential Ansible Task Config

```yaml
- name: Gather all resources for a module
  cisco.nd.<module_name>:
    state: gathered
    fabric_name: "{{ fabric_name }}"

- name: Gather selected resources for a module
  cisco.nd.<module_name>:
    state: gathered
    fabric_name: "{{ fabric_name }}"
    config:
      - name: example-resource
```

Exact keys under `config` should be module-specific and should match that module's documented identifiers and API filter capabilities.

### STEPS TO CONSIDER in gather state implemetation:
We can follow below steps for gather state implementation as well as validation of props for filtering.
it's regarding the gather state work Deeksha is doing. I was thinking of the following steps to follow in this case:

**For gathering state config properties:**
[...]

PR #312, opened 2026-06-11
Add ND 4.2 Links module (nd_links) (NDA-21)

shrsr:
Adds the nd_links module for ND 4.2 with manage/one_manage link scopes, template models, endpoints, orchestrator, strategies and fabric inventory support, plus integration tests.

## Related ND API Resource Category

* [ ] analyze
* [ ] infa
* [ ] manage
* [ ] onemanage
* [ ] other

## Checklist

* [ ] Latest commit is rebased from develop with merge conflicts resolved
* [ ] New or updates to documentation has been made accordingly
* [ ] Assigned the proper reviewers

PR #391, opened 2026-07-07
Add opt-in gathered-state filtering for loopback and local users

deekpand-cisco:
## Related Issue(s)

Related to [#340](https://github.com/CiscoDevNet/ansible-nd/issues/340).

This PR builds on the foundational framework established in
[#312](https://github.com/CiscoDevNet/ansible-nd/pull/312) 
including the state machine, base model infrastructure, orchestrator pattern,
and secrets handling. 

## Proposed Changes

This PR introduces an opt-in gathered-state filtering framework for modules
using `NDStateMachine` and enables it for:

- `cisco.nd.nd_interface_loopback`
- `cisco.nd.nd_local_user`

### Shared gathered-filtering framework

The shared implementation:

- runs filtering only when `state: gathered`;
- requires each model to opt in using `supports_gathered_filtering`;
- treats `config` as partial filter criteria for supported gathered operations;
- uses AND semantics between fields within one filter item;
- uses OR semantics between separate filter items;
- converts API responses into normalized Ansible configuration before final matching;
- allows models to declare supported filter properties through `gathered_filter_properties` and rejects unsupported properties before any API request;
- allows models to normalize filter values through `normalize_gathered_filter`;
- allows models to customize matching through `matches_gathered_filter`;
- removes duplicate matches using the model identifier;
- keeps gathered filters out of `proposed` configuration;
- prevents gathered filters from entering create, update, or delete diff processing;
- preserves existing behavior for non-gathered states;
- preserves existing behavior for models that have not opted in.

The generic local filter remains the final correctness layer. Endpoint-specific filtering is used only to reduce the number of API candidates where the endpoint safely supports it.

### Loopback-interface filtering

The loopback module supports gather-all and partial gathered filters across switches in a fabric.

Gathered loopback queries use the interface-list endpoint and its Lucene `filter` query parameter:

- every query includes `interfaceType:loopback` and `policyType:loopback`;
- `switch_ip` is resolved to a switch serial number and limits the query to that switch;
- omitting `switch_ip` queries all switches in the fabric;
- `interface_name` is translated into an exact-match Lucene term;
- separate filter items are queried independently to preserve OR semantics;
- an unfiltered loopback query for a switch supersedes narrower interface-name queries for that switch;
- pagination uses `max` and `offset` until the API reports no remaining results;
- responses from multiple switches, filter items, and pages are combined and deduplicated.

Lucene expressions are built only from structured module input. Raw Lucene expressions are not accepted from users.

The module continues to enforce its resource-ownership boundary:

- `interfaceType` must be `loopback`;
- `policyType` must be `loopback`;
- system-managed `underlayLoopback` interfaces are excluded;
- non-standard `ipfmLoopback` and `userDefined` policies are excluded.

Nested criteria such as IP address, administrative state, IPv6 address, and VRF are evaluated by the generic final local filter.

Interface names supplied as filters are normalized to lowercase to match the existing model and Nexus Dashboard API behavior.

### Local-user filtering

The local-user endpoint does not expose documented server-side Lucene filtering. Gather-all retrieves the available local users, and optional filter criteria are evaluated through the shared final local-matching layer.

Supported gathered filter properties for local users are `login_id`, `email`, `first_name`, and `last_name`. Other configuration fields (such as `user_password`, `reuse_limitation`, or `security_domains`) are rejected with a validation error before any API request.

Multiple filter items use OR semantics. String matching is case-sensitive.

Local-user passwords are not returned by Nexus Dashboard and remain excluded from gathered output. Gathered output for an existing user can be reused as configuration, but recreating a deleted user still requires a new password.

### Argument validation and gathered output

For both modules:

- `config` can be omitted for gather-all operations;
- `config` remains required for write states through `required_if`;
- the underlying Pydantic resource identifiers remain required for complete resources;
- gathered output is pruned to the module argument specification so it can be reused as `config`;
- gathered operations remain read-only and return `changed: false`.

For loopback gathered operations, `switch_ip`, `interface_name`, and nested policy fields (`admin_state`, `ip`, `ipv6`, `vrf`) can be supplied as partial filtering criteria.

For local-user gathered operations, `login_id`, `email`, `first_name`, and `last_name` can be supplied as partial filtering criteria.

### Compatibility and rollout

- Gathered filtering is disabled by default in the base model.
- Existing modules are unaffected until they explicitly opt in.
- Lucene filtering is disabled by default in the base orchestrator.
- Existing orchestrators are unaffected until they explicitly opt in.
- Custom modules using specialized runners are unaffected.
- Non-gathered state behavior remains unchanged.
- Gathered filters do not enter write-state processing.
- Gathered operations do not modify Nexus Dashboard resources.

## Test Notes

Unit coverage includes Lucene expression construction and escaping, query planning, pagination, deduplication, local matching, gathered/write-state isolation, local-user filtering, property validation, and password secrecy.

Integration coverage was added for gather-all, exact and multiple filters, OR semantics, no-match behavior, unsupported filter rejection, empty filter rejection, password secrecy, and gathered-output reuse.

Cisco Nexus Dashboard 4.2 lab environment.

- manage
- infra

The loopback-interface module uses the Manage API, while the local-user module uses the Infra AAA API.

## Checklist

- [x] New or updated documentation has been added

### DIFF
--- a/plugins/module_utils/gathered_filter.py
+++ b/plugins/module_utils/gathered_filter.py
@@ -203,13 +203,15 @@ def filter_gathered_response(
     filters: list[dict[str, Any]],
     model_class: type[NDBaseModel],
     normalize_filter: Callable[[dict[str, Any]], dict[str, Any]] | None = None,
-) -> list[dict[str, Any]]:
+) -> list[NDBaseModel]:
     """
-    Apply gathered filters to raw API objects.
+    Apply gathered filters to raw API objects and return validated models.
 
     Criteria inside one filter item use AND semantics.
     Multiple filter items use OR semantics.
-    Returned objects remain in API shape for NDConfigCollection.from_api_response().
+
+    Returns validated model instances directly so callers can construct an
+    NDConfigCollection without redundant from_response() calls.
     """
     # Validation duplicates validate_gathered_filters() intentionally so this
     # function remains safe to call standalone (e.g., from unit tests).
@@ -227,7 +229,7 @@ def filter_gathered_response(
 
         active_filters.append(normalized)
 
-    filtered: list[dict[str, Any]] = []
+    filtered_models: list[NDBaseModel] = []
     seen_identifiers = set()
 
     for response_item in response_data:
@@ -243,9 +245,9 @@ def filter_gathered_response(
             continue
 
         seen_identifiers.add(identifier)
-        filtered.append(response_item)
+        filtered_models.append(model)
 
-    return filtered
+    return filtered_models
 
 
 def validate_gathered_filters(
--- a/plugins/module_utils/gathered_filter.py
+++ b/plugins/module_utils/gathered_filter.py
@@ -108,6 +108,102 @@ def _contains_active_value(value: Any) -> bool:
     # valid loopback fields and are treated as absent criteria.
     return value not in (None, "")
 
+def _extract_active_leaf_paths(data: dict[str, Any], prefix: str = "") -> set[str]:
+    """Return dot-paths for every non-None, non-empty leaf value in a nested dict.
+
+    Given a gathered filter dict (which may be nested due to Ansible's argument
+    spec structure), this walks the tree and identifies which properties the user
+    actually supplied a value for.
+
+    Ansible pads all omitted suboptions with None after argument validation.
+    Those Nones are not user-supplied criteria and are correctly skipped.
+
+    Args:
+        data: A (possibly nested) filter dict from the user's config item.
+        prefix: Internal recursion prefix — callers should not pass this.
+
+    Returns:
+        A set of dot-path strings representing active (non-None) leaf values.
+
+    Examples:
+        >>> _extract_active_leaf_paths({"switch_ip": "10.1.1.1", "interface_name": None})
+        {"switch_ip"}
+
+        >>> _extract_active_leaf_paths({
+        ...     "config_data": {"network_os": {"policy": {"vrf": "mgmt", "ip": None}}}
+        ... })
+        {"config_data.network_os.policy.vrf"}
+
+        >>> _extract_active_leaf_paths({"login_id": "admin"})
+        {"login_id"}
+
+        >>> _extract_active_leaf_paths({"config_data": None})
+        set()  # None at any level is skipped
+    """
+    paths: set[str] = set()
+    for key, value in data.items():
+        full_path = f"{prefix}.{key}" if prefix else key
+        if isinstance(value, dict):
+            paths |= _extract_active_leaf_paths(value, full_path)
+        elif value not in (None, "", []):
+            # False and 0 ARE active values (e.g., admin_state: false)
+            paths.add(full_path)
+    return paths
+
+
+def _reject_unsupported_filter_properties(
+    filter_item: dict[str, Any],
+    supported_properties: tuple[str, ...],
+    index: int,
+) -> None:
+    """Raise ValueError if a filter item contains properties not in the supported set.
+
+    This is the enforcement function for the model's gathered_filter_properties
+    declaration. Called exactly once per filter item during pre-flight validation,
+    before any API request or normalization.
+
+    Args:
+        filter_item: One raw filter dict from the user's config list.
+        supported_properties: The model's declared tuple of allowed dot-paths.
+        index: The position of this filter item in the config list (for error messages).
+
+    Raises:
+        ValueError: With a message listing unsupported properties and the full
+            set of supported properties. Formatted for direct use in Ansible's
+            module.fail_json(msg=...).
+
+    Examples:
+        # Passes (switch_ip is supported):
+        >>> _reject_unsupported_filter_properties(
+        ...     {"switch_ip": "10.1.1.1"},
+        ...     ("switch_ip", "interface_name"),
+        ...     index=0,
+        ... )  # No error
+
+        # Fails (extra_config is not supported):
+        >>> _reject_unsupported_filter_properties(
+        ...     {"config_data": {"network_os": {"policy": {"extra_config": "cli"}}}},
+        ...     ("switch_ip", "interface_name", "config_data.network_os.policy.vrf"),
+        ...     index=0,
+        ... )
+        # ValueError: "Gathered filter at index 0 contains unsupported properties: 
+        #   'config_data.network_os.policy.extra_config'. 
+        #   Supported gathered filter properties: 'switch_ip', 'interface_name', ..."
+    """
+    supplied_paths = _extract_active_leaf_paths(filter_item)
+    supported_set = set(supported_properties)
+    unsupported = sorted(supplied_paths - supported_set)
+    if unsupported:
+        raise ValueError(
+            "Gathered filter at index {0} contains unsupported properties: {1}. "
+            "Supported gathered filter properties: {2}".format(
+                index,
+                ", ".join("'{0}'".format(p) for p in unsupported),
+                ", ".join("'{0}'".format(p) for p in supported_properties),
+            )
+        )
+
+
 
 def filter_gathered_response(
     response_data: list[dict[str, Any]],
@@ -160,24 +256,54 @@ def filter_gathered_response(
 def validate_gathered_filters(
         filters: list[Any],
         normalize_filter: Callable[[dict[str, Any]], dict[str, Any]] | None = None,
+        supported_properties: tuple[str, ...] = (),
 ) -> None:
     """ 
     Pre-flight validation for gathered filter items.
 
-    Raises ValueError if any filter item is malformed or contains no actionable 
-    criteria. Call this before making API requests so that invalid input is r
-    rejected without wasted network I/O.
+    Validates each filter item through a strict pipeline:
+      1. Type check — must be a dict
+      2. Property check — all active properties must be in the supported set
+      3. Normalize — transform values to canonical form (e.g., lowercase)
+      4. Active check — at least one non-None criterion must exist
+
+    Property validation runs BEFORE normalization because:
+      - It checks KEYS (property paths), normalization transforms VALUES.
+      - Unsupported properties should be rejected without wasting normalization effort.
+      - Error messages should show what the user wrote, not the normalized form.
+
+    Args:
+        filters: The raw config list from module.params["config"].
+            Example: [{"switch_ip": "10.1.1.1", "interface_name": "loopback0"}]
+
+        normalize_filter: Model-specific value transformer (e.g., lowercase interface_name).
+            Signature: (dict) -> dict. May be None if no normalization needed.
+
+        supported_properties: The model's gathered_filter_properties tuple.
+            Example: ("switch_ip", "interface_name", "config_data.network_os.policy.vrf")
+            Empty tuple means property validation is skipped (backward-compatible).
+
+    Raises:
+        ValueError: If any filter item fails any validation step.
     """
     for idx, filter_item in enumerate(filters):
+        # Step 1: Type check
         if not isinstance(filter_item, dict):
             raise ValueError(
                 f"Each gathered filter item must be a dictionary, "
                 f"got {type(filter_item).__name__} at index {idx}."
             )
+        
+        # Step 2: Property validation (before normalize - checks keys, not values)
+        if supported_properties:
+            _reject_unsupported_filter_properties(filter_item, supported_properties, idx)
+
+        # Step 3: Normalize values to canonical form
         normalized = deepcopy(filter_item)
         if normalize_filter is not None:
             normalized = normalize_filter(normalized)
         
+        #Step 4: Active value check
         if not _contains_active_value(normalized):
             raise ValueError(
                 f"Gathered filter item at index {idx} must contain "
--- a/plugins/module_utils/gathered_filter.py
+++ b/plugins/module_utils/gathered_filter.py
@@ -1,4 +1,4 @@
-# Copyright: (c) 2026, Deeksha Pandey (deekpand-cisco)  deekpand@cisco.com
+# Copyright: (c) 2026, Deeksha Pandey (@deekpand) <deekpand@cisco.com>
 
 """Generic local and server-side filtering helpers for gathered-state responses."""
 
@@ -191,12 +191,10 @@ def _reject_unsupported_filter_properties(
     unsupported = sorted(supplied_paths - supported_set)
     if unsupported:
         raise ValueError(
-            "Gathered filter at index {0} contains unsupported properties: {1}. "
-            "Supported gathered filter properties: {2}".format(
-                index,
-                ", ".join("'{0}'".format(p) for p in unsupported),
-                ", ".join("'{0}'".format(p) for p in supported_properties),
-            )
+            f"Gathered filter at index {index} contains unsupported properties: "
+            f"{', '.join(repr(p) for p in unsupported)}. "
+            f"Supported gathered filter properties: "
+            f"{', '.join(repr(p) for p in supported_properties)}"
         )
 
 
@@ -213,7 +211,8 @@ def filter_gathered_response(
     Multiple filter items use OR semantics.
     Returned objects remain in API shape for NDConfigCollection.from_api_response().
     """
-
+    # Validation duplicates validate_gathered_filters() intentionally so this
+    # function remains safe to call standalone (e.g., from unit tests).
     active_filters: list[dict[str, Any]] = []
     for filter_item in filters or []:
         if not isinstance(filter_item, dict):
--- a/plugins/module_utils/gathered_filter.py
+++ b/plugins/module_utils/gathered_filter.py
@@ -1,6 +1,4 @@
-# Copyright: (c) 2026, Cisco and/or its affiliates.
-
-# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+# Copyright: (c) 2026, Deeksha Pandey (deekpand-cisco)  deekpand@cisco.com
 
 """Generic local and server-side filtering helpers for gathered-state responses."""
 
@@ -77,10 +75,7 @@ def build_lucene_expressions(
     expressions: list[str] = []
 
     for filter_item in filter_items:
-        terms = [
-            f"{field_name}:{format_lucene_value(value)}"
-            for field_name, value in spec.base_terms
-        ]
+        terms = [f"{field_name}:{format_lucene_value(value)}" for field_name, value in spec.base_terms]
 
         for config_path, api_field in spec.field_map.items():
             value = get_nested_value(filter_item, config_path)
@@ -107,6 +102,7 @@ def _contains_active_value(value: Any) -> bool:
     # False and 0 can be meaningful filter values. Empty strings cannot match
     # valid loopback fields and are treated as absent criteria.
     return value not in (None, "")
+
 
 def _extract_active_leaf_paths(data: dict[str, Any], prefix: str = "") -> set[str]:
     """Return dot-paths for every non-None, non-empty leaf value in a nested dict.
@@ -186,8 +182,8 @@ def _reject_unsupported_filter_properties(
         ...     ("switch_ip", "interface_name", "config_data.network_os.policy.vrf"),
         ...     index=0,
         ... )
-        # ValueError: "Gathered filter at index 0 contains unsupported properties: 
-        #   'config_data.network_os.policy.extra_config'. 
+        # ValueError: "Gathered filter at index 0 contains unsupported properties:
+        #   'config_data.network_os.policy.extra_config'.
         #   Supported gathered filter properties: 'switch_ip', 'interface_name', ..."
     """
     supplied_paths = _extract_active_leaf_paths(filter_item)
@@ -202,7 +198,6 @@ def _reject_unsupported_filter_properties(
                 ", ".join("'{0}'".format(p) for p in supported_properties),
             )
         )
-
 
 
 def filter_gathered_response(
@@ -220,7 +215,7 @@ def filter_gathered_response(
     """
 
     active_filters: list[dict[str, Any]] = []
-    for filter_item in (filters or []):
+    for filter_item in filters or []:
         if not isinstance(filter_item, dict):
             raise ValueError("Each gathered filter item must be a dictionary.")
 
@@ -253,12 +248,13 @@ def filter_gathered_response(
 
     return filtered
 
+
 def validate_gathered_filters(
-        filters: list[Any],
-        normalize_filter: Callable[[dict[str, Any]], dict[str, Any]] | None = None,
-        supported_properties: tuple[str, ...] = (),
+    filters: list[Any],
+    normalize_filter: Callable[[dict[str, Any]], dict[str, Any]] | None = None,
+    supported_properties: tuple[str, ...] = (),
 ) -> None:
-    """ 
+    """
     Pre-flight validation for gathered filter items.
 
     Validates each filter item through a strict pipeline:
@@ -289,11 +285,8 @@ def validate_gathered_filters(
     for idx, filter_item in enumerate(filters):
         # Step 1: Type check
         if not isinstance(filter_item, dict):
-            raise ValueError(
-                f"Each gathered filter item must be a dictionary, "
-                f"got {type(filter_item).__name__} at index {idx}."
-            )
-        
+            raise ValueError(f"Each gathered filter item must be a dictionary, " f"got {type(filter_item).__name__} at index {idx}.")
+
         # Step 2: Property validation (before normalize - checks keys, not values)
         if supported_properties:
             _reject_unsupported_filter_properties(filter_item, supported_properties, idx)
@@ -302,11 +295,7 @@ def validate_gathered_filters(
         normalized = deepcopy(filter_item)
         if normalize_filter is not None:
             normalized = normalize_filter(normalized)
-        
-        #Step 4: Active value check
-        if not _contains_active_value(normalized):
-            raise ValueError(
-                f"Gathered filter item at index {idx} must contain "
-                f"at least one filtering criterion."
-            )
 
+        # Step 4: Active value check
+        if not _contains_active_value(normalized):
+            raise ValueError(f"Gathered filter item at index {idx} must contain " f"at least one filtering criterion.")
--- a/plugins/module_utils/gathered_filter.py
+++ b/plugins/module_utils/gathered_filter.py
@@ -1,0 +1,160 @@
+# Copyright: (c) 2026, Cisco and/or its affiliates.
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+"""Generic local and server-side filtering helpers for gathered-state responses."""
+
+from __future__ import annotations
+
+from copy import deepcopy
+from dataclasses import dataclass, field
+from enum import Enum
+import re
+from typing import Any, Callable, Mapping
+
+from ansible_collections.cisco.nd.plugins.module_utils.models.base import NDBaseModel
+
+FieldPath = tuple[str, ...]
+_MISSING = object()
+_SAFE_UNQUOTED = re.compile(r"^[A-Za-z0-9_.]+$")
+
+
+@dataclass(frozen=True)
+class GatheredLuceneSpec:
+    """Describe fields supported by an endpoint's Lucene filter.
+
+    ``base_terms`` are included in every generated expression. ``field_map``
+    maps structured Ansible config paths to controller-supported field names.
+    """
+
+    base_terms: tuple[tuple[str, Any], ...] = ()
+    field_map: Mapping[FieldPath, str] = field(default_factory=dict)
+
+
+def format_lucene_value(value: Any) -> str:
+    """Format one value as an exact-match Lucene literal.
+
+    Values are generated from structured module input, never accepted as raw
+    Lucene syntax. Strings containing reserved punctuation or whitespace are
+    quoted and escaped.
+    """
+    if isinstance(value, bool):
+        return str(value).lower()
+    if isinstance(value, Enum):
+        value = value.value
+    if isinstance(value, (int, float)):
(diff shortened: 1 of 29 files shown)


plugins/module_utils/gathered_filter.py
```diff
@@ -0,0 +1,160 @@
...
+) -> list[dict[str, Any]]:
+    """
+    Apply gathered filters to raw API objects.
+
+    Criteria inside one filter item use AND semantics.
+    Multiple filter items use OR semantics.
+    Returned objects remain in API shape for NDConfigCollection.from_api_response().
+    """
+
+    if not filters:
```
sivakasi-cisco: Should we still deduplicate the results when no filters are provided? Right now this returns before the deduplication logic runs.

plugins/module_utils/nd_state_machine.py
```diff
@@ -65,24 +66,67 @@ def __init__(self, module: AnsibleModule, model_orchestrator: type[NDBaseOrchest
...
+                self.model_orchestrator,
+                "supports_gathered_lucene_filtering",
+                False,
+            )
+            query_kwargs = {}
+            if lucene_candidate_filtering_enabled:
+                query_kwargs["gathered_filters"] = raw_config
+
+            response_data = self.model_orchestrator.query_all(**query_kwargs)
+
```
sivakasi-cisco: Can we validate the gathered filters before calling query_all()? Otherwise, an invalid filter may make unnecessary API calls before it fails.

plugins/module_utils/orchestrators/loopback_interface.py
```diff
@@ -224,12 +236,136 @@ def query_one(self, model_instance: LoopbackInterfaceModel, **kwargs) -> Respons
...
+            if not isinstance(result, dict):
+                break
+
+            page = result.get("interfaces", []) or []
+            candidates.extend(page)
+
+            meta = result.get("meta") or {}
+            counts = meta.get("counts") or {}
+            try:
+                remaining = int(counts.get("remaining", 0))
```
sivakasi-cisco: If remaining is missing or invalid, this treats it as zero and stops pagination. Could this return incomplete gathered results when the first page is full? Or, may be continue based on the page size or fail instead of returning partial data?

plugins/module_utils/gathered_filter.py
deekpand-cisco: Removed the early return so deduplication now runs unconditionally. The [to_config()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) function is only called when active filters exist to avoid unnecessary work.

plugins/module_utils/nd_state_machine.py
deekpand-cisco: Added  a [validate_gathered_filters()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) function as a pre-flight check in the state machine, called before [query_all()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html). Invalid filters now fail fast without making API calls.

nikhilsrikrishna:
Apologies if this was already covered—I missed the earlier discussions.

I noticed that gathered output is pruned to match `config.options`, which keeps the result reusable as module configuration. However, what if the GET response contains valuable information that is not a writable configuration option?

For example, an update-group response may include `createTime` and `modifyTime`. These fields are useful for visibility but would currently be removed because they are not part of `config.options`.

Could the framework support a declared set of gathered-only/read-only fields? These fields could:

- remain in gathered output;
- have their names and types validated when the gathered result is replayed;
- be removed before calculating diffs or creating write payloads; and
- leave unknown, undeclared fields rejected as they are today.

One option could be a model-level declaration such as `gathered_read_only_properties`, with the state machine preserving those properties for gathered output and stripping them from write processing.

@mikewiebe , could you please take a look at this? Do you see value in supporting something like this, or know of use cases or response fields across other modules that would be useful to retain in gathered output even though they are not writable configuration options?

plugins/module_utils/orchestrators/loopback_interface.py
```diff
@@ -224,15 +249,153 @@ def query_one(self, model_instance: LoopbackInterfaceModel, **kwargs) -> Respons
...
+        policy. System-managed underlay loopbacks are intentionally excluded.
+        """
+        if interface.get("interfaceType") != "loopback":
+            return False
+
+        policy_type = interface.get("configData", {}).get("networkOS", {}).get("policy", {}).get("policyType")
+
+        return policy_type == "loopback"
+
+    def _build_gathered_query_plan(self, gathered_filters: list[dict]) -> dict[str, tuple[str, set[str]]]:
```
AKDRG, line 324: ## Concern: Gathered Query Planning Can Create High REST Fanout

`_build_gathered_query_plan()` can still create high REST fanout for common unscoped gathered filters.

For example:

```yaml
- name: Gather a few loopbacks across the fabric
    fabric_name: prod_fabric
    state: gathered
    config:
      - interface_name: loopback100
      - interface_name: loopback101
      - interface_name: loopback102
```
On a 300-switch fabric, this builds three expressions per switch:
```
300 switches * 3 interface_name filters = 900 GETs
```
Because _MAX_EXPRESSIONS_PER_SWITCH is 3, the plan does not collapse to the base query. A broader per-switch query would need only:
```
300 switches * 1 base managed-loopback query = 300 GETs
```
Correctness would still be preserved because the module already applies local gathered filtering after the server-side query.

Suggested improvement:
1. split gathered filters into fabric-wide and switch-scoped groups first
2. build fabric-wide expressions once
seed all switches once only when fabric-wide filters exist
3. apply switch-scoped expressions only to their target switch
4. collapse to the base expression before expanding across every switch when the fabric-wide expression count exceeds the threshold
5. add a total request cap in addition to _MAX_EXPRESSIONS_PER_SWITCH

This would avoid filters * switches planning work and prevent gathered filtering from producing more REST calls than a broad per-switch query on large fabrics.

deekpand-cisco: Thanks Akshay for the feedback

I've restructured _build_gathered_query_plan() to implement your suggested approach:

- Split filters first — fabric-wide (no switch_ip) and switch-scoped are now classified before any expansion.
- Build fabric-wide expressions once — build_lucene_expressions is called once with all fabric-wide filters instead of per-item inside the loop.
- Switch-scoped expressions only touch their target — they're layered on after fabric-wide assignment, never multiplied.
- Collapse before expanding — the total is projected as len(switch_map) × len(expressions) and checked against the budget before the expansion loop runs.
- Total request cap — added _MAX_TOTAL_REQUESTS = 300 alongside the existing per-switch cap. Both are now defined in NDBaseInterfaceOrchestrator so vpc and ethernet orchestrators inherit them.

plugins/module_utils/models/interfaces/loopback_interface.py
```diff
@@ -443,6 +444,18 @@ class LoopbackInterfaceModel(NDBaseModel):
...
+
+    supports_gathered_filtering: ClassVar[bool] = True
+    gathered_filter_properties: ClassVar[tuple[str, ...]] = (
+        "switch_ip",
+        "interface_name",
+        "config_data.network_os.policy.admin_state",
+        "config_data.network_os.policy.ip",
+        "config_data.network_os.policy.ipv6",
+        "config_data.network_os.policy.vrf",
+    )
```
allenrobel, line 457: **Nested `config_data` gathered filters cannot succeed through the real module.**

`gathered_filter_properties` whitelists `config_data.network_os.policy.{admin_state,ip,ipv6,vrf}` but not `config_data.network_os.network_os_type` or `config_data.network_os.policy.policy_type`, while `get_argument_spec()` still declares both of those `required=True` (lines 552 and 556). Ansible enforces `required` on suboptions whenever the parent dict is supplied, regardless of `state`, so:

- `config: [{config_data: {network_os: {policy: {vrf: mgmt}}}}]` fails at `AnsibleModule()` with "missing required arguments: network_os_type" before any module code runs;
- adding `network_os_type` / `policy_type` to satisfy the argspec then fails in `_reject_unsupported_filter_properties` with "unsupported properties: 'config_data.network_os.network_os_type', 'config_data.network_os.policy.policy_type'".

Either way no `config_data` filter can pass. The unit tests do not see this because they call `filter_gathered_response()` / `validate_gathered_filters()` directly and bypass argspec validation. Suggested fix: either relax `required` on those two suboptions for `state: gathered` (they cannot be conditionally required in an argspec, so they would need `required=False` plus a model-level check for write states), or add them to the whitelist and drop them from the Lucene/local match.

tests/integration/targets/nd_interface_loopback/tasks/gathered.yaml
```diff
@@ -0,0 +1,223 @@
...
+- name: "GATHERED: Filter by nested local criterion"
+  cisco.nd.nd_interface_loopback:
+    output_level: "{{ nd_info.output_level }}"
+    fabric_name: "{{ test_fabric_name }}"
+    state: gathered
+    config:
+      - config_data:
+          network_os:
+            policy:
+              description: "Ansible integration test loopback102"
```
allenrobel, line 94: This scenario (and the ones at lines 104-116, 145-153, and 198-204) filters on `config_data.network_os.policy.description` without `network_os_type` / `policy_type`. As written it fails argspec validation ("missing required arguments: network_os_type"), and `description` is not in `gathered_filter_properties` either, so it would be rejected as unsupported even if the argspec passed. See the comment on `gathered_filter_properties` in `models/interfaces/loopback_interface.py`.

plugins/module_utils/orchestrators/loopback_interface.py
```diff
@@ -120,6 +124,16 @@ class LoopbackInterfaceOrchestrator(NDBaseInterfaceOrchestrator[LoopbackInterfac
...
+    supports_gathered_server_filtering: ClassVar[bool] = True
+    gathered_lucene_spec: ClassVar[GatheredLuceneSpec] = GatheredLuceneSpec(
+        base_terms=(
+            ("interfaceType", "loopback"),
+            ("policyType", "loopback"),
+        ),
+        field_map={
+            ("interface_name",): "interfaceName",
+        },
+    )
```
allenrobel, line 136: **`policyType:loopback` in `base_terms` drops managed loopbacks from `state: gathered`.**

Every gathered Lucene query is `interfaceType:loopback AND policyType:loopback [...]`, so ND only returns policyType `loopback`. But `_is_managed_loopback` (and every other state via `_query_all_for_management_states`) treats the union of `LoopbackPolicyTypeEnum` (`loopback`, `ipfmLoopback`, `mplsLoopback`) and `XeLoopbackPolicyTypeEnum` (`iosXeLoopback`, `iosXeLoopbackShutNoshut`, `iosXeUnderlayLoopback`, `iosXeInternalLoopback`, `csrLoopback`, `csr1kvLoopback`) as owned by this module, and DOCUMENTATION lists all of them. On the merge-base, gathered went through the unfiltered `query_all` and returned them.

Result: `gathered` silently under-reports ipfm, mpls, and every IOS-XE loopback, and a gather-then-`overridden` workflow would remove interfaces the gather never showed. Suggest dropping the `policyType` base term and keeping the policy-type check local in `_is_managed_loopback`, which already runs on every candidate.

plugins/module_utils/orchestrators/loopback_interface.py
```diff
@@ -324,17 +488,87 @@ def query_all(self, model_instance: NDBaseModel | None = None, **kwargs) -> Resp
...
+
+            result = self._request(
+                path=api_endpoint.path,
+                verb=api_endpoint.verb,
+                not_found_ok=True,
+            )
+
+            candidates = []
+            if isinstance(result, dict):
+                candidates = result.get("interfaces", []) or []
```
allenrobel, line 519: **Management-state query bypasses the shared per-switch interface cache.**

On the merge-base `query_all` called `self._switch_interfaces(switch_id)`; this method re-implements the same `EpManageInterfacesListGet` fetch with a direct `_request` and never populates `_switch_interfaces_cache`. The base-class docstring (`base_interface.py:130-154`, from PR #390 "Hoist query_all switch scoping + interface cache into NDBaseInterfaceOrchestrator") says the list GET "is issued at most once per switch per module run; the result is cached so that `query_all` and any other per-interface lookups ... share a single fetch per switch", and CLAUDE.md Performance says "Fetch each controller resource at most once per module run. If two code paths need the same data, cache it on the orchestrator or pass it through."

No duplicate GET happens today because nothing else in the loopback orchestrator reads the cache, but loopback is now the only interface orchestrator outside it (ethernet, svi, and both subinterface orchestrators still call `_switch_interfaces`). Restoring `list(self._switch_interfaces(switch_id).values())` here keeps the contract and removes the duplicated fetch code.

allenrobel (commented):
### Code review

Four inline comments (two functional defects in the loopback gathered path, one cache-contract regression, one test-scenario pointer) plus the two items below.

1. **No closing keyword in the PR body.** The body says "Related to #340". CLAUDE.md's PR template requires "Link issues with `Fixes #xxx`, `Closes #xxx`, or `Resolves #xxx`", and the team convention is that every PR carries one. #340 reads as the umbrella tracking issue for gathered support, so if this PR does not complete it, please open (or link) the issue it does complete and add `Closes #N`.

2. **Rebase note.** The branch's merge-base with `develop` is `576a681d`, before PR #312 (nd_links) landed. #312 already added `secret_field_keys` / `collect_secret_values`, the `gathered` output shape in `NDOutput`, `prune_to_spec` in `utils.py`, and `to_gathered_config()`, which is why `models/base.py`, `nd_output.py`, `nd_state_machine.py`, `utils.py` and their tests conflict. The two implementations differ in semantics: develop masks secret / `no_log` values with `NO_LOG_PLACEHOLDER` and formats gathered items via `to_gathered_config()`, while this branch drops the keys and uses `to_ansible_config()` plus a `gathered_transform` hook. `nd_interface_loopback.py` also conflicts with PR #557's `fail_from_exception()` main() standardization. When rebasing, please keep develop's conventions so the collection-wide behavior stays consistent; the PR-specific pieces (`gathered_filter.py`, the state-machine filter hooks, the loopback Lucene plan, the local-user filter properties) should carry over cleanly.

plugins/module_utils/orchestrators/loopback_interface.py
mikewiebe: Concrete example to make the failure mode and the `overridden` risk explicit:

Assume ND contains these two supported loopback interfaces:

```yaml
- interface_name: loopback10
  policyType: loopback
  ip: 10.10.10.10/32

- interface_name: loopback20
  policyType: ipfmLoopback
  ip: 10.20.20.20/32
```

The user gathers all loopbacks:

```yaml
- name: Gather all loopback interfaces
    fabric_name: fab1
    state: gathered
  register: loopback_result
```

The request sent to ND includes:

```text
interfaceType:loopback AND policyType:loopback
```

Expected result:

```yaml
gathered:
  - interface_name: loopback10
    policy_type: loopback

  - interface_name: loopback20
    policy_type: ipfmLoopback
```

Actual result:

```yaml
gathered:
  - interface_name: loopback10
    policy_type: loopback
```

`loopback20` is absent because the controller was asked to return only `policyType:loopback`. If the user later feeds this incomplete gathered result into `state: overridden`, the module can treat `loopback20` as unwanted and remove it. The server-side query therefore needs to include every supported loopback policy type, or request all loopbacks and retain the existing local ownership check.

plugins/module_utils/orchestrators/loopback_interface.py
```diff
@@ -290,7 +304,157 @@ def query_one(self, model_instance: LoopbackInterfaceModel, **kwargs) -> Respons
...
+                break
+            offset += len(page)
+        else:
+            raise RuntimeError(
+                f"Pagination limit reached ({max_pages} pages, {len(candidates)} candidates collected) "
+                f"for switch '{switch_id}'. Results may be incomplete."
+            )
+        return candidates
+
+    def query_all(
```
mikewiebe, line 473: ### Medium: Read-only gathered tasks fail when deployment is frozen

**Issue**

A fabric can be placed in deployment freeze to prevent configuration changes.
However, `state: gathered` runs the same write-safety check used by create,
update, and delete operations. The module therefore rejects a read-only
gathered task even though it would send only GET requests.

**Evidence**

- [`loopback_interface.py` lines 452-496](https://github.com/CiscoDevNet/ansible-nd/pull/391/files#diff-bfbfe608d8199f5d7c947a889c0a9b2acfa286b562db44bee753e947e6707316R452)
  calls `validate_prerequisites()` before choosing the gathered query path.
- [`base_interface.py` lines 308-321](https://github.com/CiscoDevNet/ansible-nd/pull/391/files#diff-1bdd24be418d1be85dcf48e8bd16da0b84e1c8aaaf0d001caf00e693ed4f6322R308)
  implements that method by calling `validate_for_mutation()`.
- [`fabric_context.py` lines 281-313](https://github.com/CiscoDevNet/ansible-nd/blob/20ce70491ea4ed0920ecba775cbcf1420d113f32/plugins/module_utils/fabric_context.py#L281-L313)
  rejects deployment freeze specifically because configuration cannot be
  changed.

**Practical example**

An operator freezes `fabric1` during maintenance to prevent changes. They then
try to inspect a loopback:

```yaml
- name: Inspect loopback101 during the maintenance freeze
    fabric_name: fabric1
    state: gathered
    config:
      - switch_ip: 192.0.2.11
        interface_name: loopback101
  register: loopback_result
```

This task is read-only and should only retrieve current interface information.

Expected result:

```yaml
changed: false
gathered:
  - switch_ip: 192.0.2.11
    interface_name: loopback101
```

The deployment freeze remains enabled and no configuration is changed.

Actual result: the task fails before requesting the interface information with
an error telling the user to disable deployment freeze. This happens because
the gathered task is checked as though it were going to change the fabric.

A write operation using `state: merged` should still fail during the freeze;
only the read-only `gathered` state should be allowed to continue.

**Performance requirement**

Do not add another controller GET to fix this. The module already has the
fabric information needed to detect deployment freeze. Reuse that existing
fabric context and apply different rules: allow `gathered` to continue, while
write states remain blocked.

**ND/OpenAPI applicability**

**Existing PR overlap**

No matching open issue found. [#340](https://github.com/CiscoDevNet/ansible-nd/issues/340)
was inspected and excluded as an exact owner because it covers consistent
gathered-state behavior generally but not read-only operations during
deployment freeze.

**Impact**

Operators cannot use the module for inventory or troubleshooting while a
freeze is protecting the fabric from changes, which is exactly when read-only
inspection may be most important.

**Suggested fix**

Separate read checks from write checks. Reuse the existing fabric context,
allow `state: gathered` when deployment is frozen, and keep the freeze
rejection for `merged`, `replaced`, `overridden`, and `deleted`. Add a test
proving gathered performs its GET while a write state fails before sending a
change request.

---

plugins/module_utils/gathered_filter.py
```diff
@@ -0,0 +1,302 @@
...
+        if identifier in seen_identifiers:
+            continue
+
+        seen_identifiers.add(identifier)
+        filtered_models.append(model)
+
+    return filtered_models
+
+
+def validate_gathered_filters(
```
mikewiebe, line 245: ### Medium: Invalid gathered-filter values can look like valid empty results

**Issue**

The gathered-filter code checks that the input is structured correctly, the
requested field is allowed, and the value is not empty. It does not run the
field's normal validation. Invalid IP addresses, IPv6 addresses, and choice
values can therefore be accepted as filters and produce an empty result
instead of a clear input error.

**Evidence**

- [`gathered_filter.py` lines 253-302](https://github.com/CiscoDevNet/ansible-nd/pull/391/files#diff-5431ba19ba237f0a5594aed4cc0f44713621b9037c225a340b2f670cbc07ce3aR253)
  validates shape and allowed paths but never invokes Pydantic field
  validation.
- [`loopback_interface.py` lines 502-522](https://github.com/CiscoDevNet/ansible-nd/pull/391/files#diff-2ceae40d02485f26cffe304a7f6cde1570c957a3f0005b9b23da76742d9cec94R502)
  normalizes only `interface_name` casing.
- The ordinary loopback models validate IP, IPv6, enum, and nested policy
  values, but partial filters bypass construction of those models.
- An exact-head helper probe accepted
  `config_data.network_os.policy.ip: not-an-ip` and treated it as a filter that
  matched nothing.

**Practical example**

After the nested-filter input problem from the previous issue is corrected, a
user could accidentally supply an invalid IPv4 prefix:

```yaml
- name: Gather a loopback by IP address
    fabric_name: fabric1
    state: gathered
    config:
      - config_data:
          network_os:
            policy:
              ip: 10.0.0.1/33
  register: loopback_result
```

An IPv4 prefix cannot be greater than `/32`, so `/33` is invalid.

Expected result: the module fails locally before making a GET:

```text
Invalid IPv4 address or prefix: 10.0.0.1/33
```

Actual behavior in the shared filter code: the value is present, is a string,
and is on an allowed filter path, so it is accepted and can produce:

```yaml
changed: false
gathered: []
```

The user cannot tell whether no interface has that address or the supplied
address was invalid. The same problem applies to `ip: not-an-ip`.

**Current reachability**

The conflicting nested-input rules described in the previous issue currently
stop this exact playbook earlier. However, the shared gathered-filter code
already accepts invalid field values. Once the earlier input problem is fixed,
this validation gap becomes directly reachable unless it is fixed too.

**Performance requirement**

Validation should remain local and happen before any controller GET. Fixing
this does not require an additional API request.

**ND/OpenAPI applicability**

**Existing PR overlap**

No matching existing PR comment found. The older
[validate-before-query thread](https://github.com/CiscoDevNet/ansible-nd/pull/391#discussion_r3614646296)
was inspected and excluded as an exact duplicate: it requested that validation
run before controller queries, and that ordering problem is now fixed. This
finding is different because validation runs at the correct time but does not
check whether the field's value is valid.

No matching open issue found. [#340](https://github.com/CiscoDevNet/ansible-nd/issues/340)
was inspected and excluded as an exact owner because it covers gathered
filtering generally but does not define how partial filters should reuse each
field's normal validation.

**Impact**

Typing mistakes can look like legitimate empty inventory. That is especially
risky when gathered output is later used for cleanup or `state: overridden`.

**Suggested fix**

Create a partial-filter validation path that reuses the normal field validators
without requiring a complete writable resource. Validate IP addresses, IPv6
addresses, choice values, lengths, and normalized strings before any GET. Add
tests through the real module input path, not only the lower-level helper.

---

plugins/module_utils/models/interfaces/loopback_interface.py
deekpand-cisco: Thank you for identifying the mismatch between the Ansible argument specification and the partial gathered-filter model.

I resolved this by making `network_os_type` and `policy_type` optional at the shared Ansible argument-spec layer. Partial gathered filters can now pass `AnsibleModule` validation without supplying discriminator fields that are not gathered-filter criteria. For example:

```yaml
config:
  - config_data:
      network_os:
        policy:
          vrf: mgmt
```

Validation for writable nested configuration remains strict. Resources supplied to `merged`, `replaced`, or `overridden` continue through the Pydantic discriminated models, which require the applicable `network_os_type` and `policy_type`.

I added module-level coverage using the real loopback argument specification, together with state-machine tests confirming that partial-filter validation and normalization complete before any controller query.

tests/integration/targets/nd_interface_loopback/tasks/gathered.yaml
deekpand-cisco: Thank you, Allen. I have updated the affected integration scenarios to use the supported nested `ip` criterion instead of `description`.

The updated scenarios cover nested filtering, combined server-side and local criteria, and overlapping-filter deduplication. They intentionally omit `network_os_type` and `policy_type`, so they now exercise the partial nested-filter path through the real module argument specification.

`description` remains outside `gathered_filter_properties`; if supplied as a gathered criterion, it is rejected as unsupported before any controller request.

plugins/module_utils/orchestrators/loopback_interface.py
deekpand-cisco: Thank you, Allen and Mike. 

I removed `policyType:loopback` from the gathered Lucene base terms. The server-side query now requests `interfaceType:loopback` candidates without restricting `policyType`, and `_is_managed_loopback()` remains the authoritative local ownership check.

With this change, the example now returns both `loopback10` (`policyType: loopback`) and `loopback20` (`policyType: ipfmLoopback`). The same ownership check retains all module-managed NX-OS and IOS-XE loopback policy types while continuing to exclude system-managed `underlayLoopback`, `userDefined`, non-loopback, and malformed records.

I added regression coverage confirming that the generated request contains no `policyType` restriction, all nine managed policy types are retained, and unsupported candidates are excluded locally. This prevents gathered output from silently under-reporting resources that could later be supplied to `state: overridden`.

plugins/module_utils/orchestrators/loopback_interface.py
deekpand-cisco: Thank you, Allen. I restored the management-state query path to the shared per-switch interface cache.

`_query_all_for_management_states()` now retrieves candidates through:

`list(self._switch_interfaces(switch_id).values())`

This removes the duplicated interface-list fetch logic and restores the `NDBaseInterfaceOrchestrator` contract that the interface inventory is fetched at most once per switch per orchestrator run.

The gathered path remains separate because it intentionally performs filtered and paginated Lucene queries. The normal management states now consistently reuse `_switch_interfaces_cache`, matching the behavior of the other interface orchestrators.

plugins/module_utils/orchestrators/loopback_interface.py
deekpand-cisco: Thank you, Mike. I separated read prerequisite validation from mutation prerequisite validation at the shared orchestrator level.

`state: gathered` is now classified as a read-only operation. `NDBaseInterfaceOrchestrator.validate_prerequisites()` calls `FabricContext.validate_for_read()` for gathered operations, which verifies that the fabric exists and is owned by the targeted controller without rejecting deployment freeze.

Mutation states continue through `validate_for_mutation()`, so `merged`, `replaced`, `overridden`, and `deleted` remain blocked while the fabric is frozen.

The implementation reuses the existing cached fabric summary and does not introduce an additional controller GET.

I added regression coverage confirming that:

- a gathered operation proceeds to the interface GET during deployment freeze;
- the same fabric context rejects a subsequent mutation state;
- repeated read validation and mutation validation reuse the cached fabric summary rather than querying it again.

This allows inventory and troubleshooting reads during a freeze while preserving the existing protection against configuration changes.

plugins/module_utils/gathered_filter.py
deekpand-cisco: Thank you, Mike. I added a dedicated partial-policy model for gathered loopback filters so supported filter values pass through the applicable Pydantic field constraints without requiring a complete writable resource.

Validation now covers:

- malformed IPv4 and IPv6 values;
- IPv4 prefixes greater than `/32`;
- IPv6 prefixes greater than `/128`;
- empty and overlength VRF values;
- the supported Boolean field type.

Valid IPv4 and IPv6 CIDR criteria are normalized to host form so they match the normalized values returned by gathered candidates. Fields that are not supported as gathered criteria, including discriminator and other choice-constrained fields, remain rejected by the gathered-property allowlist.

This validation runs in `NDStateMachine` before `query_all()`, so invalid values fail locally instead of producing an indistinguishable empty gathered result.

I added module-level coverage using the real loopback argument specification, together with state-machine tests that verify invalid values are rejected and `query_all()` is never called. The tested cases include malformed IPv4 and IPv6 values, invalid prefix lengths, empty and overlength VRFs, and valid CIDR normalization.

deekpand-cisco:
> **No closing keyword in the PR body.**
>
> **Rebase note.**

Thank you, Allen, for the detailed review.

I have rebased the branch onto the latest `develop`. Gathered output now continues to use `to_gathered_config()` and `prune_to_spec()`, including the collection-wide output shape. Secret and `no_log` values retain the existing `NO_LOG_PLACEHOLDER`, `secret_field_keys`, and `collect_secret_values` behavior.

I did not retain the generic `gathered_transform` hook, and `nd_interface_loopback` continues to use the standardized `fail_from_exception()` path introduced by #557. The PR-specific gathered-filter implementation, state-machine filter handling, loopback query planning, and local-user filter properties were preserved through the rebase.

Regarding the closing keyword, this PR doesn't completely covers issue #340. I’ll check once with Sivakmai to have a focused issue created for this PR’s scope, and then update the PR description with the corresponding Closes #N reference.

Thank you so much!

plugins/module_utils/models/interfaces/loopback_interface.py
allenrobel: Thanks, verified on the head commit ("Gathered filtering: fix validation, query scope, and read safety"): partial gathered filters now pass the real argument spec, and the new module-level tests close the gap that hid this.

One non-blocking follow-up. With `network_os_type` and `policy_type` now `required=False`, a merged/replaced/overridden config that omits either one still fails before any request, but the message is now Pydantic's discriminator error rather than Ansible's:

```text
config_data.network_os
  Unable to extract tag using discriminator 'network_os_type' | 'networkOSType' [type=union_tag_not_found, ...]
```

Previously the user saw `missing required arguments: network_os_type found in config -> config_data -> network_os`. Strictness is intact, so I'm fine landing this as-is. A small model-level check on the write path that raises a plain "network_os_type is required for state merged/replaced/overridden" message would restore the old clarity. Happy for that to be tracked separately rather than on this PR.

allenrobel (approved):
Verified the 2026-09-22 follow-up commit against every open thread from Mike and me: partial nested gathered filters pass the real argument spec, `policyType:loopback` is gone from the base terms so all nine managed policy types are gathered, the management-state query reuses the per-switch interface cache, gathered reads are permitted during deployment freeze, and invalid filter values now fail locally before any GET.

Full unit suite passes on the head commit (5097 tests) and CI is green. The one remaining note (discriminator error text on write states) is cosmetic and non-blocking; see the thread on `models/interfaces/loopback_interface.py`.

PR #395, opened 2026-07-08
ND Manage Networks | Parity with VRF Changes

AKDRG:
- Removed old NDModule runtime usage from Networks.
- Moved Network fabric resolution into NetworkWorkflowCoordinator, matching VRF.
- Updated Network resolver/tests to use Gen3 RestSend.
- Removed stale argspec wrapper helpers from nd_manage_networks.py.
- Cleaned legacy Optional / Union typing in Network orchestrator files.
- Fixed Network attachment missing-resource fallback to avoid per-network retry calls.
- Updated tests to enforce the new Gen3 wrapper/resolver flow.

### DIFF
--- a/plugins/module_utils/orchestrators/fabric_resolver.py
+++ b/plugins/module_utils/orchestrators/fabric_resolver.py
@@ -58,6 +58,14 @@ def _response_message(data: Any) -> str:
     return str(data)
 
 
+class FabricResolverNotFound(Exception):
+    """Raised when a resolver lookup receives an explicit controller 404."""
+
+
+class FabricResolverRequestError(Exception):
+    """Raised when a required resolver request fails for a non-404 reason."""
+
+
 def _is_error_response(response: Any) -> bool:
     """Return True when a wrapped ND response represents an HTTP/API error."""
     if not isinstance(response, dict):
@@ -180,12 +188,18 @@ class FabricResolverBase:
         """Send a resolver request through RestSend."""
         self._rest_send.path = path
         self._rest_send.verb = HttpVerbEnum(method.upper())
-        self._rest_send.commit()
+        try:
+            self._rest_send.commit()
+        except Exception as exc:
+            raise FabricResolverRequestError(f"{method.upper()} {path} failed before controller response: {exc}") from exc
 
         if ignore_not_found_error and self._rest_send.return_code == 404:
             return {}
+        if self._rest_send.return_code == 404:
+            raise FabricResolverNotFound(f"{method.upper()} {path} returned 404: {self._rest_send.error_summary}")
         if not self._rest_send.success:
-            raise Exception(f"Request failed {self._rest_send.error_summary}")
+            detail = _response_message(_response_data(self._rest_send.response_current))
+            raise FabricResolverRequestError(f"{method.upper()} {path} failed: {self._rest_send.error_summary}: {detail}")
         return self._rest_send.response_current
 
     def _controller_version(self) -> str:
@@ -388,7 +402,7 @@ class FabricResolverBase:
         enriched = dict(fabric_data or {})
         try:
             details = self._fetch_manage_fabric_details(self.fabric_name, enriched.get("clusterName"))
-        except Exception:
+        except FabricResolverNotFound:
             details = {}
 
         management = details.get("management") if isinstance(details, dict) else {}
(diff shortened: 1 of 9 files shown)


plugins/module_utils/orchestrators/network_workflow_coordinator.py
```diff
@@ -130,6 +138,41 @@ def run(self) -> dict[str, Any]:
...
+        rest_send.sender = sender
+        rest_send.response_handler = ResponseHandler()
+        if timeout is not None:
+            rest_send.timeout = timeout
+        if send_interval is not None:
+            rest_send.send_interval = send_interval
+        return rest_send
+
+    def _resolve_strategy(self, module_args: dict) -> BaseNetworkStrategy:
+        """Resolve fabric topology using the same Gen-3 REST runtime as the workflow."""
```
allenrobel: The docstring says this resolves "using **the same** Gen-3 REST runtime as the workflow," but the next line builds a distinct `RestSend` via `_new_rest_send(params=module_args, check_mode=False, timeout=1, send_interval=1)`. The workflow's own runtimes (`_new_state_machine` / `_new_network_orchestrator`) use `check_mode=self.module.check_mode` and the default timeout/interval, so it isn't the same runtime. The forced `check_mode=False` plus `timeout=1`/`send_interval=1` look like a deliberate single-attempt probe (avoiding the long retry loop on expected 404s during topology detection) — which is exactly the distinction the "same runtime" wording hides. Consider rewording to note the resolver uses a short-timeout, non-check-mode probe variant.

allenrobel (commented):
### Code review

The migration to the Gen-3 `RestSend` runtime looks clean — the `Optional`/`Union` → `X | None` modernization is complete, the resolver's `_manage_fabric_details_cache` and the unscoped-read fallback in `current_attachment_details_ignore_missing` both cut request counts, and I found no correctness bug in the changed lines. One minor docstring-accuracy nit inline.

plugins/module_utils/orchestrators/network_workflow_coordinator.py
AKDRG: Fixed

allenrobel (approved):
LGTM

plugins/module_utils/orchestrators/fabric_resolver.py
```diff
@@ -174,6 +176,22 @@ def _trace(self, event: str, **details: Any) -> None:
         entry.update(details)
         self._workflow_trace.append(entry)
 
+    def _request(self, path: str, method: str = "GET", ignore_not_found_error: bool = False) -> Any:
```
mikewiebe, line 187: ### Medium: Controller failures are rewritten as fabric-topology errors

**Issue**  
The new resolver adapter raises an ordinary `Exception` for an unsuccessful `RestSend`, but `_resolve_fabric_type()` catches every `Exception` from the initial Manage fabric GET and replaces it with an empty result. The eventual error says the fabric could not be resolved, even when ND actually returned an authentication, authorization, transport, or server error.

**Evidence**  
- [`fabric_resolver.py` lines 179-188](https://github.com/CiscoDevNet/ansible-nd/pull/395/files#diff-9d2d181e8847db1dc1d91c812759ee6347670ef02ea57d665e7daa02649c1d54R179-R188) converts every unsuccessful non-404 response into a generic exception.
- [`fabric_resolver.py` lines 415-420](https://github.com/CiscoDevNet/ansible-nd/pull/395/files#diff-9d2d181e8847db1dc1d91c812759ee6347670ef02ea57d665e7daa02649c1d54R415-R420) discards that exception and continues as if no fabric details were returned; line 451 emits the topology message.
- The replaced `NDModule.request()` path called `fail_json()` with the controller's decoded error, so a 401/403/500 did not fall through to topology classification.
- A focused probe using the real `RestSend` and `ResponseHandler` supplied HTTP 500 with `{"code": 500, "message": "controller unavailable"}`. `resolve()` returned `ValueError: Fabric 'fab1' could not be resolved from ND Manage fabric topology. Verify the fabric name and ND connectivity.` The controller message was lost.
- The changed tests mock `RestSend` in the resolver tests and mock the resolver in the coordinator tests, so no test covers this real boundary.

**Existing PR overlap**  

No matching open issue found.

**Impact**  
Users receive an invalid-fabric diagnosis for valid fabrics when credentials expire, permissions are insufficient, ND is unavailable, or ND returns a 5xx response. The task still fails, but it hides the actionable cause and sends troubleshooting in the wrong direction.

**Suggested fix**  
Suppress only the explicitly expected 404/not-found condition. Preserve and propagate all other `RestSend` failures, ideally with a typed not-found exception or result so topology fallback does not rely on broad exception handling. Add a coordinator-to-resolver test using real `RestSend` and `ResponseHandler` with only the transport mocked; cover 401, 403, 500, transport failure, and a separate 404 fabric-not-found case.

---

AKDRG: Fixed.

mikewiebe (approved):

Issue #397, opened 2026-07-13
Consolidate per-item 207 Multi-Status detection: retire remaining bespoke orchestrator detectors onto NdV1Strategy


allenrobel:
## Background

Follow-up to #295. The centralized fix for #295 landed on branch `nd_207_multistatus_per_item_status`:
`NdV1Strategy.is_success()` now returns False (and `extract_error_message()` aggregates the failing
items) for any success-code response whose `DATA.results[]` or `DATA.switchIds[]` contains an item with
`status` in {`failed`, `failure`, `error`} (case-insensitive, whitespace-tolerant, scanned regardless of
HTTP code). As part of that work the two subinterface detectors were retired onto the baseline and their
`TODO(4.2.1)` markers removed.

An inventory found **11 orchestrators** that had independently implemented their own per-item / 207
"multi-status" failure detection. The two subinterface ones are done. This issue tracks reviewing and
(where safe) retiring the **remaining 9** so the RestSend layer is the single source of truth.

## Why this needs per-orchestrator review (not a blanket sweep)

1. **Six distinct envelope keys** — the baseline handles only `results` and `switchIds`. Others in use:
   `items` (maintenance_mode), `attachUpdateGroups`/`detachUpdateGroups` (fabric_update_group),
   `policyGroups` (policy_group create).
2. **Two incompatible failure semantics**:
   - *Denylist* (`failed`/`failure`/`error` = failure) — matches the baseline: acl, maintenance_mode,
     vrfs, vrf_attachment_manager.
   - *Allowlist* (anything `!= "success"` = failure, incl. `warning`, `notexecuted`, missing status):
     fabric_update_group, network_attachment_manager, networks, policy_group. The baseline does **not**
     catch these softer cases, so those detectors must stay (or the baseline must grow an allowlist mode).
3. **Not all detectors raise** — `vrf_attachment_manager` calls `fail_json`; `policy_group.delete_bulk`
   markDelete **logs and falls back to a direct DELETE**. Retiring these changes control flow, not just messages.
4. **L3Out** additionally keys off per-item `statusCode >= 400` and a retry-on-"not found" path.

The baseline and the bespoke detectors currently **compose correctly**: the baseline preempts a detector
only for the hard-failure literals; softer/allowlist cases and non-dict/bare-list bodies still fall through
to the bespoke detector (e.g. policy_group test `00830`, a bare-list body, still uses the bespoke path).

## Tests already updated to the centralized message (behavior preserved)

[...]

Issue #396, opened 2026-07-10
nd_rest currently does not support http 207 as a success code


bdosenov:
### Ansible Version

```shell
ansible [core 2.21.1]
```

### Ansible Collection Versions

```shell
# /home/bdosenov/.ansible/collections/ansible_collections
Collection Version
---------- -------
cisco.nd   1.5.0
```

### Ansible Collection Module

```shell
cisco.nd.nd_rest
```

### Cisco Nexus Dashboard Version

```shell
4.2.1.10
```

### Cisco NX-OS Version

```shell

```

### Cisco ACI Version

```shell

```

### Expected Behavior

```shell
http 207 response with a JSON payload should be treated as a success, as there are multiple ND REST endpoints that return 207 rather than 200, 201 202 or 204 in Success-Case. Example:
/api/v1/manage/fabrics/{{ fabric-name }}/softwareUpdatePlan/actions/attachGroup
/api/v1/manage/fabrics/{{ fabric-name }}/softwareUpdatePlan/actions/detachGroup 

http 207
{"attachUpdateGroups":[{"status":"success","updateGroupName":"VXLAN_TRAINING_SITE_B_N9K_leaf_VPC_Primary","warningMessage":"Update group attached successful"}]}

should be treated as a success and the task should run through.
```

### Actual Behavior

```shell
However, nd_rest chokes on this with error: 
fatal: [nd-1]: FAILED! => {
    "changed": false,
    "msg": "Task failed: Module failed: object to sanitize can only be of type list or dict. Got <class 'NoneType'>"
}

because nd.py line 300 looks like this: 
        # 200: OK, 201: Created, 202: Accepted, 204: No Content
        if self.status in (200, 201, 202, 204):
            if output_format == "raw":
                return info.get("raw")
            return info.get("body")
```

### Ansible Playbook

```yaml
---
- name: test playbook 
  hosts: nexus_dashboard
  gather_facts: false

  tasks:
    - name: Select the target fabric
      ansible.builtin.set_fact:
        target_fabric: "{{ fabrics | selectattr('name', 'equalto', fabric) | first }}"
      tags: uri

    - name: Select the target update group within the target fabric
      ansible.builtin.set_fact:
        target_update_group: "{{ target_fabric.update_groups | selectattr('name', 'equalto', update_group) | first }}"
      tags: uri

    - name: Detach Switch from to Upgrade Group
      cisco.nd.nd_rest:
        method: post
[...]

Issue #389, opened 2026-07-02
NDStateMachine failure output misreports changed/after when an API mutation fails mid-run


allenrobel:
## Summary

`NDStateMachine` aliases its live `existing` collection into `NDOutput` as `after` at init, and `NDOutput.format()` derives `changed` by diffing `before` vs `after`. Collection mutations are not synchronized with API-call success, so when a mutating API call fails (or is swallowed by `ignore_errors=true`), the failed/ignored output misrepresents what actually happened on ND. The policy-preflight instance of this was fixed in PR #362 (commit "Run policy preflight before capability preflight and before existing-collection mutation"); this issue covers the remaining API-failure paths.

## Failure modes

1. **Create/update over-report** — `_manage_create_update_state()` applies `merge()`/`replace()`/`add()` to `existing` while categorizing, before the create/update API calls execute. A mid-run API failure raises `NDStateMachineError`; the module's handler calls `fail_json(**output.format())`, which reports `changed=True` and shows the failed item (plus any items never attempted) in `after`.
2. **Delete under-report** — `_delete_items()` executes all deletes first and removes them from `existing` (`delete_many`) only after the loop completes. A mid-loop failure means deletes that already succeeded on ND are still present in `after`, with `changed=False`.
3. **`ignore_errors=true`** — `_execute_operation()` swallows the exception, but the pre-applied collection mutation for the failed item remains, so even the "successful" output over-reports.

## Scope

All 17 NDStateMachine-driven (Gen-3) modules on develop: the 11 `nd_interface_*` modules, `nd_manage_fabric_{ebgp_vxlan,ibgp_vxlan,external}`, `nd_fabric_update_group`, `nd_local_user`, `nd_maintenance_mode`, `nd_manage_vpc_pair`.

## Non-goals / constraints

- Check-mode behavior is intentional (predicted state) and must be preserved: in check mode `_execute_operation()` skips the API call and returns `None`, and the collection mutation legitimately represents the predicted outcome.
- The preflight-failure path is already fixed (PR #362).

## Suggested direction

[...]

Issue #385, opened 2026-07-02
Tracking: NDStateMachine final-state fidelity and optional post-write verification


gmicol:
## Summary

`NDStateMachine` currently reports a predicted `after` state based on local reconciliation, not the final state refreshed from Nexus Dashboard after all write operations complete. This makes module output look authoritative even when the controller normalizes values, delays convergence, rejects part of an action, or exposes additional read-only/current fields only after a follow-up GET.

`nd_manage_vpc_pair` already has a local `verify` argument for post-write refresh. The same capability should be added to the generic Orchestrator framework so existing and future ND modules can opt into controller-verified final state.

## Current Behavior

The framework initializes state with `query_all()`, stores it as `before`, copies it into `existing`, and assigns `existing` to output. During `merged`, `replaced`, `overridden`, and `deleted`, `existing` is updated locally after the framework decides what it expects ND to contain.

For example:
- `NDStateMachine.__init__()` reads controller state once and seeds `before` / `existing`.
- `_manage_create_update_state()` updates `existing` before/around write execution.
- `_delete_items()` removes objects from `existing` after delete calls.
- `output.assign(after=self.existing)` reports the local collection as `after`.

The generic state machine does not perform a final `query_all()` after writes. Some modules also execute extra operations after `manage_state()`, such as interface `remove_pending()` / `deploy_pending()`, so a refresh inside `manage_state()` alone would still be too early for those modules.

`nd_manage_vpc_pair` has a module-local implementation that refreshes final state when `verify.enabled=true`, with retries and timeout controls. That pattern is not available to the generic Orchestrator modules.

## Problem

Users cannot choose whether the returned `after` state is:
- a predicted local result, useful for speed and check-mode style previews, or
- the actual controller state after create/update/delete/action/deploy operations.

This is especially risky for APIs with eventual consistency, post-write normalization, action endpoints, bulk operations, or deploy/save workflows.

## Suggested Fix

Add a generic verification/finalization path to the Orchestrator framework.

Recommended design:

1. Add shared `verify` argument support for Orchestrator-based modules:
   - `verify.enabled`
   - `verify.retries`
   - `verify.timeout`
   - possibly `verify.delay` if retry delay should be user-tunable

2. Add `NDStateMachine.refresh_after_state()`:
   - skipped in check mode
   - skipped when `verify.enabled=false`
   - skipped for pure no-op runs unless explicitly requested
   - retries controller refresh
[...]

PR #398, opened 2026-07-13
Detect Multi-status per-item failures in NdV1Strategy (#295)

### LINKED ISSUE #295: RestSend treats HTTP 207 Multi-Status as success even when results[].status is 'failed' or 'error'
## Summary

`plugins/module_utils/rest/response_handler_nd.py` classifies HTTP 207 as success
(`RETURN_CODE in (200, 201, 202, 204, 207)`), so per-item failures inside the response
body are silently ignored. ND uses 207 as the wrapper status for batch interface
operations regardless of whether individual items succeeded or failed.

## Reproduction (ND 4.2.1, lab SITE1, 2026-05-27)

POST `/api/v1/manage/fabrics/SITE1/switches/<sn>/interfaces` with `interfaceType: "subInterface"`
targeting a trunk-mode parent returns:

```
HTTP 207
{"results": [{"name": "Port-channel1.999",
              "status": "failed",
              "message": "Sub-interface can be created only on routed physical or
                          port-channel interfaces (discovered mode is not routed)
                          for interface Port-channel1.999[subInterface]"}]}
```

`RestSend` / `ResponseHandler` return this body up to the orchestrator without raising.
Any module hitting batch interface or interface-action endpoints (subinterface, breakout,
fabric_update_group, and likely vpc_pair too) will see "success" while ND actually
rejected every item in the batch — masking real failures and breaking idempotency.

## Status string varies by endpoint

- Subinterface POST  → `status: "failed"` (verified above)
- Breakout action    → `status: "error"`  (per existing memory `project_breakout_design.md`)

The fix needs to handle both literals.

## Suggested fix

In `response_handler_nd.py` (or the per-verb response strategies under
`plugins/module_utils/rest/response_strategies/`), treat 207 as success **only when no
item in `results[]` (or equivalent envelope) carries `status in ("failed", "error")`**.
When any item fails, propagate an error so `RestSend.commit()` raises and the
orchestrator's `_request()` surfaces a clean failure message to the user.

## Test plan

* [ ] Add a unit test: 207 with all `results[].status == "ok"` → success
* [ ] Add a unit test: 207 with mixed `results[]` (some ok, some `failed`) → raises with the failure messages aggregated
* [ ] Add a unit test: 207 with `results[].status == "error"` → raises (covers breakout style)
* [ ] Integration test: attempt to POST `nd_interface_subinterface_managed` against a trunk-mode parent and assert the module fails with the ND parent-mode message rather than returning success

## Context

Surfaced while scaffolding `nd_interface_subinterface_managed` against the ND 4.2.1 lab.
Same response shape as `project_breakout_design.md` flagged earlier, but with a different
status literal — so a fix must handle both.

allenrobel:
## Related Issue(s)

Fixes #295
Relates to #397 (follow-up: consolidate the remaining orchestrators' bespoke 207 detectors)

## Proposed Changes

ND reports per-item outcomes for batch operations in `DATA.results[]` / `DATA.switchIds[]` items carrying
`status: success|failed|failure|error`. It sends these bodies on HTTP 207 (Multi-Status) **and, for some
endpoints, on a plain HTTP 200** — the L3Out batch POST returns 200 with per-item failures inside.
`NdV1Strategy` previously classified any success-code response as success, so a per-item failure (e.g. one
switch fails while its peer succeeds, or a subinterface POST rejected because the parent is not routed) was
silently reported as `success/changed` — masking failures and breaking idempotency.

**Central fix (`NdV1Strategy`):**

- `is_success()` now scans the body of **every success-code response (200, 201, 202, 204, 207 — not only
  207)** and returns `False` when any per-item `status` in `results[]`/`switchIds[]` is
  `failed`/`failure`/`error` (case-insensitive, whitespace-tolerant).
- `extract_error_message()` aggregates the failing items as `label: message`, so the user sees which items
  failed and why. Dict-error handling factored into a helper to keep branch count in bounds.
- 12 new unit tests cover both envelope shapes, all three literals, whitespace/mixed-case, mixed
  partial-failure, a None-status control, and message aggregation.

**Subinterface retirement:** the managed and unmanaged subinterface orchestrators each carried their own
`_raise_on_multi_status_failures` guard (with `TODO(4.2.1)` markers pointing at #295). These are now
redundant — `_request()` raises via the strategy — and are removed along with their markers. Redundant
direct-call unit tests are dropped (covered centrally); the end-to-end 207 tests are kept and re-pointed to
the centralized message.

**Downstream test updates:** five tests in `manage_l3out`, `manage_acl`, and `manage_policy_group` asserted a
bespoke per-item message that the centralized path now preempts for hard-failure literals. They are updated
to the centralized message. **Behavior (raise) is preserved**, and the bespoke detectors remain in place —
they still cover the allowlist / soft-status cases (`warning`, `notexecuted`, `!= "success"`) and alternate
envelope shapes the baseline does not. Retiring those is tracked in #397.

### Affected orchestrator owners (FYI — your bespoke 207 detectors compose with this change; consolidation tracked in #397)

- @skaszlik — `manage_acl`, `manage_l3out`
- @nikhilsrikrishna — `manage_policy_group`
- @AKDRG — `networks` / `network_attachment_manager`, `vrfs` / `vrf_attachment_manager`
- @allenrobel — `subinterface_*` (retired here), `maintenance_mode`, `fabric_update_group`

## Test Notes

- Full unit suite green inside the `nd-dev` container: `ndpytest tests/unit/` — **3152 passed**.
- `nd_v1_strategy.py`: pylint 10.00/10, mypy clean, black/isort clean.
- Subinterface orchestrators: no new pylint/mypy findings vs `develop` (verified against the pristine
  baseline; pylint score improved after removing the duplicated guards).
- Verified the central fix **composes** with the remaining bespoke detectors: `manage_policy_group` test
  `00830` (bare-list body) still routes through its own detector, while `00820` (dict/`switchIds` body) is now
  caught centrally.
- Follow-up commit "Clarify per-item failure detection docs (#295)" addresses the Copilot review: docstrings
  and comments had described the detection as 207-only and omitted the `failure` literal, i.e. narrower than
  the code actually implements. Comments/docstrings only — no behavior change; suite re-run green.
- Follow-up commit "Harden per-item failure extraction from review (#295)" addresses @akinross' review.
  Adds two constant-driven helpers (`_get_typed_value`, `_first_non_empty`) and fixes three defects, two of
  which are **pre-existing on `develop`** — this PR only moved them into `_extract_dict_error_message`, and
  they sit on the path that reports an error to the user:
  - `_format_multistatus_failure` gated its label on `is not None`, so an item carrying `name: ""` rendered
    as `": <message>"`.
  - `"messages" in data_dict and len(data_dict.get("messages", []))` raised `TypeError` when ND sent the key
    with an explicit `null` value (same for `errors`) — despite the method's `## Raises: None` contract.
  - `is_success()` classifies `DATA.error` as a failure, but the extractor had no matching branch, so the
    error ND actually sent was replaced by the generic `Request failed with status <code>` fallback.

  Per-item message keys are now the `_MULTISTATUS_ITEM_MESSAGE_KEYS` constant (`message`, `warningMessage`,
  `status`) — ND is not consistent across endpoints. Note the `warningMessage` items on `fabric_update_group`
  ride an `attachUpdateGroups` envelope that is **not** in `_MULTISTATUS_ITEM_KEYS`, so they still do not
  reach the central path; the message-key seam is in place, but the envelope-key half remains #397's job.

  Six new tests (`01300`–`01350`). Full unit suite green — **3158 passed**; `nd_v1_strategy.py` pylint
  10.00/10, mypy clean, black/isort clean.
- Follow-up commit "Scan links[] Multi-Status envelope; label by linkId (PR #398 review)" addresses @shrsr's
  request for `nd_manage_links`. Adds `links` to `_MULTISTATUS_ITEM_KEYS` and `linkId` to
  `_MULTISTATUS_ITEM_LABEL_KEYS` (ahead of the generic `id`) — bulk link create (`POST /links`) and bulk link
  delete (`POST /linkActions/remove`) return HTTP 207 with `{"links": [{"linkId", "message", "status"}]}`,
  status `success|failure` (verified against the ND 4.2.1 OpenAPI). The `GET /links` list body rides the same
  envelope, but its link objects carry no top-level `status` key, so the literal-gated scan cannot
  false-positive on queries. Four new tests (`01360`–`01390`), including the OpenAPI's own `linkId: ""`
  failure example. Full unit suite green — **3182 passed**; pylint 10.00/10, mypy clean, black/isort clean.

4.2.1

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

https://claude.ai/code/session_01NAFsW8ALcvgZBSbgKjRPxn

### DIFF
--- a/plugins/module_utils/rest/response_strategies/nd_v1_strategy.py
+++ b/plugins/module_utils/rest/response_strategies/nd_v1_strategy.py
@@ -37,6 +37,9 @@ T = TypeVar("T")
 # interface / switchActions/deploy), "error" (breakout), and "failure" (acl / maintenance_mode).
 _MULTISTATUS_FAILURE_STATUSES = frozenset({"failed", "failure", "error"})
 
+# Per-item status literal that marks a successful item in a Multi-Status body.
+_MULTISTATUS_SUCCESS_STATUSES = frozenset({"success"})
+
 # DATA envelope keys whose items carry a per-item `status`. Three known shapes:
 # - `results`   -> batch interface POST / breakout action
 # - `switchIds` -> switchActions/deploy (per-switch outcome)
@@ -120,6 +123,39 @@ def _first_non_empty(mapping: Mapping[str, Any], keys: Iterable[str]) -> str | N
     return None
 
 
+def _multistatus_items_with_status(response: dict, statuses: frozenset[str]) -> list[dict[str, Any]]:
+    """
+    # Summary
+
+    Return the per-item entries in a Multi-Status body whose `status` is in `statuses`.
+
+    ## Description
+
+    Scans the known ND Multi-Status envelope arrays (`DATA.results[]`, `DATA.switchIds[]`, and `DATA.links[]`) and returns every item whose `status`
+    matches one of `statuses` (case-insensitive, whitespace-tolerant). Returns an empty list when `DATA` is not a dict, none of the arrays is
+    present, or no item matches.
+
+    ## Parameters
+
+    - response: Response dict with keys RETURN_CODE, MESSAGE, DATA, etc.
+    - statuses: The status literals to match (already lowercase)
+
+    ## Returns
+
+    - List of matching item dicts (empty list when none match)
+
+    ## Raises
+
+    None
+    """
+    matched: list[dict[str, Any]] = []
+    data = _get_typed_value(response, "DATA", dict, {})
+    for key in _MULTISTATUS_ITEM_KEYS:
+        items = _get_typed_value(data, key, list, [])
+        matched.extend(item for item in items if isinstance(item, dict) and str(item.get("status") or "").strip().lower() in statuses)
+    return matched
+
+
 def _failed_multistatus_items(response: dict) -> list[dict[str, Any]]:
     """
     # Summary
@@ -128,9 +164,8 @@ def _failed_multistatus_items(response: dict) -> list[dict[str, Any]]:
 
     ## Description
 
-    Scans the known ND Multi-Status envelope arrays (`DATA.results[]`, `DATA.switchIds[]`, and `DATA.links[]`) and returns every item whose `status`
-    is a failure literal (`failed`/`failure`/`error`, case-insensitive, whitespace-tolerant). Returns an empty list when `DATA` is not a dict, none of
-    the arrays is present, or every item succeeded.
+    Thin wrapper around `_multistatus_items_with_status` matching the failure literals in `_MULTISTATUS_FAILURE_STATUSES`
+    (`failed`/`failure`/`error`).
 
     ## Parameters
 
@@ -144,12 +179,7 @@ def _failed_multistatus_items(response: dict) -> list[dict[str, Any]]:
 
     None
     """
-    failed: list[dict[str, Any]] = []
-    data = _get_typed_value(response, "DATA", dict, {})
-    for key in _MULTISTATUS_ITEM_KEYS:
-        items = _get_typed_value(data, key, list, [])
-        failed.extend(item for item in items if isinstance(item, dict) and str(item.get("status") or "").strip().lower() in _MULTISTATUS_FAILURE_STATUSES)
-    return failed
+    return _multistatus_items_with_status(response, _MULTISTATUS_FAILURE_STATUSES)
 
 
 class NdV1Strategy:
@@ -355,6 +385,40 @@ class NdV1Strategy:
         if modified is None:
             return True
         return str(modified).lower() != "false"
+
+    def is_changed_on_failure(self, response: dict) -> bool:
+        """
+        # Summary
+
+        Report whether a failed mutation nevertheless changed controller state.
+
+        ## Description
+
+        Called for POST/PUT/DELETE responses after `is_success` has returned `False`. A Multi-Status body can mix successful and failed per-item
+        outcomes, so an aggregate failure does not imply nothing changed. The `modified` response header, when present and parseable, is
+        authoritative (mirroring `is_changed`); otherwise any per-item `status` of `success` in the recognized envelope arrays (`DATA.results[]`,
+        `DATA.switchIds[]`, `DATA.links[]`) reports `True`. Non-itemized failures (e.g. `DATA.error` on a 200) report `False`, preserving the
+        conservative historical default.
+
+        ## Parameters
+
+        - response: Response dict with keys RETURN_CODE, MESSAGE, DATA, and any HTTP response headers (lowercased) forwarded by the HttpAPI plugin.
+
+        ## Returns
+
+        - True if the `modified` header is `"true"`, or (header absent or unparseable) at least one per-item status is `success`
+        - False if the `modified` header is `"false"`, or no per-item status is `success`
+
+        ## Raises
+
+        None
+        """
+        modified = str(response.get("modified") or "").strip().lower()
+        if modified == "true":
+            return True
+        if modified == "false":
+            return False
+        return len(_multistatus_items_with_status(response, _MULTISTATUS_SUCCESS_STATUSES)) > 0
 
     def extract_error_message(self, response: dict) -> Optional[str]:
         """
--- a/plugins/module_utils/rest/response_strategies/nd_v1_strategy.py
+++ b/plugins/module_utils/rest/response_strategies/nd_v1_strategy.py
@@ -26,7 +26,10 @@ from __future__ import annotations
 
 # pylint: enable=invalid-name
 
-from typing import Any, Optional
+from collections.abc import Iterable, Mapping
+from typing import Any, Optional, TypeVar
+
+T = TypeVar("T")
 
 # Per-item status literals that mark a failure inside a Multi-Status body. ND sends these
 # on HTTP 207 and, for some endpoints, on HTTP 200 as well, so the body is scanned on any
@@ -41,6 +44,76 @@ _MULTISTATUS_ITEM_KEYS = ("results", "switchIds")
 
 # Item keys, in priority order, used to label a failing item in an error message.
 _MULTISTATUS_ITEM_LABEL_KEYS = ("name", "switchId", "serialNumber", "id")
+
+# Item keys, in priority order, carrying the per-item failure detail. ND is not consistent
+# across endpoints: `message` (batch interface / switchActions/deploy) and `warningMessage`
+# (fabric update-group attach) are both in use.
+_MULTISTATUS_ITEM_MESSAGE_KEYS = ("message", "warningMessage", "status")
+
+
+def _get_typed_value(mapping: Mapping[str, Any], key: str, expected_type: type[T], default: T) -> T:
+    """
+    # Summary
+
+    Return `mapping[key]` when it is an instance of `expected_type`, else `default`.
+
+    ## Description
+
+    ND bodies are not schema-guaranteed: a key may be absent, or present with a `null` (or otherwise unexpected) value. `dict.get(key, default)`
+    only covers the absent case -- it returns `None` for an explicit `null` -- so callers that go on to index, iterate, or call `len()` need this
+    type-checked accessor instead.
+
+    ## Parameters
+
+    - mapping: The dict to read from
+    - key: The key to read
+    - expected_type: The type the value must be an instance of
+    - default: The value returned when the key is absent or the value has the wrong type
+
+    ## Returns
+
+    - The value at `key` when it is an instance of `expected_type`, otherwise `default`
+
+    ## Raises
+
+    None
+    """
+    value = mapping.get(key)
+    return value if isinstance(value, expected_type) else default
+
+
+def _first_non_empty(mapping: Mapping[str, Any], keys: Iterable[str]) -> str | None:
+    """
+    # Summary
+
+    Return the first non-empty value in `mapping` among `keys`, as a string.
+
+    ## Description
+
+    Keys are tried in the order given. A key whose value is absent, `None`, or stringifies to an empty (or whitespace-only) string is skipped, so an
+    item such as `{"name": "", "switchId": "FDO123"}` labels as `FDO123` rather than as the empty string.
+
+    ## Parameters
+
+    - mapping: The dict to read from
+    - keys: Candidate keys, in priority order
+
+    ## Returns
+
+    - The first non-empty value as a string, or None when every candidate key is absent or empty
+
+    ## Raises
+
+    None
+    """
+    for key in keys:
+        value = mapping.get(key)
+        if value is None:
+            continue
+        text = str(value).strip()
+        if text:
+            return text
+    return None
 
 
 def _failed_multistatus_items(response: dict) -> list[dict[str, Any]]:
@@ -67,17 +140,11 @@ def _failed_multistatus_items(response: dict) -> list[dict[str, Any]]:
 
     None
     """
-    data = response.get("DATA")
-    if not isinstance(data, dict):
-        return []
     failed: list[dict[str, Any]] = []
+    data = _get_typed_value(response, "DATA", dict, {})
     for key in _MULTISTATUS_ITEM_KEYS:
-        items = data.get(key)
-        if not isinstance(items, list):
-            continue
-        for item in items:
-            if isinstance(item, dict) and str(item.get("status") or "").strip().lower() in _MULTISTATUS_FAILURE_STATUSES:
-                failed.append(item)
+        items = _get_typed_value(data, key, list, [])
+        failed.extend(item for item in items if isinstance(item, dict) and str(item.get("status") or "").strip().lower() in _MULTISTATUS_FAILURE_STATUSES)
     return failed
 
 
@@ -102,11 +169,13 @@ class NdV1Strategy:
 
     1. raw_response: Non-JSON response stored in DATA.raw_response
     2. code/message: DATA.code and DATA.message
-    3. messages array: all DATA.messages[].{code, severity, message} joined with "; "
-    4. errors array: all DATA.errors[] joined with "; "
-    5. Connection failure: No DATA with REQUEST_PATH and MESSAGE
-    6. Non-dict DATA: Stringified DATA value
-    7. Unknown: Fallback with RETURN_CODE
+    3. error: Scalar DATA.error value
+    4. messages array: all DATA.messages[].{code, severity, message} joined with "; "
+    5. errors array: all DATA.errors[] joined with "; "
+    6. Multi-Status per-item failures: DATA.results[]/DATA.switchIds[] failing items joined with "; "
+    7. Connection failure: No DATA with REQUEST_PATH and MESSAGE
+    8. Non-dict DATA: Stringified DATA value
+    9. Unknown: Fallback with RETURN_CODE
 
     ## Raises
 
@@ -207,8 +276,10 @@ class NdV1Strategy:
 
         ## Description
 
-        Labels the item by the first present identifier key (`name`, `switchId`, `serialNumber`, `id`) and appends its per-item `message` (falling
-        back to `status` when no message is present). When no identifier key is present, only the message is returned.
+        Labels the item by the first non-empty identifier key (`_MULTISTATUS_ITEM_LABEL_KEYS`) and appends the first non-empty detail key
+        (`_MULTISTATUS_ITEM_MESSAGE_KEYS`, which falls back to `status` when no message key is present). Empty and whitespace-only values are skipped
+        rather than used, so an item carrying `{"name": ""}` is not labelled with the empty string. When no identifier key is present, only the
+        detail is returned; when no detail key is present either, the generic literal `failed` is used.
 
         ## Parameters
 
@@ -222,9 +293,9 @@ class NdV1Strategy:
 
         None
         """
-        label = next((str(item[key]) for key in _MULTISTATUS_ITEM_LABEL_KEYS if item.get(key) is not None), None)
-        detail = item.get("message") or item.get("status") or "failed"
-        return f"{label}: {detail}" if label is not None else str(detail)
+        label = _first_non_empty(item, _MULTISTATUS_ITEM_LABEL_KEYS)
+        detail = _first_non_empty(item, _MULTISTATUS_ITEM_MESSAGE_KEYS) or "failed"
+        return f"{label}: {detail}" if label is not None else detail
 
     def is_not_found(self, return_code: int) -> bool:
         """
@@ -294,11 +365,12 @@ class NdV1Strategy:
         1. Connection failure (no DATA)
         2. Non-JSON response (raw_response in DATA)
         3. code/message dict
-        4. messages array with code/severity/message (all items joined)
-        5. errors array (all items joined)
-        6. Multi-Status per-item failures (results[]/switchIds[], all joined)
-        7. Unknown dict format
-        8. Non-dict DATA
+        4. Scalar error key (DATA.error)
+        5. messages array with code/severity/message (all items joined)
+        6. errors array (all items joined)
+        7. Multi-Status per-item failures (results[]/switchIds[], all joined)
+        8. Unknown dict format
+        9. Non-dict DATA
 
         ## Parameters
 
@@ -339,8 +411,11 @@ class NdV1Strategy:
 
         ## Description
 
-        Checks, in priority order: `raw_response`, `code`/`message`, the `messages[]` array, the `errors[]` array, and Multi-Status per-item
-        failures (`results[]`/`switchIds[]`). Falls back to a generic status message when no specific format matches.
+        Checks, in priority order: `raw_response`, `code`/`message`, the scalar `error` key, the `messages[]` array, the `errors[]` array, and
+        Multi-Status per-item failures (`results[]`/`switchIds[]`). Falls back to a generic status message when no specific format matches.
+
+        The `messages` and `errors` arrays are read through `_get_typed_value` because ND may send either key with an explicit `null` value, which a
+        bare `data_dict[key]` would iterate (or `len()`) into a `TypeError` on the very path that reports an error to the user.
 
         ## Parameters
 
@@ -364,18 +439,26 @@ class NdV1Strategy:
         elif "code" in data_dict and "message" in data_dict:
             msg = f"ND Error {data_dict['code']}: {data_dict['message']}"
 
+        # Scalar error key. `is_success()` already classifies a response carrying DATA.error as a
+        # failure; without this branch the error ND actually sent is dropped and the user is shown
+        # only the generic "Request failed with status <code>" fallback below.
+        if msg is None and data_dict.get("error") is not None:
+            msg = f"ND Error: {data_dict['error']}"
+
         # messages array format
-        if msg is None and "messages" in data_dict and len(data_dict.get("messages", [])) > 0:
+        if msg is None:
             parts = []
-            for m in data_dict["messages"]:
-                if all(k in m for k in ("code", "severity", "message")):
+            for m in _get_typed_value(data_dict, "messages", list, []):
+                if isinstance(m, dict) and all(k in m for k in ("code", "severity", "message")):
                     parts.append(f"ND Error {m['code']} ({m['severity']}): {m['message']}")
             if parts:
                 msg = "; ".join(parts)
 
         # errors array format
-        if msg is None and "errors" in data_dict and len(data_dict.get("errors", [])) > 0:
-            msg = f"ND Error: {'; '.join(str(e) for e in data_dict['errors'])}"
+        if msg is None:
+            errors = _get_typed_value(data_dict, "errors", list, [])
+            if errors:
+                msg = f"ND Error: {'; '.join(str(e) for e in errors)}"
 
         # Multi-Status per-item failures (results[]/switchIds[])
         if msg is None:
--- a/plugins/module_utils/rest/response_strategies/nd_v1_strategy.py
+++ b/plugins/module_utils/rest/response_strategies/nd_v1_strategy.py
@@ -28,9 +28,10 @@ from __future__ import annotations
 
 from typing import Any, Optional
 
-# Per-item status literals that mark a failure inside an HTTP 207 Multi-Status body.
-# ND is inconsistent about the literal across endpoints: "failed" (batch interface /
-# switchActions/deploy), "error" (breakout), and "failure" (acl / maintenance_mode).
+# Per-item status literals that mark a failure inside a Multi-Status body. ND sends these
+# on HTTP 207 and, for some endpoints, on HTTP 200 as well, so the body is scanned on any
+# success code. ND is also inconsistent about the literal across endpoints: "failed" (batch
+# interface / switchActions/deploy), "error" (breakout), and "failure" (acl / maintenance_mode).
 _MULTISTATUS_FAILURE_STATUSES = frozenset({"failed", "failure", "error"})
 
 # DATA envelope keys whose items carry a per-item `status`. Two known shapes:
@@ -46,7 +47,7 @@ def _failed_multistatus_items(response: dict) -> list[dict[str, Any]]:
     """
     # Summary
 
-    Return the per-item entries reporting a failure in an HTTP 207 Multi-Status body.
+    Return the per-item entries reporting a failure in a Multi-Status body.
 
     ## Description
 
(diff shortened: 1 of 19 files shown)


plugins/module_utils/rest/response_strategies/nd_v1_strategy.py
```diff
@@ -28,6 +28,57 @@
...
+
+# Item keys, in priority order, used to label a failing item in an error message.
+_MULTISTATUS_ITEM_LABEL_KEYS = ("name", "switchId", "serialNumber", "id")
+
+
+def _failed_multistatus_items(response: dict) -> list[dict[str, Any]]:
+    """
+    # Summary
+
+    Return the per-item entries reporting a failure in an HTTP 207 Multi-Status body.
```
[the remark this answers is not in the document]
allenrobel: Good catch -- fixed in "Clarify per-item failure detection docs (#295)". The detection is intentionally gated on any success code (ND sends per-item statuses on plain HTTP 200 for some endpoints; the L3Out batch POST tests cover that path), and the literal set includes "failure" alongside "failed"/"error". The prose now says so instead of implying 207-only.

plugins/module_utils/rest/response_strategies/nd_v1_strategy.py
```diff
@@ -112,6 +163,8 @@ def is_success(self, response: dict) -> bool:
 
         - Top-level `ERROR` key is present
         - `DATA.error` key is present
+        - An HTTP 207 Multi-Status body reports a per-item failure in
+          `DATA.results[]` or `DATA.switchIds[]` (status `failed`/`error`)
```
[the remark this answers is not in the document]
allenrobel: Good catch -- fixed in "Clarify per-item failure detection docs (#295)". The detection is intentionally gated on any success code (ND sends per-item statuses on plain HTTP 200 for some endpoints; the L3Out batch POST tests cover that path), and the literal set includes "failure" alongside "failed"/"error". The prose now says so instead of implying 207-only.

plugins/module_utils/rest/response_strategies/nd_v1_strategy.py
```diff
@@ -133,8 +186,42 @@ def is_success(self, response: dict) -> bool:
         data = response.get("DATA")
         if isinstance(data, dict) and data.get("error") is not None:
             return False
+        # ND returns HTTP 207 for batch operations and reports per-item outcomes in
+        # DATA.results[]/DATA.switchIds[] items carrying status: success|failed|error.
+        # The 207 status alone does not indicate every item succeeded, so a body with
+        # any failing item must not be classified as success. See issue #295.
```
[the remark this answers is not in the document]
allenrobel: Good catch -- fixed in "Clarify per-item failure detection docs (#295)". The detection is intentionally gated on any success code (ND sends per-item statuses on plain HTTP 200 for some endpoints; the L3Out batch POST tests cover that path), and the literal set includes "failure" alongside "failed"/"error". The prose now says so instead of implying 207-only.

tests/unit/module_utils/test_response_handler_nd.py
```diff
@@ -1458,6 +1458,345 @@ def test_response_handler_nd_01000():
     assert "Primary error message" in instance.error_message
 
 
+# =============================================================================
+# Test: 207 Multi-Status per-item failure detection (issue #295)
+#
+# ND returns HTTP 207 for batch operations and reports per-item outcomes in a
+# DATA envelope array whose items carry status: "success" | "failed" | "error".
```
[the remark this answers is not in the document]
allenrobel: Good catch -- fixed in "Clarify per-item failure detection docs (#295)". The detection is intentionally gated on any success code (ND sends per-item statuses on plain HTTP 200 for some endpoints; the L3Out batch POST tests cover that path), and the literal set includes "failure" alongside "failed"/"error". The prose now says so instead of implying 207-only.

tests/unit/module_utils/orchestrators/test_l3out.py
```diff
@@ -245,12 +245,13 @@ def test_l3out_00120() -> None:
     ## Test
 
     - POST returns 200 but response body has item with status "failed"
-    - `RuntimeError` is raised with "partially failed" message
+    - `RuntimeError` is raised surfacing the per-item failure message (now via the centralized
+      `NdV1Strategy` 207/multi-status handling in RestSend)
```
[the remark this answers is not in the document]
allenrobel: Good catch -- fixed in "Clarify per-item failure detection docs (#295)". The detection is intentionally gated on any success code (ND sends per-item statuses on plain HTTP 200 for some endpoints; the L3Out batch POST tests cover that path), and the literal set includes "failure" alongside "failed"/"error". The prose now says so instead of implying 207-only.

tests/unit/module_utils/orchestrators/test_l3out.py
```diff
@@ -567,12 +568,13 @@ def test_l3out_00520() -> None:
     ## Test
 
     - POST returns 200 but response body has one item with status "failed"
-    - `RuntimeError` is raised with "partially failed" message
+    - `RuntimeError` is raised surfacing the per-item failure message (now via the centralized
+      `NdV1Strategy` 207/multi-status handling in RestSend)
```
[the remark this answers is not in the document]
allenrobel: Good catch -- fixed in "Clarify per-item failure detection docs (#295)". The detection is intentionally gated on any success code (ND sends per-item statuses on plain HTTP 200 for some endpoints; the L3Out batch POST tests cover that path), and the literal set includes "failure" alongside "failed"/"error". The prose now says so instead of implying 207-only.

plugins/module_utils/rest/response_strategies/nd_v1_strategy.py
```diff
@@ -28,6 +28,58 @@
...
+
+    - List of failing item dicts (empty list when none fail)
+
+    ## Raises
+
+    None
+    """
+    data = response.get("DATA")
+    if not isinstance(data, dict):
+        return []
```
akinross: did you consider this to be handled as a generic function which could be reused by other functions that would do similar checks? something like below which can be reused below also

```
def _get_typed_value(
    mapping: Mapping[str, Any],
    key: str,
    expected_type: type[T],
    default: T,
) -> T:
    value = mapping.get(key)
    return value if isinstance(value, expected_type) else default

File: plugins/module_utils/rest/response_strategies/nd_v1_strategy.py [hunk 8]
```diff
@@ -28,6 +28,58 @@
...
+    failed: list[dict[str, Any]] = []
+    for key in _MULTISTATUS_ITEM_KEYS:
+        items = data.get(key)
+        if not isinstance(items, list):
+            continue
+        for item in items:
+            if isinstance(item, dict) and str(item.get("status") or "").strip().lower() in _MULTISTATUS_FAILURE_STATUSES:
+                failed.append(item)
+    return failed
```
akinross: if above code makes sense, I would change the function to something like below which has no continue and a single return point which makes the flow easier to understand in my opinion.

```
failed: list[dict[str, Any]] = []
data = _get_typed_value(response, "DATA", dict, {})

for key in _MULTISTATUS_ITEM_KEYS:
    items = _get_typed_value(data, key, list, [])
    failed.extend(
        item
        for item in items
        if isinstance(item, dict)
        and str(item.get("status") or "").strip().lower()
        in _MULTISTATUS_FAILURE_STATUSES
    )

return failed
```

if you want to avoid the loop a conditional with `if data:` could be added, but don't think performance wise it makes a real difference

plugins/module_utils/rest/response_strategies/nd_v1_strategy.py
```diff
@@ -133,8 +189,43 @@ def is_success(self, response: dict) -> bool:
...
+
+        ## Returns
+
+        - A human-readable `label: message` string (or just the message when unlabeled)
+
+        ## Raises
+
+        None
+        """
+        label = next((str(item[key]) for key in _MULTISTATUS_ITEM_LABEL_KEYS if item.get(key) is not None), None)
```
akinross: what happens when item.get(key) is a empty string?

plugins/module_utils/rest/response_strategies/nd_v1_strategy.py
```diff
@@ -133,8 +189,43 @@ def is_success(self, response: dict) -> bool:
...
+        detail = item.get("message") or item.get("status") or "failed"
```
akinross: is message and status a fix format for these items? because i also saw warningMessage used in some of the test fixtures, is this related?

akinross: if the intend is getting first none empty value in both why not create a generic helper function function for this, which perhaps even allows a default. this would make it extendable by just extending the _CONSTANT when more values need to be checked

plugins/module_utils/rest/response_strategies/nd_v1_strategy.py
```diff
@@ -232,34 +324,67 @@ def extract_error_message(self, response: dict) -> Optional[str]:
...
+        msg: Optional[str] = None
+        # Raw response (non-JSON)
+        if "raw_response" in data_dict:
+            msg = "ND Error: Response could not be parsed as JSON"
+        # code/message format
+        elif "code" in data_dict and "message" in data_dict:
+            msg = f"ND Error {data_dict['code']}: {data_dict['message']}"
+
+        # messages array format
+        if msg is None and "messages" in data_dict and len(data_dict.get("messages", [])) > 0:
```
akinross: i think you could have issues on data_dict.get("messages") when it exist but is None, because len(None) will typeerror, same applies to errors check below

plugins/module_utils/rest/response_strategies/nd_v1_strategy.py
```diff
@@ -232,34 +324,67 @@ def extract_error_message(self, response: dict) -> Optional[str]:
...
+
+        ## Returns
+
+        - A human-readable error message string (never None)
+
+        ## Raises
+
+        None
+        """
+        msg: Optional[str] = None
```
akinross, line 502: how is the error in DATA handled? is this returned somehow?

akinross (changes requested):

plugins/module_utils/rest/response_strategies/nd_v1_strategy.py
allenrobel: Good call -- added as `_get_typed_value(mapping, key, expected_type, default)` in "Harden per-item failure extraction from review (#295)".

It earns its keep beyond the two call sites here: it is also what fixes the `messages`/`errors` bug you spotted below. `dict.get(key, [])` covers the *absent* key but returns `None` for an explicit JSON `null`, which is exactly the case that was crashing -- the typed accessor collapses both into the default.

plugins/module_utils/rest/response_strategies/nd_v1_strategy.py
allenrobel: Taken -- `_failed_multistatus_items` is now the flat `extend`/generator form with a single return, built on `_get_typed_value`.

plugins/module_utils/rest/response_strategies/nd_v1_strategy.py
allenrobel: Real bug, thank you. `is not None` accepted `""` as a label, so an item like `{"name": "", "message": "Deploy failed"}` rendered as `": Deploy failed"`. Note the very next line already got this right (`item.get("message") or item.get("status")` uses truthiness), so the two lines disagreed with each other -- which is what your next comment picks up on.

Both now run through `_first_non_empty()`, which skips absent, `None`, empty, and whitespace-only values. Covered by new test `01300`.

plugins/module_utils/rest/response_strategies/nd_v1_strategy.py
allenrobel: No, it is not a fixed format -- good catch, and `warningMessage` is exactly the right thing to have noticed. ND is inconsistent across endpoints: batch interface / `switchActions/deploy` items carry `message`, while `fabric_update_group`'s attach items carry `warningMessage`.

So the message key is now a constant rather than a hardcoded `item.get("message")`:

```python
_MULTISTATUS_ITEM_MESSAGE_KEYS = ("message", "warningMessage", "status")
```

One scoping note so this is not mistaken for a fix that is already complete: those `fabric_update_group` items ride an `attachUpdateGroups` envelope, which is **not** in `_MULTISTATUS_ITEM_KEYS` (`results`, `switchIds`), so they do not reach the central path today -- they are still caught by that orchestrator's own detector. No regression, but the envelope-key half of the generalization is #397's job. This change puts the message-key seam in place for it. New test `01310` covers a `warningMessage` item going through the central path.

allenrobel: Agreed, and this was the most useful suggestion in the batch -- it subsumes the empty-string bug above and the `warningMessage` question below into one place. Added `_first_non_empty(mapping, keys)`, and both lookups are now constant-driven:

```python
label = _first_non_empty(item, _MULTISTATUS_ITEM_LABEL_KEYS)
detail = _first_non_empty(item, _MULTISTATUS_ITEM_MESSAGE_KEYS) or "failed"
```

Extending either key set is now a one-line change to the constant. I kept the `or "failed"` default at the call site rather than as a helper parameter, since the label case legitimately has no default (an unlabelled item renders as just the message).

plugins/module_utils/rest/response_strategies/nd_v1_strategy.py
allenrobel: Correct on both, and worth flagging that this one is pre-existing -- the same two lines are on `develop` today (this PR only moved them into `_extract_dict_error_message`). `"messages" in data_dict` passes for an explicit `null`, and `.get("messages", [])` then returns `None` rather than the default, so `len(None)` raises `TypeError` -- on the one path that runs when something has *already* gone wrong, and in a method whose docstring promises `## Raises: None`.

Both arrays now read through `_get_typed_value(data_dict, "messages", list, [])`, which drops the `in`-check-then-`len` dance entirely. I also added an `isinstance(m, dict)` guard on the items: `all(k in m for k in (...))` against a string item was doing a silent substring check rather than a key lookup. Tests `01330` (null `messages`) and `01340` (null `errors`).

plugins/module_utils/rest/response_strategies/nd_v1_strategy.py
allenrobel: It was not, and that is a real hole -- thank you. `is_success()` returns `False` when `DATA.error` is present, but the extractor had no matching branch, so the message fell through to the generic `"ND Error: Request failed with status 200"` and the text ND actually sent was dropped. Also pre-existing on `develop`; the existing test (`00670`) asserted only `success is False` and never the message, which is why it went unnoticed.

Added a branch after `code`/`message`:

```python
if msg is None and data_dict.get("error") is not None:
    msg = f"ND Error: {data_dict[error]}"
```

New test `01350` asserts the text, not just the success flag.

akinross (dismissed):

shrsr:
@allenrobel 
Following up on folding the links (207) handling into NdV1Strategy. I checked the links contract against the ND 4.2 OpenAPI, so here is exactly what nd_manage_links needs from the central detector before I can delete its own body scan.

Confirmed contract: bulk create (POST /links) and bulk delete (POST /linkActions/remove) both return HTTP 207 with a body shaped {"links": [{"linkId", "message", "status"}]}, where status is either success or failure. Both scopes (manage and one_manage) use the same shape.

What is needed:
1. Add "links" to _MULTISTATUS_ITEM_KEYS so the detector scans the links array. Adding "items" alongside it is a safe superset if you want to also cover the linkActions family.

2. Add "linkId" to _MULTISTATUS_ITEM_LABEL_KEYS so a failing link is labelled by its id in the aggregated error. Today it would fall through to id or come back unlabelled.

3. Nothing else on the literals. failure is already in _MULTISTATUS_FAILURE_STATUSES and message is already in _MULTISTATUS_ITEM_MESSAGE_KEYS, so those are covered.

4. This is the one we really have to pin down. The detector pulls the failing items out of response["DATA"], but I do not think ND wraps the links body in a DATA envelope at all. By the time my orchestrator sees the bulk response it is already a plain {"links": [...]} at the top level, with no DATA key. If that is also what is_success receives, then reading response["DATA"] hands back an empty dict, the scan finds zero items, and is_success returns True even when a link failed. In other words the central check would silently let every partial failure through for links, which is the exact bug we are trying to kill. So before I switch links over, we need to confirm where the per item array actually lands in the response object is_success is given, top level response["links"] versus response["DATA"]["links"], and point the scan there. If different endpoints expose the body at different nesting levels, the detector has to handle both.

Optional but actually nice: instead of every orchestrator editing these three module level constants, consider letting an orchestrator declare its own result keys and label keys (for example a bulk_result_keys and bulk_label_keys attribute the strategy reads). Then links, l3out, acl and policy_group can each register their envelope without touching the shared file.

Once items 1 and 2 land and the nesting question in 4 is settled, I will delete _raise_on_bulk_failures in nd_manage_links and rely on the central detector. Until then I am keeping the bespoke scan as the safe superset, with a TODO(4.2.1) marker pointing here. Happy to send links response fixtures once the shape is wired in.

allenrobel:
@shrsr Items 1 and 2 are in — commit "Scan links[] Multi-Status envelope; label by linkId (PR #398 review)". `links` is now in `_MULTISTATUS_ITEM_KEYS` and `linkId` is in `_MULTISTATUS_ITEM_LABEL_KEYS` (ahead of the generic `id`). I re-verified the contract against the ND 4.2.1 OpenAPI: both `POST /links` and `POST /linkActions/remove` return 207 with `{"links": [{"linkId", "message", "status"}]}`, status `success|failure`. Four new tests (`01360`–`01390`) cover partial failure, all-success, the spec's own `linkId: ""` failure example (empty labels are skipped rather than rendered as `": <message>"`), and a GET-shaped list body.

On item 4, the nesting question: the detector's read of `response["DATA"]` is correct, and there is no path on which the per-item array reaches `is_success()` at the top level. The HttpAPI plugin wraps every parsed JSON body under `DATA` in `_verify_response()` (`plugins/httpapi/nd.py`, `self.info["DATA"] = response_data`), and `RestSend.commit()` hands `sender.response` to the response handler unmodified — so `is_success()` sees `{"RETURN_CODE": 207, ..., "DATA": {"links": [...]}}`. The bare `{"links": [...]}` your orchestrator sees is the post-validation view: orchestrator code consumes `response["DATA"]` after the handler has already classified the response. The one practical consequence for you: unit fixtures exercising the central path need the full envelope (`DATA` wrapper), not the bare body.

Related false-positive check: `GET /links` list bodies ride the same `DATA.links[]` envelope, but link objects carry no top-level `status` key, so the literal-gated scan cannot misclassify a query — test `01390` pins this.

On the `items` superset: holding off. Envelope keys get added one at a time against a verified endpoint shape, since a speculative key risks misreading an unrelated body that happens to carry a failure literal in `status`. When the linkActions import/export family is wired up, its envelope can be added the same way. That expansion — and your per-orchestrator key-registration idea, which is a good candidate design — are both #397's scope; I'll carry the registration idea into that issue.

Links response fixtures are welcome once you switch the module over — with the above landed, you should be able to drop `_raise_on_bulk_failures` without a `TODO(4.2.1)` marker.

allenrobel:
@gmicol and @skaszlik should review this PR since it removes their 207 handlers in PRs #266, #286, and  #288  in favor of this centralized handler.

plugins/module_utils/rest/response_strategies/nd_v1_strategy.py
```diff
@@ -133,8 +262,45 @@ def is_success(self, response: dict) -> bool:
         data = response.get("DATA")
         if isinstance(data, dict) and data.get("error") is not None:
             return False
+        # ND reports per-item outcomes for batch operations in DATA.results[]/DATA.switchIds[]/
+        # DATA.links[] items carrying status: success|failed|failure|error. The HTTP status alone does not
+        # indicate every item succeeded -- ND sends these bodies on 207 and, for some endpoints,
+        # on plain 200 -- so any success-code response with a failing item must not be
+        # classified as success. See issue #295.
+        if _failed_multistatus_items(response):
```
mikewiebe, line 300: ### High: Terminal per-item failures replay the complete mutation for up to 300 seconds

**Issue**

The new item-status scan converts a success-code 200/207 response containing a
failed member into `success=False`, but `RestSend` treats every unsuccessful
result as retryable and resubmits the original mutation until its 300-second
budget expires.

**Evidence**

- [`nd_v1_strategy.py` line 270](https://github.com/CiscoDevNet/ansible-nd/pull/398/files#diff-87979e507ebd16d27dbeee457df20188efed5d44a603b787fe836f7e8e831f62R270)
  changes a success-code response containing a failed item into an unsuccessful
  response-handler result.
- [`rest_send.py` lines 330-381](https://github.com/CiscoDevNet/ansible-nd/blob/5e31239e42caffc189c3e49b19c1cc048957d581/plugins/module_utils/rest/rest_send.py#L330-L381)
  retries the same method, path, and payload whenever the result is
  unsuccessful; the default budget is 300 seconds.
- An exact-head reproduction using a mixed HTTP 207 response, a three-second
  timeout, and a one-second send interval submitted the identical POST three
  times before returning the deterministic application failure.

**Live ND 4.2 verification**

A bounded live test ran against Nexus Dashboard 4.2.1 build `4.2.1.10` using
the exact PR head. The test pre-created ACL `reg_pr398_seed_205203`, then sent
one batch containing that existing ACL and new ACL
`reg_pr398_new_205203`.

| Submission | New ACL result | Existing ACL result | HTTP |
| ---: | --- | --- | ---: |
| 1 | `success: created successfully` | `failed: already exists` | `207` |
| 2 | `failed: already exists` | `failed: already exists` | `207` |
| 3 | `failed: already exists` | `failed: already exists` | `207` |

All three requests used the identical payload, with SHA-256
`3352085c889e2d619b2fc1893f6f8e86c53c3ca003483714bce7903a858b35b3`.
The first request created the new ACL, but the existing-ACL failure made the
aggregate result unsuccessful. `RestSend` then replayed the complete payload
twice, causing the ACL that had just succeeded to fail as already existing.

The live probe was deliberately bounded to a six-second timeout, a two-second
send interval, and three submissions. The production defaults in `RestSend`
are 300 seconds and five seconds, so the same loop can submit approximately 60
times; that count is derived from the source defaults rather than a
five-minute live run. The final result was `success=False, changed=False`
despite the first request changing controller state.

The temporary ACLs and fabric were deleted successfully after the test, the
fabric deletion was confirmed by HTTP 404, and the postflight ownership audit
passed with no designated switches managed by the test controller.

**Existing PR overlap**

No matching existing PR comment found. Existing threads cover supported
envelopes, labels, null handling, error text, and refactoring, but not transport
retry ownership.

No matching open issue found. Issues
[#295](https://github.com/CiscoDevNet/ansible-nd/issues/295) and
[#397](https://github.com/CiscoDevNet/ansible-nd/issues/397) track per-item
failure detection and detector consolidation, not replay of a terminal
application failure. Issue
[#396](https://github.com/CiscoDevNet/ansible-nd/issues/396) tracks `nd_rest`
accepting successful HTTP 207 responses, not retry classification. Issue
[#456](https://github.com/CiscoDevNet/ansible-nd/issues/456) lists safe 207
handling as a dependency of a future interface workflow; it does not track the
shared `RestSend` fix.

**Impact**

A deterministic validation failure can resubmit a non-idempotent batch
mutation up to roughly 60 times. Members that already succeeded are sent again,
which can duplicate side effects, produce misleading already-exists failures,
replace the original diagnostic, and multiply controller load. Orchestrators
with their own outer retries can amplify the request count further.

**Suggested fix**

Add an explicit terminal-versus-retryable classification between response
handling and `RestSend`. A successful HTTP status containing a definitive
per-item application failure should stop after one request, while
orchestrator-specific eventual-consistency policies may retain bounded retries.
Add a production-mode regression test with `timeout > send_interval` that
asserts exactly one submission for mixed 200/207 failures, plus an outer-retry
test for L3Out.

---

mikewiebe, line 300: ### Medium: Mixed partial success is reported as `changed=False`

**Issue**

The response handler couples aggregate success to changed state. When one batch
member succeeds and another fails, the new scan correctly makes the request
unsuccessful but the failure branch reports `changed=False` even though Nexus
Dashboard changed.

**Evidence**

- [`nd_v1_strategy.py` line 270](https://github.com/CiscoDevNet/ansible-nd/pull/398/files#diff-87979e507ebd16d27dbeee457df20188efed5d44a603b787fe836f7e8e831f62R270)
  makes a mixed success/failure item response unsuccessful.
- [`response_handler_nd.py` lines 220-232](https://github.com/CiscoDevNet/ansible-nd/blob/5e31239e42caffc189c3e49b19c1cc048957d581/plugins/module_utils/rest/response_handler_nd.py#L220-L232)
  hard-codes `changed=False` whenever aggregate success is false.
- The exact-head mixed-result reproduction returned
  `success=False, changed=False` after one item succeeded and one failed.

**Live ND 4.2 verification**

A bounded live test against Nexus Dashboard 4.2.1 build `4.2.1.10` used the
exact PR head to send a batch containing one new ACL and one ACL that already
existed. The first HTTP 207 response reported:

| Batch member | Controller result |
| --- | --- |
| New ACL `reg_pr398_new_205203` | `success: created successfully` |
| Existing ACL `reg_pr398_seed_205203` | `failed: already exists` |

The next two submissions reported that both ACLs already existed, confirming
that the successful member from the first response had changed persistent
controller state. Nevertheless, the final `RestSend` result was
`success=False, changed=False`.

**Existing PR overlap**

No matching existing PR comment found. The separately posted High finding
reports the observed `changed=False` value as evidence of complete-payload
replay, but it does not identify or propose a fix for inaccurate partial-change
reporting.

[#295](https://github.com/CiscoDevNet/ansible-nd/issues/295) tracks recognizing
failed members, not truthful partial-change reporting. Issue
[#389](https://github.com/CiscoDevNet/ansible-nd/issues/389) tracks
`NDStateMachine` collection synchronization after mid-run failures, not the
shared response handler's `changed` result for one mixed 207 batch. Issues
[#385](https://github.com/CiscoDevNet/ansible-nd/issues/385) and
[#456](https://github.com/CiscoDevNet/ansible-nd/issues/456) discuss final-state
or workflow behavior but do not track this shared response-handler correction.

**Impact**

Failed Ansible output can claim that nothing changed even though successful
members were created, deleted, or deployed. Recovery automation and operators
cannot trust the result, and a retry begins from unexpectedly modified
controller state.

**Suggested fix**

Track mutation independently from aggregate success. For recognized itemized
responses, report `changed=True` when any member succeeded and
`changed=False` only when every member failed, honoring the controller modified
indicator where available. Preserve both successful and failed member evidence,
and add mixed-success and all-failed tests through `ResponseHandler`,
`RestSend`, and at least one module path.

allenrobel: Confirmed — the live ACL repro made this unambiguous. Fixed in "Classify terminal vs retryable mutation failures in ResponseHandler (#295)" and "Stop RestSend retrying terminal application failures (#295)": the response handler now marks every POST/PUT/DELETE result with `retryable`, where a failure whose RETURN_CODE is a success code (the application definitively rejected the request) is terminal, and `RestSend._commit_normal_mode()` breaks after one submission on `retryable=False`. Terminal scope is any embedded error on a success code, so the pre-existing `DATA.error`/`ERROR`-on-200 replay is fixed by the same seam. Non-success codes (5xx) keep today's bounded retries, GET semantics are untouched (no `retryable` key, default-retryable), and orchestrator-level eventual-consistency policies still work — the L3Out attach "not found" outer retry is pinned by `test_l3out_00990`. Regression tests: `test_rest_send_01100` asserts exactly one submission with `timeout > send_interval` (a sentinel second response turns any replay into a visible failure), `test_rest_send_01110` guards that 5xx still retries, and `test_l3out_00980` covers the outer-retry amplification you raised.

allenrobel: Confirmed. Fixed in "Report changed=True for mixed Multi-Status batches (#295)": the failure branch of `ResponseHandler._handle_post_put_delete_response()` now calls a new strategy method `is_changed_on_failure()` instead of hard-coding `changed=False`. The `modified` response header is authoritative in both directions when present; otherwise any per-item `status: success` in the recognized envelopes (`results[]`/`switchIds[]`/`links[]`) reports `changed=True`, and all-failed batches keep `changed=False`. Non-itemized embedded errors retain the conservative `changed=False` default. Covered by `test_response_handler_nd_01450`–`01490` (mixed, all-failed, header-override both ways, `DATA.error`).

gmicol (approved):
LGTM

mikewiebe (approved):
Re-reviewed at d2d197c3. The previously reported multi-status retry and changed-state issues are resolved, focused and full module-utils tests pass, and no blocking findings remain.

allenrobel:
Merged on approval from Mike's (@mikewiebe) and Matt's (@gmicol) teams, CI passing, and no merged dependencies.

Issue #401, opened 2026-07-15
nd_interface_loopback: ip typed as IPv4CIDR sends /32 that ND 4.2.1 rejects (wants bare IPv4)


allenrobel:
## Summary

On `develop`, `LoopbackPolicyModel.ip` (in `plugins/module_utils/models/interfaces/loopback_interface.py`) is typed `IPv4CIDR`, its description says *"in CIDR notation (e.g. 10.1.1.1/32)"*, and the `nd_interface_loopback` integration tasks supply `/32` CIDR values. Against **live ND 4.2.1**, the Manage interfaces endpoint **rejects a CIDR `ip`** on a loopback create and requires a **bare IPv4 address**.

## Evidence (live ND 4.2.1 lab)

Raw intent-only `POST /api/v1/manage/fabrics/{fabric}/switches/{switchId}/interfaces` for a `policyType: loopback` interface:

- `"ip": "10.99.99.250/32"` → `HTTP 400` — `{"message":"Policy [loopback] - Validation failed for following fields: [ip]"}`
- `"ip": "10.99.99.250"` (bare) → `HTTP 207` — `Interface added successfully`

The corresponding OpenAPI template (`intLoopbackTemplate`) declares `ip` as `format: ipv4` (a bare address), consistent with the wire behavior.

## Impact

A real `state: merged`/`replaced` create through `nd_interface_loopback` that sends a CIDR `ip` (as the field description and integration tasks encourage) is rejected by ND 4.2.1. This is latent because the integration suite is not run against a live fabric in CI.

## Suggested fix

Normalize `ip` to a bare host address before it reaches the wire — accept either bare or CIDR from the user and strip the prefix, e.g. `str(ipaddress.IPv4Interface(value).ip)`. This keeps existing `/32`-style playbooks working, sends the bare form ND accepts, and preserves idempotency (ND returns the bare address). The `ipv6` field (`IPv6CIDR`) likely has the same issue but was not verified in this probe.

## Notes

- Verified on the `nd_interface_loopback_policy_type_union` pilot branch, where this is fixed by adding an `IPv4Host` normalizing type and switching the loopback `ip` field to it. This issue tracks porting the same fix to `develop`.
- Companion read-side issue (ND injecting `linkStateRoutingTag` into the loopback `policy` on GET) is specific to strict `extra="forbid"` validation and is **not** present on `develop` (develop uses `extra="ignore"`), so no separate `develop` bug there.

[...]

Issue #386, opened 2026-07-02
Implement policy-aware deep merge semantics for nested attributes in state=merged


gmicol:
## Summary

The Orchestrator framework does not consistently perform true merge semantics for nested attributes when `state=merged`. Direct nested `NDBaseModel` fields are partially supported, but lists, dictionaries, lists of `NDNestedModel`, and lists of dictionaries are treated as atomic values. When a user supplies only part of a nested collection, the framework can replace the entire existing nested collection instead of merging the requested subset.

This can be misleading and destructive for users who expect `state=merged` to preserve unspecified nested configuration.

## Current Behavior

`NDStateMachine` uses `exclude_unset=True` for `state=merged` diffs, then calls `NDConfigCollection.merge()`, which delegates to `NDBaseModel.merge()`.

`NDBaseModel.merge()`:
- only applies fields explicitly set by the proposed model
- recursively merges only when both current and proposed values are `NDBaseModel`
- otherwise assigns the proposed value directly

The diff helper `issubset()` recursively compares dictionaries, but lists require full-list equality with the same length. This means a proposed list containing one nested item is not treated as a subset of an existing list containing multiple items.

Affected examples include:
- `FabricUpdateGroupModel.install_image_data.install_package_names`
- `FabricUpdateGroupModel.update_report_checks`
- `LocalUserModel.security_domains`
- interface trunk VLAN mapping lists
- fabric models with nested list/dict collections

## Problem

For `state=merged`, users reasonably expect omitted nested properties to remain unchanged. Today that is only reliable for direct nested model fields. Collection-like nested fields are replaced as a single property.

Example problem shape:

```yaml
state: merged
config:
  - name: existing_object
    nested_collection:
      - key: one_item_to_update
        value: new_value
```

If the controller already has other entries in `nested_collection`, the framework may send only the proposed list and drop the omitted entries.

## Suggested Fix

Implement a policy-aware deep merge engine shared by model diff and merge logic.

Recommended design:

1. Add merge policy metadata to `NDBaseModel`, for example:

```python
merge_policies: ClassVar[dict[str, MergePolicy]] = {
    "security_domains": MergePolicy.list_keyed("name"),
    "update_report_checks": MergePolicy.list_keyed("report_check_name"),
    "install_package_names": MergePolicy.list_union(),
}
```

2. Supported policies should include:
   - `replace`: current default, safest fallback
   - `model_deep`: recursive model merge
   - `dict_deep`: recursive dictionary merge
[...]

Issue #358, opened 2026-06-26
Document RETURN output for fabric modules on develop


mikewiebe:
## Summary

PR #256 surfaced that the new AI fabric modules have empty `RETURN` documentation even though they return the shared state-machine output via `nd_state_machine.output.format_with_verbosity(...)`.

Please use this issue to track a follow-up pass across **all fabric modules that are currently merged into the `develop` branch** and make sure their `RETURN` blocks accurately document the values they return.

## Scope

Review the fabric-family modules already present on `develop`, not just the new AI fabric modules from PR #256. For each merged fabric module, verify whether the module has a complete `RETURN` block and update it if it is empty, incomplete, or inconsistent with the shared state-machine output.

## Develop fabric modules to audit

Current `develop` fabric modules to include in this cleanup:

- [`plugins/modules/nd_manage_fabric_ebgp_vxlan.py`](https://github.com/CiscoDevNet/ansible-nd/blob/develop/plugins/modules/nd_manage_fabric_ebgp_vxlan.py)
- [`plugins/modules/nd_manage_fabric_ibgp_vxlan.py`](https://github.com/CiscoDevNet/ansible-nd/blob/develop/plugins/modules/nd_manage_fabric_ibgp_vxlan.py)
- [`plugins/modules/nd_manage_fabric_external.py`](https://github.com/CiscoDevNet/ansible-nd/blob/develop/plugins/modules/nd_manage_fabric_external.py)

If additional fabric modules are merged before this issue is worked, include those as well before closing the issue.

## Items to cover

- Document common result keys such as `changed`, `failed`, `before`, `after`, `proposed`, and `sent` where applicable.
- Document verbosity-dependent API details returned through `format_with_verbosity(...)`.
- Make sure fabric modules using the shared ND state machine describe the same return contract consistently.
- Consider using a reusable docs fragment or generated helper so fabric modules do not drift over time.
- Add or update tests/checks so empty `RETURN` blocks are caught for fabric modules going forward.

## Origin

This was identified while reviewing PR #256. The PR-specific inline comment covers the new AI modules, but this issue is intended to ensure the same documentation gap is closed for fabric modules already merged into `develop`.

mtarking:
I audited the `RETURN` blocks across all fabric-family modules on `develop` and in the open fabric PRs.

**`develop` fabric modules — all now have complete, consistent `RETURN` blocks:**
- `nd_manage_fabric_ibgp_vxlan`
[...]

PR #403, opened 2026-07-15
nd_interface_loopback: policy_type discriminated union

### LINKED ISSUE #452: [ND 4.2+] Extend nd_interface_loopback for NX-OS and IOS-XE policy types
## Summary

Extend `nd_interface_loopback` to support multiple NX-OS and IOS-XE loopback policy templates through `policy_type` and `network_os_type` discriminated unions.

## Related PR

- #403

## Scope

- Require and validate `policy_type` and `network_os_type`.
- Support NX-OS policy types: `loopback`, `ipfmLoopback`, and `mplsLoopback`.
- Support IOS-XE policy types: `iosXeLoopback`, `iosXeLoopbackShutNoshut`, `iosXeUnderlayLoopback`, `iosXeInternalLoopback`, `csrLoopback`, and `csr1kvLoopback`.
- Keep read handling tolerant of controller-returned fields while keeping write validation strict.
- Group bulk creates by switch and policy type.
- Validate per-item HTTP 207 responses.
- Update documentation, examples, unit tests, and integration tests.

## Acceptance Criteria

- All supported NX-OS and IOS-XE policy types validate and reconcile correctly.
- Cross-OS policy mismatches fail before controller calls.
- Merged, replaced, overridden, deleted, and gathered workflows remain idempotent.
- Mixed policy types on one switch are sent in separate API requests.
- Unit, sanity, and live integration validation pass.

## Dependencies

- Merge and rebase after #422.
- Reconcile overlapping bare-IPv4 normalization with #402.
- Related shared bulk-create issue: #409.

### LINKED ISSUE #529: nd_interface_loopback: ipv6 typed as IPv6CIDR sends a prefix that ND 4.2.1 rejects (wants bare IPv6)
## Summary

`NexusLoopbackPolicyModel.ipv6` (in `plugins/module_utils/models/interfaces/loopback_interface.py`) is typed `IPv6CIDR` and its description says *"Loopback IPv6 address in CIDR notation"*. Against **live ND 4.2.1**, the Manage interfaces endpoint **rejects a CIDR `ipv6`** on a `policyType: loopback` create or update and requires a **bare IPv6 address**. This is the IPv6 half of #401, which noted *"The `ipv6` field (`IPv6CIDR`) likely has the same issue but was not verified"*.

The same under-validation exists on the IOS-XE branches of the `policy_type` union (PR #403): `iosXeInternalLoopback` `ip`/`ipv6` and `iosXeUnderlayLoopback` `secondary_ip` are modelled as plain strings because the OpenAPI schema leaves them untyped, but ND validates all three server-side as bare addresses.

## Evidence (live ND 4.2.1.10 lab, intent-only, no deploy)

Raw `POST`/`PUT` on `/api/v1/manage/fabrics/{fabric}/switches/{switchId}/interfaces`:

| Template / field | Input | ND result |
|---|---|---|
| `loopback` `ipv6` | `2001:db8:1210::1/128` | `HTTP 400` — `Policy [loopback] - Validation failed for following fields: [ipv6]` (POST and PUT) |
| `loopback` `ipv6` | `2001:db8:1210::1` | `HTTP 207` — `Interface added successfully`; GET echoes the bare address |
| `iosXeInternalLoopback` `ip` | `10.35.210.1/32` | `HTTP 400` — validation failed `[ip]` |
| `iosXeInternalLoopback` `ipv6` | `2001:db8:210::1/128` | `HTTP 400` — validation failed `[ipv6]` |
| `iosXeInternalLoopback` `ip` / `ipv6` | bare | accepted, echoed bare |
| `iosXeInternalLoopback` `ip` / `ipv6` | `not-an-ip` / `not-an-ipv6` | `HTTP 400` — validation failed |
| `iosXeUnderlayLoopback` `secondaryIp` | `10.35.211.2/32` or `garbage` | `HTTP 400` — validation failed `[secondaryIp]` |

The ND config templates confirm the wire contract: `int_loopback` declares `ipV6Address V6IP;` and renders it through `v6_loopback_interface` as `ipv6 address $$IP$$/128` (prefix hardcoded); `ios_xe_int_loopback_internal` declares `ipV4Address IP; ipV6Address IPv6;` (also rendered with a hardcoded `/128`); `ios_xe_int_underlay_loopback` declares `ipV4Address SECONDARY_IP;`.

Schema note: `intLoopbackTemplate.ipv6` is `format: ipv6` (bare) in the OpenAPI spec — the schema was right and the model is wrong. The three IOS-XE fields are untyped `string` in the spec while ND validates them; the spec is under-typed there.

## Impact

Any `nd_interface_loopback` task that sets `ipv6` in CIDR notation on an NX-OS loopback — the form the option description recommends — is rejected by ND 4.2.1 (`IPv6CIDR` passes bare input through, so bare values happen to work). Latent because no integration task sets `ipv6` and the integration suite does not run in CI.

## Suggested fix

Add an `IPv6Host` Annotated type (mirror of `IPv4Host`: accept bare or CIDR, normalize to the bare address via `ipaddress.IPv6Interface(value).ip`) and use it for the NX-OS `loopback` `ipv6` and the IOS-XE `iosXeInternalLoopback` `ipv6`; use `IPv4Host` for `iosXeInternalLoopback` `ip` and `iosXeUnderlayLoopback` `secondary_ip`. Add an `ipv6` case to the loopback integration target so the wire contract is exercised.

4.2.1

allenrobel:
> Formerly stacked on #387 (`nd_argument_specs_relocate`). #387 has merged, this PR is retargeted to `develop` and rebased on it, and remote CI now runs (and passes) on this branch.

## Merge Ordering — #422 merged 2026-08-28; hold lifted

~~**Do not merge this PR before #422**~~ #422 (replaced/overridden field-removal detection) merged into `develop` on 2026-08-28 and this branch is rebased on it (commit "Add per-policy reverse_diff_defaults tables for the loopback union (#422 rebase)" and the three commits before it). The original reasoning, kept for the record:

- #422 adds a reverse diff pass to `NDBaseModel.get_diff` plus per-model `reverse_diff_defaults` tables (ND echoes template defaults for unset fields, so the reverse pass must strip them or replaced/overridden idempotency breaks). Its entire footprint in the loopback model is one flat table: `{"adminState": True, "routeMapTag": "12345"}`.
- This PR restructures that same model into nine per-`policy_type` policy models (3 NX-OS + 6 IOS-XE), which relocates `routeMapTag` three levels deep into `NexusLoopbackPolicyModel` and means **each policy model needs its own `reverse_diff_defaults` table** sourced from its own OpenAPI template schema. That per-policy-type work belongs here — this PR introduces the new policy types — and the `_strip_reverse_diff_defaults` recursion in #422 already supports nested per-model tables.
- If this PR merged first, #422 (a collection-wide correctness fix touching all nine interface modules plus prefix-list) would have to grow nine loopback tables — including IOS-XE templates — before it could land. Merging #422 first keeps its loopback table scoped to the single `loopback` policy type that exists on `develop` today, and confines the semantic overlap to a well-defined rebase task on this branch.

**Rebase checklist for this branch after #422 lands** (all done 2026-08-28):

- [x] Move the `"routeMapTag": "12345"` entry into `NexusLoopbackPolicyModel.reverse_diff_defaults`; keep `adminState` at the level where it lives post-restructure.
- [x] Add `reverse_diff_defaults` tables for the eight new policy models (`ipfmLoopback`, `mplsLoopback`, and the six IOS-XE templates), each sourced from its policyType's OpenAPI template schema; lab-verify echoes where possible. Any policy model without a table silently loses replaced/overridden removal detection.
- [x] Verify the read-tolerant policy models do not retain unknown server-injected keys — any retained extra key would be counted as a removal by the reverse pass and break replaced/overridden idempotency (the 2026-07-18 XE probe found no ND-injected keys in XE policy reads, but re-check on the rebased code). **Re-check outcome:** nothing is retained — the read-path validator drops undeclared keys before `extra="forbid"` sees them, so `model_extra` is always empty; the new per-policy tests assert `to_reverse_diff_dict()` retains only `policyType` for a default-echo-plus-injected-key read on all nine models. The re-check did surface one real gap: the read-tolerance only fired under the branch-local `context={"mode": "read"}` set by a `LoopbackInterfaceModel.from_response` override, so a policy model read directly (as #422's tests do) still rejected injected keys. Fixed by keying the validator off the collection-wide `"response"` context and dropping the override (commit "Align loopback read-tolerance to the collection-wide `response` context").

**Table sources (ND 4.2.1 OpenAPI template schemas):** `adminState: true` is shared by all nine templates and lives on `LoopbackPolicyStrictBase`; `intLoopbackTemplate` adds `routeMapTag` (held as the coerced `"12345"`); `intIpfmLoopbackTemplate` adds `advertiseLoopback: true`, `isServiceReflect: false`, `vrfInterface: "default"`; `intMplsLoopbackTemplate` adds `dciRoutingProtocol: "isis"`, `dciRoutingTag: "MPLS_UNDERLAY"`; `iosXeIntLoopbackInternalTemplate` adds `enablePim: false`. The other five templates (`iosXeIntLoopbackTemplate`, `iosXeIntLoopbackAdminStateTemplate`, `iosXeIntUnderlayLoopbackTemplate`, `csrIntLoopbackTemplate`, `csr1kvLoopbackTemplate`) declare no default beyond `adminState` for any field their model exposes and inherit the base table. Lab verification of the echoes for the non-`loopback` types is still outstanding and will be run with the next full lab pass.

## Related Issue(s)

Relates to #401 (develop `ip=IPv4CIDR` handling). The bare-IPv4 normalization on this branch overlaps with open PR #402 — whichever lands second will rebase and drop the duplicate change.

Closes #529 (loopback `ipv6` CIDR rejection and IOS-XE address under-validation, found during this PR's lab verification).

The design/plan docs that accompanied the POC phase have been removed from the branch and archived outside the repo.

## Proposed Changes

Consolidates the loopback interface behind discriminated unions so a single module handles every ND loopback policy template, across both switch OSes.

**`policy_type` discriminated union (NX-OS):**

- Add `LoopbackPolicyTypeEnum` and a `LoopbackPolicyBase`; make `policy_type` **required and strict** on the write path.
- Add `IpfmLoopbackPolicyModel`, `MplsLoopbackPolicyModel`, and `SecondaryIpModel`; wire the discriminated union on `policy_type`.
- Extend the loopback argspec with `policy_type` and the per-template union fields.
- `query_all` manages `loopback`, `ipfmLoopback`, and `mplsLoopback` via a set filter.
- Read path tolerates ND-injected policy keys (`extra="forbid"` was aborting reads); the write path stays strict.
- Normalize loopback `ip` to bare IPv4 (ND 4.2.1 rejects CIDR on the wire).
- Docs for `policy_type` and the per-template fields.

**Required `network_os_type` discriminator:**

- Per discussion with Mike and Matt, the platform/OS selector is a **mandatory, user-supplied** parameter — `network_os_type` (mirrors the ND API's required `networkOSType` discriminator) — so all interface parameters validate before any controller call. A future enhancement makes it optional with a per-switch `FabricContext.get_platform_type()` fallback (#404).

**IOS-XE loopback surface (`choices: [nx-os, ios-xe]`):**

- `config_data.network_os` is an outer discriminated union (`NexusLoopbackNetworkOSModel | XeLoopbackNetworkOSModel`), each OS branch carrying its own `policy_type` union: NX-OS keeps `loopback`/`ipfmLoopback`/`mplsLoopback`; IOS-XE adds `iosXeLoopback`, `iosXeLoopbackShutNoshut`, `iosXeUnderlayLoopback`, `iosXeInternalLoopback`, `csrLoopback`, `csr1kvLoopback` (all six real XE templates; `userDefined` stays excluded on both OSes). Cross-OS mismatches (`nx-os` + `iosXeLoopback`) fail structurally in Pydantic — no hand-written cross-check.
- `LoopbackPolicyStrictBase` split: `extra="forbid"` + read-tolerant validator + `admin_state` shared by all nine branches; NX common fields stay on `LoopbackPolicyBase`.
- Six IOS-XE branch models with exact per-template fields (schema-verified); new flat options `secondary_ip`, `enable_pim`.
- `query_all` filter is the union of the NX and XE policy-type enums (`XeLoopbackPolicyTypeEnum`).
- Lab probe (2026-07-18, IOS-XE device in fabric ISN) proved the OpenAPI READ schema's `csrIntLoopback` never appears on the wire — ND echoes the create-side `csrLoopback` — so the model uses the single wire-verified name (drift recorded in the team bug-tracker vault: `csr-loopback-read-schema-name-drift`); the same probe found no ND-injected keys in XE policy reads.
- DOCUMENTATION/EXAMPLES cover all four states for `ios-xe` plus a `csrLoopback` example.
- XE integration tasks gated on `nd_test_xe_switch_ip` (skipped unless supplied), including a fabric-equality-guarded mixed NX+XE `overridden` variant.

**Lab-surfaced bug fixes:**

- `create_bulk` now groups one POST per (switch, policyType) — ND rejects mixed-policyType bulk creates (207 with a failed item, nothing created; `TODO(4.2.1)` marker referencing vault note `bulk-interface-create-rejects-mixed-policy-types`). Per-item 207 `results[]` failures on both create paths are detected centrally by `NdV1Strategy.is_success` (#398, merged 2026-08-11) — the orchestrator-level stopgap this branch originally carried was removed in commit "Drop the orchestrator-level 207 results[] stopgap superseded by #398" once #398 landed on `develop`.
- Integration loopback IDs renumbered 1xx→2xx: the lab's new multisite topology auto-provisions `loopback100` (ND's default multisite loopback ID) on border gateways, colliding with the old reserved range.

## Test Notes

- **Integration:** `nd_interface_loopback` integration tests are **green** against these changes (merged/replaced/overridden/query, including the added `policy_type: loopback` tasks).
- **Unit:** all unit tests **green** — models (`test_loopback_interface.py`), orchestrator, and `nd_argument_specs` suites.
- **2026-07-16 (`network_os_type` commits):** integration tests re-run **green** against a live ND 4.2.1 lab with the now-required `network_os_type: nx-os` in every `network_os` block; full unit suite (3195 tests) **green**; `validate-modules` sanity **clean** on `nd_interface_loopback.py`.
- **2026-07-18 (IOS-XE commits):** full unit suite (3208 tests) **green**; `validate-modules` **clean**; black/isort/pylint/mypy clean on changed files; ansible-lint **clean** on the integration target. Live-lab raw-REST probe verified XE create/read/delete behavior on a real IOS-XE device (intent-only). **XE integration run against the live lab: all tests pass** (all four states + idempotency on the IOS-XE device, `iosXeLoopback` and `csrLoopback`), which also verifies the csr single-name decision on the full deploy path — the deployed read echoes `csrLoopback`, not the read-schema's `csrIntLoopback`. The first run surfaced two real module bugs (mixed-policyType bulk-create rejection masked by unchecked 207s) — fixed with TDD unit coverage, then the re-run went green. (No CI on stacked PRs — local verification per team convention.)
- **2026-08-28 (#422 rebase commits):** rebased onto `develop` @ `bb985d74` (#422 merged); one conflict in `loopback_interface.py` (the #422 table vs. the policy-model restructure) resolved by splitting the table per class. #422's new develop-side tests (`test_base_model_reverse_diff.py`, the `test_nd_state_machine.py` removal tests) adapted to the required `network_os_type` / `policy_type` discriminators. 27 new unit cases cover the nine per-policy tables (default-echo idempotency, non-default removal detection, XE-union recursion). Full unit suite **4199 passed**; black/isort/pylint/mypy **clean on the changed lines** (pre-existing argspec `dict()` R1735 hints and test-file mypy noise unchanged).
- **2026-08-27 (review-round commits):** full `nd_interface_loopback` target run from the current head against a live ND 4.2.1.10 lab (SITE1 NX-OS + ISN IOS-XE): `ok=81 changed=29 failed=0 skipped=4` (skips are the credential preflight and the conditional mixed-override block). Includes the repaired CIDR task, which now also sends a CIDR `ipv6` and asserts the bare after-state plus idempotency (#529). Model unit suite (108 tests) **green**; `validate-modules`/`yamllint` sanity, black/isort/pylint/mypy **clean** on changed files. Raw-REST lab probes behind the `IPv6Host`/`IPv4Host` retyping are tabulated in #529.

4.2.1

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist


plugins/module_utils/models/interfaces/loopback_interface.py
```diff
@@ -17,49 +17,105 @@
...
+            data = {key: value for key, value in data.items() if key in allowed}
+        return data
+
+
+class LoopbackPolicyBase(LoopbackPolicyStrictBase):
+    """
+    # Summary
+
+    Shared policy fields common to every managed NX-OS loopback template. Inherits write-strict / read-tolerant behavior and
+    `admin_state` from `LoopbackPolicyStrictBase`.
```
akinross, line 99: if this is specific to NX-OS why not reflect this in the class name? For Xe you put it in front `XeLoopbackPolicyModel`

plugins/module_utils/orchestrators/loopback_interface.py
```diff
@@ -105,11 +117,45 @@ def create(self, model_instance: LoopbackInterfaceModel, **kwargs) -> ResponseTy
...
             result = self._request(path=api_endpoint.path, verb=api_endpoint.verb, data=request_body)
+            # Validate the full envelope, not `result`: _request() returns the already-unwrapped DATA dict,
+            # on which the helper's DATA.results lookup would always no-op and mask real 207 failures.
+            self._raise_on_failed_result_items(self.rest_send.response_current)
             self._queue_deploy(model_instance.interface_name, switch_id)
             return result
         except Exception as e:
             raise RuntimeError(f"Create failed for {model_instance.get_identifier_value()}: {e}") from e
 
+    def _raise_on_failed_result_items(self, response: dict | None) -> None:
```
akinross: is this loopback specific function or could we design it to be a more generic utils helper?

plugins/module_utils/orchestrators/loopback_interface.py
```diff
@@ -152,34 +198,62 @@ def delete(self, model_instance: LoopbackInterfaceModel, **kwargs) -> None:
         self._queue_remove(model_instance.interface_name, switch_id)
         self._queue_deploy(model_instance.interface_name, switch_id)
 
+    @staticmethod
+    def _extract_policy_type(model_instance: LoopbackInterfaceModel) -> str | None:
```
akinross: did you consider exposing the policy_type as a property in the model itself? something like below

```
class LoopbackInterfaceModel(NDBaseModel):
    ...
    @property
    def policy_type(self) -> str | None:
        if (config_data := self.config_data) is None:
            return None
        if (network_os := config_data.network_os) is None:
            return None
        if (policy := network_os.policy) is None:
            return None
        return policy.policy_type

File: plugins/module_utils/orchestrators/loopback_interface.py [hunk 4]
```diff
@@ -152,34 +198,62 @@ def delete(self, model_instance: LoopbackInterfaceModel, **kwargs) -> None:
...
+            # TODO(4.2.1) bulk-interface-create-rejects-mixed-policy-types
+            # ND rejects an interfaces[] array mixing policyType values (207 with a single failed item; nothing is
+            # created), even though the create schema allows mixed arrays. One POST per (switch, policyType).
+            groups: dict[tuple[str, str | None], list[tuple[str, dict]]] = defaultdict(list)
             for model_instance in model_instances:
                 switch_id = self._resolve_switch_id(model_instance.switch_ip)
                 payload = model_instance.to_payload()
                 payload["switchId"] = switch_id
-                groups[switch_id].append((model_instance.interface_name, payload))
+                group_key = (switch_id, self._extract_policy_type(model_instance))
```
akinross: should we have a separate function that construct the groups?

plugins/module_utils/orchestrators/loopback_interface.py
```diff
@@ -152,34 +198,62 @@ def delete(self, model_instance: LoopbackInterfaceModel, **kwargs) -> None:
...
         - If any create API request fails.
+        - If a create response's `DATA.results[]` contains an item whose `status` is not exactly `"success"` (see
+          `_raise_on_failed_result_items`); no deploy is queued for that group's items in that case.
         """
         try:
-            groups: dict[str, list[tuple[str, dict]]] = defaultdict(list)
+            # TODO(4.2.1) bulk-interface-create-rejects-mixed-policy-types
+            # ND rejects an interfaces[] array mixing policyType values (207 with a single failed item; nothing is
+            # created), even though the create schema allows mixed arrays. One POST per (switch, policyType).
+            groups: dict[tuple[str, str | None], list[tuple[str, dict]]] = defaultdict(list)
```
akinross: should we leverage a small data class or named tuple to increase readability that represents the item that contains the interface_name/payload and also for the group key?

```
@dataclass(frozen=True, slots=True)
class _BulkCreateItem:
    interface_name: str
    payload: dict

@dataclass(frozen=True, slots=True)
class _BulkCreateGroupKey:
    switch_id: str
    policy_type: str | None

groups: dict[_BulkCreateGroupKey, list[_BulkCreateItem]] = defaultdict(list)
```

plugins/module_utils/orchestrators/loopback_interface.py
```diff
@@ -152,34 +198,62 @@ def delete(self, model_instance: LoopbackInterfaceModel, **kwargs) -> None:
...
                 switch_id = self._resolve_switch_id(model_instance.switch_ip)
                 payload = model_instance.to_payload()
                 payload["switchId"] = switch_id
-                groups[switch_id].append((model_instance.interface_name, payload))
+                group_key = (switch_id, self._extract_policy_type(model_instance))
+                groups[group_key].append((model_instance.interface_name, payload))
 
             results = []
-            for switch_id, items in groups.items():
+            for group_key, items in groups.items():
```
akinross, line 244: if decide the use tuple why not unpack in the loop?


```suggestion
            for (switch_id, policy_type), items in groups.items():
```

plugins/module_utils/orchestrators/loopback_interface.py
```diff
@@ -152,34 +198,62 @@ def delete(self, model_instance: LoopbackInterfaceModel, **kwargs) -> None:
...
+                switch_id = group_key[0]
                 # Guarded at runtime by @requires_bulk_support("supports_bulk_create")
                 api_endpoint = self._configure_endpoint(self.create_bulk_endpoint(), switch_sn=switch_id)  # pyright: ignore[reportOptionalCall]
                 request_body = {"interfaces": [payload for interface_name, payload in items]}
                 result = self._request(path=api_endpoint.path, verb=api_endpoint.verb, data=request_body)
+                # Validate the full envelope, not `result`: _request() returns the already-unwrapped DATA dict,
+                # on which the helper's DATA.results lookup would always no-op and mask real 207 failures.
+                self._raise_on_failed_result_items(self.rest_send.response_current)
```
akinross: why pass in self.rest_send.response_current, the function has access to it from self?

plugins/module_utils/orchestrators/loopback_interface.py
akinross, line 254: why is this big try/except needed when the error will surface up and its already properly raised / explained by the called functions? Should we consider raising them at those functions that could raise the runtime errors if we want to expand the message?

akinross (changes requested):

plugins/module_utils/models/interfaces/loopback_interface.py
allenrobel: Renamed: `LoopbackPolicyBase` → `NexusLoopbackPolicyBase`, `LoopbackPolicyModel` → `NexusLoopbackPolicyModel` (commit "Address review: Nexus prefix, policy_type property, bulk-create dataclasses"). Went with `Nexus` to match the existing `NexusLoopbackNetworkOSModel`. Left `IpfmLoopbackPolicyModel`/`MplsLoopbackPolicyModel` as-is — like the `Csr*` classes on the XE side, they mirror their wire `policyType` discriminator exactly, which already identifies the OS; the prefix is added only where the wire name carries no OS hint.

plugins/module_utils/orchestrators/loopback_interface.py
allenrobel: Held off on generalizing: `_raise_on_failed_result_items` is deliberately temporary. It carries `TODO(4.2.1) multi-status-207-status-field-inconsistent` and is slated for removal once PR #398 lands 207 handling in `NdV1Strategy`, so promoting it to a shared util would work against that removal. If #398 stalls, promoting it then is fair game.

plugins/module_utils/orchestrators/loopback_interface.py
allenrobel: Implemented as suggested — `policy_type` is now a property on `LoopbackInterfaceModel`, and the orchestrator's `_extract_policy_type` static method is removed (commit "Address review: Nexus prefix, policy_type property, bulk-create dataclasses"). Since the model is typed, only `config_data` and `policy` need None-checks.

plugins/module_utils/orchestrators/loopback_interface.py
allenrobel: Extracted into `_group_by_switch_and_policy_type()`; the mixed-policy-type TODO marker moved with it (commit "Address review: Nexus prefix, policy_type property, bulk-create dataclasses").

plugins/module_utils/orchestrators/loopback_interface.py
allenrobel: Adopted your sketch: frozen, slotted `_BulkCreateGroupKey` and `_BulkCreateItem` (commit "Address review: Nexus prefix, policy_type property, bulk-create dataclasses").

plugins/module_utils/orchestrators/loopback_interface.py
allenrobel: Superseded by the dataclass change from your other comment — the loop now reads `group_key.switch_id` / `item.payload` instead of tuples.

plugins/module_utils/orchestrators/loopback_interface.py
allenrobel: Kept the explicit parameter: the call-site comment documents why the full envelope (not `_request()`'s already-unwrapped `DATA` return) must be validated, and passing it keeps that data flow visible where it matters. The helper is also slated for removal with PR #398 (see thread above), so I'm minimizing churn on it.

plugins/module_utils/orchestrators/loopback_interface.py
allenrobel: This mirrors the established pattern in every method of this orchestrator and the sibling interface orchestrators on develop (`create`, `update`, `query_one`, `query_all` all wrap the same way) — the re-raise adds the operation + identifier context the state machine surfaces to the user. Agreed it's worth revisiting, but changing it consistently is a collection-wide refactor beyond this PR's scope (it overlaps the deferred refactor batch in #357).

allenrobel: @akinross I added a comment to #348 referencing your concern so that it's not lost as we work on the refactor.

plugins/modules/nd_interface_loopback.py
```diff
@@ -55,41 +55,131 @@
...
                     description:
                     - The administrative state of the loopback interface.
                     - It defaults to C(true) when unset during creation.
+                    - Applies to all policy_type values.
                     type: bool
                   ip:
                     description:
                     - The IPv4 address of the loopback interface.
+                    - Applies to all policy_type values except C(iosXeLoopbackShutNoshut) and C(csr1kvLoopback).
                     type: str
```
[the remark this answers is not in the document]
allenrobel: Documented in commit "Docs: note CIDR ip input is normalized to bare IPv4" — the `ip` option description now states that bare or CIDR input is accepted, and CIDR input is normalized to the bare address sent to the controller and returned in module output.

akinross (dismissed):

plugins/module_utils/models/interfaces/loopback_interface.py
```diff
@@ -108,7 +364,7 @@ class LoopbackConfigDataModel(NDNestedModel):
     """
 
     mode: Literal["managed"] = Field(default="managed", alias="mode", frozen=True)
-    network_os: LoopbackNetworkOSModel = Field(alias="networkOS")
+    network_os: NexusLoopbackNetworkOSModel | XeLoopbackNetworkOSModel = Field(alias="networkOS", discriminator="network_os_type")
```
mikewiebe, line 367: ### High: `merged` cannot transition between loopback policy types

**Issue**

The new discriminated union represents different loopback policy types with
different concrete Pydantic classes, but `NDBaseModel.merge()` rejects a nested
model when its concrete type differs from the existing model.

**Evidence**

- [`loopback_interface.py` line 367](https://github.com/CiscoDevNet/ansible-nd/pull/403/files#diff-2ceae40d02485f26cffe304a7f6cde1570c957a3f0005b9b23da76742d9cec94R367)
  adds the outer discriminated union; its nested `policy` field similarly
  selects a different concrete model for each `policy_type`.
- [`base.py` lines 269-293](https://github.com/CiscoDevNet/ansible-nd/blob/3acf5e306d9232207589dc6ce96013a2bcdad39c/plugins/module_utils/models/base.py#L269-L293)
  recursively merges direct nested models and raises `TypeError` when their
  concrete classes differ.
- An exact-head `state: merged` reproduction for an existing `loopback` policy
  and proposed `mplsLoopback` policy failed before any API request with:

  ```text
  TypeError: Cannot merge MplsLoopbackPolicyModel into NexusLoopbackPolicyModel.
  ```

**Practical example**

The same loopback interface can validly move between these policy templates:

| State | `network_os_type` | `policy_type` | Concrete policy model |
| --- | --- | --- | --- |
| Existing controller state | `nx-os` | `loopback` | `NexusLoopbackPolicyModel` |
| Desired merged state | `nx-os` | `mplsLoopback` | `MplsLoopbackPolicyModel` |

`NDStateMachine` identifies the same interface and invokes
`existing.merge(proposed)`. The merge descends into `policy`, sees the two
different classes, and raises instead of replacing the discriminated-union
branch with the requested policy.

**Existing PR overlap**

No matching existing PR comment found. The 19 current inline comments discuss
model naming, bulk grouping, and IPv4 normalization, but not merged-state
behavior when a discriminator changes.

[#386](https://github.com/CiscoDevNet/ansible-nd/issues/386) tracks
policy-aware merging for nested lists and dictionaries, not replacement of a
direct discriminated-union model when its type changes. Issue
[#452](https://github.com/CiscoDevNet/ansible-nd/issues/452) tracks the broader
loopback policy expansion but does not identify this concrete transition
failure.

**Impact**

A valid `state: merged` policy transition fails locally before Nexus Dashboard
receives a request. Users cannot reconcile an existing loopback to another
supported policy type through the normal merged workflow, despite both policy
types being accepted by the module contract.

**Suggested fix**

When a direct nested field is a discriminated union and its discriminator
changes, replace that branch with the proposed model instead of recursively
merging incompatible classes. Preserve recursive merge behavior for matching
types, and add end-to-end merged-state tests for policy-type and network-OS
transitions in both check mode and normal execution.

---

mikewiebe: @allenrobel not sure if we should support a transition like this so I am fine if you intentially reject this kind of transition.  Let me know what you think

allenrobel: Confirmed — reproduced exactly as described; the merge recursion hits the two concrete policy classes and fails before any API request.

Decision: intentionally reject the transition, per your suggestion. A field-by-field merge across two policy templates is undefined anyway — the branches have disjoint field sets — so "Reject policy_type transitions in merged state with an actionable error (#403 review)" replaces the internal `TypeError: Cannot merge MplsLoopbackPolicyModel into NexusLoopbackPolicyModel` with:

```
Cannot change policy_type from 'loopback' to 'mplsLoopback' with state: merged. Use state: replaced (or delete and re-create the resource) to change policy_type.
```

The guard keys off the union's discriminator, so it also covers the `networkOSType` level and any future discriminated unions in other interface models; non-union type mismatches keep the existing `TypeError`.

The transition itself remains available via `state: replaced`/`overridden` — those paths never call `merge()`; the proposed model goes to ND wholesale as a PUT. Lab-verified (intent-only): a PUT changing an existing loopback's policy from `loopback` to `ipfmLoopback` returned 204 and the GET echoes the new policy. Caveat: the lab currently runs ND 4.3.1.145 (dev build), not the 4.2.1 this PR targets. Same probe surfaced two 4.3.1-specific behaviors unrelated to this thread — `mplsLoopback` PUT requires a populated `MPLS_LOOPBACK_IP_POOL` even with an explicit `ip`, and the create-side gateway now enforces the discriminator enum (schema-rejecting `mplsLoopback` create, which 4.2.1 accepts) — tracking those separately.

plugins/module_utils/models/interfaces/loopback_interface.py
```diff
@@ -81,19 +144,212 @@ def coerce_route_map_tag(cls, value):
         return str(value)
 
 
-class LoopbackNetworkOSModel(NDNestedModel):
+class SecondaryIpModel(NDNestedModel):
```
mikewiebe, line 147: ### Medium: IPFM secondary addresses accept non-IPv4 values

**Issue**

`SecondaryIpModel.ip` is an unrestricted string even though the API schema
defines each IPFM secondary-list item as an IPv4 address. Malformed text and
IPv6 pass local validation and check mode and are serialized to the controller.

**Evidence**

- [`loopback_interface.py` lines 147-159](https://github.com/CiscoDevNet/ansible-nd/pull/403/files#diff-2ceae40d02485f26cffe304a7f6cde1570c957a3f0005b9b23da76742d9cec94R147)
  declares `ip: str | None`.
- The bundled OpenAPI `interfaceSecondaryIpList` schema references
  `interfaceIp`, whose format is `ipv4`.
- Exact-head probes accepted and serialized both:

  ```text
  {'ip': 'not-an-ip', 'prefix': 32}
  {'policyType': 'ipfmLoopback',
   'secondaryIpList': [{'ip': '2001:db8::1', 'prefix': 32}]}
  ```
- A user can therefore run this task successfully in check mode even though
  the secondary address is IPv6 in an IPv4-only field:

  ```yaml
  - name: Preview an IPFM loopback with an invalid secondary address
      fabric_name: lab_fabric
      config:
        - switch_ip: 192.0.2.10
          interface_name: loopback11
          config_data:
            network_os:
              network_os_type: nx-os
              policy:
                policy_type: ipfmLoopback
                ip: 10.2.2.2
                vrf: default
                secondary_ip_list:
                  - ip: 2001:db8::1
                    prefix: 32
      state: merged
    check_mode: true
  ```

  The module plans and serializes the change instead of rejecting
  `secondary_ip_list[0].ip` as non-IPv4. Without check mode, that invalid value
  is sent to Nexus Dashboard.

**Existing PR overlap**

No matching existing PR comment found. The existing IPv4 discussion covers
normalization of the primary loopback address, not validation of IPFM
secondary-list entries.

No matching open issue found. [#401](https://github.com/CiscoDevNet/ansible-nd/issues/401)
tracks the primary loopback address's bare-host normalization, and #452 is the
parent feature issue; neither records malformed or IPv6 secondary-address
acceptance.

**Impact**

Invalid configuration survives Ansible validation, appears valid in check mode,
and fails only at the controller or produces version-dependent behavior. Users
receive a later, less actionable API error.

**Suggested fix**

Use the collection's strict IPv4-host type or `IPv4Address` for
`SecondaryIpModel.ip`. Add tests accepting a valid IPv4 host and rejecting
malformed text, IPv6, and any CIDR form the item schema does not support.

---

plugins/modules/nd_interface_loopback.py
```diff
@@ -242,6 +356,124 @@
...
+          network_os:
+            network_os_type: ios-xe
+            policy:
+              policy_type: csrLoopback
+              admin_state: true
+              ip: "10.200.101.1"
+    state: merged
 """
 
 RETURN = r"""
```
mikewiebe, line 480: ### Low: The module still has an empty RETURN contract

**Issue**

The PR substantially expands the module's public contract and examples, but
`RETURN` remains empty, so generated module documentation does not describe
the state-machine result users receive.

**Evidence**

- [`nd_interface_loopback.py` line 479](https://github.com/CiscoDevNet/ansible-nd/pull/403/files#diff-d50337094881bfcdeaa3a4dcbd3933ec095d76bbe211b609f37c25f252d3e914R479)
  contains an empty `RETURN = r""" """` block.
- The sibling
  [`nd_interface_port_channel_access.py` RETURN block](https://github.com/CiscoDevNet/ansible-nd/blob/develop/plugins/modules/nd_interface_port_channel_access.py#L354)
  documents the canonical common keys.

**Existing PR overlap**

No matching open issue found. [#358](https://github.com/CiscoDevNet/ansible-nd/issues/358)
is limited to fabric-family modules and does not include
`nd_interface_loopback`.

**Impact**

Users and generated Ansible docs cannot determine the shape, availability, or
verbosity conditions of `before`, `after`, `diff`, `proposed`, and
diagnostic output without reading implementation code.

**Suggested fix**

Populate `RETURN` using the port-channel-access sibling as the canonical
pattern. Document `changed`, `output_level`, `before`, `after`, `diff`,
`proposed`, `logs`, and failure `msg`, with loopback-specific nested
samples and accurate `returned`, `type`, and `elements` metadata.

plugins/module_utils/models/interfaces/loopback_interface.py
allenrobel: Fixed in commit "Validate IPFM secondary IPs as bare IPv4; populate RETURN (#403 review)". `SecondaryIpModel.ip` now uses a new `IPv4HostStrict` Annotated type. Went strict-bare rather than reusing `IPv4Host`: the item schema is `format: ipv4` with the mask length in the sibling `prefix` field, so CIDR input is ambiguous (`10.2.2.3/24` vs `prefix: 32`) and is rejected rather than normalized — unlike the primary `ip`, which has no companion prefix field. Tests cover valid bare IPv4 accepted, plus malformed text, IPv6, and CIDR rejected. The XE underlay `secondaryIp` intentionally stays an unvalidated string — its schema (`interfaceSecondaryIp`) declares no `ipv4` format, matching the `iosXeInternalLoopback` precedent.

plugins/modules/nd_interface_loopback.py
allenrobel: Populated in commit "Validate IPFM secondary IPs as bare IPv4; populate RETURN (#403 review)", using `nd_interface_port_channel_access` as the canonical pattern — `changed`, `output_level`, `before`, `after`, `diff`, `proposed`, `logs`, and failure `msg`, with loopback-specific samples and matching `returned`/`type`/`elements` metadata. `validate-modules` sanity passes.

mikewiebe:
I’m holding approval on this PR for three reasons:

1. **The 207-response handling regressed.** The previous loopback-specific check accepted only an explicit `status: success`. The shared handler now treats missing, `warning`, `notexecuted`, and unknown statuses as successful. This can make the module report success and queue deployment for an interface that ND did not accept. [Issue #397](https://github.com/CiscoDevNet/ansible-nd/issues/397) explicitly says these allowlist checks must remain until the shared handler supports equivalent behavior.

   **Practical example**

   An operator requests creation and deployment of `loopback10`:

   ```yaml
   - name: Create loopback10
       fabric_name: lab-fabric
       config:
         - switch_ip: 10.122.84.181
           interface_name: loopback10
           config_data:
             network_os:
               network_os_type: nx-os
               policy:
                 policy_type: loopback
                 ip: 10.10.10.10
       state: merged
       config_actions:
         deploy: true
   ```

   ND rejects the interface but returns HTTP 207 without an item-level `status`:

   ```json
   {
     "RETURN_CODE": 207,
     "MESSAGE": "Multi-Status",
     "DATA": {
       "results": [
         {
           "name": "loopback10",
           "message": "invalid policyType"
         }
       ]
     }
   }
   ```

   Because the item has no recognized failure status, the current denylist classifies the response as successful, queues `loopback10` for deployment, and may report `changed: true`. It should instead fail with `loopback10: invalid policyType` and never queue deployment.

   Please restore the loopback-specific check or add shared exact-success handling, with tests for missing and non-success status values.

2. **`overridden` can target fabric-created MPLS loopbacks for deletion.** The module includes `mplsLoopback` in its owned-resource set. Consequently, an MPLS loopback returned by ND but omitted from the desired `overridden` configuration is queued for removal and deployment.

   MPLS loopbacks are fabric-created and historically update-only. They are also absent from the reviewed OpenAPI create discriminator. Please treat them as update-only: allow updates to known MPLS loopbacks, reject explicit creation/deletion, and exclude them from `overridden` deletion scope. Add a regression test proving an omitted MPLS loopback is preserved. [Issue #452](https://github.com/CiscoDevNet/ansible-nd/issues/452) is related, but does not currently define this ownership protection.

3. **The declared #422 dependency is incomplete.** The PR says it must wait for [PR #422](https://github.com/CiscoDevNet/ansible-nd/pull/422), which is still open. Combining the current branch heads also produces a conflict in `loopback_interface.py`.

   Please merge #422, rebase this PR, resolve the conflict, rerun the tests, and request another review.

Once these items are addressed, I can re-review promptly. The other tracked follow-up work and low-severity RETURN documentation issue are not blocking approval.

tests/integration/targets/nd_interface_loopback/tasks/merged.yaml
```diff
@@ -134,46 +136,46 @@
...
     state: deleted
 
 # --- MERGED WITH CIDR-format ip ---
 # Regression guard for #401: live ND 4.2.1 rejects a CIDR ip on a loopback create (400) and stores/returns the bare
 # host form. The model normalizes CIDR input to bare before the wire, so a CIDR ip must (a) create successfully with a
 # bare ip in the after-state and (b) be idempotent on re-apply against ND's bare wire form.
 
-- name: "MERGED CIDR: Create loopback104 with a CIDR-format ip (normal mode)"
-  cisco.nd.nd_interface_loopback: &merge_lb104_cidr
+- name: "MERGED CIDR: Create loopback204 with a CIDR-format ip (normal mode)"
```
mikewiebe, line 153: ### Medium: The newly rebased CIDR integration workflow cannot pass argument validation

**Issue**

The CIDR workflow inherited during the latest rebase still uses the old defaulted loopback shape. PR #403 requires both nested discriminators, so this task fails locally before contacting ND.

**Evidence**

- [`merged.yaml` supplies `network_os.policy` without either required discriminator](https://github.com/CiscoDevNet/ansible-nd/pull/403/files#diff-4c73904284c4aa034f34815d65616287aefeb6ea3f91dfba895b8f2f52457102R153).
- [The current argspec requires `network_os_type`](https://github.com/CiscoDevNet/ansible-nd/pull/403/files#diff-2ceae40d02485f26cffe304a7f6cde1570c957a3f0005b9b23da76742d9cec94R474) and [requires `policy_type`](https://github.com/CiscoDevNet/ansible-nd/pull/403/files#diff-2ceae40d02485f26cffe304a7f6cde1570c957a3f0005b9b23da76742d9cec94R478).
- Exact module invocation failed first with `missing required arguments: network_os_type found in config -> config_data -> network_os`. Adding only `network_os_type: nx-os` then failed with `missing required arguments: policy_type found in config -> config_data -> network_os -> policy`.
- On `develop`, the same task works because those values default to `nx-os` and `loopback`. PR #403 deliberately makes them required. The latest commit only renumbered `loopback104` to `loopback204`; it did not adapt the inherited task.
- GitHub's `Integration Tests` job is skipped, so the otherwise green CI matrix does not execute this failure.

**Existing PR overlap**

No matching open issue found. Issue [#452](https://github.com/CiscoDevNet/ansible-nd/issues/452) generally requires live integration to pass, but it does not identify this concrete invalid task introduced by the rebase.

**Impact**

The committed integration target always fails when it reaches this task, so the PR cannot currently produce a successful live target run and the CIDR/idempotency regression guard is ineffective.

**Suggested fix**

Add `network_os_type: nx-os` under `network_os` and `policy_type: loopback` under `policy`, then run the complete target against live ND. A unit or static contract check should also validate module calls embedded in integration YAML so this is caught when integration CI is skipped.

---

plugins/module_utils/models/interfaces/loopback_interface.py
```diff
@@ -81,19 +144,212 @@ def coerce_route_map_tag(cls, value):
...
+    ## Raises
+
+    None
+    """
+
+    policy_type: Literal["iosXeUnderlayLoopback"] = Field(alias="policyType", description="IOS-XE underlay loopback policy template discriminator")
+    description: AsciiDescription = Field(default=None, alias="description", min_length=1, max_length=254, description="Interface description")
+    extra_config: str | None = Field(default=None, alias="extraConfig", description="Additional CLI for the interface")
+    ip: IPv4Host = Field(default=None, alias="ip", description="Loopback IPv4 address (bare host form; CIDR input is accepted and normalized)")
+    secondary_ip: str | None = Field(default=None, alias="secondaryIp", description="Secondary IP address of the NVE interface loopback")
```
akinross: why is this input a string and not a ip or SecondaryIpModel type? or not marked similar to XeInternalLoopbackPolicyModel ip and ipv6 fields, unvalidated string per the ND schema for this template

another questions is even when the unvalidated string per the ND schema for this template is there should we not enforce the validation to prevent non ip values being set? could the ND schema be faulty in this case? I think we should test this behaviour and based on that determine what we should validate

plugins/modules/nd_interface_loopback.py
```diff
@@ -55,41 +55,134 @@
...
                   admin_state:
                     description:
                     - The administrative state of the loopback interface.
                     - It defaults to C(true) when unset during creation.
+                    - Applies to all policy_type values.
                     type: bool
                   ip:
                     description:
                     - The IPv4 address of the loopback interface.
+                    - Accepts bare (C(10.1.1.1)) or CIDR (C(10.1.1.1/32)) input. CIDR input is normalized to the bare address, which is
```
akinross, line 100: iosXeInternalLoopback accepts an arbitrary string without normalization so this doc is not entirely correct i would say

plugins/modules/nd_interface_loopback.py
akinross: should we specify clearly that any resource is related to the defined loopback policy types from this module?

plugins/modules/nd_interface_loopback.py
```diff
@@ -124,8 +217,12 @@
...
-- The IP Fabric for Media (C(ipfmLoopback)) and user-defined (C(userDefined)) loopback policies will be managed by dedicated modules.
+- This module supports both NX-OS and IOS-XE loopback interfaces (interface_type C(loopback)), selected via
+  O(config[].config_data.network_os.network_os_type).
+- This module manages three NX-OS loopback policy templates, selected via O(config[].config_data.network_os.policy.policy_type)
+  C(loopback), C(ipfmLoopback) (IP Fabric for Media), and C(mplsLoopback), plus six IOS-XE managed templates
+  C(iosXeLoopback), C(iosXeLoopbackShutNoshut), C(iosXeUnderlayLoopback), C(iosXeInternalLoopback), C(csrLoopback), and
+  C(csr1kvLoopback). The user-defined (C(userDefined)) loopback policy is not yet supported for either network OS.
 """
 
 EXAMPLES = r"""
```
akinross, line 241: some of the examples do not define deploy=true but do not explicitly mention without deploying as some examples do, minor inconsistency that requires cleanup

plugins/modules/nd_interface_loopback.py
```diff
@@ -55,41 +55,134 @@
...
+                    - Additional CLI configuration commands to apply to the interface.
+                    - Applies to all policy_type values except C(iosXeLoopbackShutNoshut).
                     type: str
                   vrf:
                     description:
                     - The VRF to which the loopback interface belongs.
                     - Maximum 32 characters.
+                    - Applies when policy_type is C(loopback), C(ipfmLoopback), C(iosXeLoopback), C(iosXeInternalLoopback), or C(csrLoopback).
+                    type: str
+                  ipv6:
```
akinross, line 121: NX-OS ipv6 requires CIDR notation, while IOS-XE internal IPv6 is unvalidated which should be added to description

plugins/modules/nd_interface_loopback.py
```diff
@@ -55,41 +55,134 @@
...
+                type: str
+                required: true
+                choices: [ nx-os, ios-xe ]
               policy:
                 description:
                 - The policy configuration for the loopback interface.
+                - The policy fields present depend on O(config[].config_data.network_os.policy.policy_type).
                 type: dict
                 suboptions:
+                  policy_type:
```
akinross, line 74: is it intended that not all policy types have examples?

plugins/modules/nd_interface_loopback.py
```diff
@@ -261,9 +376,211 @@
...
+  description: Whether the module changed, or in check mode would change, the interface configuration.
+  returned: always
+  type: bool
+  sample: true
+output_level:
+  description: The output verbosity level in effect for the run, echoing the O(output_level) parameter.
+  returned: always
+  type: str
+  sample: normal
+before:
```
akinross, line 610: interface_type and mode are hardcoded model defaults which are included in the sample always

plugins/modules/nd_interface_loopback.py
```diff
@@ -261,9 +376,211 @@
...
+    interface_name: loopback10
+    config_data:
+      network_os:
+        network_os_type: nx-os
+        policy:
+          policy_type: loopback
+          admin_state: true
+          ip: 10.1.1.1
+          vrf: management
+after:
```
akinross, line 630: interface_type and mode are hardcoded model defaults which are included in the sample always

plugins/modules/nd_interface_loopback.py
```diff
@@ -261,9 +376,211 @@
...
+    interface_name: loopback10
+    config_data:
+      network_os:
+        network_os_type: nx-os
+        policy:
+          policy_type: loopback
+          admin_state: true
+          ip: 10.1.1.2
+          vrf: management
+diff:
```
akinross, line 650: while testing I noticed that diff is always empty, is this intended because this will be implemented later? i also do not see any tests that specify assert the diffing.

similar behaviour appears to be the case for log, but requires more testing

akinross (changes requested):

tests/integration/targets/nd_interface_loopback/tasks/merged.yaml
allenrobel: Good catch — the CIDR task came over from develop in the rebase with the old defaulted shape. Fixed in "Integration: add required discriminators to the CIDR merge task (#403 review)". Full `nd_interface_loopback` target run from the current head against ND 4.2.1.10 (SITE1 + ISN): `ok=81 changed=29 failed=0 skipped=4` (skips are the credential preflight and the conditional mixed-override block). While lab-verifying this I also found that ND rejects a CIDR `ipv6` on NX loopbacks — filed #529 and fixed in "Validate loopback ipv6 and IOS-XE internal/underlay addresses as bare hosts"; the same CIDR task now exercises `ipv6` too. Agree a static check of module calls embedded in integration YAML would have caught this while the Integration Tests job is skipped — tracking that separately rather than growing this PR.

plugins/module_utils/models/interfaces/loopback_interface.py
allenrobel: Lab-verified on ND 4.2.1.10: the schema is under-typed. ND validates `secondaryIp` server-side as a bare IPv4 address (template `ipV4Address SECONDARY_IP`) — CIDR and non-IP input both 400. Now typed `IPv4Host` (bare or CIDR in, normalized to bare) in "Validate loopback ipv6 and IOS-XE internal/underlay addresses as bare hosts". The same probe showed `iosXeInternalLoopback` `ip`/`ipv6` are validated too, so those are now `IPv4Host`/`IPv6Host` as well — and, as a bonus, NX `loopback.ipv6` must be *bare*, not CIDR as the model had since #220. Evidence table in #529.

plugins/modules/nd_interface_loopback.py
allenrobel: Reworded — with the validation change above, every policy type's `ip` now accepts bare or CIDR and normalizes to bare, and the description says so.

plugins/modules/nd_interface_loopback.py
allenrobel: Reworded: `ipv6` accepts bare or CIDR and is normalized to bare for both `loopback` and `iosXeInternalLoopback`; the controller applies /128 itself (lab-verified — CIDR on the wire is rejected, see #529).

plugins/modules/nd_interface_loopback.py
allenrobel: Yes — the `overridden` text now says the scope is the loopback policy types this module manages, fabric-wide across all switches, and that other policy types (e.g. system-provisioned `underlayLoopback`, `userDefined`) are never modified or deleted. That matches `query_all`'s managed-policy-type filter.

plugins/modules/nd_interface_loopback.py
allenrobel: Made consistent: every example now sets `config_actions.deploy: true` except the one explicitly titled "without deploying".

plugins/modules/nd_interface_loopback.py
allenrobel: Not intended — added examples for `iosXeLoopbackShutNoshut`, `iosXeUnderlayLoopback`, `iosXeInternalLoopback` and `csr1kvLoopback`; every `policy_type` choice now has one.

plugins/modules/nd_interface_loopback.py
allenrobel: Right — `interface_type` and `config_data.mode` are always present in `before`/`after` (`to_config()` doesn't exclude them). The `before`, `after` and `proposed` samples now include them.

plugins/modules/nd_interface_loopback.py
allenrobel: Same fix as the `before` sample above — `interface_type` and `config_data.mode` are now shown in the `after` (and `proposed`) samples.

plugins/modules/nd_interface_loopback.py
allenrobel: Not intended, and not specific to this module — good catch. `NDStateMachine` never passes `diff` or `logs` to `NDOutput` (it only assigns `before`/`after`/`proposed`), and nothing in `module_utils` collects log records into a list, so every NDStateMachine-backed module returns `diff: []` and `logs: []` today, even though ~24 modules document `diff` as "returned: always" (this module's RETURN block inherited that from the port-channel-access one). Filed #530 to fix it at the framework layer: derive `diff` in `NDOutput.format()` from `before`/`after` (shape to be settled in #304, where the resource-manager module already emits a dcnm-style diff), add an in-memory log handler for `output_level: debug`, and add unit tests asserting the diff shapes — coordinated with #389/#525, which also touch `nd_output.py`. Keeping it out of this PR since it affects all 25 Gen-3 modules.

plugins/modules/nd_interface_loopback.py
```diff
@@ -55,41 +55,138 @@
...
+                  secondary_ip_list:
                     description:
-                    - Additional CLI configuration commands to apply to the interface.
+                    - A list of secondary IPv4 addresses configured on the loopback interface.
+                    - Applies when policy_type is C(ipfmLoopback).
+                    type: list
+                    elements: dict
+                    suboptions:
+                      ip:
+                        description:
```
akinross, line 156: should config be marked as a list? not sure this is the correct notation but seems to be done in other parts
```suggestion
                        - The mask length is set separately via O(config[].config_data.network_os.policy.secondary_ip_list.prefix); CIDR notation is rejected.
```

akinross (changes requested):

plugins/modules/nd_interface_loopback.py
allenrobel: Yes — `config[]` is the right notation, and it was the lone unbracketed per-item reference in this module (lines 69/216/229/230 already use it). Fixed in "Docs: bracket list levels in the secondary_ip_list prefix O() reference (#403 review)". One step further than the suggestion: `secondary_ip_list` is itself a list, so per the convention in nd_manage_policy (e.g. `O(config[].switch[].policies[].name)`) it is now `O(config[].config_data.network_os.policy.secondary_ip_list[].prefix)`. validate-modules passes.

akinross (approved):

plugins/module_utils/orchestrators/loopback_interface.py
```diff
@@ -152,37 +197,58 @@ def delete(self, model_instance: LoopbackInterfaceModel, **kwargs) -> None:
...
         """
         try:
-            groups: dict[str, list[tuple[str, dict]]] = defaultdict(list)
-            for model_instance in model_instances:
-                switch_id = self._resolve_switch_id(model_instance.switch_ip)
-                payload = model_instance.to_payload()
-                payload["switchId"] = switch_id
-                groups[switch_id].append((model_instance.interface_name, payload))
-
+            groups = self._group_by_switch_and_policy_type(model_instances)
```
mikewiebe, line 242: ### High: A later grouped-create failure strands an earlier successful group

**Issue**

`create_bulk()` sends `(switch, policy_type)` groups sequentially and queues successful groups for deferred deployment. If a later group fails, the module exits before flushing the earlier queue. On retry, the earlier interface is already present and becomes `no_diff`, so it is not queued for deployment again.

**Evidence**

- This loop sends and queues one group at a time.
- `remove_pending()` and `deploy_pending()` run only after `manage_state()` succeeds.
- The state machine skips unchanged resources on retry.
- An exact-head two-run probe reproduced the failure: group A succeeded, group B failed, A remained staged, and the retry queued only B.

Practical sequence:

1. A task requests `loopback10` as `loopback` and `loopback20` as `ipfmLoopback`, with `deploy: true`.
2. ND accepts the first POST and rejects the second.
3. The task fails before `deploy_pending()`, leaving `loopback10` staged.
4. A retry sees `loopback10` as unchanged and does not rebuild its deploy queue.
5. The retry can succeed while `loopback10` remains pending rather than deployed.

**Existing PR overlap**

Issues #385, #389, #409, and #481 cover related final-state, failure-output, grouping, and staged-intent behavior, but none defines recovery of an earlier successful group after a later group fails.

**Impact**

A routine retry can finish successfully while controller intent and switch running state remain divergent.

**Suggested fix**

Make partial outcomes recoverable: roll back earlier accepted groups, deliberately finalize and accurately report the successful subset before raising, or reconstruct pending deployment from controller state on retry. Add two-run convergence tests for first-group-success/later-group-failure and mixed-result 207 responses.

plugins/modules/nd_interface_loopback.py
```diff
@@ -55,41 +55,138 @@
...
+                    description:
+                    - The loopback policy template to apply. This is a discriminator that determines which of the
+                      remaining C(policy) suboptions are applicable.
+                    - Use V(loopback) for a standard NX-OS loopback interface.
+                    - Use V(ipfmLoopback) for an IP Fabric for Media loopback interface.
+                    - Use V(mplsLoopback) for an MPLS loopback interface.
+                    - Use V(iosXeLoopback) for a general-purpose IOS-XE loopback.
+                    - Use V(iosXeLoopbackShutNoshut) to manage only the admin state of an IOS-XE loopback.
+                    - Use V(iosXeUnderlayLoopback) for an IOS-XE underlay (NVE source) loopback.
+                    - Use V(iosXeInternalLoopback) for an IOS-XE internal loopback (unvalidated ip/ipv6, PIM option).
```
mikewiebe: ### Low: This policy description still says validated IP fields are unvalidated

**Issue**

The `iosXeInternalLoopback` summary says its `ip` and `ipv6` fields are unvalidated, but the current model validates and normalizes both.

**Evidence**

- This description still says `unvalidated ip/ipv6`.
- `IosXeInternalLoopbackPolicyModel` now uses `IPv4Host` and `IPv6Host`.
- Unit tests cover CIDR normalization and malformed-input rejection.

**Existing PR overlap**

Related to thread r3861618727. The runtime validation concern was fixed and that thread resolved, but this summary text was not updated.

Related to #529, which supplied the address-validation fix but does not specifically track this remaining sentence.

**Impact**

The public documentation contradicts the detailed field descriptions and actual runtime behavior.

**Suggested fix**

Replace `IOS-XE internal loopback (unvalidated ip/ipv6, PIM option)` with wording such as `IOS-XE internal loopback (IPv4/IPv6 and PIM options)`.

plugins/module_utils/orchestrators/loopback_interface.py
allenrobel: Confirmed — a later group's failure exited before flushing the earlier group's queued deploys, and the retry classified the earlier interfaces as no_diff, so they stayed staged. The update path had the same shape (a successful PUT followed by any later failure).

Fixed in "Finalize controller-accepted mutations when a later operation fails (#403 review)" by finalizing rather than rolling back or reconstructing: on the module failure path, a new NDBaseInterfaceOrchestrator.deploy_accepted_mutations() deploys exactly the queued pairs whose mutation the controller already accepted, and the failure message names the finalized subset. On retry the earlier group is then genuinely converged, and the retry deploys only the remaining work. Details:

- Pairs still present in _pending_removes are excluded — the delete path queues its deploy before interfaceActions/remove is sent, so deploying those would ship pending intent whose removal never reached the controller.
- config_actions.deploy=false is honored on the failure path too; staged intent remains the documented contract there.
- Rollback was rejected (destructive deletes on a failure path) and reconstruct-on-retry was rejected (requires reading ND pending-config state we can't do cheaply or reliably).

Tests per your suggestion: a two-run first-group-success/later-group-failure convergence test and a mixed-result-207-as-failing-group test (test_loopback_interface 00450/00460), plus deploy_accepted_mutations coverage in test_base_interface (00640–00670). One deliberate limit: within-group recovery of a mixed 207's reported-success items is NOT attempted — whether ND actually creates that subset is not lab-characterized, and the per-item status field is known-inconsistent (vault: multi-status-207-status-field-inconsistent) — so a failed group still queues nothing. The fix is verified by the unit suite reproducing your two-run probe shape; I have not yet re-run the live probe against the lab and will fold that into the next lab pass on this branch.

plugins/modules/nd_interface_loopback.py
allenrobel: Fixed in "Docs: iosXeInternalLoopback ip/ipv6 are validated since the #529 fix (#403 review)" — the summary now reads "IOS-XE internal loopback (IPv4/IPv6 and PIM options)" per your suggested wording, matching the detailed field descriptions and the IPv4Host/IPv6Host validation from the #529 fix.

plugins/module_utils/orchestrators/loopback_interface.py
allenrobel: Live re-run of the two-run probe against ND 4.2.1.10 (SITE1 / S1_LE1) from the current head:

Run 1 — loopback210 (`loopback`) + loopback1024 (`mplsLoopback`, server-rejected: HTTP 500 "Out of Range : loopback1024 Id [1024] is not within the range of 0 and 1023"; passes client-side validation, so ND is the rejecter). Task failed with the 500 and the message ended "NOTE: before the failure, the controller had already accepted changes for interface(s) [loopback210 (switchId …)]; those changes were deployed." Verified on the switch via ND showCommands: `show run interface loopback210` present; ND switch diff shows loopback210 as `equal` (no pending delta); loopback1024 GET 404.

Run 2 (retry) — loopback210 unchanged + loopback211 (`ipfmLoopback`). `before` classified loopback210 as existing/unchanged (your retry premise), the task succeeded, and both interfaces are in running config with the switch diff reporting zero non-equal entries — controller intent and switch state converged across the two runs.

Side notes from the run: the retry used `ipfmLoopback` rather than `mplsLoopback` because 4.2.1 `mplsLoopback` create consults MPLS_LOOPBACK_IP_POOL even with an explicit `ip` and 500s on SITE1's empty pool (a separate ND behavior, not part of this fix); that failed attempt also confirmed the finalizer stays silent when nothing was accepted. Probe interfaces cleaned up via `state: deleted`.

allenrobel:
Status on the three hold items:

**1. 207 handling — fixed in "Trust only exact success in 207 Multi-Status items (#397, #403 review)".** Went with shared exact-success handling rather than restoring the loopback-local check, per #397's consolidation direction: `NdV1Strategy.is_success` now applies an exact-`success` allowlist to the envelope arrays (`results[]`/`switchIds[]`/`links[]`) on HTTP 207 specifically — missing, `None`, empty, `warning`, `notexecuted`, and unknown statuses all classify as failure — and `extract_error_message` mirrors the split, so your example now fails with `loopback10: invalid policyType` and never queues deployment. Plain-200 bodies keep the failure-literal denylist because ND ships legitimately status-less item arrays on 200 (e.g. the GET /links list envelope). Tests cover your missing-status example plus the non-success variants, the all-success 207, the 200-scoping guard, and the error-message shape; two #398-era tests that pinned the old tolerance (response_handler 01290, manage_route_map 00215 — synthetic bodies, not lab-verified wire shapes) are inverted to the new semantics.

**2. mplsLoopback update-only — pushing back with lab evidence.** A four-cell matrix on ND 4.2.1 (SITE1, borderGateway and leaf × MPLS Handoff on/off, 2026-06-29) shows `mplsLoopback` create is user-driven, not fabric-created: with fabric MPLS Handoff enabled (which provisions MPLS_LOOPBACK_IP_POOL) the create succeeds — even on a plain leaf, role is not enforced at create — and with it disabled the create fails 500 "[MPLS_LOOPBACK_IP_POOL] is an empty pool". Its absence from the create discriminator is schema under-documentation (the create gateway accepts it on 4.2.1); both behaviors are recorded in the team bug-tracker vault. Given creates are user-initiated, `overridden` removing an omitted mplsLoopback follows the same documented managed-policy-type ownership as every other type this module manages (the docs now state that scope explicitly). If the concern is the opaque empty-pool 500, I'd rather add a fail-fast MPLS-Handoff preflight than make the type update-only — but if you still want it excluded from `overridden` deletion scope after the above, let's settle that direction here and I'll implement it.

**3. #422 dependency — done.** #422 merged 2026-08-28; this branch is rebased on it with the conflict resolved and the per-policy `reverse_diff_defaults` rebase checklist completed (PR body updated). Full unit suite is green and the complete `nd_interface_loopback` integration target passed against ND 4.2.1.10 (ok=81 changed=29 failed=0), plus a live two-run partial-failure probe on 2026-09-01 (evidence in the r3905703008 thread).

Your 09-01 inline findings (stranded-deploy High, docs Low) are also addressed in their threads. Ready for re-review.

mikewiebe (approved):

allenrobel:
Merged on approvals from @akinross and @mikewiebe

PR #405, opened 2026-07-16
Enhancing Fabric Type Support for Switches Module

AKDRG:
## Proposed Changes
This PR improves nd_manage_switches support for fabric-specific switch onboarding by adding fabric capability validation, exposing supported switch platform types, and improving lifecycle output visibility for config-save and deploy operations.

Changes

- Added fabric capability validation before switch discovery/add workflows.
- Enforced fabric-specific support for:
1. switch role
2. preserve_config
3. platform_type
- Exposed platform_type in the playbook argspec with supported choices:
1. nx-os
2. ios-xe
3. ios-xr
5. other

- Registered finalize operations in module output:
1. config_save
2. deploy_switches
3. deploy_config

- Preserved full controller response metadata for finalize API calls where available.
- Added/updated unit tests for fabric validation, platform input handling, and finalize output visibility.

## Test Notes
Validated against live controller
pytest tests/unit/module_utils/test_nd_switch_resources.py -q
44 passed

4.2(1)

* [ ] analyze
* [ ] infa
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist


plugins/module_utils/manage_switches/fabric_switch_capabilities.py
```diff
@@ -0,0 +1,249 @@
+# -*- coding: utf-8 -*-
```
akinross, line 1: I see another implementation of fabric type retrieval in https://github.com/CiscoDevNet/ansible-nd/pull/338. The use case might be slightly different ( do not know full context )but I think we should align before we end up with a lot of potential similar functionality being scattered everywhere.

akinross (changes requested):

allenrobel:
@AKDRG — a heads-up and a question, not a request to change anything here. This PR is doing the right thing; the collision is on my side.

**Situation.** In #404 I added a `PlatformTypeEnum` to `plugins/module_utils/enums.py` for `FabricContext.get_platform_type()`, which reads `additionalData.platformType` off `GET /fabrics/{fabric}/switches`. I did that without noticing your `PlatformType` in `models/manage_switches/enums.py` already covers the same ground — same values, except yours also has `SONIC`, which mine is missing (a real gap: my lookup would silently return `None` for a SONiC switch). Two enums one token apart in the name is a trap for the next reader, so I'd rather converge than ship the duplicate.

**Proposal.** Promote `PlatformType` from `models/manage_switches/enums.py` up to `module_utils/enums.py`, and have `FabricContext` use it. Per CLAUDE.md, `module_utils/enums.py` is for "Enums commonly used by most module utilities" — with both `manage_switches` and `fabric_context` needing it, that's what it's become. The reverse (a top-level util importing from `models/manage_switches/`) would be a layering inversion.

**This shouldn't cost you anything.** Done as a move + re-export — leaving `PlatformType` re-exported from `models/manage_switches/enums.py` (via `__all__`, to keep pylint quiet) — every existing import site keeps working untouched, nothing in this PR changes, and you don't need to rebase. `module_utils/enums.py` imports only `from enum import Enum`, so there's no cycle, and the re-export lives in a named module rather than `__init__.py`, so it's clear of the empty-init sanity rule. Two of your changed files (`nd_switch_resources.py`, `config_models.py`) are `PlatformType` consumers, which is exactly why I'd rather not touch import sites while you're in flight.

**The question I actually want your read on.** Your enums look deliberately endpoint-scoped — `PlatformType` documented as AddSwitches (POST switches), `ShallowDiscoveryPlatformType` split out because shallowDiscovery excludes `apic`. My use is a third context: the GET switches *response*. Is that response set known to match the AddSwitches request set, or would you rather it stay a separate enum on principle? If they can legitimately diverge, sharing one is the wrong call and I'll keep a distinct read-side enum — properly named and documented, with `SONIC` added. You know this API surface better than I do.

One detail either way: `normalize()` defaults `None -> NX_OS`, which suits the write path but not mine, where "switch reports no `platformType`" has to stay `None`. So I'd use the bare enum rather than `normalize()` regardless.

Happy to do the move in #404 if you're good with it, or leave it to you if you'd rather own that file.

allenrobel:
@AKDRG  thanks for moving PlatformType to `plugins/module_utils/enums.py` 👍

I'll wait for #405 to merge to leverage PlatformType at its new home.

plugins/module_utils/utils.py
```diff
@@ -7,6 +7,9 @@
 from copy import deepcopy
 from typing import Any
 
+from ansible_collections.cisco.nd.plugins.module_utils.endpoints.v1.manage.manage_fabrics import (
+    EpManageFabricsGet,
```
allenrobel: This adds a third module-scope `endpoints/v1/manage` import to the shared `utils.py`. Because `nd.py` imports `issubset` from this file, every legacy `nd_*`/`ndi_*` module transitively loads the manage endpoints package at import time — an import-time failure there would break the whole collection, not just `nd_manage_switches`. And since `module_utils` is public API once released, relocating `FabricUtils.get_fabric_info()`/`get_fabric_type()` later is a breaking change, while moving them now is mechanical. This is the same coupling concern raised on PR #198 for the first two endpoint imports in this file. Suggest housing the fabric-info retrieval (and this import) in `plugins/module_utils/manage_switches/` — which also dovetails with the open thread about aligning fabric-type retrieval with PR #338.

plugins/module_utils/manage_switches/fabric_capabilities.py
```diff
@@ -0,0 +1,259 @@
...
+    for cfg in configs:
+        prefix = f"{cfg.seed_ip}"
+        platform_type = cfg.platform_type
+        role = SwitchRole.normalize(cfg.role) if cfg.role else None
+        if platform_type not in capability.platform_types:
+            errors.append(
+                f"{prefix}: platform_type '{_enum_value(cfg.platform_type)}' is not supported for {capability.family} fabric "
+                f"'{fabric_name}' (type '{fabric_type}'). Supported platform_type values: {_enum_values(capability.platform_types)}."
+            )
+        if role not in capability.roles:
```
allenrobel: `role` normalizes to `None` when unset, and `None` is never a member of any `capability.roles` frozenset — so a switch config with `role` unspecified is unconditionally rejected for every fabric type. That contradicts the model contract in `config_models.py` (`role: SwitchRole | None` — "None means not specified (uses controller default)") and `normalize_role`'s "Returns None when not specified". This is reachable from a playbook with an explicit `role: null` (argspec defaults apply only when the key is omitted entirely). The `"<unspecified>"` display string below shows the `None` case was anticipated — suggest treating `None` as valid (skip the role check and let the controller default apply) rather than hard-failing.

plugins/module_utils/manage_switches/nd_switch_resources.py
```diff
@@ -1414,18 +1467,14 @@ def post_add_processing(
...
-        wait_kwargs: dict[str, Any] = {}
-        if spec.all_preserve_config:
-            wait_kwargs["all_preserve_config"] = True
-        if spec.skip_greenfield_check:
-            wait_kwargs["skip_greenfield_check"] = True
-
-        success = spec.wait_utils.wait_for_switch_manageable(
-            all_serials,
-            **wait_kwargs,
+        success = spec.wait_utils.wait_for_post_add_switches(
```
allenrobel, line 1522: With this refactor, `PostAddProcessingSpec.all_preserve_config` is never read anymore — reload behavior is now derived per switch by `_split_post_add_wait_sets()`. But the call site still computes `all_preserve_config = all(...)`, still logs "All switches brownfield (preserve_config=True) — reload detection skipped" (no longer accurate: preserve-config switches now go through the `nxos_preserve` wait policy, nothing is skipped), and the `post_add_processing()` docstring still documents `all_preserve_config` as controlling wait behavior. Suggest removing the dead field and its computation, and updating the log line and docstring to match the new per-switch splitting.

plugins/module_utils/models/manage_switches/config_models.py
```diff
@@ -503,7 +503,24 @@ def normalize_role(cls, v: str | SwitchRole | None) -> SwitchRole | None:
     @classmethod
     def normalize_platform_type(cls, v: str | PlatformType | None) -> PlatformType:
         """Normalize platform_type for case-insensitive matching (NX_OS, nx-os, etc.)."""
-        return PlatformType.normalize(v)
+        platform_type = cls._normalize_config_platform_type(v)
+        if platform_type in (PlatformType.SONIC, PlatformType.APIC):
+            raise ValueError(f"platform_type '{platform_type.value}' is not supported. " "Supported platform_type values: nx-os, ios-xe, ios-xr, other.")
```
allenrobel, line 610: `normalize_platform_type` now raises `ValueError` for `sonic`/`apic`, but its docstring only describes normalization. CLAUDE.md: "Docstrings MUST contain at last a '# Summary' section and a '## Raises' section" — worth documenting the rejection under `## Raises` so callers know this validator does more than coerce case.

allenrobel (commented):
### Code review

Four findings, each as an inline comment: endpoint-import coupling in shared `utils.py`, `role: None` rejected by the new fabric-capability validation, a now-dead `all_preserve_config` flag, and an undocumented `ValueError` in `normalize_platform_type`.

plugins/module_utils/utils.py
AKDRG: Added a new file for this, fabric_details_cache. The file is kept under module_utils as Fabric Details are queried by a couple of modules and is expected to be used by more.

plugins/module_utils/manage_switches/fabric_capabilities.py
AKDRG: Fixed.

plugins/module_utils/manage_switches/nd_switch_resources.py
AKDRG: Fixed.

plugins/module_utils/models/manage_switches/config_models.py
AKDRG: Fixed

plugins/module_utils/manage_switches/fabric_switch_capabilities.py
AKDRG: Hey Akini,
This PR will introduce the broader GET of the Fabric Details through fabric_details_cache and return the type. I have discussed with Sivakami on using this. She will make those changes once this PR is merged.

allenrobel (dismissed):
LGTM after comments were addressed.

akinross (dismissed):

sivakasi-cisco:
https://github.com/CiscoDevNet/ansible-nd/issues/438

plugins/module_utils/manage_switches/nd_switch_resources.py
```diff
@@ -2608,6 +2771,27 @@ def _proposed_to_config_list(self, configs: list["SwitchConfigModel"]) -> list[d
                 self.log.warning("Could not convert config %s for output: %s", cfg.seed_ip, exc)
         return result
 
+    def _validate_fabric_capabilities(self, configs: list["SwitchConfigModel"]) -> None:
```
mikewiebe, line 2775: ### High: Gathered non-NX switch configuration cannot be replayed

**Issue**

The module reads `additionalData.platformType` from live inventory but removes it from gathered output. Revalidating that output applies the `nx-os` default, so the new fabric-capability validation rejects otherwise valid non-NX switches.

**Evidence**

- [`nd_switch_resources.py` lines 2774-2785](https://github.com/CiscoDevNet/ansible-nd/pull/405/files#diff-6dd07c8d609b04466f34b90a9b295c8710d66a6d17802f5277faf98290274e19R2774) introduces fabric capability validation over the reconstructed `SwitchConfigModel`.
- [`config_models.py` lines 670-678](https://github.com/CiscoDevNet/ansible-nd/blob/7fbfa7c8c5b0b2553c53c403596bcdafbb512c4b/plugins/module_utils/models/manage_switches/config_models.py#L670-L678) correctly reads the live platform into a config model.
- [`config_models.py` lines 680-698](https://github.com/CiscoDevNet/ansible-nd/blob/7fbfa7c8c5b0b2553c53c403596bcdafbb512c4b/plugins/module_utils/models/manage_switches/config_models.py#L680-L698) explicitly says gathered output is immediately usable as `config` input after credentials are replaced, while intentionally removing `platform_type`.
- [`config_models.py` lines 328-332](https://github.com/CiscoDevNet/ansible-nd/blob/7fbfa7c8c5b0b2553c53c403596bcdafbb512c4b/plugins/module_utils/models/manage_switches/config_models.py#L328-L332) defaults the missing platform back to `nx-os`.
- [`fabric_switch_capabilities.py` lines 117-125](https://github.com/CiscoDevNet/ansible-nd/blob/7fbfa7c8c5b0b2553c53c403596bcdafbb512c4b/plugins/module_utils/manage_switches/fabric_switch_capabilities.py#L117-L125) permits `leaf` for Campus IOS-XE but not Campus NX-OS.
- An exact-head probe built a live Campus IOS-XE leaf record with `additionalData.platformType: ios-xe`, serialized it through `_inventory_to_config_list()`, and revalidated it. The gathered entry omitted the platform, defaulted back to `nx-os`, and the new matrix rejected the leaf role:

  ```text
  192.0.2.10: role 'leaf' is not supported for platform_type 'nx-os'
  in Campus VXLAN ... Supported role values:
  borderGateway, borderGatewaySpine, borderGatewaySuperSpine.
  ```

- For example, consider a managed Campus leaf at `192.0.2.10` whose live inventory reports `platform_type: ios-xe` and `role: leaf`. A `state: gathered` task reads the IOS-XE platform but returns a config entry containing only the address and role. If the user supplies the required credentials and replays that gathered entry with `state: merged`, the missing platform defaults to `nx-os`. The capability validator therefore rejects the otherwise valid switch:

  ```yaml
  # Live controller inventory
  ip_address: 192.0.2.10
  platform_type: ios-xe
  role: leaf

  # Gathered and replayed configuration
  ip_address: 192.0.2.10
  role: leaf
  # platform_type was removed, so replay defaults it to nx-os
  ```

  ```text
  192.0.2.10: role 'leaf' is not supported for platform_type 'nx-os'
  in Campus VXLAN.
  ```

- A model unit test currently codifies the platform omission instead of checking a replay through capability validation.

**Existing PR overlap**

No matching existing PR comment found. The resolved omitted-role thread fixed `role: None`; it did not address loss of `platform_type`.

No matching open issue found. Issue [#438](https://github.com/CiscoDevNet/ansible-nd/issues/438) tracks `config_actions` and resource-level deployment, not gathered platform identity.

**Impact**

A standard gather/edit/reapply workflow fails for a valid Campus IOS-XE leaf. External IOS-XR and IOS-XE switches likewise lose their platform identity, and gathered output is an incomplete audit source for the platform support introduced by this PR.

**Suggested fix**

Include the inventory-reported `platform_type` in gathered, before/after, proposed, diff, and check-mode serializers. Add round-trip tests for at least a Campus IOS-XE leaf and an External IOS-XR switch.

---

plugins/module_utils/manage_switches/nd_switch_resources.py
```diff
@@ -2849,8 +3036,12 @@ def manage_state(self) -> None:
             self.log.info("Overridden state with no config — deleting all switches from fabric")
             return self._handle_deleted_state(None)
 
+        if self.state not in ("merged", "replaced", "overridden"):
+            self.nd.module.fail_json(msg=f"Unsupported state: {self.state}")
+
         # --- Validate & classify ------------------------------------------------
```
mikewiebe, line 3050: ### High: Onboarding-only `preserve_config` validation rejects existing no-op switches

**Issue**

Capability validation runs before `compute_changes()`, so it validates `preserve_config` for every desired switch, including switches already managed and ultimately classified as idempotent. Because the model injects `preserve_config: false`, an existing External switch fails unless every reconciliation playbook repeats the historical onboarding choice.

**Evidence**

- [`nd_switch_resources.py` lines 3042-3059](https://github.com/CiscoDevNet/ansible-nd/pull/405/files#diff-6dd07c8d609b04466f34b90a9b295c8710d66a6d17802f5277faf98290274e19R3042) runs the new capability validation before diff classification and idempotency.
- [`config_models.py` lines 323-327](https://github.com/CiscoDevNet/ansible-nd/blob/7fbfa7c8c5b0b2553c53c403596bcdafbb512c4b/plugins/module_utils/models/manage_switches/config_models.py#L323-L327) defaults omitted `preserve_config` to false.
- [`fabric_switch_capabilities.py` lines 243-247](https://github.com/CiscoDevNet/ansible-nd/blob/7fbfa7c8c5b0b2553c53c403596bcdafbb512c4b/plugins/module_utils/manage_switches/fabric_switch_capabilities.py#L243-L247) rejects that value unconditionally for External fabrics.
- [`nd_switch_resources.py` lines 2748-2756](https://github.com/CiscoDevNet/ansible-nd/blob/7fbfa7c8c5b0b2553c53c403596bcdafbb512c4b/plugins/module_utils/manage_switches/nd_switch_resources.py#L2748-L2756) also hard-codes `preserve_config: false` in gathered inventory.
- An exact-head probe classified an existing External switch as a complete no-op, with every work bucket empty, but the pre-classification capability check still raised:

  ```text
  preserve_config 'false' is not supported for External ...
  Supported preserve_config values: true.
  ```

- For example, an External switch is initially onboarded with `preserve_config: true`. On a later idempotency run, the playbook describes the same managed switch but omits this onboarding-only option:

  ```yaml
  - cisco.nd.nd_manage_switches:
      state: merged
      config:
        - ip_address: 192.0.2.20
          username: admin
          password: password
  ```

  The model inserts `preserve_config: false`, and capability validation rejects it before `compute_changes()` can establish that the operation is a no-op:

  ```text
  preserve_config 'false' is not supported for External.
  Supported preserve_config values: true.
  ```

  Had change computation run first, every work bucket would have been empty:

  ```text
  switches_to_add: []
  switches_to_update: []
  switches_to_remove: []
  ```

  Gathered External configurations have the same problem because the serializer explicitly emits `preserve_config: false`.

**Existing PR overlap**

No matching existing PR comment found. Earlier comments discuss a dead preserve-config wait flag but do not cover pre-diff validation of an onboarding-only setting.

No matching open issue found. Focused searches for External switch and `preserve_config` idempotency did not identify an owner.

**Impact**

Existing External switch playbooks that were idempotent before this PR can now fail before diffing, discovery, or writes. Gathered External configuration is inherently unreplayable because it carries the rejected default.

**Suggested fix**

Classify the plan before enforcing onboarding-only fields and validate `preserve_config` only for switches that will actually be discovered, added, or re-added. Preserve input explicitness if omission must differ from false. Add a regression test for an existing External no-op switch with omitted `preserve_config`.

---

mikewiebe: Can you clarify why External switches require `preserve_config: true`? Is this an ND controller requirement or an intended module policy?

`preserve_config` appears to control initial onboarding behavior. Requiring it to remain `true` during every subsequent reconciliation means an omitted value—and the module’s own gathered value of `false`—causes an already-managed no-op switch to fail.

If this is a controller requirement, could we document the rationale and limit its validation to discovery, add, and re-add operations? Otherwise, should omission be accepted for switches that are already managed?

plugins/modules/nd_manage_switches.py
```diff
@@ -117,11 +120,24 @@
...
+                description:
+                - Platform type of the switch.
+                - Supported values depend on the target fabric type. See C(notes).
+                type: str
+                default: nx-os
+                choices:
+                - nx-os
+                - ios-xe
+                - ios-xr
+                - other
```
mtarking, line 141: What is other?

plugins/modules/nd_manage_switches.py
```diff
@@ -267,7 +283,25 @@
...
   its C(seed_ip) already exists in the fabric inventory with no configuration
-  drift (same role).
+  drift. If C(role) is omitted, role drift is ignored and the controller
+  default applies on add.
+- The module validates switch platform, provided C(role), and C(preserve_config)
+  against the resolved fabric type before discovery or write operations.
+- 'NX-OS only fabrics: Routed, DataCenter VXLAN, Enhanced Classic LAN, AI VXLAN,
+  AI Routed, and IPFM.'
+- 'NX-OS and IOS-XE fabrics: Campus VXLAN.'
+- 'NX-OS, IOS-XE, IOS-XR, and other platform fabrics: External.'
```
mtarking: Is there a reason these are in single quotes?

plugins/modules/nd_manage_switches.py
```diff
@@ -267,7 +283,25 @@
...
+- C(preserve_config) must be C(false) for Routed, Campus VXLAN, AI VXLAN,
+  AI Routed, and IPFM fabrics.
+- C(preserve_config) must be C(true) for External fabrics.
+- C(preserve_config) can be C(true) or C(false) for DataCenter VXLAN and
+  Enhanced Classic LAN fabrics.
+- Campus VXLAN supports C(leaf) and C(spine) roles for IOS-XE switches.
+- Campus VXLAN supports C(border_gateway), C(border_gateway_spine), and
+  C(border_gateway_super_spine) roles for NX-OS switches.
+- IPFM supports C(leaf), C(spine), and C(tier2_leaf) switch roles.
```
mtarking: We shouldn't expose any support for IPFM yet.

plugins/modules/nd_manage_switches.py
```diff
@@ -267,7 +283,25 @@
...
-  drift (same role).
+  drift. If C(role) is omitted, role drift is ignored and the controller
+  default applies on add.
+- The module validates switch platform, provided C(role), and C(preserve_config)
+  against the resolved fabric type before discovery or write operations.
+- 'NX-OS only fabrics: Routed, DataCenter VXLAN, Enhanced Classic LAN, AI VXLAN,
+  AI Routed, and IPFM.'
+- 'NX-OS and IOS-XE fabrics: Campus VXLAN.'
+- 'NX-OS, IOS-XE, IOS-XR, and other platform fabrics: External.'
+- C(preserve_config) must be C(false) for Routed, Campus VXLAN, AI VXLAN,
```
mtarking: I do not think this statement is true. Lets discuss in standup if needed.

plugins/modules/nd_manage_switches.py
```diff
@@ -267,7 +283,25 @@
...
+  AI Routed, and IPFM fabrics.
+- C(preserve_config) must be C(true) for External fabrics.
```
mtarking: This is implicit true for external fabrics and not user configurable and in general defaults to true.

plugins/modules/nd_manage_switches.py
```diff
@@ -117,11 +120,24 @@
                 - edge_router
                 - core_router
                 - tor
+                - tier2_leaf
```
mtarking: We shouldn't expose yet imho.

mtarking (changes requested):

plugins/modules/nd_manage_switches.py
AKDRG: Other is used for Switches from other vendors such as Arista.

plugins/modules/nd_manage_switches.py
AKDRG: Fixed

plugins/modules/nd_manage_switches.py
AKDRG: Removed.

plugins/modules/nd_manage_switches.py
AKDRG: Fixed

plugins/modules/nd_manage_switches.py
AKDRG: Defaults derived and validated depending on fabric type.

plugins/modules/nd_manage_switches.py
AKDRG: Preserve config is disabled(false) for these fabrics.

plugins/module_utils/manage_switches/nd_switch_resources.py
AKDRG: External Fabric has fixed preserve config "true". Omission has been now handled through the default validator which will fix it to true.

plugins/module_utils/manage_switches/nd_switch_resources.py
AKDRG: Fixed.

allenrobel (commented):
### Code review

Incremental review of the three commits since my last review (`Fix Preserve Config, Platform Type Validation + Support Docs`, `Remove support for IPFM/Tier2`, `Refine switch preserve config capabilities`). One finding, posted inline — a follow-up to the earlier (resolved) IPFM thread.

plugins/module_utils/manage_switches/fabric_switch_capabilities.py
```diff
@@ -0,0 +1,289 @@
...
+        preserve_config_values=frozenset({False}),
+        default_preserve_config=False,
+    ),
+    FabricSwitchCapability(
+        family="IPFM",
+        fabric_types=frozenset({"ipfm", "ipfmenhanced"}),
+        platform_types=frozenset({PlatformType.NX_OS}),
+        roles=IPFM_ROLES,
+        preserve_config_values=frozenset({False}),
+        default_preserve_config=False,
```
allenrobel, line 178: Follow-up to the earlier IPFM thread ("We shouldn't expose any support for IPFM yet" / "Removed."): the removal in `Remove support for IPFM/Tier2` covers the `tier2_leaf` argspec choice, the DOCUMENTATION notes, and the unit test, but this capability entry still makes IPFM fully functional. Nothing gates the `fabric` parameter, so a task targeting an `ipfm`/`ipfmenhanced` fabric with `role: leaf` or `spine` still validates and onboards end-to-end. `IPFM_ROLES` (line 93) also still contains `SwitchRole.TIER2_LEAF`, so the enforcement layer continues to accept the role that was removed from the argspec. If IPFM should not be exposed yet, this entry (and `IPFM_ROLES`) needs to be removed or gated as well; if the entry stays, the DOCUMENTATION notes should re-document IPFM's constraints, including the newly added `default_preserve_config=False` derivation.

AKDRG: Gated the onboarding of switches for supported fabrics through SUPPORTED_SWITCH_ONBOARDING_FABRIC_TYPES. Documentation has been modified to announce the support of only DC VXLAN and Campus.

plugins/module_utils/manage_switches/nd_switch_resources.py
mikewiebe: # Following are 3 more issues introduced after the last set of changes

### 1. High: Omitted platform defaults still break an existing Campus IOS-XE no-op

**Issue**

@AKDRG while verifying this update, found the following regession/behavior.  Are you planning to handle this as part of https://github.com/CiscoDevNet/ansible-nd/issues/488?

Capability validation applies the NX-OS add default before determining whether an existing switch needs onboarding work. An idempotent Campus IOS-XE leaf therefore fails when a normal reconciliation playbook omits the newly introduced `platform_type`.

This is a regression introduced by PR #405. Before this PR, `platform_type` was not part of the module's public switch configuration, and current `develop` treats the same input as an unchanged no-op.

**Evidence**

- [`nd_switch_resources.py` lines 3051-3067](https://github.com/CiscoDevNet/ansible-nd/pull/405/files#diff-6dd07c8d609b04466f34b90a9b295c8710d66a6d17802f5277faf98290274e19R3051) validates capabilities before `compute_changes()` can classify the existing switch as idempotent.
- [`fabric_switch_capabilities.py` lines 283-299](https://github.com/CiscoDevNet/ansible-nd/pull/405/files#diff-cba745ac1866c79c2b655f2cdb85f37762b0570ffaddc8bd922aed7eda5adcc7R283) changes an omitted platform to NX-OS and validates the role against that assumed platform.
- [`config_models.py` lines 283-294](https://github.com/CiscoDevNet/ansible-nd/pull/405/files#diff-56b5c0c375eba2e5d66ef3b8a903e186dc96dd78e352c5d8bbdb2ff9a8333d42R283) excludes `platform_type` from normal-switch diffing, confirming it is not a drift field for an existing no-op.
- Practical example: assume `Campus_AK` already contains this switch:

  ```text
  Management IP: 192.0.2.10
  Platform:      ios-xe
  Role:          leaf
  ```

  A pre-PR reconciliation playbook can legitimately omit the newly introduced platform field:

  ```yaml
  - name: Reconcile the existing Campus IOS-XE leaf
    cisco.nd.nd_manage_switches:
      fabric: Campus_AK
      state: merged
      config:
        - seed_ip: 192.0.2.10
          username: admin
          password: "{{ switch_password }}"
          role: leaf
          # platform_type omitted
  ```

  Expected:

  ```yaml
  changed: false
  ```

  Actual result under PR #405:

  ```text
  role 'leaf' is not supported for platform_type 'nx-os'
  in Campus VXLAN
  ```

- The exact-head diff engine correctly classified this input as:

  ```text
  idempotent=1
  to_add=0
  to_update=0
  ```

  The actual module never reaches that result because capability validation runs first.

- Current `develop` at `3026bddc147861e07aa89b26be1277f884c6def4` treats the same input as a no-op. The regression exists at both the previously reviewed PR head `7fbfa7c` and current head `d0e213c`.

**Existing PR overlap**

This is a follow-up to the existing pre-diff validation thread. Omitted `preserve_config` derivation was addressed, but this confirms that onboarding-only defaults are still applied to existing idempotent switches.

Related to [#488](https://github.com/CiscoDevNet/ansible-nd/issues/488), which says NX-OS should be the default only for add operations. It does not defer this PR-specific existing-switch regression.

**Impact**

A valid Campus IOS-XE leaf that previously returned `changed: false` now fails before discovery, diffing, or writes. Adding `platform_type: ios-xe` avoids the failure, but requiring newly introduced onboarding metadata in every existing-switch reconciliation is the compatibility regression.

**Suggested fix**

Classify entries against live inventory before applying onboarding defaults. For an existing switch, resolve an omitted platform from its inventory record; apply the NX-OS default only to actual add or re-add operations. Add a `manage_state()` regression test proving this Campus IOS-XE input performs no writes.

mikewiebe: ### 2. High: The derived iBGP preserve default lets POAP and swap finish before reboot

**Issue**

`preserve_config` no longer has one static default. When omitted, PR #405 derives `true` for Data Center VXLAN iBGP and `false` for eBGP and Campus. The iBGP value is applied to every operation, including POAP and swap.

Those operations are consequently placed in the preserve-config wait set, which can accept the first pre-reboot `normal` response instead of requiring the documented `normal -> unreachable -> ok` reboot transition.

**Evidence**

- [`fabric_switch_capabilities.py` lines 113-118](https://github.com/CiscoDevNet/ansible-nd/pull/405/files#diff-cba745ac1866c79c2b655f2cdb85f37762b0570ffaddc8bd922aed7eda5adcc7R113) sets the omitted iBGP value to `true`.
- [`fabric_switch_capabilities.py` lines 283-300](https://github.com/CiscoDevNet/ansible-nd/pull/405/files#diff-cba745ac1866c79c2b655f2cdb85f37762b0570ffaddc8bd922aed7eda5adcc7R283) applies it without considering whether the operation is normal onboarding, POAP, or swap.
- The corresponding omitted defaults are `false` for eBGP and Campus. Changing every fabric default to `false` would route this omitted-value example through the reboot wait, but it would also change normal iBGP brownfield onboarding and would not protect an explicit `preserve_config: true` POAP/swap input.
- [`nd_switch_resources.py` lines 1546-1562](https://github.com/CiscoDevNet/ansible-nd/pull/405/files#diff-6dd07c8d609b04466f34b90a9b295c8710d66a6d17802f5277faf98290274e19R1546) places every NX-OS entry with `preserve_config: true` in `nxos_preserve`.
- [`utils.py` lines 792-807](https://github.com/CiscoDevNet/ansible-nd/pull/405/files#diff-1846ae50f0a5d3c8903c0ba0b65de49dec6da09f972884db0c7cd7ac92121b3bR792) explains that POAP devices must be observed through the reboot transition, but [lines 863-866](https://github.com/CiscoDevNet/ansible-nd/pull/405/files#diff-1846ae50f0a5d3c8903c0ba0b65de49dec6da09f972884db0c7cd7ac92121b3bR863) accept a preserve-config device as soon as `systemMode == normal`.

For example:

```yaml
- name: Bootstrap an iBGP leaf
  cisco.nd.nd_manage_switches:
    fabric: FABRIC_IBGP
    state: merged
    config:
      - seed_ip: 192.0.2.20
        username: admin
        password: "{{ switch_password }}"
        role: leaf
        poap:
          serial_number: FDO12345ABC
          hostname: leaf1
        # preserve_config omitted
```

PR #405 derives:

```text
preserve_config=True
nxos_reload=[]
nxos_preserve=['FDO12345ABC']
```

Immediately after import, ND can still return the pre-reboot state:

```text
systemMode=normal
discoveryStatus=ok
```

The preserve-config branch accepts that first response:

```text
wait_returned=True
calls=['fetch']
```

Expected behavior is to keep waiting until the reboot transition has been observed:

```text
normal -> unreachable -> ok
```

**Existing PR overlap**

Related to the [resolved wait-set discussion](https://github.com/CiscoDevNet/ansible-nd/pull/405#discussion_r3616756776) and the active preserve-config thread. Neither discussion tests the derived iBGP default through POAP or swap readiness.

Related to [#488](https://github.com/CiscoDevNet/ansible-nd/issues/488), which owns fabric-derived preserve defaults. It does not cover operation-specific POAP/swap reboot behavior.

**Impact**

The module can proceed with credentials, save, deploy, and report success while the switch has not completed—or even begun—its expected reboot. Subsequent finalization may fail or act against a switch that is not ready.

**Suggested fix**

Make readiness operation-aware instead of changing every fabric default. Normal onboarding can retain its fabric-derived value, while NX-OS POAP and swap imports should always use the reload-observation policy regardless of `preserve_config`.

If preserve mode is not meaningful for those operations, reject or ignore it explicitly. Add omitted-value and explicit-`true` regression tests that require an observed `unreachable -> ok` transition before finalization.

plugins/module_utils/manage_switches/fabric_switch_capabilities.py
mikewiebe: ### High: Unsupported fabrics still reach destructive removal paths

**Issue**

The new supported-fabric gate runs after the early `deleted` and empty-`overridden` branches. Those states can therefore remove selected switches—or the entire inventory—from IPFM or any other fabric the module declares unsupported.

**Evidence**

- [`nd_switch_resources.py` lines 3033-3045](https://github.com/CiscoDevNet/ansible-nd/pull/405/files#diff-6dd07c8d609b04466f34b90a9b295c8710d66a6d17802f5277faf98290274e19R3033) returns directly into deletion for both states.
- [`nd_switch_resources.py` lines 3050-3052](https://github.com/CiscoDevNet/ansible-nd/pull/405/files#diff-6dd07c8d609b04466f34b90a9b295c8710d66a6d17802f5277faf98290274e19R3050) invokes capability validation only after those returns.
- [`fabric_switch_capabilities.py` lines 184-195](https://github.com/CiscoDevNet/ansible-nd/pull/405/files#diff-cba745ac1866c79c2b655f2cdb85f37762b0570ffaddc8bd922aed7eda5adcc7R184) permits only iBGP, eBGP, and Campus onboarding.
- [`nd_manage_switches.py` lines 289-293](https://github.com/CiscoDevNet/ansible-nd/pull/405/files#diff-342b604bca79f054258da9349010dd7bc8ee600a70a11edfde997399a0a09067R289) says validation occurs before write operations.

For example, the module correctly rejects a nonempty write configuration against IPFM:

```yaml
- cisco.nd.nd_manage_switches:
    fabric: IPFM_A
    state: merged
    config:
      - seed_ip: 192.0.2.10
        username: admin
        password: "{{ switch_password }}"
```

```text
Fabric type 'ipfm' is not supported.
```

However, changing only the state to `deleted` bypasses that gate:

```yaml
- name: Remove one switch from the unsupported IPFM fabric
  cisco.nd.nd_manage_switches:
    fabric: IPFM_A
    state: deleted
    config:
      - seed_ip: 192.0.2.10
```

The exact-head probe called the bulk-remove path with that switch's serial number.

More seriously, an empty overridden configuration also bypasses validation:

```yaml
- name: Remove every switch from the unsupported IPFM fabric
  cisco.nd.nd_manage_switches:
    fabric: IPFM_A
    state: overridden
    config: []
```

That path calls `_handle_deleted_state(None)`, which treats the whole existing inventory as the deletion target.

**Existing PR overlap**

This expands the active IPFM fail-closed thread. The latest change blocks onboarding, but destructive states remain reachable because they return before the new validation call.

Tracked by [#488](https://github.com/CiscoDevNet/ansible-nd/issues/488), which requires fabric validation before write operations. [#477](https://github.com/CiscoDevNet/ansible-nd/issues/477) is also related because it requires explicit validation of switch-removal safety.

**Impact**

A module advertised as unsupported for IPFM can still remove a selected IPFM switch. Empty `overridden` can deboard every switch in the fabric, making the fail-closed boundary inconsistent specifically on destructive operations.

**Suggested fix**

Resolve and validate the fabric support boundary before every mutating state, including `deleted` and empty `overridden`.

Keep onboarding-field validation separate so deletion does not require credentials, role, platform, or preserve settings. If deboarding from unsupported fabrics is intentionally allowed, document and test that narrower contract explicitly instead of bypassing the gate implicitly.

plugins/module_utils/manage_switches/nd_switch_resources.py
mikewiebe: ### High: Gathered output still cannot be replayed through the module argspec

**Issue**

The prior missing-`platform_type` defect is fixed, but gathered output still uses controller enum spellings instead of the module's public argspec spellings.

Every gathered switch contains lowercase `auth_proto`, while multiword roles use camelCase. Ansible rejects those values before Pydantic or capability validation runs.

**Evidence**

- [`config_models.py` lines 680-699](https://github.com/CiscoDevNet/ansible-nd/pull/405/files#diff-56b5c0c375eba2e5d66ef3b8a903e186dc96dd78e352c5d8bbdb2ff9a8333d42R680) says gathered output is immediately usable as `config` after replacing credentials, but returns internal/controller enum values without converting them to the public input format.
- [`config_models.py` lines 730-773](https://github.com/CiscoDevNet/ansible-nd/pull/405/files#diff-56b5c0c375eba2e5d66ef3b8a903e186dc96dd78e352c5d8bbdb2ff9a8333d42R730) accepts uppercase authentication choices such as `MD5` and snake_case roles such as `border_gateway`.

For example, gathering an NX-OS border gateway returns:

```yaml
gathered:
  - seed_ip: 192.0.2.10
    auth_proto: md5
    role: borderGateway
    platform_type: nx-os
    username: <username>
    password: <password>
```

After replacing the credential placeholders, a user follows the documented gather/edit/reapply workflow:

```yaml
- name: Reapply the gathered switch configuration
  cisco.nd.nd_manage_switches:
    fabric: FABRIC_IBGP
    state: merged
    config:
      - seed_ip: 192.0.2.10
        auth_proto: md5
        role: borderGateway
        platform_type: nx-os
        username: admin
        password: "{{ switch_password }}"
```

The real Ansible argument validator rejects it:

```text
value of auth_proto must be one of: MD5, SHA, ...; got: md5

value of role must be one of: ..., border_gateway, ...;
got: borderGateway
```

Even a simple role such as `leaf` fails because every gathered entry includes `auth_proto: md5`.

The new gathered test confirms only that `platform_type` appears. It never serializes the complete gathered result and validates it through the module argspec.

**Existing PR overlap**

Related to the resolved gathered non-NX replay thread. The precise missing-platform defect is fixed, but this separate enum-serialization defect leaves the same end-to-end workflow broken.

No matching open issue found. The complete open-issue inventory was searched for the switch gathered, role, and authentication-protocol round-trip symptom. Issue #488 covers fabric/platform capability validation, not gathered serialization.

**Impact**

No gathered switch entry can be fed back to `state: merged` after credentials are supplied because the authentication value is invalid. Switches with multiword roles receive a second validation error.

**Suggested fix**

Serialize gathered fields using the public argspec representation:

- `md5` → `MD5`
- `borderGateway` → `border_gateway`
- Apply equivalent mappings to all authentication protocols and multiword roles.

Add an end-to-end test that JSON-serializes gathered output, replaces credentials, validates it through the real module argspec, and then validates the resulting model for simple and multiword roles.

plugins/module_utils/manage_switches/nd_switch_resources.py
mikewiebe: ### 3. High: No switches can be added to External and Inter-Fabric Connectivity fabrics

**Issue**

PR #405’s new fabric exposure gate excludes the fabric type shown in ND as **External and Inter-Fabric Connectivity** (`externalConnectivity`).

Consequently, `cisco.nd.nd_manage_switches` rejects every switch addition to this fabric type before discovery or any write. This affects **NX-OS, IOS-XE, IOS-XR, and `other`** platforms.

This is a regression: current `develop` allows the request to proceed to the controller, while PR #405 rejects it locally.

**Practical example**

```yaml
- name: Add an NX-OS switch to External and Inter-Fabric Connectivity
  cisco.nd.nd_manage_switches:
    fabric: EXTERNAL_A
    state: merged
    config:
      - seed_ip: 192.0.2.30
        username: admin
        password: "{{ switch_password }}"
        platform_type: nx-os
        role: leaf
```

PR #405 fails with the equivalent of:

```text
Fabric type 'externalConnectivity' is not supported.
```

Changing `platform_type` does not help because the fabric type—not the switch platform—is rejected.

**Impact**

No switches can be added to **External and Inter-Fabric Connectivity fabrics** through `cisco.nd.nd_manage_switches`. Read-only gathering still works, and legacy `cisco.dcnm.dcnm_inventory` or direct API workflows are unaffected.

**Suggested fix**

Add normalized `externalConnectivity` to the supported-fabric gate and test onboarding with all four supported platform types. This is also explicitly required by open issue [#488](https://github.com/CiscoDevNet/ansible-nd/issues/488).

plugins/module_utils/manage_switches/fabric_switch_capabilities.py
AKDRG: ### Fix

Moved fabric support validation ahead of every mutating state path, including destructive paths.

The following states now validate the fabric support boundary before any delete or reconciliation handler can run:

- `deleted`
- empty `overridden`
- non-empty `overridden`
- `replaced`

Unsupported fabrics such as IPFM remain readable through `gathered`, but mutating states fail closed before selected delete or delete-all behavior can execute.

plugins/module_utils/manage_switches/nd_switch_resources.py
AKDRG: ## 1. Omitted platform defaults still break an existing Campus IOS-XE no-op

### Fix

Separated existing no-op reconciliation from onboarding/write validation.

The module now classifies switches against live inventory first. If an existing switch is already idempotent, it is not forced through onboarding capability validation just because `platform_type` was omitted in the playbook.

For actionable existing configs, omitted `platform_type` is resolved from inventory when safe, such as IOS-XE, IOS-XR, or other. Read-side-only platforms like `sonic` and `apic` are not defaulted to NX-OS for write operations.

## 2. Derived iBGP preserve default lets POAP and swap finish before reboot

### Fix

Made post-add readiness operation-aware instead of relying only on `preserve_config`.

Data Center VXLAN iBGP can still derive `preserve_config: true` for normal onboarding. However, NX-OS POAP and swap workflows are forced into the reload-observation wait path, even when `preserve_config` is true or derived true.

This ensures the module waits for the expected reload transition instead of accepting the first pre-reboot `normal/ok` response.

## 3. No switches can be added to External and Inter-Fabric Connectivity fabrics

### Fix

Added External / `externalConnectivity` to the explicit supported-fabric gate.

The support matrix now allows External and Inter-Fabric Connectivity fabrics with these supported platforms:

- `nx-os`
- `ios-xe`
- `ios-xr`
- `other`

mikewiebe (approved):

plugins/module_utils/manage_switches/fabric_switch_capabilities.py
allenrobel: Verified at head 041a2b62: `_validate_supported_fabric()` now runs in `manage_state()` ahead of the `deleted` and empty-`overridden` returns as well as the merged/replaced/overridden classify path, `gathered` stays readable for unsupported fabrics, and `test_manage_state_blocks_unsupported_fabric_before_destructive_delete_paths` covers all four mutating states against `ipfm`. Keeping the IPFM matrix entry behind the exposure gate rather than deleting it is fine given the comment on `SUPPORTED_SWITCH_ONBOARDING_FABRIC_TYPES`. Resolving.

allenrobel (approved):
Re-verified the post-08-13 delta (`d0e213cc`, `041a2b62`): the fabric exposure gate is now enforced before every mutating state including `deleted` and empty `overridden`, DOCUMENTATION matches the supported set, and the switch-resources unit tests pass locally (74 passed). LGTM.

mikewiebe:
Merging with 2 required approvals

Issue #353, opened 2026-06-26
Consolidate per-model interface-name normalization into shared helpers


allenrobel:
## Summary

Interface-name normalization is copy-pasted across interface models, each model carrying its own
`field_validator` that rewrites a user-supplied interface name to ND's canonical wire form. The logic
should live in one shared place so every `nd_interface_*` module (and any other module that round-trips
interface names) shares a single source of truth.

There are currently **three distinct behaviors**, keyed by interface type, because ND's canonical wire
casing differs per type:

| Behavior | Example | Where today |
|---|---|---|
| Physical ethernet: expand any case-insensitive prefix of `Ethernet` to the canonical form; preserve digits/separators; Title-case an unrecognized prefix | `eth1/1` → `Ethernet1/1`, `e1/1` → `Ethernet1/1` | `ethernet_access_interface.normalize_interface_name`; port-channel **members** via `port_channel_*.normalize_ports` → `_normalize_member_name` |
| Port-channel name: lowercase | `Port-Channel501` → `port-channel501` | `port_channel_access_interface.normalize_interface_name`, `port_channel_trunk_host_interface.normalize_interface_name` |
| Loopback name: lowercase | `Loopback0` → `loopback0` | `loopback_interface.normalize_interface_name` |

## Why it matters

The wire key is matched **exactly** for idempotency. A name that is not canonicalized never matches ND's
returned form, so the resource reports `changed` and re-deploys on every run — the exact bug fixed in #269
(`nd_interface_port_channel_access`) and again in #270 (`nd_interface_port_channel_trunk_host`). Each new
interface module re-implements the same normalizer, so the bug can (and did) regress independently per
branch (the #270 copy had been weakened to a first-char Title-case before review). Centralizing it removes
that whole class of per-branch regression.

There is also a **naming collision** today that a consolidation should resolve: `normalize_interface_name`
already means *different things* in different models — in `ethernet_access` it expands to `Ethernet…`,
while in `port_channel_*`/`loopback` it lowercases. Same name, opposite transform.

## Affected files

- `plugins/module_utils/models/interfaces/ethernet_access_interface.py` (`normalize_interface_name`, `_CANONICAL_INTERFACE_TYPE`)
[...]

PR #411, opened 2026-07-21
Pair-scoped vPC interface identity: support same vPC id on two pairs

### LINKED ISSUE #356: vPC name-only identifier breaks when two pairs reuse the same vPC id in one fabric
## Summary

vPC interface models identify an interface by **`interface_name` alone** (e.g. `vpc100`), deliberately excluding
`switch_ip` from identity. This is the `vpc-interface-dual-peer-duplicate` workaround: ND echoes a single vPC
interface from **both** peers' per-switch `GET .../interfaces`, so a composite `(switch_ip, interface_name)` identity
made the peer-side copy look orphaned and `_manage_override_deletions` queued a spurious delete.

The name-only identity is correct for the dual-peer dedup, but it breaks down when **two different vPC pairs in the
same fabric reuse the same vPC id** (legal in NX-OS — vPC ids are per-pair):

1. **Hard failure (CONFIRMED):** listing both same-name interfaces in one task `config` collides in
   `NDConfigCollection.add()` ("Item with identifier `vpc100` already exists") and fails the entire task, even though
   the two interfaces live on distinct pairs.
2. **Silent override gap (PLAUSIBLE):** under `state=overridden`, `query_all`'s name-keyed dedup collapses the two
   pairs' interfaces to one entry, so the second pair's interface is hidden — it is neither deleted (override no
   longer authoritative) nor correctly matched, and an update may target the wrong pair.

## Why it matters

The two requirements are in tension: the dual-peer dedup **needs** name-only identity, but multi-pair same-id
**needs** identity to include the pair. Resolving this requires distinguishing "same interface seen on both peers of
one pair" from "two interfaces with the same id on two different pairs" — which the per-switch GET does not make
trivial (both look like duplicate names).

## Affected PRs / modules

- #279 `nd_interface_vpc_access` (`accessVpcHost`)
- #280 `nd_interface_vpc_trunk_host` (`trunkVpcHost`)
- Any future `vpc_*` interface type sharing `VpcInterfaceBaseOrchestrator`.

## Proposed approach

1. **Lab-verify** whether ND 4.2.1 permits two vPC pairs in one fabric to own the same vPC id, and whether the
   per-switch GET exposes a stable pair discriminator (e.g. the `{switchId, peerSwitchId}` set) that distinguishes
   the two cases. Record findings in the vault.
2. If supported, key dedup on **`interface_name` + the unordered peer-pair `{switchId, peerSwitchId}`** instead of
   `interface_name` alone: two records for one pair (peers swapped) collapse; two records on different pairs stay
   distinct. Carry the pair discriminator into the config-collection identity so same-name-different-pair items no
   longer collide. Add idempotency + override tests for the multi-pair case.
3. If unsupported by ND, document the single-pair-per-id assumption in the module `notes:` and validate against it.

allenrobel:
## Merge Ordering — #422 merged 2026-08-28; rebased and lab-verified

#422 merged into `develop` on 2026-08-28 and this branch is rebased on it (clean replay, no conflicts). The replaced/overridden re-run it called for is done — see the 2026-08-28 entry under Test Notes. Original reasoning kept for the record:

~~**Prefer merging #422 (replaced/overridden field-removal detection) before this PR.**~~ Not a textual concern — the two
branches touch different regions of the vPC model files (identifier block here vs. policy-model defaults tables there)
and auto-merge cleanly. The coupling is semantic:

- Review of #422 surfaced a vPC-specific replaced/overridden idempotency bug in its reverse diff pass: ND echoes the
  orchestrator-injected `peerSwitchId` inside the policy block, which the proposed config can never express, so the
  reverse pass reports a removal on every run (permanent `changed=true` + re-PUT). The fix lands in #422 and will touch
  the same vPC model files this PR modifies.
- This PR changes vPC identity and `query_all` dedup — exactly the machinery that produces the `before[]` set #422's
  reverse pass classifies against.

**After #422 merges:** rebase this branch and re-run the vPC `replaced`/`overridden` integration scenarios (including
the new `multi_pair.yaml`) against the live lab before merging, to confirm idempotency holds with the reverse pass
active on the rebased code.

## Related Issue(s)

Fixes #356

Added NaC-CapSet1 since Issue #356 also contains this label.

## Proposed Changes

- Restore the composite (`switch_ip`, `interface_name`) identifier on both vPC interface models
  (`accessVpcHost`, `trunkVpcHost`), matching every other interface module. Two vPC pairs in one fabric
  can now legally own the same vPC id (e.g. `vpc100` on two pairs) in a single task `config` — previously
  this collided at parse time ("Item with identifier `vpc100` already exists").
- Move the `vpc-interface-dual-peer-duplicate` workaround from the model identity into
  `VpcInterfaceBaseOrchestrator.query_all`: dedup is now keyed on `interfaceName` + the unordered
  `{switchId, peerSwitchId}` pair set, so one pair's two peer echoes collapse while same-name interfaces
  on different pairs stay distinct (fixes the `state: overridden` silent-collapse gap).
- Add a `preflight()` guard that fails fast when one config lists the same `interface_name` under both
  peers of the same vPC pair (two items targeting one ND resource). Pair resolution only runs for
  duplicated names, so runs with unique names issue zero extra requests.
- Document the pair-scoped identity and the same-pair-duplicate rejection in both modules' `notes:`.
- Integration: new `multi_pair.yaml` scenario in both vPC targets (same name + id created on both lab
  pairs in one task, idempotency, fail-fast guard, `overridden` reconciliation, cleanup), plus a
  backward-compatible `nd_test_manage_vpc_pairs` gate (default `true`) so test runs can treat hand-built
  lab vPC pairs as substrate instead of creating/unpairing them. The multi-pair scenario deliberately
  runs after `deleted.yaml`: its `state: overridden` step is fabric-wide, so running it earlier would
  delete the `vpc201`/`vpc101` artifacts the deleted-state tests assert against.

Design/spec: `docs/superpowers/specs/2026-07-20-vpc-multi-pair-identity-design.md`. Lab premise directly
confirmed on ND 4.2.1: the `vpcId` resource pool is `devicePair`-scoped, and creating the same-name/same-id
vPC interface intent on two pairs succeeds (findings recorded in the team bug-tracker vault note
`vpc-interface-dual-peer-duplicate`).

## Test Notes

- **2026-08-28 (post-#422 rebase):** rebased onto `develop` @ `bb985d74`; full unit suite **4146 passed**; CI green. Live integration against the rebuilt ND 4.2.1.10 lab (SITE1, pair S1_LE1/S1_LE2, `nd_test_manage_vpc_pairs=false`): the `setup` → `merged` → `replaced` → `overridden` → `deleted` scenarios of both `nd_interface_vpc_trunk_host` and `nd_interface_vpc_access` are **green** (each `ok=30 changed=8 failed=0`), and every `REPLACED IDEMPOTENT` / `OVERRIDDEN IDEMPOTENT` re-apply reports no change — confirming replaced/overridden idempotency holds with #422's reverse pass (`peerSwitchId` exclusion) active on this branch's pair-scoped identity and dedup. Then, after rebuilding the second lab pair (S1_LE3/S1_LE4, domain 2, physical peer link — the rebuilt lab had left them unpaired), **`multi_pair.yaml` re-run green on both targets** (each `ok=20 changed=4 failed=0`): same-name/same-id create on both pairs, idempotent re-apply, the same-pair fail-fast guard, cross-pair `overridden` reconciliation plus its idempotent re-apply, and cleanup. Post-run raw-API check: no vPC interfaces left on any of the four leafs, both pairs intact.
- Unit: full suite green via `ndpytest tests/unit/` (3184 passed), including new multi-pair dedup,
  missing-`peerSwitchId` fallback, and preflight-guard tests (TDD; fixtures modeled on captured live wire
  data from both lab pairs).
- Sanity: `ndtest` — all checks relevant to the changed files pass; 2 pre-existing failures on develop
  (`action-plugin-docs`, `pylint` on `manage_fabric_base.py`) are in files untouched by this branch.
- Live integration (ND 4.2.1, SITE1, two vPC pairs — S1_LE1/S1_LE2 domain 1, S1_LE3/S1_LE4 domain 2, both
  physical peer link): `ansible-test network-integration` for `nd_interface_vpc_trunk_host` and
  `nd_interface_vpc_access`, run with `nd_test_manage_vpc_pairs=false`. Both targets green end-to-end
  (each `ok=49 changed=12 failed=0`), including the new multi-pair blocks; post-run raw-API check
  confirmed no leftover `vpcId` pool entries and both lab pairs intact.
- validate-modules pass for both modules.

4.2.1

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

https://claude.ai/code/session_01BHGFDqqU37PRqNHhKw1FCQ

### DIFF
--- a/plugins/module_utils/orchestrators/vpc_interface_base.py
+++ b/plugins/module_utils/orchestrators/vpc_interface_base.py
@@ -410,7 +410,9 @@ class VpcInterfaceBaseOrchestrator(NDBaseInterfaceOrchestrator[ModelType]):
         Return the unordered vPC pair discriminator for an interface record: `frozenset({switchId, peerSwitchId})`. The two peer copies
         of one vPC interface carry the same set (only the orientation swaps), while same-name interfaces on different pairs carry
         disjoint sets (issue #356). Falls back to `frozenset({switch_id})` when `peerSwitchId` is absent so a malformed record degrades
-        to per-switch identity instead of raising.
+        to per-switch identity instead of raising. Note: if BOTH peer copies of one interface ever lacked `peerSwitchId`, they would get
+        disjoint keys, both would survive dedup, and `state: overridden` could queue a spurious delete of the unconfigured copy —
+        acceptable only because the wire always carries the field (lab-verified 2026-07-20).
 
         ## Raises
 
--- a/plugins/module_utils/orchestrators/vpc_interface_base.py
+++ b/plugins/module_utils/orchestrators/vpc_interface_base.py
@@ -324,25 +324,25 @@ class VpcInterfaceBaseOrchestrator(NDBaseInterfaceOrchestrator[ModelType]):
         except Exception as e:
             raise RuntimeError(f"Query failed for {model_instance.get_identifier_value()}: {e}") from e
 
-    def _configured_switch_ip_by_interface_name(self) -> dict[str, str]:
+    def _configured_switch_ips_by_interface_name(self) -> dict[str, set[str]]:
         """
         # Summary
 
-        Map each config `interface_name` to the user-supplied `switch_ip`. A vPC interface spans two peers and ND
-        returns it on both; this lets `query_all` pick the peer whose IP the user actually configured so the existing
-        state's composite identifier `(switch_ip, interface_name)` matches the proposed identifier and the run stays
-        idempotent regardless of which peer the user names.
+        Map each config `interface_name` to the set of user-supplied `switch_ip` values. A vPC interface spans two peers and ND returns
+        it on both; `query_all` uses this to keep the peer copy whose IP the user actually configured, so the existing state's composite
+        identifier `(switch_ip, interface_name)` matches the proposed identifier and the run stays idempotent regardless of which peer
+        the user names. A set (not a single IP) because the same name may legally appear on multiple pairs (issue #356).
 
         ## Raises
 
         None
         """
-        mapping: dict[str, str] = {}
+        mapping: dict[str, set[str]] = {}
         for item in self.rest_send.params.get("config") or []:
             name = item.get("interface_name")
             switch_ip = item.get("switch_ip")
             if name and switch_ip:
-                mapping[name] = switch_ip
+                mapping.setdefault(name, set()).add(switch_ip)
         return mapping
 
     @staticmethod
@@ -361,6 +361,28 @@ class VpcInterfaceBaseOrchestrator(NDBaseInterfaceOrchestrator[ModelType]):
         network_os = config_data.get("networkOS") or {}
         policy = network_os.get("policy") or {}
         return policy.get("policyType")
+
+    @staticmethod
+    def _pair_key(iface: dict, switch_id: str) -> frozenset[str]:
+        """
+        # Summary
+
+        Return the unordered vPC pair discriminator for an interface record: `frozenset({switchId, peerSwitchId})`. The two peer copies
+        of one vPC interface carry the same set (only the orientation swaps), while same-name interfaces on different pairs carry
+        disjoint sets (issue #356). Falls back to `frozenset({switch_id})` when `peerSwitchId` is absent so a malformed record degrades
+        to per-switch identity instead of raising.
+
+        ## Raises
+
+        None
+        """
+        config_data = iface.get("configData") or {}
+        network_os = config_data.get("networkOS") or {}
+        policy = network_os.get("policy") or {}
+        peer_switch_id = policy.get("peerSwitchId")
+        if peer_switch_id:
+            return frozenset({switch_id, peer_switch_id})
+        return frozenset({switch_id})
 
     def _managed_vpc_interfaces(self, switch_ip: str, switch_id: str, managed_types: set[str]) -> list[dict]:
         """
@@ -399,6 +421,9 @@ class VpcInterfaceBaseOrchestrator(NDBaseInterfaceOrchestrator[ModelType]):
         Each returned interface dict is enriched with a `switchIp` field so that the model can be constructed with the
         composite identifier `(switch_ip, interface_name)`.
 
+        Dedup is keyed on `(interfaceName, frozenset({switchId, peerSwitchId}))`, not `interfaceName` alone, so that two vPC pairs in
+        the same fabric may legally reuse the same vPC interface name (issue #356).
+
         ## Raises
 
         ### RuntimeError
@@ -410,23 +435,27 @@ class VpcInterfaceBaseOrchestrator(NDBaseInterfaceOrchestrator[ModelType]):
         managed_types = self._managed_policy_types()
         try:
             self.validate_prerequisites()
-            configured_ip_by_name = self._configured_switch_ip_by_interface_name()
+            configured_ips_by_name = self._configured_switch_ips_by_interface_name()
             # TODO(4.2.1) vpc-interface-dual-peer-duplicate
-            # ND returns each vPC interface TWICE — once per peer switch — with identical configData. Dedupe by
-            # interfaceName. When the interface is in the user config, keep the peer whose switchIp the user supplied so
-            # idempotency holds regardless of which peer they name; otherwise keep the alphabetically-lower switchId for
-            # a stable representative. Without this dedupe, `_manage_override_deletions` would see the peer-side copy as
-            # "not in proposed" and queue a spurious delete.
-            interfaces_by_name: dict[str, tuple[str, dict]] = {}
+            # ND returns each vPC interface TWICE — once per peer switch — with identical configData. Dedupe on
+            # (interfaceName, frozenset({switchId, peerSwitchId})): one pair's two copies share the unordered set and
+            # collapse, while same-name interfaces on DIFFERENT pairs have disjoint sets and stay distinct (two pairs
+            # may legally reuse a vPC id — the vpcId pool is devicePair-scoped; issue #356). When the interface is in
+            # the user config, keep the peer whose switchIp the user supplied so idempotency holds regardless of which
+            # peer they name; otherwise keep the alphabetically-lower switchId for a stable representative. Without
+            # this dedupe, `_manage_override_deletions` would see the peer-side copy as "not in proposed" and queue a
+            # spurious delete.
+            interfaces_by_key: dict[tuple[str, frozenset[str]], tuple[str, dict]] = {}
             for switch_ip, switch_id in self._switches_to_query().items():
                 for iface in self._managed_vpc_interfaces(switch_ip, switch_id, managed_types):
                     name = iface.get("interfaceName")
                     if name is None:
                         continue
-                    existing = interfaces_by_name.get(name)
-                    if self._prefers_candidate(name, switch_id, switch_ip, existing, configured_ip_by_name):
-                        interfaces_by_name[name] = (switch_id, iface)
-            return [entry[1] for entry in interfaces_by_name.values()]
+                    key = (name, self._pair_key(iface, switch_id))
+                    existing = interfaces_by_key.get(key)
+                    if self._prefers_candidate(name, switch_id, switch_ip, existing, configured_ips_by_name):
+                        interfaces_by_key[key] = (switch_id, iface)
+            return [entry[1] for entry in interfaces_by_key.values()]
         except Exception as e:
             raise RuntimeError(f"Query all failed: {e}") from e
 
@@ -436,15 +465,15 @@ class VpcInterfaceBaseOrchestrator(NDBaseInterfaceOrchestrator[ModelType]):
         switch_id: str,
         switch_ip: str,
         existing: tuple[str, dict] | None,
-        configured_ip_by_name: dict[str, str],
+        configured_ips_by_name: dict[str, set[str]],
     ) -> bool:
         """
         # Summary
 
-        Decide whether a newly-seen per-peer copy of a vPC interface should replace the one already kept for `name`.
-        Prefer the peer whose `switch_ip` the user configured (idempotency); when neither peer is the configured one
-        (or the interface is not in config, e.g. an `overridden` deletion candidate), prefer the alphabetically-lower
-        `switch_id` for a stable representative.
+        Decide whether a newly-seen per-peer copy of a vPC interface should replace the one already kept for its dedup key. Prefer the
+        peer whose `switch_ip` the user configured (idempotency); when neither peer is the configured one (or the interface is not in
+        config, e.g. an `overridden` deletion candidate), prefer the alphabetically-lower `switch_id` for a stable representative. The
+        two candidates for one key are always the two peers of one pair (issue #356 keys dedup on the unordered pair set).
 
         ## Raises
 
@@ -452,10 +481,10 @@ class VpcInterfaceBaseOrchestrator(NDBaseInterfaceOrchestrator[ModelType]):
         """
         if existing is None:
             return True
-        configured_ip = configured_ip_by_name.get(name)
-        if configured_ip is not None:
-            if switch_ip == configured_ip:
+        configured_ips = configured_ips_by_name.get(name, set())
+        if configured_ips:
+            if switch_ip in configured_ips:
                 return True
-            if existing[1].get("switchIp") == configured_ip:
+            if existing[1].get("switchIp") in configured_ips:
                 return False
         return switch_id < existing[0]
--- a/plugins/module_utils/orchestrators/vpc_interface_base.py
+++ b/plugins/module_utils/orchestrators/vpc_interface_base.py
@@ -371,7 +371,8 @@ class VpcInterfaceBaseOrchestrator(NDBaseInterfaceOrchestrator[ModelType]):
         Map each config `interface_name` to the set of user-supplied `switch_ip` values. A vPC interface spans two peers and ND returns
         it on both; `query_all` uses this to keep the peer copy whose IP the user actually configured, so the existing state's composite
         identifier `(switch_ip, interface_name)` matches the proposed identifier and the run stays idempotent regardless of which peer
-        the user names. A set (not a single IP) because the same name may legally appear on multiple pairs (issue #356).
+        the user names. A set (not a single IP) because the same name may legally appear on multiple pairs (issue #356). Keys are
+        canonical (lowercase) names — see `_canonical_interface_name`.
 
         ## Raises
 
@@ -382,8 +383,24 @@ class VpcInterfaceBaseOrchestrator(NDBaseInterfaceOrchestrator[ModelType]):
             name = item.get("interface_name")
             switch_ip = item.get("switch_ip")
             if name and switch_ip:
-                mapping.setdefault(name, set()).add(switch_ip)
+                mapping.setdefault(self._canonical_interface_name(name), set()).add(switch_ip)
         return mapping
+
+    @staticmethod
+    def _canonical_interface_name(name: str) -> str:
+        """
+        # Summary
+
+        Return the canonical (lowercase) form of a vPC interface name. The vPC models lowercase `interface_name` on validation
+        (`VPC501` -> `vpc501`, matching what ND echoes), but `query_all` reads the user config raw from `rest_send.params`, so both
+        the configured names and the response names must pass through this one canonicalizer before they are compared; otherwise a
+        mixed-case config never matches its own echo and the dedup falls back to the wrong peer (PR #411 review).
+
+        ## Raises
+
+        None
+        """
+        return name.lower()
 
     @staticmethod
     def _policy_type(iface: dict) -> str | None:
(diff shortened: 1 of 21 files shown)


plugins/modules/nd_interface_vpc_trunk_host.py
```diff
@@ -346,6 +346,11 @@
 - The primary switch supplied in O(config[].switch_ip) must already be in a vPC pair (managed by
   M(cisco.nd.nd_manage_vpc_pair)). The peer serial is auto-resolved from the pair record.
 - C(peer1) refers to the switch supplied in O(config[].switch_ip); C(peer2) refers to the auto-resolved peer.
+- A vPC interface is identified by the combination of O(config[].switch_ip) and O(config[].interface_name). Two different vPC
+  pairs in the same fabric may reuse the same vPC id (for example C(vpc100) on two pairs); list each under its own pair's
+  O(config[].switch_ip).
+- Listing the same O(config[].interface_name) under both peers of the same vPC pair is rejected; list each vPC interface once,
+  under either peer.
```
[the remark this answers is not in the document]
allenrobel: Agreed — the duplicate-peer guard lives in `preflight()`, which the state machine runs only for `merged`/`replaced`/`overridden`; the `deleted` path skips it. Scoped the note to those states and documented that for `deleted` duplicate entries simply resolve to the same vPC interface, in commit "Scope duplicate-peer rejection note to states that run preflight (#356)".

plugins/modules/nd_interface_vpc_access.py
```diff
@@ -303,6 +303,11 @@
 - The primary switch supplied in O(config[].switch_ip) must already be in a vPC pair (managed by
   M(cisco.nd.nd_manage_vpc_pair)). The peer serial is auto-resolved from the pair record.
 - C(peer1) refers to the switch supplied in O(config[].switch_ip); C(peer2) refers to the auto-resolved peer.
+- A vPC interface is identified by the combination of O(config[].switch_ip) and O(config[].interface_name). Two different vPC
+  pairs in the same fabric may reuse the same vPC id (for example C(vpc100) on two pairs); list each under its own pair's
+  O(config[].switch_ip).
+- Listing the same O(config[].interface_name) under both peers of the same vPC pair is rejected; list each vPC interface once,
+  under either peer.
```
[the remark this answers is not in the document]
allenrobel: Agreed — the duplicate-peer guard lives in `preflight()`, which the state machine runs only for `merged`/`replaced`/`overridden`; the `deleted` path skips it. Scoped the note to those states and documented that for `deleted` duplicate entries simply resolve to the same vPC interface, in commit "Scope duplicate-peer rejection note to states that run preflight (#356)".

tests/integration/targets/nd_interface_vpc_trunk_host/tasks/multi_pair.yaml
```diff
@@ -0,0 +1,141 @@
...
+    config:
+      - switch_ip: "{{ test_switch_ip_peer1 }}"
+        interface_name: "{{ vpc_multi_pair_name }}"
+    state: deleted
+  register: mp_cleanup
+
+- name: "MULTI-PAIR CLEANUP: Verify deletion"
+  ansible.builtin.assert:
+    that:
+      - mp_cleanup is changed
```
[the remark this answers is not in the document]
allenrobel: Good catch — added a best-effort delete of pair 2's interface to the CLEANUP section (`failed_when: false`, no `changed` assert, so it is a no-op when the overridden step already removed it) in commit "Best-effort pair-2 delete in multi-pair cleanup (#356)".

tests/integration/targets/nd_interface_vpc_access/tasks/multi_pair.yaml
```diff
@@ -0,0 +1,141 @@
...
+    config:
+      - switch_ip: "{{ test_switch_ip_peer1 }}"
+        interface_name: "{{ vpc_multi_pair_name }}"
+    state: deleted
+  register: mp_cleanup
+
+- name: "MULTI-PAIR CLEANUP: Verify deletion"
+  ansible.builtin.assert:
+    that:
+      - mp_cleanup is changed
```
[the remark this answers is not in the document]
allenrobel: Good catch — added a best-effort delete of pair 2's interface to the CLEANUP section (`failed_when: false`, no `changed` assert, so it is a no-op when the overridden step already removed it) in commit "Best-effort pair-2 delete in multi-pair cleanup (#356)".

docs/superpowers/plans/2026-07-20-vpc-multi-pair-identity.md
```diff
@@ -0,0 +1,1161 @@
+# Pair-Aware vPC Interface Identity (#356) Implementation Plan
```
mikewiebe: Need to remove these docs

docs/superpowers/specs/2026-07-20-vpc-multi-pair-identity-design.md
```diff
@@ -0,0 +1,122 @@
+# Design: pair-aware vPC interface identity (issue #356)
+
+Date: 2026-07-20
```
mikewiebe: Need to remove these docs

docs/superpowers/plans/2026-07-20-vpc-multi-pair-identity.md
allenrobel: Done

docs/superpowers/specs/2026-07-20-vpc-multi-pair-identity-design.md
allenrobel: Done

plugins/module_utils/orchestrators/vpc_interface_base.py
```diff
@@ -362,6 +402,30 @@ def _policy_type(iface: dict) -> str | None:
...
+
+        None
+        """
+        config_data = iface.get("configData") or {}
+        network_os = config_data.get("networkOS") or {}
+        policy = network_os.get("policy") or {}
+        peer_switch_id = policy.get("peerSwitchId")
+        if peer_switch_id:
+            return frozenset({switch_id, peer_switch_id})
+        return frozenset({switch_id})
```
mikewiebe: ### High: Missing peer metadata can make `overridden` delete the vPC it should retain

**Issue**

The new pair discriminator falls back to the queried switch alone when an ND
interface echo omits `peerSwitchId`. If only one of the two peer echoes omits
that field, the two records get different pair keys and both survive
deduplication. `state: overridden` then treats the unconfigured peer copy as a
different resource and deletes it, even though both records represent the same
pair-wide vPC interface.

**Evidence**

- [`vpc_interface_base.py` line 427](https://github.com/CiscoDevNet/ansible-nd/pull/411/files#diff-e8ad326e0cde4b80ec8422d231810a48322151232983290010bfb0ee7957b61cR427)
  returns a singleton `{switch_id}` key when `peerSwitchId` is missing. The
  preceding comment says both echoes must omit the field, but one missing value
  is enough: peer A becomes `{A}` while peer B becomes `{A, B}`.
- [`vpc_interface_base.py` lines 490-500](https://github.com/CiscoDevNet/ansible-nd/pull/411/files#diff-e8ad326e0cde4b80ec8422d231810a48322151232983290010bfb0ee7957b61cR490)
  deduplicates strictly by `(interfaceName, pair_key)`, so those two keys remain
  as separate existing resources.
- An exact-head state-machine probe produced existing identifiers
  `(192.0.2.1, vpc100)` and `(192.0.2.2, vpc100)` for the two echoes. With only
  `(192.0.2.1, vpc100)` proposed, the override deletion set contained
  `(192.0.2.2, vpc100)`.
- Practical example: suppose leaf A (`SERIAL-A`, `10.122.84.181`) and leaf B
  (`SERIAL-B`, `10.122.84.182`) share `vpc100`, while the playbook uses
  `state: overridden` and lists only `(10.122.84.181, vpc100)`. ND echoes the
  interface from leaf A without `peerSwitchId` and from leaf B with
  `peerSwitchId: SERIAL-A`. The fallback computes `{SERIAL-A}` for the first
  echo and `{SERIAL-A, SERIAL-B}` for the second. The module consequently sees
  the leaf-B copy as an unwanted second resource and sends
  `DELETE /switches/SERIAL-B/interfaces/vpc100`. Because `vpc100` is pair-wide,
  that removes the interface from both peers, including the copy the playbook
  intended to retain.
- The bundled Manage 1.1.411 OpenAPI schemas expose `peerSwitchId` for both vPC
  access and trunk-host responses but do not mark it required. The fallback
  therefore handles a schema-valid response shape, not an impossible state.

**Existing PR overlap**

No matching existing PR comment found. The earlier duplicate-peer comments
cover user config that lists both peers; they do not cover an incomplete
controller echo during deduplication.

Related open issue [#356](https://github.com/CiscoDevNet/ansible-nd/issues/356)
defines the pair-scoped identity work implemented by this PR, but it does not
track the missing-`peerSwitchId` destructive fallback.

**Impact**

A valid `overridden` run can send DELETE through the second peer for the same
shared vPC interface that the playbook intends to keep. That is incorrect
resource targeting and can remove live pair-wide interface intent.

**Suggested fix**

Resolve missing peer identity from the authoritative vPC-pair endpoint or a
validated pair cache. If the pair still cannot be identified, fail closed
before computing destructive override deletions. Do not use a singleton
per-switch key as the fallback for `overridden`. Add access and trunk-host tests
where one peer echo, and then both peer echoes, omit `peerSwitchId`; neither case
may queue a delete of the desired vPC.

plugins/module_utils/orchestrators/vpc_interface_base.py
```diff
@@ -324,25 +364,25 @@ def query_one(self, model_instance: ModelType, **kwargs) -> ResponseType:
...
         None
         """
-        mapping: dict[str, str] = {}
+        mapping: dict[str, set[str]] = {}
         for item in self.rest_send.params.get("config") or []:
             name = item.get("interface_name")
             switch_ip = item.get("switch_ip")
             if name and switch_ip:
-                mapping[name] = switch_ip
+                mapping.setdefault(name, set()).add(switch_ip)
```
mikewiebe: ### Medium: Mixed-case vPC names select the wrong peer during `overridden`

**Issue**

The access and trunk-host models normalize an interface name such as `VPC501`
to `vpc501`, but the new configured-peer preference map indexes raw Ansible
input without the same normalization. During the fabric-wide `overridden`
query, the normalized response misses that map and the code may retain the
alphabetically lower switch rather than the switch named by the user.

**Evidence**

- [`vpc_interface_base.py` line 385](https://github.com/CiscoDevNet/ansible-nd/pull/411/files#diff-e8ad326e0cde4b80ec8422d231810a48322151232983290010bfb0ee7957b61cR385)
  stores the raw `interface_name` as the preference-map key.
- [`vpc_access_interface.py` lines 379-393](https://github.com/CiscoDevNet/ansible-nd/pull/411/files#diff-fc0d76021a429b07196ccfc093d05609d28284084411ce37e2d92c36fcd0fe33R379)
  and [`vpc_trunk_host_interface.py` lines 540-554](https://github.com/CiscoDevNet/ansible-nd/pull/411/files#diff-d1d2f61f4360961f191c824d5a3da29c8459b747986de7c63626cf96a343f47dR540)
  lowercase that same value in the validated models.
- An exact-head probe with `VPC501` configured on peer `.2` built the raw map
  `{'VPC501': {'.2'}}`, selected the response representative
  `('.1', 'vpc501')`, and produced proposed identifier `('.2', 'vpc501')`.
  The state machine therefore classified the operation as a create plus an
  override deletion instead of no change.
- Practical example: leaf A (`10.122.84.181`, `SERIAL-A`) and leaf B
  (`10.122.84.182`, `SERIAL-B`) share `vpc501`. A valid `state: overridden`
  playbook identifies leaf B with `interface_name: VPC501`. The model converts
  the desired identifier to `(10.122.84.182, vpc501)`, but the raw preference
  map remains `{'VPC501': {'10.122.84.182'}}`. ND returns the canonical
  `interfaceName: vpc501`; that lowercase lookup misses the uppercase map key,
  so the stable fallback selects leaf A. The resulting plan is CREATE
  `(10.122.84.182, vpc501)` plus DELETE `(10.122.84.181, vpc501)`, even though
  both identities describe the same pair-wide interface. The create can fail as
  a duplicate; if accepted, the delete can remove the interface the playbook
  intended to retain. Check mode previews the same false operations.

**Existing PR overlap**

Related open issues [#353](https://github.com/CiscoDevNet/ansible-nd/issues/353)
and [#356](https://github.com/CiscoDevNet/ansible-nd/issues/356) cover interface
normalization and pair-scoped identity respectively, but neither tracks this
raw-map/model normalization mismatch.

**Impact**

Supported mixed-case input becomes non-idempotent in `overridden`: check mode
reports a false create/delete plan, and normal mode can fail on a duplicate
create or, if accepted, proceed to delete the existing shared resource through
the wrong representative peer. This is a regression from the base name-only
identity, which still matched the normalized name.

**Suggested fix**

Build the preference map from validated model data, or normalize both raw input
and response names through the same shared canonicalizer before lookup and
deduplication. Add access and trunk-host tests that configure an uppercase or
mixed-case name on the higher-serial peer and require `overridden` to remain
idempotent in normal and check mode.

tests/integration/targets/nd_interface_vpc_trunk_host/tasks/multi_pair.yaml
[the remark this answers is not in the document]
mikewiebe: ### Medium: Failed multi-pair integration runs skip the added cleanup

**Issue**

Both integration targets unconditionally run the new four-switch scenario
against a pre-existing second vPC pair. The added pair-2 delete is ordinary tail
work, not an `always` cleanup. Any failed module call or assertion after the
MERGED create skips both cleanup tasks, while the target's outer `always` can
only unpair pair 1 and never cleans pair 2.

**Evidence**

- [`nd_interface_vpc_access/tasks/multi_pair.yaml` lines 143-154](https://github.com/CiscoDevNet/ansible-nd/pull/411/files#diff-2bef862c2e0a72a18b202ce32b2e24735cba68e84729204298774e4978acb003R143)
  describes the pair-2 delete as abort cleanup, but places it at the end of the
  normal task sequence. The trunk-host target has the same structure in
  [`multi_pair.yaml` lines 143-154](https://github.com/CiscoDevNet/ansible-nd/pull/411/files#diff-8912143a67c76fa6a3ed2a31c81adeae162c5bdd47c58435ec1c5ee170b80618R143).
- [`nd_interface_vpc_access/tasks/main.yaml` lines 51-63](https://github.com/CiscoDevNet/ansible-nd/pull/411/files#diff-5d45216f0503901fd71ecb038dcdb1d2057d5131ba1423bba3903ed6a0bd26c5R51)
  includes the scenario unconditionally; its `always` block only deletes pair
  1 and only when `test_manage_vpc_pairs` is true. The trunk-host target mirrors
  this behavior.
- If an assertion fails after the MERGED create, Ansible jumps directly to the
  outer `always`. Pair 2's interface remains. The next run can then fail the
  assertion that `nm_mp_merged is changed` before reaching tail cleanup,
  repeatedly poisoning the target.

**Existing PR overlap**

Already raised in Copilot's trunk-host
[r3618865559](https://github.com/CiscoDevNet/ansible-nd/pull/411#discussion_r3618865559)
and access
[r3618865579](https://github.com/CiscoDevNet/ansible-nd/pull/411#discussion_r3618865579)
cleanup threads. The current commit adds best-effort tail deletes, but this
re-review confirms that the failure path requested by those comments remains
unhandled. Both threads should be reopened or followed up despite their current
resolved state.

Related open issue [#508](https://github.com/CiscoDevNet/ansible-nd/issues/508)
tracks deterministic cleanup for failed integration runs generally, but it does
not implement this PR's exact two-pair cleanup boundary.

**Impact**

A failed CI or live run can leave vPC interface intent on a shared, externally
prepared second pair. That contaminates the testbed, makes reruns unreliable,
and can alter interfaces outside the original pair-1 teardown scope.

**Suggested fix**

Wrap the multi-pair mutations and assertions in a `block` with an `always`
section that best-effort deletes the exact test interface from both pairs, with
no changed assertions in cleanup. Add a pre-clean for stale artifacts and gate
the four-switch scenario behind an explicit opt-in defaulting to false when the
second pair is externally owned.

tests/integration/targets/nd_interface_vpc_access/tasks/multi_pair.yaml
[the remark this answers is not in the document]
mikewiebe: ### Medium: Failed multi-pair integration runs skip the added cleanup

**Issue**

Both integration targets unconditionally run the new four-switch scenario
against a pre-existing second vPC pair. The added pair-2 delete is ordinary tail
work, not an `always` cleanup. Any failed module call or assertion after the
MERGED create skips both cleanup tasks, while the target's outer `always` can
only unpair pair 1 and never cleans pair 2.

**Evidence**

- [`nd_interface_vpc_access/tasks/multi_pair.yaml` lines 143-154](https://github.com/CiscoDevNet/ansible-nd/pull/411/files#diff-2bef862c2e0a72a18b202ce32b2e24735cba68e84729204298774e4978acb003R143)
  describes the pair-2 delete as abort cleanup, but places it at the end of the
  normal task sequence. The trunk-host target has the same structure in
  [`multi_pair.yaml` lines 143-154](https://github.com/CiscoDevNet/ansible-nd/pull/411/files#diff-8912143a67c76fa6a3ed2a31c81adeae162c5bdd47c58435ec1c5ee170b80618R143).
- [`nd_interface_vpc_access/tasks/main.yaml` lines 51-63](https://github.com/CiscoDevNet/ansible-nd/pull/411/files#diff-5d45216f0503901fd71ecb038dcdb1d2057d5131ba1423bba3903ed6a0bd26c5R51)
  includes the scenario unconditionally; its `always` block only deletes pair
  1 and only when `test_manage_vpc_pairs` is true. The trunk-host target mirrors
  this behavior.
- If an assertion fails after the MERGED create, Ansible jumps directly to the
  outer `always`. Pair 2's interface remains. The next run can then fail the
  assertion that `nm_mp_merged is changed` before reaching tail cleanup,
  repeatedly poisoning the target.

**Existing PR overlap**

Already raised in Copilot's trunk-host
[r3618865559](https://github.com/CiscoDevNet/ansible-nd/pull/411#discussion_r3618865559)
and access
[r3618865579](https://github.com/CiscoDevNet/ansible-nd/pull/411#discussion_r3618865579)
cleanup threads. The current commit adds best-effort tail deletes, but this
re-review confirms that the failure path requested by those comments remains
unhandled. Both threads should be reopened or followed up despite their current
resolved state.

Related open issue [#508](https://github.com/CiscoDevNet/ansible-nd/issues/508)
tracks deterministic cleanup for failed integration runs generally, but it does
not implement this PR's exact two-pair cleanup boundary.

**Impact**

A failed CI or live run can leave vPC interface intent on a shared, externally
prepared second pair. That contaminates the testbed, makes reruns unreliable,
and can alter interfaces outside the original pair-1 teardown scope.

**Suggested fix**

Wrap the multi-pair mutations and assertions in a `block` with an `always`
section that best-effort deletes the exact test interface from both pairs, with
no changed assertions in cleanup. Add a pre-clean for stale artifacts and gate
the four-switch scenario behind an explicit opt-in defaulting to false when the
second pair is externally owned.

plugins/module_utils/orchestrators/vpc_interface_base.py
allenrobel: Good catch — one missing echo was indeed enough (`{A}` vs `{A, B}`), and the old docstring's "both copies" reasoning was wrong.

Fixed in *Resolve missing peerSwitchId via vpcPair and canonicalize names in vPC overridden dedup*: `_pair_key` now resolves a missing peer through the existing `_resolve_peer_switch_id` (authoritative `vpcPair` GET, cached per switch, so only the degraded path pays one extra request) and `query_all` fails closed if the pair can't be established — no singleton fallback remains.

Tests added:
- base orchestrator: one-missing (`01020`), both-missing (`01030`), and unpaired-404 fail-closed (`01040`), each pinning the exact request sequence via `Results.path`
- `NDStateMachine`-level access and trunk-host tests (`00450`/`00455`) asserting no delete is queued in either missing case and `changed` stays `False`

plugins/module_utils/orchestrators/vpc_interface_base.py
allenrobel: Confirmed — the preference map was keyed on raw input while the models lowercase.

Fixed in *Resolve missing peerSwitchId via vpcPair and canonicalize names in vPC overridden dedup*: a shared `_canonical_interface_name()` (mirrors the models' `normalize_interface_name`) is applied to both the config keys and the response names before dedup/lookup.

Tests added:
- base orchestrator `00970`: `VPC501` configured on the higher-serial peer keeps that peer as the representative
- `NDStateMachine`-level access and trunk-host tests (`00440`/`00445`) with `VPC100`/`VPC500` on the higher-serial peer, asserting `overridden` plans no create/delete and reports `changed: False` in both normal and check mode

mikewiebe (approved):
Approved.

akinross (approved):

PR #422, opened 2026-07-23
Detect field removals in replaced/overridden states (one-way subset diff)

### LINKED ISSUE #410: replaced/overridden states cannot detect field removals (one-way subset diff)
# `replaced`/`overridden` states cannot detect field removals (one-way subset diff)

## Summary

For `replaced` and `overridden` states, the Gen-3 state machine classifies a proposed item that merely
*removed* a field (relative to existing device config) as `no_diff`. No update request is sent and the
module reports `changed: false`, even though standard Ansible `replaced` semantics say unspecified
attributes must be reset to their defaults.

Discovered while working the storm-control mutex thread on PR #360:
<https://github.com/CiscoDevNet/ansible-nd/pull/360#discussion_r3611622647>. PR #360 fixes the specific
storm-control case for **merged** state (via the `merge_would_change` hook) and adds a scoped reverse
pass to `NDConfigCollection.get_diff_collection`; this issue covers the general case for the non-merged
states, which was explicitly declared out of scope there.

## Root cause

`NDStateMachine._manage_create_update_state` calls `get_diff_config(proposed_item, exclude_unset=False)`
for `replaced`/`overridden`. The chain:

1. `NDBaseModel.get_diff` dumps both sides via `to_diff_dict()`, which always passes `exclude_none=True`.
   Any optional field the user omits (defaulting to `None`) vanishes from the proposed dump — regardless
   of `exclude_unset`.
2. `utils.issubset(proposed_data, existing_data)` iterates only the *proposed* dict's keys. Fields present
   on existing but absent from proposed are never examined.

So when the only difference is a removal, the one-way subset test passes and the item is classified
`no_diff`. The `merge_would_change` cross-check added in PR #360 runs only when `exclude_unset=True`
(merged path) and deliberately does not cover this.

## Reproduction

Against a switch whose Ethernet interface currently has `storm_control_pct: 80.0` configured:

```yaml
- name: Replace interface config, omitting storm control
  cisco.nd.nd_interface_ethernet:
    state: replaced
    config:
      - fabric_name: SITE1
        switches:
          - switch_ip: 192.168.7.11
            interfaces:
              - name: Ethernet1/10
                admin_state: up
                # storm_control_pct intentionally omitted -- replaced should reset it
```

Expected: a PUT is issued, `storm_control_pct` is cleared on the device, module reports `changed: true`.

Actual: classified `no_diff`, no request sent, `changed: false`, device keeps `storm_control_pct: 80.0`.

Any optional field behaves the same way (descriptions, freeform config, speed overrides, ...); storm
control is just the lab-verified example.

## Notes for the fix

- **The wire side already behaves correctly once a diff triggers.** The replace path sends the full
  proposed payload, and PR #360's integration work verified on ND 4.2.1 that PUT clears omitted fields.
  The defect is purely diff classification: the PUT that would fix the device is never issued.
- **Naive symmetric equality is not the fix.** The existing model carries server-populated fields
  (switch IDs, operational echoes, ND-injected defaults) that proposed config can never contain; a plain
  two-way comparison would report `changed` on every run and break idempotency. A reverse pass must be
  scoped to user-settable/payload fields — the same reasoning behind the scoped reverse pass PR #360
  added to `NDConfigCollection.get_diff_collection`.
- **Blast radius**: `NDBaseModel.get_diff` is shared by all Gen-3 modules (`nd_interface_*`,
  `nd_local_user`, `nd_manage_fabric_*`). Both `replaced` and the update-classification phase of
  `overridden` are affected. Existing integration targets likely under-test this (removal-only replaced
  tasks currently pass as no-ops), so fixing it will need test additions alongside the diff change.

## Affected code

- `plugins/module_utils/models/base.py` — `NDBaseModel.get_diff`, `NDBaseModel.to_diff_dict`
- `plugins/module_utils/utils.py` — `issubset`
- `plugins/module_utils/nd_state_machine.py` — `_manage_create_update_state` (`exclude_unset = self.state == "merged"`)
- `plugins/module_utils/nd_config_collection.py` — `get_diff_config`

## Environment

- Collection: `cisco.nd` 2.0.0 (develop)
- ND version: 4.2.1 (wire behavior lab-verified during PR #360)

allenrobel:
## Related Issue(s)

Fixes #410

## Merge order

- This PR
- #366
- #403
- #291
- Future PR to fix issue #447

## Proposed Changes

- `NDBaseModel.get_diff` gains a **reverse pass** on the `exclude_unset=False` (replaced/overridden) path: after the forward subset check, the new pure util `utils.has_removals()` walks the existing side's payload-scoped dump (`exclude_from_diff | payload_exclude_fields`) and classifies any non-empty field absent from proposed as a difference, so the full-payload PUT that resets omitted fields is actually issued. Merged-state semantics are unchanged.
- Empty existing-side values (`""`, `[]`, `{}`) normalize to absent, generalizing the `PrefixListModel` description precedent; its ad-hoc `get_diff` override is removed (its tests pass against the inherited behavior).
- **Default-echo normalization**: ND echoes the OpenAPI template default for every field the user never set (lab-verified: ~22 concrete fields on trunkHost; ND injects `routeMapTag: 12345` even on user-created loopbacks). New per-model `reverse_diff_defaults` ClassVars (alias → template default, sourced from the ND 4.2.1 OpenAPI template schemas, values in the model's dumped form) let `to_reverse_diff_dict()` strip default-valued echoes recursively, keeping replaced/overridden runs idempotent. Tables added for all nine interface policy models (loopback, ethernet access/trunkHost, port-channel access/trunkHost, vPC access/trunkHost, SVI, managed subinterface).
- Known residual: `TrunkVpcHostPolicyModel` — the schema defaults for `peer1AllowedVlans`/`peer2AllowedVlans` have no model fields (per-peer→collapsed `access_vlan` workaround); vPC families were not lab-verified in this PR.

**Post-review hardening** (three follow-up commits addressing all five findings from Claude's pre-merge code review):

- Commit *"Fix reverse-pass idempotency for server-populated existing-side data"* — three confirmed bugs, one mechanism: the reverse pass fired on existing-side data the proposed config can never express, permanently classifying items as changed.
  - New alias-keyed `reverse_diff_exclude` ClassVar, applied at each nested model's **own** level during the reverse scrub; both vPC policy models exclude the orchestrator-injected `peerSwitchId` (not in the argspec, injected only at payload-build time, echoed by ND on reads).
  - The reverse scrub drops `extra="allow"` server keys (`model_extra`) at every nesting level, so undeclared ND GET keys on the fabric models (top-level or nested management) never count as removals.
  - `LocalUserModel` gains a `reverse_diff_defaults` table for ND's falsy echoes (`xLaunch=false`, `reuseLimitation=0`, `timeIntervalLimitation=0` — the module's own integration tests assert these `before` values); `False`/`0` are real values, not empty markers, so empty-normalization alone could not cover them.
- Commit *"Add TODO(4.2.1) workaround markers to all reverse_diff_defaults tables"* — all ten table sites (nine interface policy models + `LocalUserModel`) carry `TODO(4.2.1) get-echoes-schema-defaults-for-unset-fields`, backed by a new bug-tracker vault note of the same id (resolvable via `get_bug_by_id`) documenting the defaults-echo behavior across both the interfaces and localUsers endpoint families, so `/nd-workaround-audit` can surface the tables for re-verification against future ND releases.
- Commit *"Derive reverse-pass dicts from the forward dumps (halve get_diff cost)"* — `to_reverse_diff_dict` now derives from `to_diff_dict` plus in-place scoping, and `get_diff` reuses its already-computed forward dumps: 2 dumps per no-diff comparison instead of 4, pinned by a dump-count regression test. Side benefit: subclass `to_diff_dict` overrides (e.g. the AI-eBGP `nxapiHttp` pop) now scope the reverse pass symmetrically. `NDOutput`'s two-directional changed loop is deliberately unchanged — payload-excluded-but-diff-compared fields (loopback `switch_ip`, SVI/subinterface `oper_data`, prefix-list `ip_version`) rely on the second direction, so removing it would be a semantics change, not an optimization (documented in the commit body).

## Test Notes

- New unit tests: `tests/unit/module_utils/test_utils.py` (first coverage for `issubset` + `has_removals`), `tests/unit/module_utils/models/test_base_model_reverse_diff.py` (removal detection, empty/default normalization, per-family default tables incl. the verbatim lab-captured trunkHost echo), state-machine replaced/overridden/merged classification tests, and the previously missing `exclude_unset=False` storm-control case.
- `test_loopback_interface_00620` updated: proposed-with-fewer-fields is now a difference by default (the old assertion documented the #410 bug).
- Post-review commits add a 005xx test section (per-bug phantom-removal reproductions built TDD-first, each paired with a genuine-removal guard proving the fix does not over-strip) and a 006xx efficiency section (dump-count contract: exactly one `model_dump` per side on the no-diff replaced path, plus standalone `to_reverse_diff_dict` behavior parity).
- Full unit suite passes: 3865 tests via `ndpytest tests/unit/` (nd-dev container machine).
- **Review re-verification:** the pre-merge code review's five findings (three confirmed idempotency bugs, the missing workaround markers, the redundant reverse dumps) were re-verified fixed at branch HEAD by re-running the original reproduction scripts — phantom changes now classify `no_diff`, genuine removals still classify `changed` — with no new findings.
- black/isort/pylint/mypy clean on changed files; `ndtest` sanity passes except the pre-existing `action-plugin-docs` findings on `plugins/action/tests/integration/*` (untouched by this PR).
- Lab verification (SITE1): the issue's repro passes — `state: replaced` omitting `storm_control_broadcast_level: 80.0` reports `changed: true`, issues the PUT, and clears the value on ND; replaced double-runs on `nd_interface_ethernet_trunk_host` and `nd_interface_loopback` report `changed: false` on the second run.

4.2.1

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

https://claude.ai/code/session_01WqAuV2pWYJTno2bfdcNZCm

### DIFF
--- a/plugins/module_utils/models/base.py
+++ b/plugins/module_utils/models/base.py
@@ -332,7 +332,16 @@ class NDBaseModel(BaseModel, ABC):
                 difference. Empty existing values (``""``, ``[]``, ``{}``) are
                 normalized to absent so ND-echoed empty markers keep runs
                 idempotent.
+
+        Raises:
+            TypeError: If ``other`` is not an instance of this model's type
+                (same contract as ``merge``). The reverse pass applies
+                ``other``'s own ``reverse_diff_*`` declarations, so a
+                cross-type comparison is a programming error, not a diff.
         """
+        if not isinstance(other, type(self)):
+            raise TypeError(f"Cannot diff {type(other).__name__} against {type(self).__name__}. Both must be the same type.")
+
         self_data = self.to_diff_dict()
         other_data = other.to_diff_dict(exclude_unset=exclude_unset)
         is_subset = issubset(other_data, self_data)
--- a/plugins/module_utils/models/base.py
+++ b/plugins/module_utils/models/base.py
@@ -248,19 +248,34 @@ class NDBaseModel(BaseModel, ABC):
         config and must not count as removals). Like `to_diff_dict`, the top-level exclusion sets apply at the top
         level only; nested exclusions are declared on the nested model via `reverse_diff_exclude`.
 
+        Derived from `to_diff_dict` (one dump, then in-place scoping) rather than a second `model_dump`, so subclass
+        `to_diff_dict` overrides scope the reverse pass too, and `get_diff` can reuse its forward dumps instead of
+        re-dumping both models.
+
         ## Raises
 
         None
         """
-        data = self.model_dump(
-            by_alias=True,
-            exclude_none=True,
-            exclude=(self.exclude_from_diff | self.payload_exclude_fields) or None,
-            mode="json",
-            **kwargs,
-        )
-        self._scrub_reverse_diff_dict(data)
+        data = self.to_diff_dict(**kwargs)
+        self._apply_reverse_diff_scope(data)
         return data
+
+    def _apply_reverse_diff_scope(self, data: Dict[str, Any]) -> None:
+        """
+        # Summary
+
+        Convert a `to_diff_dict` export into the reverse-pass shape, in place: pop the aliases of top-level
+        `payload_exclude_fields` (already absent when a field is also in `exclude_from_diff`), then run
+        `_scrub_reverse_diff_dict` so each nested model applies its own exclusions/extras/defaults declarations.
+
+        ## Raises
+
+        None
+        """
+        for field_name in self.payload_exclude_fields:
+            field_info = type(self).model_fields.get(field_name)
+            data.pop(field_info.alias if field_info is not None and field_info.alias else field_name, None)
+        self._scrub_reverse_diff_dict(data)
 
     def _scrub_reverse_diff_dict(self, data: Dict[str, Any]) -> None:
         """
@@ -320,8 +335,14 @@ class NDBaseModel(BaseModel, ABC):
         is_subset = issubset(other_data, self_data)
         if is_subset and exclude_unset and self.merge_would_change(other):
             return False
-        if is_subset and not exclude_unset and has_removals(self.to_reverse_diff_dict(), other.to_reverse_diff_dict()):
-            return False
+        if is_subset and not exclude_unset:
+            # Reuse the forward dumps for the reverse pass (they are not read again): scoping them in place
+            # avoids a second full model_dump per side (PR #422 review, efficiency finding).
+            self._apply_reverse_diff_scope(self_data)
+            # Same-class access; pylint cannot infer `other` shares this NDBaseModel API.
+            other._apply_reverse_diff_scope(other_data)  # pylint: disable=protected-access
+            if has_removals(self_data, other_data):
+                return False
         return is_subset
 
     def merge_would_change(self, other: "NDBaseModel") -> bool:
--- a/plugins/module_utils/models/base.py
+++ b/plugins/module_utils/models/base.py
@@ -77,6 +77,9 @@ class NDBaseModel(BaseModel, ABC):
     # value equal to its declared default is normalized to absent during removal detection -- omitting it from
     # proposed config is not a pending reset. Source the values from the ND OpenAPI template schema for the
     # model's policyType (see the `nd-openapi` MCP); a wrong value here breaks replaced/overridden idempotency.
+    # Values MUST be in the model's DUMPED form, not the schema-declared form: when a validator coerces a field
+    # on read (e.g. loopback `routeMapTag` schema integer 12345 stored as string "12345"), the table must hold
+    # the coerced value or the default never matches and the field silently reopens issue #410 for that model.
     reverse_diff_defaults: ClassVar[Dict[str, Any]] = {}
 
     # Keys (field ALIASES / wire keys) stripped from this model's level of the reverse-pass dump regardless of
--- a/plugins/module_utils/models/base.py
+++ b/plugins/module_utils/models/base.py
@@ -8,7 +8,7 @@ from abc import ABC
 from typing import Any, ClassVar, Dict, List, Literal, Optional, Set, Tuple, Union
 
 from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import BaseModel, ConfigDict
-from ansible_collections.cisco.nd.plugins.module_utils.utils import issubset
+from ansible_collections.cisco.nd.plugins.module_utils.utils import has_removals, issubset
 
 
 def _strip_none_values(data):
@@ -71,6 +71,13 @@ class NDBaseModel(BaseModel, ABC):
     # Fields to explicitly exclude per mode
     payload_exclude_fields: ClassVar[Set[str]] = set()
     config_exclude_fields: ClassVar[Set[str]] = set()
+
+    # ND template defaults for the reverse pass of `get_diff` (issue #410), keyed by field ALIAS (wire key).
+    # ND echoes the schema-declared template default for every field the user never set, so an existing-side
+    # value equal to its declared default is normalized to absent during removal detection -- omitting it from
+    # proposed config is not a pending reset. Source the values from the ND OpenAPI template schema for the
+    # model's policyType (see the `nd-openapi` MCP); a wrong value here breaks replaced/overridden idempotency.
+    reverse_diff_defaults: ClassVar[Dict[str, Any]] = {}
 
     # --- Subclass Validation ---
 
@@ -221,6 +228,53 @@ class NDBaseModel(BaseModel, ABC):
             **kwargs,
         )
 
+    def to_reverse_diff_dict(self, **kwargs) -> Dict[str, Any]:
+        """
+        # Summary
+
+        Export for the reverse pass of `get_diff` (issue #410), scoped to payload shape: fields in `exclude_from_diff` or
+        `payload_exclude_fields` are excluded, so only fields the PUT body can express participate in removal detection.
+        Existing-side values equal to their `reverse_diff_defaults` entry are stripped (recursively, each nested model
+        applying its own table) because ND echoes template defaults for unset fields. Like `to_diff_dict`, the exclusion
+        sets apply at the top level only; models needing nested exclusions override `get_diff`.
+
+        ## Raises
+
+        None
+        """
+        data = self.model_dump(
+            by_alias=True,
+            exclude_none=True,
+            exclude=(self.exclude_from_diff | self.payload_exclude_fields) or None,
+            mode="json",
+            **kwargs,
+        )
+        self._strip_reverse_diff_defaults(data)
+        return data
+
+    def _strip_reverse_diff_defaults(self, data: Dict[str, Any]) -> None:
+        """
+        # Summary
+
+        Remove keys from `data` (an aliased dump of `self`) whose value equals the model's declared `reverse_diff_defaults`
+        entry, then recurse into nested `NDBaseModel` fields so each nested model applies its own table.
+
+        ## Raises
+
+        None
+        """
+        for alias, default in self.reverse_diff_defaults.items():
+            if alias in data and data[alias] == default:
+                del data[alias]
+        for field_name, field_info in type(self).model_fields.items():
+            value = getattr(self, field_name, None)
+            if isinstance(value, NDBaseModel):
+                alias = field_info.alias or field_name
+                nested = data.get(alias)
+                if isinstance(nested, dict):
+                    # Same-class recursion; pylint cannot infer `value` is an NDBaseModel from getattr.
+                    value._strip_reverse_diff_defaults(nested)  # pylint: disable=protected-access
+
     def get_diff(self, other: "NDBaseModel", exclude_unset: bool = False) -> bool:
         """Diff comparison.
 
@@ -234,11 +288,22 @@ class NDBaseModel(BaseModel, ABC):
                 to catch merge side effects the one-way subset test cannot see
                 (e.g. mutually exclusive counterpart fields that the merge
                 would clear).
+
+                When False (the ``replaced``/``overridden`` path), a subset
+                match is additionally cross-checked with ``has_removals`` over
+                the payload-scoped dumps (issue #410): a field present on
+                ``self`` (device) but absent from ``other`` (proposed) means
+                the full-payload PUT would reset it, so it must classify as a
+                difference. Empty existing values (``""``, ``[]``, ``{}``) are
+                normalized to absent so ND-echoed empty markers keep runs
+                idempotent.
         """
         self_data = self.to_diff_dict()
         other_data = other.to_diff_dict(exclude_unset=exclude_unset)
         is_subset = issubset(other_data, self_data)
         if is_subset and exclude_unset and self.merge_would_change(other):
+            return False
+        if is_subset and not exclude_unset and has_removals(self.to_reverse_diff_dict(), other.to_reverse_diff_dict()):
             return False
         return is_subset
 
--- a/plugins/module_utils/models/base.py
+++ b/plugins/module_utils/models/base.py
@@ -79,6 +79,13 @@ class NDBaseModel(BaseModel, ABC):
     # model's policyType (see the `nd-openapi` MCP); a wrong value here breaks replaced/overridden idempotency.
     reverse_diff_defaults: ClassVar[Dict[str, Any]] = {}
 
+    # Keys (field ALIASES / wire keys) stripped from this model's level of the reverse-pass dump regardless of
+    # value. For server-populated fields the proposed config can never express (e.g. the orchestrator-injected
+    # `peerSwitchId` on vPC policy models): their presence on the existing side is not a pending reset, and
+    # unlike `reverse_diff_defaults` they have no single constant value to match against. Applies at the
+    # declaring model's own nesting level, so nested models scope their own exclusions.
+    reverse_diff_exclude: ClassVar[Set[str]] = set()
+
     # --- Subclass Validation ---
 
     def __init_subclass__(cls, **kwargs):
@@ -234,9 +241,12 @@ class NDBaseModel(BaseModel, ABC):
 
         Export for the reverse pass of `get_diff` (issue #410), scoped to payload shape: fields in `exclude_from_diff` or
         `payload_exclude_fields` are excluded, so only fields the PUT body can express participate in removal detection.
-        Existing-side values equal to their `reverse_diff_defaults` entry are stripped (recursively, each nested model
-        applying its own table) because ND echoes template defaults for unset fields. Like `to_diff_dict`, the exclusion
-        sets apply at the top level only; models needing nested exclusions override `get_diff`.
+        The dump is then scrubbed recursively, each nested model applying its own declarations: values equal to their
+        `reverse_diff_defaults` entry are stripped (ND echoes template defaults for unset fields), keys in
+        `reverse_diff_exclude` are stripped unconditionally (server-populated fields the proposed config can never
+        express), and keys retained by `extra="allow"` are stripped (undeclared server keys are not expressible in
+        config and must not count as removals). Like `to_diff_dict`, the top-level exclusion sets apply at the top
+        level only; nested exclusions are declared on the nested model via `reverse_diff_exclude`.
 
         ## Raises
 
@@ -249,20 +259,27 @@ class NDBaseModel(BaseModel, ABC):
             mode="json",
             **kwargs,
         )
-        self._strip_reverse_diff_defaults(data)
+        self._scrub_reverse_diff_dict(data)
         return data
 
-    def _strip_reverse_diff_defaults(self, data: Dict[str, Any]) -> None:
+    def _scrub_reverse_diff_dict(self, data: Dict[str, Any]) -> None:
         """
         # Summary
 
-        Remove keys from `data` (an aliased dump of `self`) whose value equals the model's declared `reverse_diff_defaults`
-        entry, then recurse into nested `NDBaseModel` fields so each nested model applies its own table.
+        Scrub `data` (an aliased dump of `self`) for removal detection: drop keys listed in `reverse_diff_exclude`, keys
+        retained only via `extra="allow"` (undeclared server keys), and keys whose value equals the model's declared
+        `reverse_diff_defaults` entry, then recurse into nested `NDBaseModel` fields so each nested model applies its own
+        declarations.
 
         ## Raises
 
         None
         """
+        for alias in self.reverse_diff_exclude:
+            data.pop(alias, None)
+        # `model_extra` keys are stored under their wire spelling, matching the aliased dump.
+        for extra_key in getattr(self, "model_extra", None) or {}:
+            data.pop(extra_key, None)
         for alias, default in self.reverse_diff_defaults.items():
             if alias in data and data[alias] == default:
                 del data[alias]
@@ -273,7 +290,7 @@ class NDBaseModel(BaseModel, ABC):
                 nested = data.get(alias)
                 if isinstance(nested, dict):
                     # Same-class recursion; pylint cannot infer `value` is an NDBaseModel from getattr.
-                    value._strip_reverse_diff_defaults(nested)  # pylint: disable=protected-access
+                    value._scrub_reverse_diff_dict(nested)  # pylint: disable=protected-access
 
     def get_diff(self, other: "NDBaseModel", exclude_unset: bool = False) -> bool:
         """Diff comparison.
--- a/plugins/module_utils/models/interfaces/ethernet_access_interface.py
+++ b/plugins/module_utils/models/interfaces/ethernet_access_interface.py
@@ -26,7 +26,7 @@ wrapping or flattening.
 from __future__ import annotations
 
 import re
-from typing import ClassVar, Literal
+from typing import Any, ClassVar, Literal
 
 from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import (
     Field,
@@ -70,6 +70,32 @@ class EthernetAccessPolicyModel(StormControlMutexMixin):
 
     None
     """
+
+    # ND 4.2.1 `int_access_host` template defaults (schema-sourced via nd-openapi `intAccessHostTemplate`). ND echoes these
+    # for every field the user never set; the reverse pass of `get_diff` normalizes existing-side matches to absent
+    # so replaced/overridden removal detection (issue #410) stays idempotent against default echoes.
+    reverse_diff_defaults: ClassVar[dict[str, Any]] = {
+        "adminState": True,
+        "bpduFilter": "default",
+        "bpduGuard": "enable",
+        "cdp": True,
+        "debounceTimer": 100,
+        "duplexMode": "auto",
+        "errorDetectionAcl": True,
+        "fec": "auto",
+        "linkType": "auto",
+        "monitor": False,
+        "mtu": "jumbo",
+        "negotiateAuto": True,
+        "netflow": False,
+        "orphanPort": False,
+        "pfc": False,
+        "portTypeEdgeTrunk": True,
+        "qos": False,
+        "speed": "auto",
+        "stormControl": False,
+        "stormControlAction": "default",
+    }
 
     admin_state: bool | None = Field(default=None, alias="adminState", description="Enable or disable the interface")
     access_vlan: int | None = Field(default=None, alias="accessVlan", ge=1, le=4094, description="VLAN for this access port")
--- a/plugins/module_utils/models/interfaces/ethernet_access_interface.py
+++ b/plugins/module_utils/models/interfaces/ethernet_access_interface.py
@@ -71,6 +71,7 @@ class EthernetAccessPolicyModel(StormControlMutexMixin):
     None
     """
 
+    # TODO(4.2.1) get-echoes-schema-defaults-for-unset-fields
     # ND 4.2.1 `int_access_host` template defaults (schema-sourced via nd-openapi `intAccessHostTemplate`). ND echoes these
     # for every field the user never set; the reverse pass of `get_diff` normalizes existing-side matches to absent
     # so replaced/overridden removal detection (issue #410) stays idempotent against default echoes.
--- a/plugins/module_utils/models/interfaces/loopback_interface.py
+++ b/plugins/module_utils/models/interfaces/loopback_interface.py
@@ -48,6 +48,7 @@ class LoopbackPolicyModel(NDNestedModel):
     None
     """
 
+    # TODO(4.2.1) get-echoes-schema-defaults-for-unset-fields
     # ND 4.2.1 `int_loopback` template defaults (schema-sourced via nd-openapi `intLoopbackTemplate`). ND echoes these
     # for every field the user never set; the reverse pass of `get_diff` normalizes existing-side matches to absent
     # so replaced/overridden removal detection (issue #410) stays idempotent against default echoes.
--- a/plugins/module_utils/models/interfaces/loopback_interface.py
+++ b/plugins/module_utils/models/interfaces/loopback_interface.py
@@ -26,7 +26,7 @@ work via standard Pydantic serialization with no custom wrapping or flattening.
 
 from __future__ import annotations
 
-from typing import ClassVar, Literal
+from typing import Any, ClassVar, Literal
 
 from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import (
     Field,
@@ -47,6 +47,16 @@ class LoopbackPolicyModel(NDNestedModel):
 
     None
     """
+
+    # ND 4.2.1 `int_loopback` template defaults (schema-sourced via nd-openapi `intLoopbackTemplate`). ND echoes these
+    # for every field the user never set; the reverse pass of `get_diff` normalizes existing-side matches to absent
+    # so replaced/overridden removal detection (issue #410) stays idempotent against default echoes.
+    # Values must be in the model's DUMPED form: the schema declares routeMapTag as integer 12345, but
+    # `coerce_route_map_tag` stores it as a string (ND 4.2.1 GET-side type drift), so the table holds "12345".
+    reverse_diff_defaults: ClassVar[dict[str, Any]] = {
+        "adminState": True,
+        "routeMapTag": "12345",
+    }
 
     admin_state: bool | None = Field(default=None, alias="adminState", description="Enable or disable the interface")
     ip: IPv4Host = Field(default=None, alias="ip", description="Loopback IPv4 address (bare host form, e.g. 10.1.1.1; CIDR input is accepted and normalized)")
--- a/plugins/module_utils/models/interfaces/port_channel_trunk_host_interface.py
+++ b/plugins/module_utils/models/interfaces/port_channel_trunk_host_interface.py
@@ -32,7 +32,7 @@ interfaces inherit trunk-mode settings from the port-channel; users do not pre-c
 from __future__ import annotations
 
 import re
-from typing import Annotated, ClassVar, Literal, Optional  # Optional needed for Annotated runtime expr (see types.py)
+from typing import Annotated, Any, ClassVar, Literal, Optional  # Optional needed for Annotated runtime expr (see types.py)
 
 from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import (
     BeforeValidator,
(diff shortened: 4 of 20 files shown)


plugins/module_utils/models/interfaces/port_channel_trunk_host_interface.py
```diff
@@ -230,6 +230,37 @@ class PortChannelTrunkHostPolicyModel(StormControlMutexMixin):
     - If the wrapped model serializer receives a non-`dict` from the handler (see `_strip_policy_type_in_config`).
     """
 
+    # TODO(4.2.1) get-echoes-schema-defaults-for-unset-fields
+    # ND 4.2.1 `int_port_channel_trunk_host` template defaults (schema-sourced via nd-openapi `intPortChannelTrunkHostTemplate`). ND echoes these
+    # for every field the user never set; the reverse pass of `get_diff` normalizes existing-side matches to absent
+    # so replaced/overridden removal detection (issue #410) stays idempotent against default echoes.
+    reverse_diff_defaults: ClassVar[dict[str, Any]] = {
```
mikewiebe, line 237: ### High: Missing `ptp: false` default breaks port-channel idempotency

**Issue**

The new reverse-diff normalization omits Nexus Dashboard's normal `ptp: false`
echo from the port-channel trunk-host default table, so unchanged
`replaced`/`overridden` configurations are classified as changed.

**Evidence**

- [`port_channel_trunk_host_interface.py` line 237](https://github.com/CiscoDevNet/ansible-nd/pull/422/files#diff-d2ab524cc8cd4727b8449727f59773a79668d8ac08e5617de6638ef13c4c74a4R237)
  adds `reverse_diff_defaults` for the controller's normal schema defaults but
  omits `ptp`.
- [`port_channel_trunk_host_interface.py` line 302](https://github.com/CiscoDevNet/ansible-nd/pull/422/files#diff-d2ab524cc8cd4727b8449727f59773a79668d8ac08e5617de6638ef13c4c74a4R302)
  declares `ptp` as a user-configurable policy field, and normal Nexus
  Dashboard responses populate it as `false` when the user did not set it.
- An exact-head reproduction using the existing controller-response fixture
  retained `{"ptp": false}` in the reverse dictionary and reported a
  difference against an otherwise unchanged desired configuration.

**Practical example**

Assume this port-channel is already configured correctly and PTP was never
enabled:

```yaml
- name: Maintain the server trunk
  cisco.nd.nd_interface_port_channel_trunk_host:
    fabric_name: FABRIC1
    state: replaced
    config:
      - switch_ip: 192.0.2.11
        interface_name: port-channel10
        config_data:
          network_os:
            policy:
              admin_state: true
              allowed_vlans: "100-110"
              ports:
                - Ethernet1/1
                - Ethernet1/2
              # ptp is intentionally omitted
```

Nexus Dashboard returns the existing policy with schema defaults populated:

```yaml
adminState: true
allowedVlans: "100-110"
ports:
  - Ethernet1/1
  - Ethernet1/2
ptp: false
```

The desired and actual PTP state are effectively identical: PTP is disabled.
However, the new reverse comparison retains `ptp: false` because that normal
controller echo is missing from `reverse_diff_defaults`.

| Run | Actual configuration | Module result | Controller actions |
| ---: | --- | --- | --- |
| 1 | Already correct; PTP disabled | `changed: true` | PUT and deploy |
| 2 | Still correct; PTP disabled | `changed: true` | Same PUT and deploy |
| 3 | Still correct; PTP disabled | `changed: true` | Same PUT and deploy |

The controller continues returning `ptp: false`, so every run detects the same
false difference.

**Existing PR overlap**

No matching existing PR comment found. PR #422 currently has no inline or
top-level conversation comments identifying this port-channel default.

[#410](https://github.com/CiscoDevNet/ansible-nd/issues/410) tracks the broader
one-way removal-detection defect that PR #422 implements; it does not track
this missing-default regression in the proposed fix.

**Impact**

An idempotent `replaced` or `overridden` task can report `changed=True`, issue a
redundant PUT, and deploy the same port-channel on every playbook run. Across
many port-channels, this produces unnecessary controller and switch churn and
makes change reporting unreliable.

**Suggested fix**

Add `"ptp": False` to
`PortChannelTrunkHostPolicyModel.reverse_diff_defaults`. Add a regression test
built from a normal controller response containing `ptp: false`, with desired
configuration omitting PTP, and assert that both `replaced` and `overridden`
remain unchanged and schedule no PUT or deployment.

---

allenrobel: Confirmed — the reverse pass retained the ND-injected `ptp` and misclassified unchanged port-channels as changed. Fixed in "Strip ND-injected ptp from the port-channel reverse pass", but via `reverse_diff_exclude` rather than the defaults table, for two reasons: (1) `intPortChannelTrunkHostTemplate` declares no `ptp` property, and the `reverse_diff_defaults` tables are schema-sourced (slug `get-echoes-schema-defaults-for-unset-fields`) — this injection is the sibling deviation `interface-get-undocumented-ptp-field`; (2) the injected value isn't constant: after a fabric-PTP deploy, ND rewrites all physical/port-channel records to `ptp: true` fabric-wide (lab-verified, see the vault note), so normalizing only `false` would re-break idempotency there. The unconditional strip follows the `peerSwitchId` precedent on the vPC models. Regression tests cover the `ptp: false` echo, the post-deploy `ptp: true` rewrite, and that a user-set `ptp` still forward-diffs.

plugins/module_utils/models/base.py
```diff
@@ -72,6 +72,20 @@ class NDBaseModel(BaseModel, ABC):
     payload_exclude_fields: ClassVar[Set[str]] = set()
     config_exclude_fields: ClassVar[Set[str]] = set()
 
+    # ND template defaults for the reverse pass of `get_diff` (issue #410), keyed by field ALIAS (wire key).
+    # ND echoes the schema-declared template default for every field the user never set, so an existing-side
+    # value equal to its declared default is normalized to absent during removal detection -- omitting it from
+    # proposed config is not a pending reset. Source the values from the ND OpenAPI template schema for the
+    # model's policyType (see the `nd-openapi` MCP); a wrong value here breaks replaced/overridden idempotency.
+    reverse_diff_defaults: ClassVar[Dict[str, Any]] = {}
```
mtarking, line 83: Design concern (non-blocking, already partially acknowledged via the `TODO(4.2.1)` markers): these per-model `reverse_diff_defaults` tables are hand-maintained and pinned to the ND 4.2.1 OpenAPI schema. The two failure modes on an ND upgrade that changes a template default are asymmetric, and one is dangerous:

- Stale/wrong value → the existing-side key survives → false `changed=True` + redundant PUT (noisy but safe).
- **User sets a field on-device to the table's (now-wrong) value and omits it from proposed → the reverse pass strips it → the reset is silently skipped.** That's a correctness regression with no test to catch it.

The `/nd-workaround-audit` tooling and vault note mitigate but don't eliminate this. Could we track a firm follow-up to derive these defaults at runtime from the GET-echoes schema rather than transcribing them by hand? That would remove the version-pinning fragility across all ten tables.

plugins/module_utils/utils.py
```diff
@@ -75,6 +75,53 @@ def issubset(subset: Any, superset: Any) -> bool:
...
+    """
+    if not isinstance(existing_data, dict) or not isinstance(proposed_data, dict):
+        return False
+
+    for key, value in existing_data.items():
+        if _is_effectively_empty(value):
+            continue
+        if key not in proposed_data:
+            return True
+        if isinstance(value, dict) and has_removals(value, proposed_data[key]):
```
mtarking, line 126: `has_removals` recurses into nested **dicts** present on both sides but not into **lists**, so a removal expressed inside a list element (e.g. a nested dict in a policy/`ports` list that loses a key) won't be detected by the reverse pass. In practice the forward `issubset` catches most list divergences, so this is likely fine — but the dict-vs-list asymmetry is surprising. Worth a one-line comment here so a future reader doesn't assume deep list coverage.

plugins/module_utils/utils.py
```diff
@@ -75,6 +75,53 @@ def issubset(subset: Any, superset: Any) -> bool:
...
+    # Summary
+
+    Return `True` when `value` carries no user-visible content: `None`, `""`, `[]`, `{}`, or a dict whose values are all themselves effectively empty.
+    ND echoes such empty markers for fields the user never configured, so `has_removals` must not treat them as removable values.
+
+    ## Raises
+
+    None
+    """
+    if value is None or value == "" or value == [] or value == {}:
```
mtarking, line 92: Minor: a dict of only-empty values is collapsed to empty (the recursive branch below), but a list of empty markers (`[""]`, `[{}]`) is not — only an exact `== []` counts. That's probably intentional, but the asymmetry with the dict case is surprising; a short comment noting lists are only empty-normalized when literally `[]` would help.

plugins/module_utils/models/interfaces/loopback_interface.py
```diff
@@ -48,6 +48,17 @@ class LoopbackPolicyModel(NDNestedModel):
...
 
+    # TODO(4.2.1) get-echoes-schema-defaults-for-unset-fields
+    # ND 4.2.1 `int_loopback` template defaults (schema-sourced via nd-openapi `intLoopbackTemplate`). ND echoes these
+    # for every field the user never set; the reverse pass of `get_diff` normalizes existing-side matches to absent
+    # so replaced/overridden removal detection (issue #410) stays idempotent against default echoes.
+    # Values must be in the model's DUMPED form: the schema declares routeMapTag as integer 12345, but
+    # `coerce_route_map_tag` stores it as a string (ND 4.2.1 GET-side type drift), so the table holds "12345".
+    reverse_diff_defaults: ClassVar[dict[str, Any]] = {
+        "adminState": True,
+        "routeMapTag": "12345",
```
mtarking, line 59: This `routeMapTag: "12345"` (int→str GET-side drift) is a good illustration of the broader brittleness: `reverse_diff_defaults` values must match the **serialized/dumped** form exactly. A subtle form mismatch (int vs str, enum vs value, casing) silently reopens #410 for that field with no failing test. Each entry effectively needs lab verification in dumped form — worth calling out in the review/merge notes as a maintenance cost, not just for loopback.

plugins/module_utils/models/interfaces/vpc_trunk_host_interface.py
```diff
@@ -223,6 +223,41 @@ class TrunkVpcHostPolicyModel(StormControlMutexMixin):
...
+    # `peerSwitchId` is orchestrator-injected (resolved from the vPC pair record; not in the argspec), so the
+    # proposed config can never express it. ND echoes it on reads; without this exclusion the reverse pass of
+    # `get_diff` would count it as a removal on every replaced/overridden run, breaking idempotency.
+    reverse_diff_exclude: ClassVar[set[str]] = {"peerSwitchId"}
+
+    # TODO(4.2.1) get-echoes-schema-defaults-for-unset-fields
+    # ND 4.2.1 `int_vpc_trunk_host` template defaults (schema-sourced via nd-openapi `intVpcTrunkHostTemplate`). ND echoes these
+    # for every field the user never set; the reverse pass of `get_diff` normalizes existing-side matches to absent
+    # so replaced/overridden removal detection (issue #410) stays idempotent against default echoes.
+    reverse_diff_defaults: ClassVar[dict[str, Any]] = {
```
mtarking, line 235: The PR body notes the vPC families were **not lab-verified** in this PR, and that `TrunkVpcHostPolicyModel`'s `peer1AllowedVlans`/`peer2AllowedVlans` schema defaults have no corresponding model fields (per-peer → collapsed `access_vlan` workaround). Since vPC is the most complex interface family and this table drives replaced/overridden idempotency, could we file an explicit tracked follow-up to lab-verify both vPC tables (access + trunk-host) against a real ND read before users rely on replaced/overridden here? Flagging so the residual doesn't get lost after merge.

mtarking (commented):
Reviewed the reverse-pass diff logic — the core design is correct and well-tested, and I agree with the approach. The forward/reverse separation is clean (`has_removals` keys off presence only, values stay with the forward `issubset` pass), stripping default-valued existing keys is semantically justified for replaced/overridden (resetting an already-default field is a genuine no-op), and the dump-reuse efficiency refactor is nicely pinned by a regression test. CI is green and the prior `ptp` finding is resolved.

My comments are non-blocking. The one theme worth a firm follow-up is the hand-maintained, ND-4.2.1-pinned `reverse_diff_defaults` tables: the failure modes on a future ND default change are asymmetric and one (silently skipped reset) is a correctness risk with no test to catch it. Runtime-deriving those defaults from the GET-echoes schema would remove that fragility. The rest are small clarifying-comment nits (list non-recursion in `has_removals`, list-of-empties normalization) plus an ask to file an explicit follow-up to lab-verify the unverified vPC tables.

Overall: approve-with-follow-ups from my side.

plugins/module_utils/models/base.py
allenrobel: The silent-skip direction (user resets a field to the table's now-stale value, omits it from proposed, reverse pass strips it) is the dangerous one, agreed — noisy false-changed is self-announcing, this isn't. Filed #497 to track deriving these defaults at runtime from the template schema ND serves, replacing the hand-maintained tables. The issue captures the asymmetry, the dumped-form coercion requirement, and the schema-fetch-failure fallback question. Note `reverse_diff_exclude` stays regardless: it covers fields ND echoes but does not declare in any schema (`ptp`, `peerSwitchId`), which no runtime derivation would return.

plugins/module_utils/utils.py
allenrobel: Comment added at the recursion site in "Document reverse-pass list handling and dumped-form table requirement". Worth stating here too: the dict-vs-list asymmetry is benign by construction, not just "likely fine" — the forward `issubset` pass matches list elements bidirectionally (`issubset(item, candidate) and issubset(candidate, item)`), so a list element that loses a key fails the forward pass and classifies as changed; a wholly-omitted list key is caught by the key-presence check in `has_removals`. List coverage is complete across the two passes — it just wasn't visible at this call site, which is what the new comment fixes.

plugins/module_utils/utils.py
allenrobel: Comment added in "Document reverse-pass list handling and dumped-form table requirement". The asymmetry is intentional conservatism: ND has only been observed echoing literal `[]` for never-configured list fields, so only that form is empty-normalized. Collapsing `[""]` / `[{}]` without lab evidence of ND echoing those shapes could mask a real pending removal, so lists stay literal-`[]`-only until we see otherwise.

plugins/module_utils/models/interfaces/loopback_interface.py
allenrobel: Generalized in "Document reverse-pass list handling and dumped-form table requirement": the base-class comment over `reverse_diff_defaults` now states that table values must be in the model's dumped form, not the schema-declared form, citing `routeMapTag` (schema integer `12345`, stored `"12345"`) as the failure shape — a form mismatch silently reopens #410 for that field with no failing test. Will also call this out in the merge notes as a maintenance cost applying to all ten tables; #497 (runtime derivation) is the longer-term fix for the transcription fragility.

plugins/module_utils/models/interfaces/vpc_trunk_host_interface.py
allenrobel: We'll lab-verify both vPC tables (access + trunk-host) against a real ND read today and reply here with the results by end of today.

allenrobel: Lab verification completed as promised — both vPC `reverse_diff_defaults` tables checked against live ND 4.2.1 reads (SITE1, clean greenfield rebuild earlier today; physical-peer-link vPC pair S1_LE1/S1_LE2).

**Method:** created minimal `trunkVpcHost` and `accessVpcHost` interfaces sending only identity + port-channel ids + member ports (every optional field unset), then compared the GET echo field-by-field against the tables.

**Results:**

- **All 23 trunk-table and 22 access-table entries match the live echoes value-for-value.** No stale or wrong-valued entries in either table.
- **One gap found and fixed:** ND echoes a collapsed `allowedVlans: "none"` on trunkVpcHost GETs when the user never set it. The trunk table had no entry for it, so `replaced`/`overridden` reported `changed` on every run — lab-reproduced with the module (merged idempotent, replaced never converged). Fixed in commit *"Cover the collapsed allowedVlans echo in the vPC trunk reverse pass (#422 lab-verify)"*.
- **The fix required the dumped-form keys** — exactly the subtlety the dumped-form table comment on this line warns about. A collapsed-form `"allowedVlans"` entry does NOT work (lab-reproduced): the write-side dump fans the field out per `vpc-interface-peer-vlan-collapse`, so the entries are `peer1AllowedVlans`/`peer2AllowedVlans` = `"none"`. Regression unit test added (`test_base_model_reverse_diff_00590`) covering both the unset-echo no-diff and explicit-value change-detection paths. Post-fix lab rerun: merged ×2 and replaced ×2 all idempotent.
- **`nativeVlan` and `accessVlan` are NOT echoed when unset** (verified on both types), so their absence from the tables is correct — the asymmetry vs `allowedVlans` is now recorded in the vault note (`vpc-interface-peer-vlan-collapse`).
- **`ptp`:** ND injects `ptp: false` into vPC GETs too (same as port-channel), but the vPC policy models don't declare `ptp`, so `extra="ignore"` drops it at parse time — no exclude needed, unlike port-channel where it's a declared field.

**Incidental lab observation:** a physical-peer-link vPC pair created with the bare pair payload (no `vpcPairDetails`, ND auto-selects the peer-link) config-saved and deployed cleanly on 4.2.1 / NX-OS 10.6(2) — peer adjacency ok, Po500 up. The `vpc-physical-peer-link-configsave-fails` 500 documented in the integration vars appears specific to explicitly supplying member ports in `vpc_pair_details`; worth a follow-up before reverting the integration substrate to physical.

Full suite: 3912 unit tests green; black/isort/pylint/mypy clean on the changed files.

plugins/module_utils/models/interfaces/port_channel_trunk_host_interface.py
mikewiebe: ### High: Unconditional `ptp` exclusion silently defeats replaced/overridden removal

**Issue**

The fix for ND's injected `ptp` echo unconditionally removes `ptp` from the
reverse comparison even though the module still exposes it as a writable
port-channel field. A `replaced` or `overridden` task that omits a previously
enabled PTP value is classified as unchanged, so the requested reset is never
sent.

**Evidence**

- [`port_channel_trunk_host_interface.py` lines 264-270](https://github.com/CiscoDevNet/ansible-nd/pull/422/files#diff-d2ab524cc8cd4727b8449727f59773a79668d8ac08e5617de6638ef13c4c74a4R264)
  adds `reverse_diff_exclude = {"ptp"}`, stripping either `false` or `true`
  from removal detection.
- The same model still declares `ptp` as “Enable Precision Time Protocol on
  the interface” at
  [line 310](https://github.com/CiscoDevNet/ansible-nd/blob/dce04e8b33bc8386c03c15508132d057600761ee/plugins/module_utils/models/interfaces/port_channel_trunk_host_interface.py#L310),
  includes it in the argspec, and the module documents it as writable at
  [`nd_interface_port_channel_trunk_host.py` line 195](https://github.com/CiscoDevNet/ansible-nd/blob/dce04e8b33bc8386c03c15508132d057600761ee/plugins/modules/nd_interface_port_channel_trunk_host.py#L195).
- The current regression at
  [`test_base_model_reverse_diff.py` lines 708-730](https://github.com/CiscoDevNet/ansible-nd/blob/dce04e8b33bc8386c03c15508132d057600761ee/tests/unit/module_utils/models/test_base_model_reverse_diff.py#L708)
  deliberately asserts that existing `ptp: true` plus a proposed replacement
  omitting `ptp` is “no difference.” Test 00580 proves only that an explicitly
  supplied opposite value still forward-diffs.
- Exact-head probe:

  ```text
  existing response: ptp=true
  proposed replaced config: ptp omitted
  get_diff(..., exclude_unset=False): True  # interpreted as no_diff
  ```
- A practical two-run workflow demonstrates the user-visible failure. First,
  the user enables PTP through the documented field:

  ```yaml
  - name: Enable PTP on the server port-channel
    cisco.nd.nd_interface_port_channel_trunk_host:
      fabric_name: FABRIC1
      state: replaced
      config:
        - switch_ip: 192.0.2.11
          interface_name: port-channel10
          config_data:
            network_os:
              policy:
                admin_state: true
                allowed_vlans: "100-110"
                ports:
                  - Ethernet1/1
                  - Ethernet1/2
                ptp: true
  ```

  Later, the user relies on `replaced` semantics and omits `ptp` to reset it:

  ```yaml
  - name: Replace the policy and reset omitted settings
    cisco.nd.nd_interface_port_channel_trunk_host:
      fabric_name: FABRIC1
      state: replaced
      config:
        - switch_ip: 192.0.2.11
          interface_name: port-channel10
          config_data:
            network_os:
              policy:
                admin_state: true
                allowed_vlans: "100-110"
                ports:
                  - Ethernet1/1
                  - Ethernet1/2
                # ptp omitted: reset it to the default
  ```

  | Outcome | Expected | Current PR behavior |
  | --- | --- | --- |
  | Module result | `changed: true` | `changed: false` |
  | Controller action | PUT resetting PTP | No PUT |
  | Final PTP state | Disabled/defaulted | Remains `true` |

  The forward subset test passes because desired omits `ptp`. The reverse pass
  should detect the existing-only field, but `reverse_diff_exclude` removes it,
  so the state machine reports no difference and never sends the reset.

**Existing PR overlap**

Related to @mikewiebe's active
[thread r3681578271](https://github.com/CiscoDevNet/ansible-nd/pull/422#discussion_r3681578271).
That thread identified the original `ptp: false` false-positive loop. The
current finding shows that the unconditional-exclusion repair creates the
inverse false negative. The thread is **not close-ready**.

Related open issues: [#410](https://github.com/CiscoDevNet/ansible-nd/issues/410)
broadly tracks omitted-field false negatives for `replaced`/`overridden`, but
does not resolve the new writable-field versus controller-owned `ptp` contract
conflict introduced by this repair. [#497](https://github.com/CiscoDevNet/ansible-nd/issues/497)
tracks schema-derived default tables and explicitly leaves undeclared `ptp`
handling to `reverse_diff_exclude`; it does not track this conflict either.

**Impact**

A documented configuration field can remain enabled after a replacement says
to remove omitted settings. This is a silent convergence failure: Ansible
reports no change and never sends the reset. The author's lab evidence that ND
rewrites `ptp` fabric-wide explains why a simple `False` default is
insufficient, but it also demonstrates that the ownership contract must be
resolved rather than hiding the field in only one comparison direction.

**Suggested fix**

Choose one consistent contract. If `ptp` is controller/fabric-owned telemetry,
remove it from the model's writable fields, argspec, documentation, and payload.
If it is user-writable, preserve real `replaced`/ `overridden` reset
semantics while separately normalizing controller-injected fabric state. Add
tests for omitted reset, explicit true/false transitions, and both observed GET
echoes.

plugins/module_utils/utils.py
mikewiebe: ### Medium: Reverse scrubbing skips models inside lists

**Issue**

`_scrub_reverse_diff_dict` removes controller-only extras from the current model and directly nested models, but it does not recurse into model instances inside lists, tuples, or dictionaries. A controller-added key in a nested list can therefore make an otherwise identical `replaced` or `overridden` resource appear changed forever.

**Evidence**

- [`base.py` lines 304–311](https://github.com/CiscoDevNet/ansible-nd/pull/422/files#diff-de188588baa5d72cce7c8f2cd0de0d83e0b769a6e26ff5eba9fc50875b24b902R304) recurses only when the direct field value is an `NDBaseModel`.
- [`utils.py` lines 50–62](https://github.com/CiscoDevNet/ansible-nd/pull/422/files#diff-f3e3720de5c754cfeb714aa1ed0f9d0a1a043f8c0005944d58565decefe1817bR50) requires list elements to match in both directions, so a server-only key fails the forward comparison before reverse scrubbing occurs.
- `NetflowExporterModel` permits response extras and is held in `netflowExporterCollection: list[NetflowExporterModel]` at [`manage_fabric_common.py` lines 110–127](https://github.com/CiscoDevNet/ansible-nd/blob/9eab6c54c8092288d821bda3a5f07d9fdae854c7/plugins/module_utils/models/manage_fabric/manage_fabric_common.py#L110).
- The current-head probe retains the controller-only field:

  ```text
  existing: {'children': [{'name': 'same', 'controllerOnly': 'x'}]}
  reverse:  {'children': [{'name': 'same', 'controllerOnly': 'x'}]}
  get_diff: False  # interpreted as changed
  ```

Practical example:

```yaml
# Desired configuration
netflowExporterCollection:
  - exporterName: COLLECTOR1
    exporterIp: 192.0.2.50
    vrf: management
    sourceInterfaceName: loopback0
    udpPort: 2055
```

ND returns the same configuration with a generated identifier:

```yaml
netflowExporterCollection:
  - exporterName: COLLECTOR1
    exporterIp: 192.0.2.50
    vrf: management
    sourceInterfaceName: loopback0
    udpPort: 2055
    controllerGeneratedId: exporter-74
```

Although every user-managed value matches, the list item still contains `controllerGeneratedId` during comparison. The module consequently reports `changed: true` and can repeat the same PUT/deploy on every run.

**Existing PR overlap**

Related to @mtarking’s active [thread r3731410339](https://github.com/CiscoDevNet/ansible-nd/pull/422#discussion_r3731410339). That thread discusses list recursion, and the reply claims the forward and reverse passes provide complete list coverage. This counterexample shows that user-field removals are covered, but controller-only extras inside list-contained models are not. The thread is not close-ready.

Related open issue [#450](https://github.com/CiscoDevNet/ansible-nd/issues/450) broadly tracks nested-resource idempotency when controller responses contain additional fields, primarily for merged nested resources. It does not identify this PR’s forward-before-scrub failure on the `replaced`/`overridden` path. [#386](https://github.com/CiscoDevNet/ansible-nd/issues/386) concerns nested-list merge semantics and is a different operation.

**Impact**

Any shared Gen-3 model that allows response extras in child models held by a container can repeatedly issue writes and deployments for an unchanged resource. Check mode also predicts a false change.

**Suggested fix**

Recursively scrub `NDBaseModel` instances inside lists, tuples, and dictionary values before the forward comparison, while preserving container shape and ordering. Add a NetFlow-exporter regression plus a control proving a real user-managed list change is still detected.

plugins/module_utils/models/interfaces/port_channel_trunk_host_interface.py
allenrobel: Confirmed — the exclusion traded the false-positive loop for the inverse false negative, exactly as described. The contract conflict is now resolved by removing `ptp` from the writable surface entirely (commit "Remove inert ptp field from port-channel trunk-host (#422 review)"), your first option. Two findings drove that choice over preserving reset semantics:

1. `intPortChannelTrunkHostTemplate` declares no `ptp` property (nor does `intTrunkHostTemplate` — no non-IPFM template does).
2. Lab probe, ND 4.2.1.10 (2026-08-12, intent-only, fabric PTP disabled): two trunkPoHost port-channels created identically except one sent `ptp: true`. ND persisted and echoed the sent value, but `GET .../pendingConfig` showed byte-identical generated CLI for both — no PTP command. On a non-PTP fabric a user-set `ptp` was persist-but-inert: it configured nothing while the echo faked success. Rendering with fabric PTP *deployed* is untested, but our 2026-06-10 lab evidence shows the fabric-PTP deploy rewrites the stored value fabric-wide regardless of per-interface intent — so per-interface `ptp` cannot durably express user intent in either regime, and there was no working "reset" to preserve.

The injected GET echo now drops at parse time (`extra="ignore"`), matching the sibling ethernet/vPC models; `reverse_diff_exclude` is gone, so no writable field is hidden from removal detection. Since the module has not shipped, the option removal has no user impact. Tests cover idempotency against both the `ptp: false` echo and the post-fabric-PTP `true` rewrite, plus a pin that `ptp` stays out of `model_fields` and every dump. The bug-tracker vault note (interface-get-undocumented-ptp-field) is updated with the probe evidence.

plugins/module_utils/utils.py
allenrobel: Confirmed with the real models — existing `NetflowExporterModel` list item carrying a controller-only key vs. an identical proposed item: forward `issubset` returns False before any scrub runs, permanent `changed=true`. Your counterexample also corrects my earlier reply: "complete list coverage" holds for user-field *removals*, not for controller *extras* inside list items.

One scoping fact: the failure is in the forward pass's bidirectional list-element match (`issubset(item, candidate) and issubset(candidate, item)`), and `issubset` is byte-identical to develop — this PR neither introduced nor touched it. The same false-`changed` fires on develop today, on the merged path as well; the reverse scrub this PR adds runs only after a forward match, so it cannot reach the defect (which is also why recursing the scrub into lists would not fix the user-visible symptom). Rather than grow #422 into forward-diff semantics, I have posted the counterexample and a proposed fix direction to #450 (nested-resource idempotency when controller responses carry additional fields): https://github.com/CiscoDevNet/ansible-nd/issues/450#issuecomment-5269175029. Happy to pick it up there as a fast-follow.

plugins/module_utils/models/interfaces/port_channel_trunk_host_interface.py
allenrobel: Follow-up closing the caveat: we deployed fabric PTP on SITE1 (`ptpVlanId: 1000`, ToR source SVIs, `feature ptp` live on all 7 switches) and re-ran the probe in the enabled-not-deployed and fully-deployed regimes. A client-sent `ptp` is persist-but-inert in **all three** regimes — echoed back, zero pending-CLI difference vs an identical control. Two structural confirmations from the deploy: ND's generated PTP CLI lands only on physical interfaces, never under port-channels (matching NX-OS's per-physical-port PTP model, so a port-channel-level `ptp` knob cannot render anything), and the fabric-wide record rewrite reproduced on 4.2.1.10 — every pre-deploy physical/port-channel record now stores `ptp: true` including `port-channel500`, which received no PTP CLI. The removal rationale is now unconditional; code comments and the vault note are updated (commit "Close the ptp probe caveat: inert in all three fabric-PTP regimes").

plugins/module_utils/models/base.py
allenrobel: Resolving comment after 6 days.

plugins/module_utils/utils.py
allenrobel: Resolving comment after 6 days.

plugins/module_utils/models/interfaces/loopback_interface.py
allenrobel: Resolving comment after 6 days.

plugins/module_utils/models/interfaces/vpc_trunk_host_interface.py
allenrobel: Resolving comment after 6 days.

mikewiebe (approved):
Re-reviewed at 9873648d7a2e1c47395db8058746f22b4cef08f5. The prior PTP blocker is resolved, the remaining nested-list behavior predates this PR and is tracked by #450, and the focused/full unit verification is green.

plugins/module_utils/models/base.py
```diff
@@ -221,6 +238,78 @@ def to_diff_dict(self, **kwargs) -> Dict[str, Any]:
...
+                del data[alias]
+        for field_name, field_info in type(self).model_fields.items():
+            value = getattr(self, field_name, None)
+            if isinstance(value, NDBaseModel):
+                alias = field_info.alias or field_name
+                nested = data.get(alias)
+                if isinstance(nested, dict):
+                    # Same-class recursion; pylint cannot infer `value` is an NDBaseModel from getattr.
+                    value._scrub_reverse_diff_dict(nested)  # pylint: disable=protected-access
+
```
akinross, line 312: is get_diff supposed to be limited to the same model? if so we should probably add type validation similar to merge and error or return false

plugins/module_utils/models/interfaces/ethernet_access_interface.py
```diff
@@ -71,6 +71,33 @@ class EthernetAccessPolicyModel(StormControlMutexMixin):
     None
     """
 
+    # TODO(4.2.1) get-echoes-schema-defaults-for-unset-fields
+    # ND 4.2.1 `int_access_host` template defaults (schema-sourced via nd-openapi `intAccessHostTemplate`). ND echoes these
+    # for every field the user never set; the reverse pass of `get_diff` normalizes existing-side matches to absent
+    # so replaced/overridden removal detection (issue #410) stays idempotent against default echoes.
+    reverse_diff_defaults: ClassVar[dict[str, Any]] = {
```
akinross, line 78: Did you consider having these defaults defined at Field level with `json_schema_extra={"reverse_diff_default": True}` or Annotated?

plugins/module_utils/models/base.py
allenrobel: Good catch — `merge` had the guard and `get_diff` didn't, and this PR made the gap matter: the reverse pass calls `other._apply_reverse_diff_scope()`, which assumes `other` carries the same `reverse_diff_*` declarations as `self`. Added the same `isinstance(other, type(self))` check with a `TypeError` in "Reject cross-type get_diff with TypeError (#422 review)". Chose raise over `return False` for consistency with `merge` and because a cross-type comparison is a programming error — returning `False` would surface it as a spurious "changed" instead of failing loudly. It applies on both the merged and replaced/overridden paths. The only Gen-3 caller (`NDConfigCollection.get_diff_config`) always compares items from the same collection, so no real path changes behavior. Test `test_base_model_reverse_diff_00620` covers both paths.

plugins/module_utils/models/interfaces/ethernet_access_interface.py
allenrobel: Considered it, and kept the per-model `ClassVar` table for this PR for three reasons:

1. **The values are not the schema defaults verbatim** — they must be in the model's *dumped* form (e.g. loopback `routeMapTag` is schema integer `12345` but a validator stores `"12345"`, see the `base.py` comment above the table). A per-field `json_schema_extra={"reverse_diff_default": ...}` would carry the same caveat, just spread across ~20 fields per model instead of one commented block, and the table's `TODO(4.2.1)` marker + schema-source note would have nowhere obvious to live.
2. **Sanity environment** — `pydantic_compat.py` ships a `Field` shim for the no-pydantic sanity run; reading `json_schema_extra` back off `model_fields` at scrub time is one more surface that shim would have to fake. Nothing under `plugins/module_utils/models` uses `json_schema_extra` today, so it would be a new convention introduced by a bug-fix PR.
3. **#403 follow-on** — the loopback policy_type discriminated-union PR moves these tables per-policy-model on rebase; a `ClassVar` is the simplest unit to override in a subclass.

`Annotated` has the same trade-offs plus a per-field type-alias explosion (one alias per default value). I'm happy to revisit as part of the model refactor (#348) once the tables are settled per policy type — that's the point where a declarative Field-level form would pay for itself across all 10 models at once.

plugins/module_utils/models/interfaces/port_channel_trunk_host_interface.py
allenrobel: Resolving since Mike has approved.

plugins/module_utils/utils.py
allenrobel: Resolving after 2 weeks.

akinross (approved):

PR #423, opened 2026-07-23
Added validation for Gather filter supported properties

jeetugangwar11:
## Proposed Changes

- Define and document the supported `state: gathered` filter properties:
  - `entity_name`
  - `pool_name`
  - `switches`
  - `resource`
  - `scope_type`
  - `pool_type`
- Use `pool_name` and `switches` as GET API query parameters.
- Apply all six supported properties during local resource matching.
- Do not use Lucene filtering because normalized entity matching for device-pair and link resources must be performed locally.
- Reject unsupported gathered filter properties before making API requests.
- Reject empty gathered filter items such as `config: [{}]`.
- Reject explicitly null supported properties, including when a null property is combined with another valid filter.
- Preserve gather-all behavior when `config` is omitted or specified as an empty list.
- Update module documentation with the supported filter matrix and validation behavior.
- Add unit tests for supported, unsupported, empty, null-only, and mixed-null gathered filters.
- Add invalid gathered-filter integration coverage to the root, BASE, iBGP, eBGP, and External resource-manager test suites.

## Test Notes

The following validation was completed:

- Focused gathered validation tests: `22 passed`
- Complete resource-manager unit suite: `291 passed`
- Parsed all five updated invalid_params.yaml files successfully.
- Verified register names are unique within each integration test file.
- Ran `ansible-test sanity --test validate-modules --venv plugins/modules/nd_manage_resource_manager.py` successfully.
- Verified no VS Code diagnostics in the changed Python and YAML files.

Cisco Nexus Dashboard 4.3x

* [ ] analyze
* [ ] infa
* [x] manage
* [ ] onemanage
* [ ] other

Resource category: Manage resource manager API (`/manage/api/v1`).

## Checklist

### DIFF
--- a/plugins/module_utils/models/manage_resource_manager/resource_manager_config_model.py
+++ b/plugins/module_utils/models/manage_resource_manager/resource_manager_config_model.py
@@ -17,7 +17,7 @@ Fields map directly to the module's config suboptions:
   switch           → switch          (list of switch IPs/serials; required for non-fabric scopes)
 
 State-aware validation is supported when model_validate() is called with
-context={"state": "merged|deleted|query|gathered"}.
+context={"state": "merged|deleted|gathered"}.
 """
 
 from __future__ import annotations
@@ -50,14 +50,15 @@ class ResourceManagerConfigModel(NDBaseModel):
 
     Provides full per-field and cross-field validation for resource allocation
     configuration. Supports state-aware validation when model_validate() is
-    called with context={"state": "merged|deleted|query|gathered"}.
+    called with context={"state": "merged|deleted|gathered"}.
 
     Field requirements by state:
       merged:          entity_name, pool_type, pool_name, scope_type required;
                        switch required for non-fabric scopes.
       deleted:         entity_name, pool_type, pool_name, scope_type required;
                        switch required for non-fabric scopes.
-      query/gathered:  all fields optional (used as filters).
+      gathered:        config may be omitted or empty; each provided filter item
+                       requires at least one supported, non-null property.
 
     Note: The nd_manage_resource_manager module performs its own mandatory field
     checks before calling from_config(). This model adds per-field normalization
@@ -149,8 +150,8 @@ class ResourceManagerConfigModel(NDBaseModel):
         if state != "gathered" or not isinstance(data, dict):
             return data
 
-        supplied_properties = set(data)
-        unsupported = sorted(supplied_properties - set(cls.GATHERED_FILTER_PROPERTIES))
+        active_properties = {key for key, value in data.items() if value is not None}
+        unsupported = sorted(active_properties - set(cls.GATHERED_FILTER_PROPERTIES))
         if unsupported:
             raise ValueError(
                 "unsupported gathered filter properties: {0}. Supported gathered filter properties: {1}".format(
@@ -159,11 +160,6 @@ class ResourceManagerConfigModel(NDBaseModel):
                 )
             )
 
-        null_properties = sorted(key for key, value in data.items() if value is None)
-        if null_properties:
-            raise ValueError("gathered filter properties cannot be null: {0}".format(", ".join("'{0}'".format(key) for key in null_properties)))
-
-        active_properties = {key for key, value in data.items() if value is not None}
         if not active_properties:
             raise ValueError(
                 "gathered filter item must contain at least one supported property: {0}".format(
@@ -485,8 +481,9 @@ class ResourceManagerConfigModel(NDBaseModel):
         When model_validate(context={"state": "merged"}) or
         model_validate(context={"state": "deleted"}) is passed, the model
         enforces that entity_name, pool_type, pool_name, and scope_type are
-        all provided. For 'query' / 'gathered' state (or no context), all
-        fields remain optional and serve as filters.
+        all provided. For 'gathered' state, individual fields are optional, but
+        each provided filter item must contain at least one supported, non-null
+        property.
         """
         state = (info.context or {}).get("state") if info else None
         if state in ("merged", "deleted"):
--- a/plugins/module_utils/models/manage_resource_manager/resource_manager_config_model.py
+++ b/plugins/module_utils/models/manage_resource_manager/resource_manager_config_model.py
@@ -161,11 +161,7 @@ class ResourceManagerConfigModel(NDBaseModel):
 
         null_properties = sorted(key for key, value in data.items() if value is None)
         if null_properties:
-            raise ValueError(
-                "gathered filter properties cannot be null: {0}".format(
-                    ", ".join("'{0}'".format(key) for key in null_properties)
-                )
-            )
+            raise ValueError("gathered filter properties cannot be null: {0}".format(", ".join("'{0}'".format(key) for key in null_properties)))
 
         active_properties = {key for key, value in data.items() if value is not None}
         if not active_properties:
--- a/plugins/module_utils/models/manage_resource_manager/resource_manager_config_model.py
+++ b/plugins/module_utils/models/manage_resource_manager/resource_manager_config_model.py
@@ -36,6 +36,7 @@ from ansible_collections.cisco.nd.plugins.module_utils.models.manage_resource_ma
 from ansible_collections.cisco.nd.plugins.module_utils.models.manage_fabric.constants import is_pool_supported
 from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat import (
     Field,
+    ValidationError,
     field_validator,
     model_validator,
 )
@@ -64,6 +65,14 @@ class ResourceManagerConfigModel(NDBaseModel):
     """
 
     identifiers: ClassVar[list[str]] = []
+    GATHERED_FILTER_PROPERTIES: ClassVar[tuple[str, ...]] = (
+        "entity_name",
+        "pool_name",
+        "switches",
+        "resource",
+        "scope_type",
+        "pool_type",
+    )
 
     # Fields excluded from diff — operational input flags not present in gathered state.
     # entity_name, pool_type, pool_name, scope_type, resource and vrf_name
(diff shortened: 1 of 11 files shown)


plugins/module_utils/models/manage_resource_manager/resource_manager_config_model.py
```diff
@@ -132,6 +141,77 @@ class ResourceManagerConfigModel(NDBaseModel):
...
+
+        supplied_properties = set(data)
+        unsupported = sorted(supplied_properties - set(cls.GATHERED_FILTER_PROPERTIES))
+        if unsupported:
+            raise ValueError(
+                "unsupported gathered filter properties: {0}. Supported gathered filter properties: {1}".format(
+                    ", ".join("'{0}'".format(key) for key in unsupported),
+                    ", ".join("'{0}'".format(key) for key in cls.GATHERED_FILTER_PROPERTIES),
+                )
+            )
```
allenrobel: `supplied_properties = set(data)` counts a key as "supplied" even when its value is `None` — and on a real module run, Ansible's argument-spec validation populates **every** declared `config` suboption with `None` for keys the user omitted (the module declares `config` with `elements="dict", options={...}`). So every gathered item arrives here with all 8 keys present, including `vrf_name` and `is_pre_allocated`, and this check raises `unsupported gathered filter properties: 'is_pre_allocated', 'vrf_name'` for **every** real `state: gathered` invocation — including a perfectly valid `config: [{entity_name: foo}]`. Reproduced against ansible-core 2.18.17's `ArgumentSpecValidator`:

```python
validator.validate({"config": [{"entity_name": "foo"}]})
# -> {'config': [{'entity_name': 'foo', 'pool_type': None, ..., 'vrf_name': None, 'is_pre_allocated': None}]}
```

The unit tests don't catch this because they inject sparse dicts directly into `nd.params`, bypassing Ansible's suboption normalization; the new integration tests will fail with the wrong message against a real ND (e.g. the empty-filter case expects `must contain at least one supported property` but will get the unsupported-properties error, and the mixed-null case expects `cannot be null`).

Suggested direction: compute `unsupported` from non-`None` keys only (e.g. `{k for k, v in data.items() if v is not None}`). Note this also means the `null_properties` check below can never distinguish an explicit `vrf_name: null` from an omitted key — Ansible collapses both to `None` — so the "explicitly null" error path is likely undeliverable as designed and may need rethinking (the existing `active_properties` check already covers the all-empty case).

plugins/modules/nd_manage_resource_manager.py
```diff
@@ -133,6 +137,11 @@
   - Idempotence checking compares the existing resource value to the desired value.
   - Entity name matching is order-insensitive for tilde-separated serial numbers.
   - For C(state=gathered), partial filter entries are supported (e.g., just C(entity_name) or just C(switches)) to retrieve specific subsets of resources.
+  - "Gathered filter support matrix (GET query / Lucene / local match): C(entity_name) (no / no / yes),
+    C(pool_name) (yes / no / yes), C(switches) (yes / no / yes), C(resource) (no / no / yes),
+    C(scope_type) (no / no / yes), and C(pool_type) (no / no / yes)."
+  - Gathered filters use the C(poolName) and C(switchId) GET query parameters. Lucene filtering is not used because exact server-side matching
+    cannot preserve normalized entity matching for device-pair and link resources.
```
allenrobel: These two notes document ND API internals/workaround rationale — which GET query parameters exist vs. don't, and why Lucene filtering is avoided ("exact server-side matching cannot preserve normalized entity matching for device-pair and link resources"). Per team policy, ND API quirk/workaround rationale should not appear in module `DOCUMENTATION` — it belongs in `TODO(X.Y.Z)` code markers (with a bug-tracker vault note) and the PR body. Please drop these notes from `DOCUMENTATION`; user-facing docs should describe behavior (which filters are supported), not the API mechanics behind it — the preceding notes already cover that.

allenrobel (commented):
### Code review

Two inline comments below (one confirmed functional regression, one documentation-policy issue), plus one finding that lands on lines outside the diff:

1. **Stale "all fields optional" docstrings** — the new `validate_gathered_properties` validator rejects a gathered filter item with zero active properties, so "all fields optional" is no longer accurate for gathered state. The module `DOCUMENTATION` was updated with the new "at least one supported property" constraint, but the model's own docstrings still make the old claim in two places: the class docstring ([`query/gathered:  all fields optional (used as filters).`](https://github.com/CiscoDevNet/ansible-nd/blob/7087756512d791b2ca67c8d04aa92b23fe751132/plugins/module_utils/models/manage_resource_manager/resource_manager_config_model.py#L59-L61)) and the `apply_state_validation` docstring ([`all fields remain optional and serve as filters`](https://github.com/CiscoDevNet/ansible-nd/blob/7087756512d791b2ca67c8d04aa92b23fe751132/plugins/module_utils/models/manage_resource_manager/resource_manager_config_model.py#L487-L490)). Please update both to reflect the new constraint.

mikewiebe:
I have completed my review and have no additional findings. Once the outstanding review comments from @allenrobel are resolved, this PR is good to merge from my perspective.

plugins/modules/nd_manage_resource_manager.py
jeetugangwar11: Hi @allenrobel , I have made the changes, could you please have a look.

plugins/module_utils/models/manage_resource_manager/resource_manager_config_model.py
jeetugangwar11: Hi @allenrobel , I have made the changes. could you please have a look.

allenrobel (approved):
LGTM

The TODO is in the wrong format for an ND workaround since the version is not specified.  But we can fix that later.  Suggested fix:

```python
# TODO(4.2.1) resource-manager-gathered-entity-order-lucene
# ND echoes canonical endpoint order for multi-switch entity names, so exact Lucene filters on
# user-supplied entity_name can miss matches; gathered filtering matches entity names locally instead.
```

allenrobel:
@jeetugangwar11 when you get a chance later, below is the workaround note for the TODO(4.2.1) with the link back slug (`id:` field in the Markdown frontmatter) which links back to this note.  Once we figure out how to make the MCP server that serves these notes shareable, your LLM will be able to reference this, and other, workarounds.

Thanks

``` markdown
---
id: resource-manager-gathered-entity-order-lucene
endpoints:
  - GET /api/v1/manage/fabrics/{fabricName}/resources
  - POST /api/v1/manage/fabrics/{fabricName}/resources
tags:
  - deviation
found: 4.2.1
fixed:
status: workaround
severity: medium
guidance: "Do not filter resources server-side by entityName (Lucene `filter=entityName:...`): ND echoes multi-switch entity names (devicePair/link scopes) in canonical endpoint order, so an exact match against the user-supplied tilde-separated serial order can silently miss resources. Fetch broader (poolName/switchId query params + pagination) and match entityName locally, order-insensitively on the tilde-separated segments."
---

# Summary

For multi-switch resources (`scopeType: devicePair` and `link`), `entityName` is a tilde-joined composite of switch serials plus an entity suffix (spec example: `FDO222419B7~FDO222419B8~vpc55`). ND normalizes these names to a **canonical endpoint order** on read: the order of the serial segments echoed by `GET /api/v1/manage/fabrics/{fabricName}/resources` is ND's, not necessarily the order the client supplied at allocation time.

The endpoint's `filter` query parameter is a Lucene **exact-match** filter (the spec documents it only as "Lucene format filter" with no field list). An exact `filter=entityName:<user-supplied-name>` therefore silently excludes resources whose stored name differs from the query only by endpoint order — the resource exists and would match after order-insensitive normalization, but the server-side filter never returns it.

## Impact

- Any client that pushes `entityName` filtering to the server via Lucene gets **silently incomplete results** for devicePair/link resources — no error, just missing rows.
- Clients must instead over-fetch and match locally, which costs extra transfer/pagination on large fabrics. In `cisco.nd`'s `nd_manage_resource_manager` (`state: gathered`), this is why `entity_name` is a local-match-only filter while only `poolName` and `switchId` are pushed as GET query parameters.
- Single-switch scopes (`fabric`, `device`, `deviceInterface`) have no serial-order ambiguity in `entityName`, but the collection matches locally across the board for consistency.

## Workaround

### Full

Fetch with the documented first-class query parameters (`poolName`, `switchId`) plus `max`/`offset` pagination to bound the result set, then match `entityName` client-side, comparing tilde-separated segments order-insensitively. This is what `nd_manage_resource_manager` does (see `_build_gathered_resource_criteria` in `plugins/module_utils/manage_resource_manager/nd_manage_resource_manager_resources.py`; workaround marker `TODO(4.2.1) resource-manager-gathered-entity-order-lucene`).

## Reproduction

Not yet independently lab-reproduced. Sourced from `cisco.nd` resource-manager module development (jeetugangwar11, PRs [[#333](https://github.com/CiscoDevNet/ansible-nd/pull/333)](https://github.com/CiscoDevNet/ansible-nd/pull/333) and [[#423](https://github.com/CiscoDevNet/ansible-nd/pull/423)](https://github.com/CiscoDevNet/ansible-nd/pull/423)): the original implementation carried the in-code observation that "ND may return canonical endpoint order for multi-switch resources, and some simple entity-only Lucene filters can exclude resources that should match after local normalization", and the module documents "Entity name matching is order-insensitive for tilde-separated serial numbers" as a consequence.

To reproduce on the lab (SITE1, 192.168.7.7):

1. Allocate a devicePair resource (e.g. a `vpcId`) naming the entity with serials in a chosen order: `SERIAL-B~SERIAL-A~vpc10`.
2. `GET /api/v1/manage/fabrics/{fabricName}/resources` and inspect the echoed `entityName` — expect canonical (reordered) serials.
3. `GET /api/v1/manage/fabrics/{fabricName}/resources?filter=entityName:SERIAL-B~SERIAL-A~vpc10` — expect the resource to be absent from the filtered response while present unfiltered.

Update this section (and consider promoting `tags` to include `bug`) once the lab run confirms steps 2–3.

## Expected behavior

Either the read echoes `entityName` exactly as allocated, or the server-side `filter` matches entity names order-insensitively — so a client can round-trip the name it allocated into a Lucene filter and find the resource.

## Actual behavior

The stored/echoed `entityName` is canonicalized, and the Lucene filter is exact, so a round-tripped user-supplied name can silently match nothing.

## Related

- [[[vPC per-peer VLAN fields collapse on read](https://github.com/CiscoDevNet/ansible-nd/pull/vPC%20per-peer%20VLAN%20fields%20collapse%20on%20read)]] — same family: ND normalizes the echo of what was written.
```

jeetugangwar11:
Hi @allenrobel, I have made the changes for TODO, could you please have a look and verify if this looks good.

allenrobel (dismissed):
LGTM

allenrobel (approved):
No changes since my last approval except an (unnecessary) whitespace modification in manage_fabrics.py (unnecessary since this was not the cause of the sanity failure (pylint complaining about disallowed name in a unit test file that was actually fixed due to the rebase).

I'd ask that the change to manage_fabrics.py be reverted since this PR doesn't otherwise touch that file, but it's a cosmetic change so won't hold up the PR merge over it.

tests/integration/targets/nd_resource_manager/tests/nd/base/invalid_params.yaml
```diff
@@ -42,3 +42,67 @@
...
+          - invalid_gathered_empty_item is failed
+          - invalid_gathered_empty_item.msg is defined
+          - invalid_gathered_empty_item.msg is search("must contain at least one supported property")
+
+    - name: BASE gathered filters - Null-only property
+      cisco.nd.nd_manage_resource_manager:
+        state: gathered
+        fabric_name: "{{ ansible_it_fabric }}"
+        config:
+          - pool_name: null
```
sivakasi-cisco, line 85: Is the explicit-null accepted?

jeetugangwar11: @sivakasi-cisco , This is for negative scneraio validation. It should fail the module as per validation

mikewiebe (approved):

Issue #304, opened 2026-06-01
ND module's result output mismatch


jeetugangwar11:
Problem statement:
We are seeing different results output for ND modules in cases of using verbosity or module is using orchestrator layer. We need to create results and and output in a way, we must have common properties in results. If CLI is using verbosity Level-1, Level-2 or Level-3, we should have only addition field instead different properties like its prefixing api_* right now.
current results from Fabric VRF module:

<img width="467" height="414" alt="Image" src="https://github.com/user-attachments/assets/e82880db-8621-4b54-8b80-aa9c1bf07af1" />

And Resource Manager Mudule:

<img width="467" height="414" alt="Image" src="https://github.com/user-attachments/assets/a2427c89-30fa-4eeb-9669-7e8c0a31dd39" />

jeetugangwar11:
I have tested out merge, delete and gathered state in case of -vvv and -vv and attched sample response for the same

[Gathered-vv.json](https://github.com/user-attachments/files/30495571/Gathered-vv.json)
[Delete-vvv.json](https://github.com/user-attachments/files/30495570/Delete-vvv.json)
[delete-vv.json](https://github.com/user-attachments/files/30495569/delete-vv.json)
[Merge-vvv.json](https://github.com/user-attachments/files/30495567/Merge-vvv.json)
[Merge-vv.json](https://github.com/user-attachments/files/30495568/Merge-vv.json)
[gathered-vvv.json](https://github.com/user-attachments/files/30495566/gathered-vvv.json)

Not sure if fabric_type and workflow, populating these two props make sense for resource manager module as its not handling different workflows. 

Please let me know if this makes sense or I need to include these props as well.

Created Below PR for the same:
https://github.com/CiscoDevNet/ansible-nd/pull/455

cc: @mikewiebe , @allenrobel , @AKDRG , @sivakasi-cisco

PR #333, opened 2026-06-15
Ansible ND 4.X | Resource Manager Module with Pydantic Models + Smart Endpoints

jeetugangwar11:
## Related Issue(s)

- CiscoDevNet/ansible-nd#233

**Description:**
This PR adds the nd_manage_resource_manager module for resource allocation management in the cisco.nd collection, including the core resource handler, endpoint wrappers, Pydantic validation/response models, and shared constants.

**What's Included:**
Resource handler — [nd_manage_resource_manager_resources.py] supports gathered, merged and deleted states. 
The module pre-fetches all existing resources and switches at init time, translates playbook switch management IPs to switchId serial numbers automatically, and uses a single-batch POST for creates and a POST to actions/remove for deletes.
A ResourceManagerDiffEngine stateless helper class handles entity-name normalisation (tilde-order-insensitive), IPv4/IPv6-aware value comparison, to_add/to_update/to_delete/idempotent diff classification, and partial-match mismatch diagnostics (same entity_name, differing pool_name/scope_type/switch) logged to the debugs output bucket.

**Endpoint Definitions:**
Three endpoint classes in manage_fabrics_resources.py:
EpManageFabricResourcesGet — GET /api/v1/manage/fabrics/{fabricName}/resources — supports switchId, poolName query params and Lucene-style filter/max/offset/sort.
EpManageFabricResourcesPost — POST /api/v1/manage/fabrics/{fabricName}/resources — batch resource allocation; optional tenantName query param.
EpManageFabricResourcesActionsRemovePost — POST /api/v1/manage/fabrics/{fabricName}/resources/actions/remove — batch release of allocated resource IDs.

One additional endpoint in manage_fabrics_switches.py:
EpManageFabricSwitchesGet — GET /api/v1/manage/fabrics/{fabricName}/switches — used at init to build the fabricManagementIp to switchId translation map.

**Pydantic Model Layer:**
Model files in plugins/module_utils/models/manage_resource_manager/:
resource_manager_config_model.py — ResourceManagerConfigModel: validates a single playbook config entry including per-field normalisation (pool_type uppercase, scope_type underscore form, IPv4/v6/CIDR resource value), cross-field rules (pre-allocated requires resource, switch required for non-fabric scopes, pool_name/scope_type compatibility via POOL_SCOPE_MAP), and state-aware context validation.

[...]

PR #455, opened 2026-07-29
Added fixes for output mismatch and returning status code got from ND

jeetugangwar11:
Issues:
-------
1. Output mismatch in case of mudule using NDStateMachine and Orchestraror.
2. Use RETURN_CODE returned by ND instead of hardcoding 200 for successful operations.

Changes:
----------
Use the HTTP RETURN_CODE provided by the ND response envelope for successful Resource Manager API operations instead of hardcoding 200.
Preserve the response metadata for resource create, update-delete, and delete operations.
Continue using a synthetic 200 response for check-mode and logical operations where no API request is sent.
Preserve per-resource operation status and messages in the response DATA.
Retain the actual ND response code for partial-success failures, such as HTTP 207 Multi-Status.

Test Notes:
-----------
Ran the complete Resource Manager unit test suite:
Verified that partial delete responses expose the ND-provided RETURN_CODE: 207.
Verified that API path, verb, payload, response, result, metadata, and diff are returned correctly at -vvv.

Nexus Dashboard: 4.3.1.135.51166

Related ND API Resource Category:
------------------------------------
 manage

Checklist
 Latest commit is rebased from develop with merge conflicts resolved
 New or updates to documentation has been made accordingly
 Assigned the proper reviewers


tests/integration/targets/nd_resource_manager/tasks/rm_tasks.yaml
```diff
@@ -128,7 +128,14 @@
...
   ansible.builtin.include_tasks: vpc_pair_once.yaml
-  loop: "{{ test_items | map(attribute='fabric_type') | unique | reject('equalto', 'external') | list }}"
+  loop: >-
+    {{
+      test_items
+      | rejectattr('testcase', 'equalto', 'gathered')
+      | map(attribute='fabric_type')
+      | unique
+      | list
+    }}
```
allenrobel, line 138: This rewrite adds the `gathered` testcase exclusion but drops the previous `reject('equalto', 'external')` fabric-type filter, which PR #333 added to gate vPC-pair setup to fabric types where vPC applies. Non-gathered external testcases (`external/merge.yaml`, `external/delete.yaml`, `external/sanity.yaml`) still exist and `vpc_pair_once.yaml` has no `when:` guard, so a `fabric_type=external` (or `all`) run will now attempt `nd_manage_vpc_pair` against the external fabric, which has no vPC concept.

Suggest keeping both filters:

```yaml
  loop: >-
    {{
      test_items
      | rejectattr('testcase', 'equalto', 'gathered')
      | map(attribute='fabric_type')
      | unique
      | reject('equalto', 'external')
      | list
    }}
```

allenrobel (commented):
### Code review

One finding, on the vPC-pair setup loop in `rm_tasks.yaml`. The RETURN_CODE/verbosity changes themselves look correct (status captured immediately after each request, no cross-request leakage; unit and integration assertions match the new behavior).

tests/integration/targets/nd_resource_manager/tasks/rm_tasks.yaml
jeetugangwar11: Hi @allenrobel 

Creating a VPC pair from the ND dashboard UI is working in case of external fa fabric.
image
image

Please let me know if I am missing anything here.

Thanks.

allenrobel: Hi @jeetugangwar11 — thanks for checking the UI behavior.

I traced the history and confirmed the `reject('equalto', 'external')` filter was yours to begin with: it landed in PR #333 via your commit "[Jeet] Added test cases for each fabric type" — the same commit that added the `external/` test suites — so you own both the original exclusion and this removal. Given that, plus your observation that vPC-pair creation works on an external fabric, I'm fine dropping the filter.

One confirmation before we close this out: have you run the integration suite with `fabric_type=external` (or `all`) end-to-end against a live setup since removing the filter — i.e., with `vpc_pair_once.yaml` now executing against the external fabric? The non-gathered external testcases (`external/merge.yaml`, `external/delete.yaml`, `external/sanity.yaml`) will now trigger `nd_manage_vpc_pair` setup there, which is the path the old filter was gating. If that run is green, please resolve this thread.

jeetugangwar11: Hi @allenrobel , yeah earlier I excluded it, but later I found out it is supported, so I removed it.

allenrobel (dismissed):
LGTM

tests/integration/targets/nd_resource_manager/tests/nd/base/delete.yaml
```diff
@@ -123,16 +123,6 @@
...
-      # ND accepts VPC_DOMAIN_ID in the create response, but the resources GET
-      # inventory exposes an additional networkVlan allocation on the
-      # second switch instead of a VPC_DOMAIN_ID resourceId.
-      - entity_name: "base_network_vlan_device"
-        pool_type: "ID"
-        pool_name: "networkVlan"
-        scope_type: "device"
-        switches:
-          - "{{ ansible_switch2 }}"
-
```
allenrobel, line 135: Removing this entry (commit "[Jeet | issue-304] delete assertion fix") appears to regress the test in three ways:

1. **Resource leak** — this was the only entry that cleaned up the resource created by `base_vpc_domain_fabric` (`VPC_DOMAIN_ID`, fabric scope). Per the deleted comment, ND's resources GET exposes that allocation as a phantom `networkVlan` entry on the *second* switch, never as a `VPC_DOMAIN_ID` resourceId. The remaining switch1 `base_network_vlan_device` entry cannot match it (`_make_resource_key` includes the switch IP), and the `always:` Final cleanup reuses the same `*rm_delete` anchor — so the resource now leaks in the fabric on every run of this target.
2. **Count mismatch** — `rm_test_delete_config` drops to 14 entries while [`rm_expected_resource_count` stays 15](https://github.com/CiscoDevNet/ansible-nd/blob/5f30860bc254f7abb964d61746dc6187163614f4/tests/integration/targets/nd_resource_manager/tests/nd/base/delete.yaml#L205-L207), so the [delete assertion `(result["diff"][0]["deleted"] | length) == rm_expected_resource_count`](https://github.com/CiscoDevNet/ansible-nd/blob/5f30860bc254f7abb964d61746dc6187163614f4/tests/integration/targets/nd_resource_manager/tests/nd/base/delete.yaml#L245-L247) should fail 14 != 15 on the next lab run.
3. **Issue reference** — the commit cites issue-304, but #304 is about verbose-output field naming across modules, not delete fixtures.

If ND behavior changed in your lab (e.g. the phantom `networkVlan` allocation no longer appears, so this entry stopped matching), could you share that evidence? In that case the removal may be right, but the delete-scoped count needs to drop to 14 and some cleanup path for the `VPC_DOMAIN_ID`-created resource is still needed (or its create entry removed as well).

tests/integration/targets/nd_resource_manager/tests/nd/base/gathered.yaml
```diff
@@ -292,8 +292,8 @@
     - name: BASE common pools - Assert gathered resources
       ansible.builtin.assert:
         that:
-          - '(result["gathered"] | length) == rm_expected_resource_count'
-          - '(result["gathered"] | length) == (rm_expected_entity_names | length)'
+          - '(result["gathered"] | map(attribute="entity_name") | unique | list | length) == rm_expected_resource_count'
+          - '(result["gathered"] | map(attribute="entity_name") | unique | list | length) == (rm_expected_entity_names | length)'
```
allenrobel: Deduping the gathered count by `entity_name` collapses genuinely distinct resources: the switch1 and switch2 `networkVlan` allocations share `entity_name: "base_network_vlan_device"`, so this assertion can no longer detect an extra or leaked resource — including the `VPC_DOMAIN_ID`-created one that `base/delete.yaml` no longer cleans up (see comment there). It also treats `entity_name` as a unique identity, which the module itself documents as untrue (GAP-5 comment in `nd_manage_resource_manager_resources.py`: "Entity name alone is not unique: Resource Manager supports allocating multiple pools to the same entity."). Same pattern in `base/sanity.yaml` line 292.

Suggest reverting to plain `length` once the delete cleanup question is settled, or adding a comment reconciling the dedup with the GAP-5 non-uniqueness invariant.

(Noting this hunk predates the earlier approval — raising it only because it interacts directly with the `base/delete.yaml` removal above.)

allenrobel (commented):
### Re-review (post-approval changes)

Re-reviewed the changes pushed since the earlier approval. The only substantive post-approval change is the removed cleanup entry in `base/delete.yaml` (commit "[Jeet | issue-304] delete assertion fix"); the unit-test file changes are formatting-only. Two inline comments below — one on the removal itself, one on the related gathered/sanity count dedup it interacts with.

tests/integration/targets/nd_resource_manager/tests/nd/base/delete.yaml
jeetugangwar11: Hi @allenrobel , 

**1. Resource leak:** I'm creating a resource with 

- entity_name: "base_network_vlan_device"
 pool_type: "ID"
 pool_name: "networkVlan"
 scope_type: "device"
 switches:
     - "{{ ansible_switch2 }}"
 resource: "2451"

 and it's creating two resources in ND, one with ansible_switch1 and another one with ansible_switch2 (the Controller created one).

When I was adding the resource IDs from ansible_switch1 and ansible_switch2 to delete the resources, I encountered an issue with an invalid resource ID. I confirmed with Srivalli (from the ND team), and she mentioned to send only one resource ID which is part of ansible_switch1, and it will delete both resources.

I had missed removing VPC_DOMAIN_ID from the create resource, which I have now removed while addressing this comment.

**2. Count mismatch:** As I have removed VPC_DOMAIN_ID from create resources now count will be same in delete and create resources config.

**3. Issue reference:** This PR is having change for issue-304 as well as additional change for return status code from ND and test cases fixes.

tests/integration/targets/nd_resource_manager/tests/nd/base/gathered.yaml
jeetugangwar11: @allenrobel , I have made fix for this by 
```
| map('split', '~')
| map('sort')
| map('join', '~')
| unique
| sort
```

tests/integration/targets/nd_resource_manager/tests/nd/base/delete.yaml
jeetugangwar11: Hey @allenrobel , please let me know if I addressed your comment and if we can mark it as resolved, or if there's anything else I need to take care of.

allenrobel: Hi @jeetugangwar11 — thanks for the detailed explanation and the fix.

I've verified the changes in "[Jeet] Test cases fix | Removed resource creation with pool_name: VPC_DOMAIN_ID": the `VPC_DOMAIN_ID` create entry is removed across the whole target (delete, merge, gathered, and sanity), both `rm_test_config` and `rm_test_delete_config` are down to 14 entries, and `rm_expected_resource_count` is 14 in all four files — so the leak and the count mismatch from my original comment are both addressed. Thanks also for the context from Srivalli on the controller-created switch2 `networkVlan` twin being deleted via switch1's resource ID; the counts are consistent under that behavior.

Resolving this thread. One heads-up: the companion commit "[Jeet] Gathered assertion fix for entity Name" has an issue that will fail in the lab — see my comment on the `gathered.yaml` thread.

tests/integration/targets/nd_resource_manager/tests/nd/base/gathered.yaml
allenrobel: Hi @jeetugangwar11 — the normalization approach in "[Jeet] Gathered assertion fix for entity Name" is a genuine improvement: the `split('~') | sort | join('~')` canonicalization plus the equality check against `rm_expected_entity_names` catches unexpected or leaked entity names, which the old count-only dedup couldn't. Nice.

However, the first assertion references a variable that is never defined:

```yaml
- '(rm_gathered_entity_names | length) == (merge_result["diff"][0]["merged"] | length)'
```

Every task in this file uses `register: result`, and the gather task overwrites the create task's `result` before this assert runs — there is no `register: merge_result` anywhere in the target. This assert will fail with an undefined-variable error on every lab run, in both `base/gathered.yaml` and `base/sanity.yaml` (same line in each). It also suggests this exact revision hasn't been run against a live ND — could you re-run the `gathered` and `sanity` testcases after fixing?

Two easy fixes — either register the create task as `merge_result`, or drop the count assertion entirely and rely on the equality check (the expected-names comparison already implies the count). I'd lean toward the latter since it's one less moving part.

I've un-resolved this thread so the `merge_result` issue isn't missed — it's a blocker for approval since the testcase can't pass as committed.

jeetugangwar11: Thank you so much @allenrobel  for flagging this out. This assertion was meant to be removed after adding the normalization assertion, but I somehow missed it.

I have removed it now. I hope this should be fine now. could you please have a look again and resolve the comment if it looks good. Thanks

mikewiebe:
@jeetugangwar11 conflicts need to be resolved

tests/integration/targets/nd_resource_manager/tests/nd/base/gathered.yaml
allenrobel: Verified at head 14ce5961 — the `merge_result` assertion is gone from both `base/gathered.yaml` and `base/sanity.yaml`, and the normalization + equality check that remains evaluates correctly. Resolving this thread. I will hold the approval until the develop conflicts from #423 (`nd_manage_resource_manager.py`, `nd_manage_resource_manager_resources.py`, `test_manage_resource_manager.py`) are resolved, since that merge touches module code — ping me once that push is up.

allenrobel (approved):
### Re-approval after the #423 conflict resolution

Compared the PR's net diff at the last review (head "Update test_manage_resource_manager.py" era, 14ce5961) against the current head "[Jeet | Issue-304] Removed undefined assertion" (b55cfdc1), each taken against its own merge-base with develop. The PR's delta is byte-identical apart from one expected difference: the deleted flat `tests/nd/invalid_params.yaml` now also drops the gathered-filter validation tests that #423 added there, and those tests are present in all four per-fabric `invalid_params.yaml` files, so nothing is lost. The #423 changes in `nd_manage_resource_manager.py`, `nd_manage_resource_manager_resources.py`, and `test_manage_resource_manager.py` survived the rebase intact. All review threads are resolved and CI (sanity, units, galaxy-importer, build) is green on this head.

allenrobel:
Approved.  @jeetugangwar11 can you check if this PR closes #304 ?  If it does, can you add "Closes #304" to the PR description so that issue is closed automatically when this PR merges?

PR #486, opened 2026-08-04
Issue-406: Added changes for command timeouts  (NDA-34)

### LINKED ISSUE #406: Remove ND_TIMEOUT/module timeout override and use ANSIBLE_PERSISTENT_COMMAND_TIMEOUT
## Ansible Version

```shell
ansible [core 2.16.3]
python version = 3.12.13
jinja version = 3.1.6
libyaml = True
```

## Ansible Collection Versions

```shell
Collection  Version
----------  -------
cisco.nd    1.5.0
```

Reproduced on `develop` commit `9b68f916c77bed4b919d7ad4f5c0638bf36a043c`.

## Ansible Collection Module

```shell
All cisco.nd modules using the shared nd_argument_spec()/nd_argument_spec_v2()
and the cisco.nd.nd HTTPAPI plugin.
```

```shell
Nexus Dashboard 4.2.1
Build version 4.2.1.10
```

This is a client-side connection configuration issue and is not specific to an ND release.

## Cisco NX-OS Version

```shell
Not applicable
```

## Cisco ACI Version

```shell
Not applicable
```

## Expected Behavior

`ANSIBLE_PERSISTENT_COMMAND_TIMEOUT` should be the authoritative environment
variable for overriding Ansible's default 30-second persistent command timeout.

When a user sets:

```shell
export ANSIBLE_PERSISTENT_COMMAND_TIMEOUT=1000
```

the ND collection should preserve that value. The collection should not require
the additional `ND_TIMEOUT` variable or a module-level `timeout` argument.

## Actual Behavior

Every ND module receives `timeout: 30` from the shared argument specification
when no module timeout or `ND_TIMEOUT` is supplied. The HTTPAPI plugin then
writes that value to `persistent_command_timeout`, replacing the native Ansible
value supplied through `ANSIBLE_PERSISTENT_COMMAND_TIMEOUT`.

Current sequence:

```text
ANSIBLE_PERSISTENT_COMMAND_TIMEOUT=1000
        |
        v
Ansible initializes persistent_command_timeout=1000
        |
        v
Shared ND argument specification supplies timeout=30
        |
        v
ND HTTPAPI plugin sets persistent_command_timeout=30
```

## Ansible Playbook

```yaml
---
- name: Reproduce native persistent timeout override
  hosts: nd
  gather_facts: false

  tasks:
    - name: Query ND without a module-level timeout
      cisco.nd.nd_version:
        state: query
```

## Steps to Reproduce

```shell
unset ND_TIMEOUT
export ANSIBLE_PERSISTENT_COMMAND_TIMEOUT=1000

ansible-config dump --only-changed |
    grep PERSISTENT_COMMAND_TIMEOUT

ansible-playbook -i inventory.yml reproduce.yml -vvv
```

## Relevant Debug Output

```shell
PERSISTENT_COMMAND_TIMEOUT(env: ANSIBLE_PERSISTENT_COMMAND_TIMEOUT) = 1000
```

Despite that configuration, the shared argument specifications declare:

```python
timeout=dict(
    type="int",
    default=30,
    fallback=(env_fallback, ["ND_TIMEOUT"]),
)
```

The HTTPAPI plugin subsequently executes:

```python
if self.params.get("timeout") is not None:
    self.connection.set_option(
        "persistent_command_timeout",
        self.params.get("timeout"),
    )
```

Because the argument always defaults to `30`, the condition is always true and
the native Ansible setting is overwritten.

## Required Implementation

1. Remove `timeout` and the `ND_TIMEOUT` fallback from:
   - `plugins/module_utils/nd.py`
   - `plugins/module_utils/nd_v2.py`
2. Remove the HTTPAPI code that sets `persistent_command_timeout` from module parameters.
3. Remove `ND_TIMEOUT` and module-level `timeout` from the shared documentation fragment.
4. Update module documentation, examples, integration tests, and `module_defaults`
   to use `ANSIBLE_PERSISTENT_COMMAND_TIMEOUT` or `ansible_command_timeout`.
5. Leave resource-specific settings such as `verify.timeout` unchanged; those
   control internal query/retry behavior and are not connection timeouts.
6. Leave `ANSIBLE_PERSISTENT_CONNECT_TIMEOUT` unchanged; it controls persistent
   connection lifetime and is outside this issue.

## Compatibility and Migration

Removing the public module-level `timeout` argument is a breaking change. It
should be deprecated before removal or scheduled for the next major release.

Migration:

```yaml
# Before
module_defaults:
  cisco.nd.nd_manage_vpc_pair:
    timeout: 300

# After
vars:
  ansible_command_timeout: 300
```

Environment-based users should migrate from:

```shell
export ND_TIMEOUT=300
```

to:

```shell
export ANSIBLE_PERSISTENT_COMMAND_TIMEOUT=300
```

## Acceptance Criteria

- `ND_TIMEOUT` is no longer supported or documented.
- ND modules no longer reset `persistent_command_timeout`.
- The Ansible default remains 30 seconds when no native override is supplied.
- `ANSIBLE_PERSISTENT_COMMAND_TIMEOUT` and `ansible_command_timeout` work without
  also specifying a module timeout.
- Unit tests verify that the HTTPAPI plugin preserves the native connection option.
- Documentation and integration tests no longer use module-level `timeout`.
- Resource-specific options such as `verify.timeout` continue to work unchanged.
- A deprecation or major-version migration path covers existing playbooks using
  the module-level `timeout` argument.

## Existing Issue Check

GitHub reports 88 open issues. No open or closed issue matches `ND_TIMEOUT`,
`PERSISTENT_COMMAND_TIMEOUT`, `ansible_command_timeout`, or this root cause.

Related but non-overlapping:

- #385 tracks generic post-write verification controls.
- #349 tracks interface query fan-out and controller-side timeout risk.

jeetugangwar11:
## Related Issue(s)
Fixes #406 

## Proposed Changes
1. Remove timeout from nd_argument_spec() in nd_argument_specs.py. 
2. Remove the persistent_command_timeout assignment from nd.py
3. Remove timeout and ND_TIMEOUT from the shared documentation fragment.

Now we can use timeout by providing below ways:
1. By playbook vars:
vars:
  ansible_command_timeout: 300
2. By adding below props in inventory:
ansible_command_timeout=100
3. By exporting variable:
export ANSIBLE_PERSISTENT_COMMAND_TIMEOUT=300

## NOTE:
Variable precedence order - module reads from:
 1. Playbook facts
 2. Playbook vars
 3. Inventory
 4. Env varialble ANSIBLE_PERSISTENT_COMMAND_TIMEOUT

## Test Notes
I have added test case for the same. 
I tested above precedence order in local by adding logs.

4.3.1

NA

## Checklist


tests/integration/targets/nd_backup/tasks/main.yml
```diff
@@ -5,9 +5,9 @@
 # GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
 - name: Set vars
   ansible.builtin.set_fact:
+    ansible_command_timeout: 90 # Backup tasks commonly require 10-15 minutes or more.
```
mtarking: **Blocking:** `ansible_command_timeout` is a *connection* variable read by `ansible-connection` when the persistent socket is initialized. Setting it via `set_fact` after an ND task has already opened the persistent connection may not reconfigure the already-running socket, so the effective `persistent_command_timeout` can stay at the default 30s.

The old `timeout: 90` worked because the httpapi plugin called `connection.set_option("persistent_command_timeout", ...)` on every request (dynamic). That path is now gone, so timing here depends entirely on native connection-var resolution.

Please confirm empirically that a >30s operation in this target still gets the extended window. If not, move this to play-level `vars:` / inventory `[nd:vars]` (as the maintenance-mode and interface targets already do), which is resolved at connection setup and is reliable. Same concern applies to the other `set_fact: ansible_command_timeout` targets: `nd_backup_restore`, `nd_service`, `nd_service_instance`, `nd_rest/*`, and `nd_vpc_pair/*`.

tests/integration/targets/nd_vpc_pair/tasks/nd_vpc_pair_delete.yaml
```diff
@@ -8,9 +8,9 @@
 
 - name: DELETE - Set nd_info defaults
   ansible.builtin.set_fact:
+    ansible_command_timeout: 90
```
mtarking: **Scope regression:** `set_fact` persists for the remainder of the play and is per-host, whereas the replaced `module_defaults:`/inline `timeout:` was scoped to this specific task file. The block-scoped `vars:` conversions elsewhere in this PR (e.g. `nd_interface_port_channel_*`, `nd_maintenance_mode`) preserve the original scoping and are the better pattern.

Recommend converting these `set_fact` cases to block/task-scoped `vars:` for consistency and to avoid leaking a long timeout into unrelated later tasks. Also note this previously honored `nd_vpc_pair_module_timeout | default(600)` — the new hardcoded `90` drops that override knob and lowers the ceiling from 600→90; confirm that's intended.

tests/unit/test_httpapi.py
```diff
@@ -0,0 +1,24 @@
...
+    }
+    connection.get_option.side_effect = options.get
+    connection.send.return_value = (None, '{"ok": true}')
+    httpapi = HttpApi(connection)
+    httpapi.get_option = MagicMock(return_value="DefaultAuth")
+    httpapi.params = {"timeout": 30}
+
+    httpapi.send_request("GET", "/api/v1/test")
+
+    assert connection.get_option("persistent_command_timeout") == 1000
```
mtarking: This assertion is tautological: `connection.get_option` is wired to `options.get` via `side_effect`, so it returns `1000` regardless of the plugin's behavior — it doesn't prove anything. The meaningful check is the next line (that `set_option` was never called with `persistent_command_timeout`). Suggest dropping this line or replacing it with something that verifies the plugin left the option untouched.

Also: this new file has no GPL/copyright header, unlike the other test files — `ansible-test sanity` will likely flag a missing license/copyright header. Please add one at the top.

plugins/module_utils/orchestrators/vrf_attachment_manager.py
```diff
@@ -849,11 +849,6 @@ def wait_for_vrfs_delete_ready(
             time.sleep(self.delete_wait_delay)
 
     def _delete_wait_timeout(self, module_args: dict, item_count: int) -> int:
-        explicit_timeout = module_args.get("timeout")
```
mtarking, line 852: This removes a real user override, not just a connection timeout: previously an explicit module `timeout` raised the internal delete-wait ceiling. That's *internal query/retry* behavior — arguably the same category as `verify.timeout`, which #406 said to leave unchanged. Removing the argspec key forces this branch to change (it'd always be `None`), so the edit is justified, but please confirm the adaptive formula `min(900, 30 + extra_chunks * delete_wait_delay)` actually covers the large-VRF-delete cases this override existed for. The new unit test asserts the formula but not the dropped-override intent.

plugins/module_utils/nd_argument_specs.py
```diff
@@ -32,7 +32,6 @@ def nd_argument_spec() -> dict[str, Any]:
         "username": {"type": "str", "fallback": (env_fallback, ["ND_USERNAME", "ANSIBLE_NET_USERNAME"])},
         "password": {"type": "str", "required": False, "no_log": True, "fallback": (env_fallback, ["ND_PASSWORD", "ANSIBLE_NET_PASSWORD"])},
         "output_level": {"type": "str", "default": "normal", "choices": ["debug", "info", "normal"], "fallback": (env_fallback, ["ND_OUTPUT_LEVEL"])},
-        "timeout": {"type": "int", "default": 30, "fallback": (env_fallback, ["ND_TIMEOUT"])},
```
mtarking, line 35: Removing the argspec key is the right fix. One verification request: issue #406 explicitly named `plugins/module_utils/nd.py` and `nd_v2.py`. This PR changes `nd_argument_specs.py` (the current shared source) but I don't see those files touched. Please grep the tree for `timeout` and `ND_TIMEOUT` to confirm nothing else still reads the removed param — e.g. `nd_v2.py`, action plugins, or `sender_nd.set_params()` — so it can't silently resolve to `None`/absent at runtime.

mtarking (changes requested):
Thanks for tackling this — the core fix is exactly right and minimal: dropping the argspec `timeout` default that unconditionally clobbered `persistent_command_timeout`, and stopping the httpapi plugin from writing that option at all. `test_nd_argument_specs.py` (`"timeout" not in spec`) and the doc-fragment removal are good regression guards, and `verify.timeout` / `persistent_connect_timeout` are correctly left untouched.

Requesting changes on a few items before merge (details inline):

1. **`set_fact: ansible_command_timeout` may not reconfigure an already-open persistent socket** — it's a connection var resolved at connection setup. The old `timeout:` applied per-request dynamically; this doesn't. Please verify the migrated targets still get the extended window, or move to play-level `vars:` / inventory `[nd:vars]` (as the interface and maintenance-mode targets do). Affects `nd_backup`, `nd_backup_restore`, `nd_service`, `nd_service_instance`, `nd_rest/*`, `nd_vpc_pair/*`.
2. **Scope regression** — `set_fact` is play-lifetime and per-host; prefer block-scoped `vars:` to match the original `module_defaults` scoping. Also confirm the `nd_vpc_pair` change from `nd_vpc_pair_module_timeout | default(600)` to a hardcoded `90` is intended.
3. **Verify no residual `timeout` / `ND_TIMEOUT` readers** — #406 named `nd.py`/`nd_v2.py`; please grep to confirm nothing else (e.g. `nd_v2.py`, `sender_nd.set_params()`, action plugins) still reads the removed param.

Non-blocking:
4. The `vrf_attachment_manager._delete_wait_timeout` removal drops a user override of an internal wait ceiling — confirm the adaptive formula covers the large-delete cases it existed for.
5. `test_httpapi.py` has a tautological assertion and is missing a license/copyright header (sanity may flag it).

Once 1–3 are addressed this looks ready. Nice, well-scoped change overall.

plugins/module_utils/orchestrators/vrf_attachment_manager.py
mtarking: cc @AKDRG

tests/unit/module_utils/test_nd_argument_specs.py
```diff
@@ -33,7 +33,7 @@ def test_nd_argument_specs_00000() -> None:
 
     - The key set matches the historical spec from nd.py exactly
```
allenrobel: This bullet ("The key set matches the historical spec from nd.py exactly") is now stale: with `timeout` removed from the asserted key set (and the bullet two lines below updated to say the option is absent), the key set intentionally no longer matches the historical nd.py spec exactly. Suggest rewording, e.g. "The key set matches the historical spec from nd.py, minus the removed `timeout` option".

allenrobel (commented):
### Code review

One docstring-consistency finding, inline below.

plugins/module_utils/orchestrators/vrf_attachment_manager.py
AKDRG: The removal of explicit_timeout is acceptable. However, the adaptive formula needs more tuning and seems insufficient when scale increases. I have opened https://github.com/CiscoDevNet/ansible-nd/issues/492 for tracking and fixing it.

tests/integration/targets/nd_vpc_pair/tasks/nd_vpc_pair_delete.yaml
jeetugangwar11: Hi @mtarking , I have made the changes, please have a look once.

tests/integration/targets/nd_backup/tasks/main.yml
jeetugangwar11: Hi @mtarking , I have made the changes, please have a look once.

tests/unit/test_httpapi.py
jeetugangwar11: Hi @mtarking , I have made the changes, please have a look once.

plugins/module_utils/orchestrators/vrf_attachment_manager.py
jeetugangwar11: Hi @mtarking , as these changes are acceptable, so resolving this comment.

plugins/module_utils/nd_argument_specs.py
jeetugangwar11: Hi @mtarking , I have made the changes, please have a look once.

tests/unit/module_utils/test_nd_argument_specs.py
jeetugangwar11: Hi @allenrobel , I have made the changes, please have a look once.

sivakasi-cisco (approved):
Changes look good

mtarking (approved):

PR #491, opened 2026-08-06
Address fabric module issues and fix save without switches issue (NDA-35)

### LINKED ISSUE #272: Add frozen=True to class_name fields on pre-existing endpoints in manage_fabrics.py
## Summary

The `class_name` field on six endpoint classes in `plugins/module_utils/endpoints/v1/manage/manage_fabrics.py` is missing `frozen=True`, while every newer endpoint file in the package (`manage_interfaces.py`, `manage_switches.py`) sets it consistently. This is a small consistency cleanup — the `Literal[...]` annotation already constrains the value at type-check time, but `frozen=True` also blocks reassignment at runtime, making the class identity marker truly immutable.

## Affected fields

In `plugins/module_utils/endpoints/v1/manage/manage_fabrics.py`:

- `EpManageFabricsGet.class_name` (line 170)
- `EpManageFabricsListGet.class_name` (line 282)
- `EpManageFabricsPost.class_name` (line 344)
- `EpManageFabricsPut.class_name` (line 398)
- `EpManageFabricsDelete.class_name` (line 442)
- `EpManageFabricsSummaryGet.class_name` (line 487)

`EpManageFabricsDeploymentFreezeGet` (line 540) was just updated to `frozen=True` in commit 23fd3e7 on the `nd_interface_loopback` branch (PR #220), establishing the in-file precedent.

## Proposed change

Add `frozen=True` as the second keyword argument to each `Field(...)` call, matching the pattern below:

```python
class_name: Literal["EpManageFabricsGet"] = Field(
    default="EpManageFabricsGet", frozen=True, description="Class name for backward compatibility"
)
```

## Acceptance

- [ ] All six `class_name` fields use `frozen=True`
- [ ] `python -m pytest tests/unit/module_utils/endpoints/test_endpoints_api_v1_manage_fabrics.py` passes
- [ ] No other behavior changes

## Context

Originally surfaced during code review of PR #220 (`nd_interface_loopback`). The PR's review item #3 noted the inconsistency; we addressed only the endpoint added by that PR (`EpManageFabricsDeploymentFreezeGet`) so the change stayed scoped to the new branch. This issue tracks bringing the rest of `manage_fabrics.py` in line.

### LINKED ISSUE #292: _EpManageFabricsBase does not URL-encode fabric_name in path
## Summary

`_EpManageFabricsBase.path` in `plugins/module_utils/endpoints/v1/manage/manage_fabrics.py` appends `fabric_name` to the path segments **without URL-encoding**, unlike the sibling endpoint files which wrap it in `urllib.parse.quote(..., safe="")`.

## Affected Code

`plugins/module_utils/endpoints/v1/manage/manage_fabrics.py:113-115`

```python
segments = ["fabrics"]
if self.fabric_name is not None:
    segments.append(self.fabric_name)   # raw, not URL-encoded
```

This base class is inherited by every fabrics endpoint, so all of these are affected:

- `EpManageFabricsGet`
- `EpManageFabricsPut`
- `EpManageFabricsDelete`
- `EpManageFabricsSummaryGet`
- `EpManageFabricsDeploymentFreezeGet`

(`EpManageFabricsListGet` and `EpManageFabricsPost` set `_require_fabric_name = False`, so they are not affected.)

## Inconsistency With Sibling Endpoints

The other manage endpoints that embed `fabric_name` in the path **do** encode it:

- `manage_interfaces.py:92` — `quote(self.fabric_name, safe="")`
- `manage_interfaces.py:324,376,428` — `quote(self.fabric_name, safe="")`
- `manage_switches.py:77,184` — `quote(self.fabric_name, safe="")`

`manage_fabrics.py` does not import `quote` at all.

## Impact

A `fabric_name` containing characters that require percent-encoding (space, `/`, `#`, `%`, `?`, etc.) produces a malformed request path. A `/` in particular would be interpreted as an extra path segment rather than part of the fabric name, causing the request to hit the wrong endpoint.

## Suggested Fix

Encode `fabric_name` in `_EpManageFabricsBase.path` to match the sibling endpoints:

```python
from urllib.parse import quote

segments = ["fabrics"]
if self.fabric_name is not None:
    segments.append(quote(self.fabric_name, safe=""))
```

## ND Version

4.2.1

### LINKED ISSUE #490: Fabric modules need check for switches in fabric before config save allowed
### Ansible Version

```shell
N/A
```

```shell
N/A
```

### Ansible Collection Module

```shell

```

```shell
4.2.1
```

### Cisco NX-OS Version

```shell

```

### Cisco ACI Version

```shell

```

### Expected Behavior

```shell
Fabric modules need check for switches in fabric before config save allowed.
```

### Actual Behavior

```shell
Fails day0 with save as true.
```

### Ansible Playbook

```yaml
N/A
```

### Steps to Reproduce

```shell
N/A
```

### Relevant Debug Output

```shell

```

mtarking:
## Related Issue(s)

Resolves #292  
Resolves #272  
Resolves #490 

## Proposed Changes

`_EpManageFabricsBase.path` appended `fabric_name` to the path segments without URL-encoding, unlike the sibling endpoint files (`manage_interfaces.py`, `manage_switches.py`) which wrap it in `quote(..., safe="")`. A `fabric_name` containing characters that require percent-encoding (space, `/`, `#`, `%`, `?`, etc.) produced a malformed request path — a `/` in particular would be interpreted as an extra path segment and hit the wrong endpoint.

- Added `from urllib.parse import quote` to `plugins/module_utils/endpoints/v1/manage/manage_fabrics.py`.
- `_EpManageFabricsBase.path` now appends `quote(self.fabric_name, safe="")`. Because the base class is inherited by every fabrics endpoint that embeds a fabric name in the path, this fixes them all: `EpManageFabricsGet`, `EpManageFabricsPut`, `EpManageFabricsDelete`, `EpManageFabricsSummaryGet`, `EpManageFabricsDeploymentFreezeGet`, and the four `EpManageFabricsMembers*` / `EpManageFabricsMemberCandidatesGet` endpoints. (`EpManageFabricsListGet` and `EpManageFabricsPost` set `_require_fabric_name = False` and are unaffected.)
- Also encoded `fabric_name` in `EpManageFabricConfigDeployPost.path`, which overrides `path` and had the same unencoded bug (not called out in the issue).

This is backward compatible: normal fabric names (alphanumeric, `-`, `_`) are unchanged by `quote`.

## Test Notes

- Added two unit tests to `tests/unit/module_utils/endpoints/test_endpoints_api_v1_manage_fabrics.py` asserting that reserved characters (`/`, space, `#`) in `fabric_name` are percent-encoded in both the base path builder and the `EpManageFabricConfigDeployPost` path override.
- Full module passes (41 existing + 2 new = 43 passed):

4.2.1

* [ ] analyze
* [ ] infa
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

### DIFF
--- a/plugins/module_utils/endpoints/v1/manage/manage_fabrics_switches.py
+++ b/plugins/module_utils/endpoints/v1/manage/manage_fabrics_switches.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 __author__ = "Akshayanat C S"
 
 from typing import Literal
+from urllib.parse import quote
 
 from ansible_collections.cisco.nd.plugins.module_utils.enums import HttpVerbEnum
 from ansible_collections.cisco.nd.plugins.module_utils.endpoints.mixins import (
@@ -120,7 +121,7 @@ class _EpManageFabricsSwitchesBase(FabricNameMixin, NDEndpointBaseModel):
         """Build the base endpoint path."""
         if self.fabric_name is None:
             raise ValueError("fabric_name must be set before accessing path")
-        return BasePath.path("fabrics", self.fabric_name, "switches")
+        return BasePath.path("fabrics", quote(self.fabric_name, safe=""), "switches")
 
 
 class EpManageFabricsSwitchesGet(_EpManageFabricsSwitchesBase):
--- a/plugins/module_utils/orchestrators/config_actions_mixin.py
+++ b/plugins/module_utils/orchestrators/config_actions_mixin.py
@@ -32,6 +32,8 @@ Or use the convenience method to process a batch::
 
 from __future__ import annotations
 
+import logging
+
 from ansible_collections.cisco.nd.plugins.module_utils.endpoints.v1.manage.manage_fabrics_actions_config_save import (
     EpFabricConfigSavePost,
 )
@@ -46,6 +48,8 @@ from ansible_collections.cisco.nd.plugins.module_utils.endpoints.v1.manage.manag
 )
 from ansible_collections.cisco.nd.plugins.module_utils.enums import OperationType
 from ansible_collections.cisco.nd.plugins.module_utils.orchestrators.types import ResponseType
+
+logger = logging.getLogger("nd.ConfigActionsMixin")
 
 
 class ConfigActionsMixin:
@@ -85,6 +89,21 @@ class ConfigActionsMixin:
             verb=ep.verb,
             operation_type=OperationType.UPDATE,
         )
+
+    def _fabric_has_switches(self, fabric_name: str) -> bool:
+        """Return True if the fabric currently has any switches on the controller.
+
+        ND rejects configSave/deploy on a switchless fabric ("Fabric ... cannot be
+        deployed without any switches"), so callers must skip save when this is False.
+        """
+        ep = EpManageFabricsSwitchesGet(fabric_name=fabric_name)
+        result = self._request(
+            path=ep.path,
+            verb=ep.verb,
+            not_found_ok=True,
+            operation_type=OperationType.QUERY,
+        )
+        return bool(result.get("switches")) if result else False
 
     def config_deploy(self, fabric_name: str, deploy_type: str = "global") -> ResponseType:
         """Deploy fabric configuration.
@@ -216,6 +235,13 @@ class ConfigActionsMixin:
             return
 
         for fabric_name in fabric_names:
+            # ND rejects save/deploy on a switchless fabric. Skip (not an error) so
+            # a day-0 fabric create — or the fabric step on a first run, before the
+            # switches step adds switches — does not fail. Once switches exist,
+            # saves (including idempotent fabric-setting updates) and deploys proceed.
+            if not self._fabric_has_switches(fabric_name):
+                logger.info("Skipping config save/deploy for '%s': fabric has no switches.", fabric_name)
+                continue
             if save:
                 self.config_save(fabric_name)
             if deploy:
--- a/plugins/module_utils/orchestrators/config_actions_mixin.py
+++ b/plugins/module_utils/orchestrators/config_actions_mixin.py
@@ -32,8 +32,6 @@ Or use the convenience method to process a batch::
 
 from __future__ import annotations
 
-import logging
-
 from ansible_collections.cisco.nd.plugins.module_utils.endpoints.v1.manage.manage_fabrics_actions_config_save import (
     EpFabricConfigSavePost,
 )
@@ -48,8 +46,6 @@ from ansible_collections.cisco.nd.plugins.module_utils.endpoints.v1.manage.manag
 )
 from ansible_collections.cisco.nd.plugins.module_utils.enums import OperationType
 from ansible_collections.cisco.nd.plugins.module_utils.orchestrators.types import ResponseType
-
-logger = logging.getLogger("nd.ConfigActionsMixin")
 
 
 class ConfigActionsMixin:
@@ -90,11 +86,13 @@ class ConfigActionsMixin:
             operation_type=OperationType.UPDATE,
         )
 
-    def _fabric_has_switches(self, fabric_name: str) -> bool:
-        """Return True if the fabric currently has any switches on the controller.
+    def _get_fabric_switches(self, fabric_name: str) -> list[dict]:
+        """Return the fabric's switches from the controller (empty list if none).
 
         ND rejects configSave/deploy on a switchless fabric ("Fabric ... cannot be
-        deployed without any switches"), so callers must skip save when this is False.
+        deployed without any switches"), so callers skip save/deploy when this is
+        empty. Fetched once per fabric per run and passed to the deploy path so the
+        switches endpoint is queried at most once per fabric.
         """
         ep = EpManageFabricsSwitchesGet(fabric_name=fabric_name)
         result = self._request(
@@ -103,15 +101,18 @@ class ConfigActionsMixin:
             not_found_ok=True,
             operation_type=OperationType.QUERY,
         )
-        return bool(result.get("switches")) if result else False
+        return result.get("switches", []) if result else []
 
-    def config_deploy(self, fabric_name: str, deploy_type: str = "global") -> ResponseType:
+    def config_deploy(self, fabric_name: str, deploy_type: str = "global", switches: list[dict] | None = None) -> ResponseType:
         """Deploy fabric configuration.
 
         Args:
             fabric_name: Name of the fabric to deploy.
             deploy_type: ``"global"`` for fabric-wide deploy, ``"switch"`` for
                 switch-level deploy targeting only out-of-sync switches.
+            switches: Optional pre-fetched switch list (from
+                ``_get_fabric_switches``) reused for switch-level deploy to avoid a
+                redundant switches query. Fetched on demand when None.
 
         Returns:
             API response data, or None if no switches need deployment
@@ -124,7 +125,7 @@ class ConfigActionsMixin:
         if deploy_type == "global":
             return self._deploy_global(fabric_name)
         elif deploy_type == "switch":
-            return self._deploy_switches(fabric_name)
+            return self._deploy_switches(fabric_name, switches)
         else:
             raise ValueError(f"Invalid deploy_type '{deploy_type}'. Must be 'global' or 'switch'.")
 
@@ -137,16 +138,18 @@ class ConfigActionsMixin:
             operation_type=OperationType.UPDATE,
         )
 
-    def _deploy_switches(self, fabric_name: str) -> ResponseType:
+    def _deploy_switches(self, fabric_name: str, switches: list[dict] | None = None) -> ResponseType:
         """Deploy to switches that need configuration deployment.
 
-        Queries the fabric's switches, identifies those with a
-        ``configSyncStatus`` that is not ``inSync``, and deploys
-        to them via the switchActions/deploy endpoint.
+        Identifies switches with a ``configSyncStatus`` that is not ``inSync`` and
+        deploys to them via the switchActions/deploy endpoint. When ``switches`` is
+        provided it is reused; otherwise the switch list is fetched.
 
         Returns None if all switches are already in sync.
         """
-        switch_ids = self._get_switches_needing_deploy(fabric_name)
+        if switches is None:
+            switches = self._get_fabric_switches(fabric_name)
+        switch_ids = self._filter_switches_needing_deploy(switches)
         if not switch_ids:
             return None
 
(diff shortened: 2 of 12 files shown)


plugins/module_utils/orchestrators/config_actions_mixin.py
```diff
@@ -216,6 +235,13 @@ def execute_config_actions(
             return
 
         for fabric_name in fabric_names:
+            # ND rejects save/deploy on a switchless fabric. Skip (not an error) so
+            # a day-0 fabric create — or the fabric step on a first run, before the
+            # switches step adds switches — does not fail. Once switches exist,
+            # saves (including idempotent fabric-setting updates) and deploys proceed.
+            if not self._fabric_has_switches(fabric_name):
```
allenrobel: This new unconditional GET breaks the existing suite in `tests/unit/module_utils/orchestrators/test_config_actions_mixin.py` (not in this PR's diff). The mock `Sender`/`ResponseGenerator` replays fixtures strictly in call order, so the extra switches GET consumes fixtures meant for the save/deploy calls. Running that suite against this branch:

- `test_execute_save_and_global_deploy` fails (`assert 2 == 4`)
- `test_execute_save_and_switch_deploy` fails (`assert 1 == 3`)
- `test_execute_save_only` passes for the wrong reason: both fabrics are classified as switchless (the switches GET consumes the save-POST fixture), so no save is issued, yet the task count coincidentally still matches.

The fixtures/comments in that test file need updating to account for the new per-fabric switches GET.

plugins/module_utils/orchestrators/config_actions_mixin.py
```diff
@@ -86,6 +90,21 @@ def config_save(self, fabric_name: str) -> ResponseType:
...
+        deployed without any switches"), so callers must skip save when this is False.
+        """
+        ep = EpManageFabricsSwitchesGet(fabric_name=fabric_name)
+        result = self._request(
+            path=ep.path,
+            verb=ep.verb,
+            not_found_ok=True,
+            operation_type=OperationType.QUERY,
+        )
+        return bool(result.get("switches")) if result else False
```
allenrobel: With `deploy_type="switch"`, a single `execute_config_actions()` iteration now fetches `/fabrics/{fabric_name}/switches` twice: once here, then again in `_get_switches_needing_deploy()`, with no caching between them. Team convention is to fetch each controller resource at most once per module run and cache/pass the result (same concern mikewiebe raised on the Resource Manager PR #333).

Suggest fetching the switch list once per fabric in `execute_config_actions()` and reusing it for both the has-switches check and the needing-deploy filter - that restructuring would also resolve the test breakage flagged at line 242.

plugins/module_utils/orchestrators/config_actions_mixin.py
```diff
@@ -86,6 +90,21 @@ def config_save(self, fabric_name: str) -> ResponseType:
             operation_type=OperationType.UPDATE,
         )
 
+    def _fabric_has_switches(self, fabric_name: str) -> bool:
+        """Return True if the fabric currently has any switches on the controller.
+
+        ND rejects configSave/deploy on a switchless fabric ("Fabric ... cannot be
+        deployed without any switches"), so callers must skip save when this is False.
+        """
+        ep = EpManageFabricsSwitchesGet(fabric_name=fabric_name)
```
allenrobel, line 97: `EpManageFabricsSwitchesGet` builds its path in `manage_fabrics_switches.py` without `quote()`, so a fabric name containing `/`, space, or `#` still produces a malformed path through this new call site - the same bug class this PR fixes in `manage_fabrics.py` for #292. The same applies to `EpFabricConfigSavePost` (`manage_fabrics_actions_config_save.py`) and `EpFabricDeployPost` (`manage_fabrics_actions_deploy.py`) used by `config_save()`/`config_deploy()` in this mixin. Worth extending the `quote(self.fabric_name, safe="")` fix to those endpoint files here (or in a fast follow-up) so the mixin's whole execution path is covered.

allenrobel (commented):
### Code review

Reviewed the fabric_name URL-encoding fix and the switchless-fabric save/deploy skip. Four inline comments below, plus one finding on an unchanged line that can't be anchored in the diff:

1. The `execute_config_actions()` docstring ([lines 220-231](https://github.com/CiscoDevNet/ansible-nd/blob/5877776d708cf9755987eceb1e192fa8918b4867/plugins/module_utils/orchestrators/config_actions_mixin.py#L219-L232)) still documents unconditional validate/save/deploy per fabric and doesn't mention the new switchless-fabric skip. Worth a one-line update so the docstring matches the new behavior.

plugins/module_utils/orchestrators/config_actions_mixin.py
```diff
@@ -216,6 +235,13 @@ def execute_config_actions(
             return
 
         for fabric_name in fabric_names:
+            # ND rejects save/deploy on a switchless fabric. Skip (not an error) so
+            # a day-0 fabric create — or the fabric step on a first run, before the
+            # switches step adds switches — does not fail. Once switches exist,
+            # saves (including idempotent fabric-setting updates) and deploys proceed.
+            if not self._fabric_has_switches(fabric_name):
+                logger.info("Skipping config save/deploy for '%s': fabric has no switches.", fabric_name)
```
allenrobel: `logger.info()` is invisible in a normal playbook run - the collection disables logging output unless `ND_LOGGING_CONFIG` is set (`common/log.py`) - so a user whose day-0 save/deploy is skipped gets no signal in the module output. Suggest `self.rest_send.warn(...)` instead: its docstring covers exactly this case (surfacing non-fatal preflight skips without converting them into module errors), and there's precedent in `base_interface.py` for the check-mode capability-preflight skip.

plugins/module_utils/orchestrators/config_actions_mixin.py
mtarking: Fixed. Resolving the double-fetch (comment below) meant the per-fabric switches GET is now issued exactly once and reused, so I updated `test_config_actions_mixin.py` accordingly: `test_execute_save_and_global_deploy` now expects 6 calls (switches + save + deploy × 2 fabrics), `test_execute_save_only` expects 4 (switches + save × 2), and `test_execute_save_and_switch_deploy` fixtures were reordered to switches → save → switchActions/deploy (still 3). `test_execute_save_only` now passes for the right reason. Added `test_execute_skips_switchless_fabric` to cover the empty-switches skip + warning path. Full mixin suite is green.

plugins/module_utils/orchestrators/config_actions_mixin.py
mtarking: Done. Replaced `_fabric_has_switches()` with `_get_fabric_switches()` which returns the switch list. `execute_config_actions()` now fetches it once per fabric and threads it into `config_deploy(..., switches=switches)` → `_deploy_switches(switches=...)`, so the switches endpoint is queried at most once per fabric per run. Extracted the sync-status filter into `_filter_switches_needing_deploy()`. The public `config_deploy()`/`_deploy_switches()` signatures stay backward compatible — `switches` defaults to `None` and falls back to an on-demand fetch for any other caller.

plugins/module_utils/orchestrators/config_actions_mixin.py
mtarking: Extended the fix in this PR. Added `quote(self.fabric_name, safe="")` to `EpManageFabricsSwitchesGet` (`manage_fabrics_switches.py`), `EpFabricConfigSavePost` (`manage_fabrics_actions_config_save.py`), and `EpFabricDeployPost` (`manage_fabrics_actions_deploy.py`). I also encoded the `switchActions` base path (`manage_fabrics_switchactions.py`), which `EpManageFabricsSwitchActionsDeployPost` uses in the switch-level deploy path. Added percent-encoding unit tests for all four so the mixin's whole execution path is covered.

plugins/module_utils/orchestrators/config_actions_mixin.py
mtarking: Switched to `self.rest_send.warn(...)` so the skip is surfaced in module output, matching the `base_interface.py` check-mode precedent. Removed the now-unused `import logging` and module-level `logger`. The new `test_execute_skips_switchless_fabric` asserts the warning is emitted.

mtarking:
@allenrobel Re: the `execute_config_actions()` docstring — good catch, updated it to document the switchless-fabric skip (surfaced as a warning via `rest_send.warn`, not an error) and the single per-fabric switch fetch. Also updated the `config_deploy()`/`_deploy_switches()` docstrings for the new optional pre-fetched `switches` argument. All four inline threads addressed in the latest commit; full endpoints + orchestrators unit suites pass (1304).

plugins/module_utils/orchestrators/config_actions_mixin.py
allenrobel: Verified the fixes on the updated branch - the mixin suite and all the new percent-encoding tests pass, and the switches endpoint is now fetched once per fabric. Nice work.

One follow-up this change surfaces: the full unit suite now fails in `tests/unit/module_utils/test_manage_vpc_pair_actions.py` (`test_manage_vpc_pair_actions_00090_resolve_fabric_type_uses_management_type_and_encoded_endpoint`):

```text
assert '/api/v1/manage/fabrics/fab%252Fname' == '/api/v1/manage/fabrics/fab%2Fname'
```

`manage_vpc_pair/runtime_endpoints.py` line 106 pre-quotes the fabric name before handing it to `EpManageFabricsGet` - a caller-side workaround from before the endpoint encoded. With the endpoint-side `quote()` added in this PR, the name is now double-encoded (`%2F` -> `%252F`), which would send a malformed request for any fabric name with reserved characters. The fix is to drop the caller-side `quote()` there and pass the raw name, since the endpoint now owns encoding. `manage_vpc_pair/query.py` line 218 is unaffected - it hand-builds its path without the endpoint class, and I found no other pre-quoting callers. Re-running that test file should confirm.

mtarking: Good catch. Reverted.

allenrobel: Confirmed - verified the revert and ran the full unit suite (3867 tests) against the latest commit: all green, including test_manage_vpc_pair_actions_00090, which now produces the correctly single-encoded path. All findings from this review are addressed. Thanks Matt!

allenrobel (approved):

plugins/module_utils/endpoints/v1/manage/manage_fabrics_switches.py
AKDRG, line 325: Shouldn't we do the same here?
return BasePath.path("fabrics", quote(self.fabric_name, safe=""), "switches", quote(self.switch_sn, safe=""), "actions")

AKDRG: I think self.switch_sn need not be enclosed in the safe quotes though.

PR #495, opened 2026-08-07
[ND 4.2+] Add Interface Groups Module

### LINKED ISSUE #326: [ND 4.2+] Add Interface Groups Module

```shell
Add Interface Groups Module
```

manage

```shell

```

```yaml
TBD
```

### Ansible Version

```shell

```

```shell

```

```shell
4.2.1
```

### Cisco NX-OS Version

```shell

```

### Cisco ACI Version

```shell

```

nikhilsrikrishna:
## Related Issue(s)

Closes #326

## Proposed Changes

Adds the `cisco.nd.nd_manage_interface_group` module for managing Interface Groups in Cisco Nexus Dashboard fabrics.

### Module capabilities

- Supports `merged`, `replaced`, `overridden`, `deleted`, and `gathered` states.
- Supports all Interface Group types: `any`, `ethernetCustom`, `ethernetWithPolicy`, `ethernetWithoutPolicy`, `portChannel`, and `vpc`.
- Supports network associations and switch-interface membership.
- Supports predictive check mode and standard `NDOutput`, including verbosity-based request diagnostics.
- Supports resource-level interface deployment, switch-level deployment, and staging changes with `deploy: false`.
- Clears network and interface associations before deleting associated groups.
- Supports paginated gathered queries with local filtering for nested networks, switch interfaces, normalized group types, and custom-template configuration.
- Extends the shared state machine with an optional pre-reconciliation mutation hook for controller operations that require ordered prerequisite changes.

### State behavior

- `merged` is additive for networks, switch interfaces, Ethernet attributes, and custom-template inputs. Existing values omitted from the task are preserved.
- `replaced` authoritatively manages explicitly supplied fields while preserving omitted network and interface collections. Explicit empty lists clear those collections.
- `overridden` applies the requested authoritative configuration and removes Interface Groups omitted from the task.
- Interface moves under `replaced` and `overridden` are validated before mutation and performed in the required remove-then-add order.
- An `any` group can contain Ethernet, port-channel, and vPC members. Member additions are submitted in homogeneous batches where required by the controller.

The additive nested-list handling is implemented specifically for Interface Groups so that `state=merged` does not remove existing networks or members omitted from the task.

### Shared state-machine extension

This PR adds an optional `prepare_mutations()` hook to the shared orchestrator state machine.

The existing state-machine flow performs validation and then proceeds directly to normal create/update reconciliation. Interface Groups require an additional ordered step when moving an interface between groups:

1. Validate the complete requested configuration and move.
2. Remove the interface from its current Interface Group.
3. Add the interface to the target Interface Group.
4. Deploy the affected interfaces or switches when requested.

Adding the interface to the target first can fail because the controller still considers it owned by the source group. Performing the removal during normal target reconciliation is also too late.

The new hook runs after all preflight validation succeeds but before create/update reconciliation begins. This ensures that no prerequisite mutation is attempted until the complete task has passed validation.

The base `NDBaseOrchestrator` implementation is a no-op, so existing modules retain their current behavior. Interface Groups override the hook to perform the required source-group detachment.

In check mode, the hook does not call the controller. It updates only the in-memory state so that the predicted `after`, `changed`, and subsequent reconciliation results represent the complete move.

The hook can also support future modules that require ordered prerequisite mutations before their normal resource updates, without requiring those modules to replace or duplicate the shared reconciliation flow.

### Validation and error handling

- Uses Pydantic models for argument specification, normalization, and state-specific validation.
- Rejects duplicate group names and duplicate member ownership within the same task.
- Validates member-name kinds against the selected Interface Group type.
- Verifies that newly referenced networks exist before making changes.
- Validates custom Ethernet templates as user-defined Ethernet policy templates eligible for shared-interface editing.
- Validates custom-template keys, required values, basic types, and enumerated values against the referenced template.
- Handles per-item results from bulk create, delete, and deployment responses. A successful HTTP response is not treated as successful when an individual item reports failure.
- Recognizes HTTP 207 in the legacy ND response helper so the response body reaches the resource-specific per-item validator.
- Preserves actionable controller and validation error messages with `changed: false` when no mutation occurred.

### Deployment behavior

- Resource-level deployment deploys only affected interfaces.
- Switch-level deployment deploys pending configuration for the affected switches and can therefore include changes outside the current task.
- Switch-level deployment on vPC-paired switches depends on #573. NDFC can automatically include the vPC peer in the deployment response. When the requested switch has pending commands but its peer is already in sync, the peer is returned with `status: notExecuted` and `message: No Commands to execute`. PR #573 classifies this exact result as a successful no-op while continuing to report genuine per-switch failures.
- Resource-level deployment does not deploy referenced networks. The module returns a warning directing users to `cisco.nd.nd_manage_networks` when network deployment must be handled separately.
- `deploy: false` stages the Interface Group intent without issuing a deployment request.

### Integration-test harness

The Interface Groups integration suite follows the check-mode, apply, idempotency, and read-only controller-validation pattern proposed in #337.

Because #337 is still in flight, this PR includes a lightweight interim version of the harness containing the functionality required by the Interface Groups suite. It will be aligned with the final shared implementation if #337 changes before merging.

For negative scenarios, this suite provides a lightweight task-level equivalent of the proposed per-phase message validation by asserting that the returned failure message contains the expected text. The reusable `msg_contains` harness option and non-GET call-count expectation proposed on #337 are not introduced as shared harness features in this PR.

## Test Notes

Unit-test coverage includes:

- Interface Group endpoint construction and payloads
- Pydantic input and response models
- Type-specific member validation
- Additive merge and authoritative replacement behavior
- Custom-template eligibility and input validation
- Network-reference validation
- Member moves and ordered prerequisite mutations
- Per-item multi-status failure handling
- Resource-level and switch-level deployment selection
- Standard module output and verbosity handling
- Integration validation action plugins

The integration-test target includes:

- Testbed preflight, fixture creation, and cleanup
- Every Interface Group type
- Ethernet, port-channel, mixed `any`, and optional vPC membership
- Custom Ethernet templates
- Network associations
- Resource-level and switch-level deployment
- `deploy: false`
- `merged`, `replaced`, `overridden`, `deleted`, and `gathered`
- Check mode and idempotency
- Gathered filtering and pagination behavior
- Cross-group member moves
- Negative validation and actionable error-message checks
- Delete cleanup and delete idempotency

Environment-dependent vPC and custom-template cases can be enabled through the integration configuration when the required fixtures are available.

* [ ] analyze
* [ ] infa
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist


mikewiebe:
@nikhilsrikrishna Please resolve the merge conflicts

plugins/module_utils/orchestrators/manage_interface_group.py
```diff
@@ -0,0 +1,1181 @@
...
+        proposed: NDConfigCollection,
+        check_mode: bool = False,
+    ) -> None:
+        """Remove cross-group members from source groups before final updates."""
+        for source_name, intermediate in sorted(self._move_plan.items()):
+            before = existing.get(source_name)
+            if before is None:
+                continue
+            if not check_mode:
+                self._put_group(intermediate)
```
allenrobel: **Accepted mutation stranded on the failure path (move hook).** This PUT removes the member from the source group and is a real, controller-accepted mutation. It runs before `_manage_create_update_state()` adds the member to the target. If the target `update()`/`create_bulk()` then fails, the module `fail_json`s with the member in neither group, and the source-side deploy queued on the next line is never sent (see the comment on `main()` in the module). A retry does not recover it: the source group now reads back exactly as intended, classifies as no-diff, and the pending switch config for those interfaces is never deployed.

This is the class CLAUDE.md describes under *Interface Module Mutation Stages and Queue Invariants*: "A stranded accepted mutation is invisible to a retry (it reads back as intended, classifies as no-diff, and is never re-queued). That is why the finalizer exists (issue #546)." Same finding class as the #547 review. The move needs either a failure-path finalizer that deploys what was accepted, or the source detach must be reported (`changed: true`, and the source interfaces listed) when the target step fails.

allenrobel (commented):
### Code review

Four findings, all inline: (1) controller-accepted mutations can be stranded undeployed on the failure path at the move hook, the delete-path detach, and a mixed create 207, with no finalizer in `main()`; (2) three ND-workaround sites lack `TODO(X.Y.Z)` markers; (3) per-network existence check is an N+1 GET loop; (4) EXAMPLES omit `overridden`.

Checked against the ND 4.2.1 OpenAPI spec for the interfaceGroups create/remove 207 schemas.

plugins/module_utils/orchestrators/manage_interface_group.py
```diff
@@ -0,0 +1,1181 @@
...
+
+    def delete_bulk(self, model_instances: list[InterfaceGroupConfigModel], **kwargs) -> ResponseType:
+        """Clear associations, then delete Interface Groups in one bulk request."""
+        for item in model_instances:
+            before = self._existing_groups.get(item.interface_group_name, item)
+            if before.networks or before.switch_interfaces:
+                detached = deepcopy(before)
+                detached.networks = []
+                detached.switch_interfaces = []
+                self._put_group(detached)
```
allenrobel: **Same class: detach PUT is accepted before the bulk remove is attempted.** Each iteration PUTs the group with its associations cleared and queues the affected interfaces for deploy, then the bulk `actions/remove` is sent after the loop. If the remove raises, or returns a mixed 207 (the 4.2.1 spec example shows exactly that: one `success`, one `failed` "associated with networks/interfaces"), the detach is accepted on the controller but never deployed, and a retry sees a group with no associations and re-queues nothing.

Note the delete 207 items *do* carry `interfaceGroupName`, so per-accepted-item bookkeeping is possible here: pop the accepted names from `_existing_groups` and keep their deploy pairs, and only raise for the names reported `failed`.

plugins/module_utils/orchestrators/manage_interface_group.py
```diff
@@ -0,0 +1,1181 @@
...
+                create_models.append(item)
+
+        payload = InterfaceGroupsCreateRequestModel(interface_groups=create_models).to_payload()
+        response = self._request(
+            path=endpoint.path,
+            verb=endpoint.verb,
+            data=payload,
+            operation_type=OperationType.CREATE,
+        )
+        self._validate_create_response_contract(response, expected_count=len(create_models))
```
allenrobel: **Mixed 207 discards the groups that were created.** `_validate_create_response_contract` raises on any non-`success` item before the loop below records `_existing_groups` or queues deploy, so groups the controller did create are neither reported nor deployed, and on a retry they classify as no-diff and are never deployed (the PR's own `test_manage_interface_group_00126` locks this in).

Per the 4.2.1 spec (`POST /fabrics/{fabricName}/interfaceGroups`, 207 schema), the per-item result carries only `type`, `status`, `message`, with no group name, so per-name correlation is not possible here. Options: correlate by index against the request order (verify in the lab that ND preserves it), or on a mixed 207 re-query the fabric groups and record whichever requested names now exist, then raise for the rest. Either way the error message should name the group, not just `type=ethernet`, when several groups of the same type were requested.

CLAUDE.md asks, for any bulk stage: "what a mixed 207 does to its queue; whether the finalizer could resubmit its request; whether the except-path reader sees a response that belongs to this request; and what the result items are keyed by." Worth answering for create, delete, and the any-group cumulative PUTs in the PR body.

plugins/modules/nd_manage_interface_group.py
```diff
@@ -0,0 +1,627 @@
...
+                    diff=[],
+                    changed=False,
+                    gathered=gathered,
+                )
+            )
+            return
+
+        nd_state_machine.manage_state()
+        if not module.check_mode:
+            nd_state_machine.model_orchestrator.deploy_pending()
```
allenrobel, line 799: **No failure-path finalizer.** `deploy_pending()` runs only when `manage_state()` returns. All three `except` handlers go straight to `fail_json`, so anything queued in `_pending_interfaces`/`_pending_switches` by a mutation that the controller already accepted (the move-hook source PUT, the delete-path detach PUTs, the created groups in a mixed 207, the earlier batches of an any-group cumulative PUT) is dropped. The `nd_interface_*` modules handle this with `finalize_accepted_intent()` in their handlers (issue #546, PR #547); this module has the same queue shape and needs the same treatment, with the `config_actions.deploy: false` case remaining a no-op.

plugins/module_utils/orchestrators/manage_interface_group.py
```diff
@@ -0,0 +1,1181 @@
...
+        except ValueError as exc:
+            raise RuntimeError(f"Interface Group delete returned invalid per-item results: {exc}") from exc
+        if parsed.failures:
+            messages = [f"{item.interface_group_name or '?'}: {item.message or 'unknown error'}" for item in parsed.failures]
+            raise RuntimeError(f"Interface Group delete returned a non-success per-item status: {'; '.join(messages)}")
+
+    def create_bulk(self, model_instances: list[InterfaceGroupConfigModel], **kwargs) -> ResponseType:
+        """Create Interface Groups and populate ``any`` membership in batches.
+
+        For controller compatibility, create ``any`` groups without members,
```
allenrobel: **Missing `TODO(X.Y.Z) <slug>` marker.** "For controller compatibility" describes working around a create-endpoint limitation (an `any` group cannot be created with its members; they have to be added afterwards in homogeneous cumulative PUTs). CLAUDE.md: "Any code that works around an ND API discrepancy (wire shape vs. schema, wrong HTTP status, silent field drops, etc.) MUST be flagged with a `TODO(X.Y.Z) <slug>` marker". `TODO(4.2.1) TBD` is fine for now; we will backfill the vault slug. Same applies at the two sites in the models (comments on `validators.py` and `config_models.py`).

plugins/module_utils/models/manage_interface_groups/validators.py
```diff
@@ -0,0 +1,272 @@
...
+    @staticmethod
+    def normalize_response_ethernet_attributes(value: dict | None) -> dict | None:
+        """Keep supported shared-policy attributes from an ND response."""
+        if not isinstance(value, dict):
+            return value
+        normalized = {}
+        for key, item in value.items():
+            if key not in _ETHERNET_ATTRIBUTE_KEYS:
+                continue
+            # Existing controller-created policies can echo an empty description.
```
allenrobel: **Missing `TODO(X.Y.Z) <slug>` marker.** "Existing controller-created policies can echo an empty description" is a wire-echo discrepancy handled in code. Per CLAUDE.md *Workaround Markers* this site needs a `TODO(4.2.1) <slug>` line (or `TODO(4.2.1) TBD`) above the comment so it can be audited and removed on the release cadence.

plugins/module_utils/models/manage_interface_groups/config_models.py
```diff
@@ -0,0 +1,786 @@
...
+        return merged
+
+    @staticmethod
+    def _normalize_controller_omitted_empty_attributes(
+        first: dict[str, Any],
+        second: dict[str, Any],
+    ) -> None:
+        """Treat controller-omitted empty policy strings as equivalent.
+
+        ND may omit empty optional shared-policy strings from a subsequent GET.
```
allenrobel: **Missing `TODO(X.Y.Z) <slug>` marker.** "ND may omit empty optional shared-policy strings from a subsequent GET" is a silent field drop worked around here. Per CLAUDE.md *Workaround Markers*, add a `TODO(4.2.1) <slug>` (or `TODO(4.2.1) TBD`) marker at this site.

plugins/module_utils/orchestrators/manage_interface_group.py
```diff
@@ -0,0 +1,1181 @@
...
+                template.get("parameters") or [],
+                input_label="template_config",
+            )
+            failures.extend(f"'{after.interface_group_name}': {message}" for message in input_errors)
+
+        if failures:
+            raise RuntimeError("; ".join(failures))
+
+    def _validate_networks_exist(self, networks: set[str]) -> None:
+        missing = sorted(name for name in networks if not self._network_exists(name))
```
allenrobel: **One GET per referenced network.** `_network_exists` issues `GET /fabrics/{fabricName}/networks/{networkName}` for every newly referenced network, so a task that attaches N networks costs N requests before any mutation. `EpManageFabricsNetworksGet` (`endpoints/v1/manage/manage_fabrics_networks.py`) already exists; one list request plus an in-memory set check covers all groups in the task. CLAUDE.md *Performance*: "No N+1 requests. Prefer one `query_all` plus in-memory filtering over a loop of `query_one` calls." and "Fetch each controller resource at most once per module run."

plugins/modules/nd_manage_interface_group.py
```diff
@@ -0,0 +1,627 @@
...
+  interfaces only, and returns a warning for referenced networks. Use
+  M(cisco.nd.nd_manage_networks) to deploy those networks separately.
+- With O(config_actions.deploy=true) and O(config_actions.type=switch), the selected
+  switch-level deploy is intentionally broad and can deploy unrelated pending switch configuration.
+- O(state=gathered) performs one paginated fabric read and applies the complete
+  filter contract locally so nested network/member filters and normalized
+  Ethernet types produce consistent results.
+"""
+
+EXAMPLES = r"""
```
allenrobel, line 313: **EXAMPLES do not cover `state: overridden`.** The block has merged, replaced, deleted, and gathered, but `overridden` (one of the five `state` choices) has no example. Team convention is that EXAMPLES include every supported state. An overridden example is also the natural place to show the "source group omits the member in the same task" move that the notes describe.

plugins/module_utils/orchestrators/manage_interface_group.py
nikhilsrikrishna: Thanks, Allen! I had overlooked how a controller-accepted mutation could be stranded when a later operation failed. I have addressed this in `9ea553a2`. Source-group member removal now uses the same accepted-mutation bookkeeping as normal updates. If the target-group operation subsequently fails, the confirmed source-side change and its deployment targets are retained, reported in `after`, and finalized when `deploy: true`. Unit coverage includes a successful source detach followed by target failure.

plugins/module_utils/orchestrators/manage_interface_group.py
nikhilsrikrishna: The association-clearing updates are now recorded individually before bulk removal. Keyed delete results are reconciled by `interfaceGroupName`, while malformed or incomplete results use bounded readback. Confirmed detach/delete mutations retain their deployment targets even when another delete item fails.

plugins/module_utils/orchestrators/manage_interface_group.py
nikhilsrikrishna: Mixed or ambiguous create results no longer rely on response order or group type. The module performs bounded paginated readback and records only requested groups whose names and normalized configurations match. Failure output identifies the requested and confirmed group names, and only confirmed mutations are queued for deployment.

plugins/modules/nd_manage_interface_group.py
nikhilsrikrishna: All module exception paths now finalize controller-accepted mutations when `deploy: true`. The finalizer is skipped in check mode and when `deploy: false`, does not resubmit a normal deployment that was already attempted, preserves the primary error, and reports controller-confirmed `after` and `changed` state.

plugins/module_utils/orchestrators/manage_interface_group.py
nikhilsrikrishna: Addressed in `5746a73e`. I added a `TODO(4.2.1)` marker at the NDFC compatibility batching path, explicitly noting that which interface member types can be combined in a single request remains to be determined.

plugins/module_utils/models/manage_interface_groups/validators.py
nikhilsrikrishna: Addressed in `5746a73e` by removing the workaround rather than retaining it. The built-in `ethernetWithPolicy` description field is no longer exposed or serialized; user input is rejected and a controller-returned description is ignored. The obsolete workaround site therefore no longer needs a TODO marker.

plugins/module_utils/models/manage_interface_groups/config_models.py
nikhilsrikrishna: Addressed in `5746a73e`. I added a `TODO(4.2.1)` marker directly above the normalization for NDFC-omitted empty optional shared-policy fields.

plugins/module_utils/orchestrators/manage_interface_group.py
nikhilsrikrishna: Addressed in `5746a73e`. The per-network GET loop is replaced by one cached, paginated fabric Network-list read followed by in-memory set validation. The cache is reused for the module run, and unit tests cover missing names, reuse, and pagination.

plugins/modules/nd_manage_interface_group.py
nikhilsrikrishna: Addressed in `5746a73e`. `EXAMPLES` now includes `state: overridden`, showing the complete desired fabric set and a same-task member move from the source group to the destination. The state documentation also explains supplied collections, omitted collections, and explicit empty lists.

plugins/module_utils/models/manage_interface_groups/config_models.py
```diff
@@ -0,0 +1,783 @@
...
+            interface_names = item.get("interface_names", item.get("interfaceNames"))
+            if not isinstance(interface_names, list):
+                return value
+            merged.setdefault(switch_id, []).extend(interface_names)
+        return [
+            {"switch_id": switch_id, "interface_names": interface_names}
+            for switch_id, interface_names in sorted(merged.items(), key=lambda entry: str(entry[0]))
+        ]
+
+    @model_validator(mode="after")
```
mikewiebe, line 258: ### High: A sparse merged update can reset omitted Ethernet policy attributes

**Issue**

When an existing `ethernetWithPolicy` group is updated with `state: merged` and
the user repeats the type but omits `ethernet_attributes`, validation
materializes a complete default Ethernet policy. The wrapper serializes those
injected values back into `module.params`, and the second model parse treats
them as explicitly supplied. The additive merge then replaces every existing
nondefault policy attribute with the defaults.

**Evidence**

- [`config_models.py` lines 258-295](https://github.com/CiscoDevNet/ansible-nd/pull/495/files#diff-69851252d92838d69bd76b65e8e9a62a5be18bed46a79dca4973467a81ffc9e2R258)
  inject an `InterfaceGroupEthernetAttributesModel` whenever type is
  `ethernetWithPolicy` and `ethernet_attributes` is omitted.
- [`validators.py` lines 42-55](https://github.com/CiscoDevNet/ansible-nd/pull/495/files#diff-cbc2519dbdd663a292ad00f5346f521f5cdf0684aaf1735d1b9d3c967f96c623R42)
  define the injected values, including `adminState=true`,
  `allowedVlans=none`, `cdp=true`, `mtu=jumbo`, and `speed=auto`.
- [`nd_manage_interface_group.py` lines 545-559](https://github.com/CiscoDevNet/ansible-nd/pull/495/files#diff-201ddd3d9004630a94ad1c82fff4aa0e0cf25b89b1eef154c241f5d2fbbf48c7R545)
  serialize every non-`None` model value into `module.params` rather than
  preserving the sparse input field set.
- [`config_models.py` lines 432-458](https://github.com/CiscoDevNet/ansible-nd/pull/495/files#diff-69851252d92838d69bd76b65e8e9a62a5be18bed46a79dca4973467a81ffc9e2R432)
  apply every field present in the reparsed nested model during merged
  reconciliation.
- An exact-head probe reproduced the reset shown in the practical example
  below.

**Practical example**

Assume the existing `servers` Interface Group has a deliberately nondefault
Ethernet policy:

```yaml
interface_group_name: servers
type: ethernetWithPolicy
ethernet_attributes:
  admin_state: false
  allowed_vlans: "10-20"
  cdp: false
  mtu: default
  speed: 1Gb
networks:
  - existing-net
```

The user submits a sparse merged task that is intended only to add `new-net`:

```yaml
- name: Add a network without changing the Ethernet policy
  cisco.nd.nd_manage_interface_group:
    fabric_name: fab1
    state: merged
    config:
      - interface_group_name: servers
        type: ethernetWithPolicy
        networks:
          - new-net
```

Expected result: `new-net` is associated with the group and every omitted
`ethernet_attributes` value remains unchanged.

Actual proposed result:

| Ethernet attribute | Existing value | Value after sparse merge |
| --- | --- | --- |
| `admin_state` | `false` | `true` |
| `allowed_vlans` | `10-20` | `none` |
| `cdp` | `false` | `true` |
| `mtu` | `default` | `jumbo` |
| `speed` | `1Gb` | `auto` |

The network-only update therefore also rewrites the group's Ethernet policy.

**ND/OpenAPI applicability**

**Existing PR overlap**

No matching open issue found.

**Impact**

A task intended only to add a Network association can silently rewrite the
group's live Ethernet policy. With deployment enabled, the reset can reach
member switch interfaces; with deployment disabled, incorrect pending intent
is still staged on the controller.

**Suggested fix**

Apply Ethernet defaults only while building a create payload, not by mutating a
sparse playbook model. Preserve `model_fields_set` or use `exclude_unset`
through the wrapper-to-state-machine handoff. Add an end-to-end
wrapper/state-machine test for an existing nondefault group using merged input
that supplies `type` and `networks` but omits `ethernet_attributes`; cover check
mode, normal mode, `deploy=false`, and `deploy=true`.

---

nikhilsrikrishna: Thanks for identifying this; I had overlooked that validator-injected defaults were being serialized as explicit merged input. I have addressed this in [`c0f316ff`](https://github.com/CiscoDevNet/ansible-nd/pull/495/commits/c0f316ff7c6ae5b64c5f9121069a2b441fcfb856).

For `state: merged`, the wrapper now preserves the user input sparse field set through the state-machine reparse using `exclude_unset`, so validator-injected Ethernet defaults are not treated as explicitly supplied input. Create operations still materialize the required defaults.

I added end-to-end wrapper → state machine → orchestrator coverage for an existing group with nondefault policy attributes, covering check mode, apply, idempotency, and both `deploy: false` and resource-level `deploy: true`. The tests verify that adding a Network preserves omitted Ethernet attributes and existing member/Network associations. Explicit empty and partially supplied attributes are also covered.

The targeted live integration regression passed on NDFC for both deployment modes.

allenrobel (approved):
All nine of my review threads are addressed at c0f316ff and I verified each against the code:

- Move-hook source detach, delete-path detach, and mixed-207 create now record only controller-confirmed mutations (per-item by `interfaceGroupName` for delete, bounded readback for create) and keep their deploy targets.
- All three `main()` exception paths call the failure-path finalizer, guarded by check mode, `deploy: false`, and `_deploy_attempted`.
- Per-network GET loop replaced by one cached, paginated network list.
- `state: overridden` EXAMPLE added with a same-task member move.
- Description-echo workaround removed; both remaining workaround sites carry `TODO(4.2.1)` markers.

Unit tests for the module, orchestrator, models, and action plugin pass locally (236 passed). Approving.

PR #498, opened 2026-08-10
Config Actions: Add Common Framework for Repository-Wide Save and Deploy Handling

AKDRG:
## Related Issue(s)

Related to #368

## Proposed Changes

- Add a repository-wide `config_actions` framework for parsing, validating, planning, and executing save/deploy intent.
- Introduce policy-driven config action behavior for module families:
  - interface
  - fabric
  - switch
  - vPC pair
  - resource-oriented modules
  - legacy compatibility
- Add shared support for:
  - defaulting omitted `config_actions` through module-family policies
  - rejecting explicit empty `config_actions: {}`
  - validating unsupported options and unsupported deploy types
  - enforcing `deploy=true` / `save=false` restrictions where required
  - read-only state handling for gathered workflows
  - resource-level deploy gating through `config_actions.type: resource`
  - per-resource deploy override tracking
- Add `ConfigActionsController` and backend protocol so module-specific integrations can plug in their own save/deploy endpoint implementations without changing the common framework.
- Add `ConfigActionsMixin` facade for orchestrators that need to execute normalized config action plans.
- Update shared `nd_argument_specs.config_actions_spec()` to delegate to the policy-based implementation while preserving compatibility with existing call patterns.
- Add focused unit coverage for argument specs, parser behavior, raw Ansible argument detection, validation, policies, controller behavior, and mixin delegation.

## Test Notes

- Ran focused unit tests:

  `PYENV_VERSION=ndfclab PYTHONPATH=$PWD/collections python -m pytest -q tests/unit/module_utils/config_actions tests/unit/module_utils/orchestrators/test_config_actions_mixin.py`

- Result:

  `52 passed in 0.25s`

N/A - common framework only. Module-specific endpoint integrations and live controller validation will be handled in follow-up PRs.

- [ ] analyze
- [ ] infa
- [ ] manage
- [ ] onemanage
- [x] other

## Checklist


plugins/module_utils/config_actions/parser.py
```diff
@@ -0,0 +1,174 @@
...
+        deploy=deploy,
+        type=action_type,
+        provided=provided,
+        explicit_options=explicit_options,
+        resource_deploy_provided=bool(resource_indexes),
+        resource_deploy_indexes=resource_indexes,
+        resource_deploy_overrides=resource_overrides,
+    )
+
+    if state in policy.read_only_states and not provided:
```
allenrobel, line 161: The read-only-state force-off only fires when `config_actions` is omitted entirely (`not provided`). If a task in `gathered` state supplies `config_actions` with only a non-write suboption — e.g. `config_actions: {type: switch}` — then `provided` is true, this branch is skipped, and Ansible's default fill-in leaves `save`/`deploy` at the policy defaults (`True` for five of the six shipped policies). `validate_config_actions()` doesn't catch it either: its read-only check only fires when `save`/`deploy` are *explicitly named* in the raw args (`explicit_options & {"save", "deploy"}`), not when they're merely default-filled.

So `parse_config_actions()` can return `ConfigActions(deploy=True)` in a read-only state with no error — a caller trusting `actions.deploy` would deploy during `gathered`. Suggest forcing `save`/`deploy` off in read-only states regardless of `provided` (unless explicitly set, which validation already rejects), and adding a test for the `type`-only-in-gathered case — the new suite only covers the omitted-dict and explicitly-named cases.

plugins/module_utils/orchestrators/config_actions_mixin.py
```diff
@@ -64,6 +73,38 @@ class ConfigActionsMixin:
...
+        if selected_backend is None:
+            if self.config_actions_backend_class is None:
+                raise ValueError("No config actions backend is configured for this orchestrator.")
+            selected_backend = self.config_actions_backend_class(self)
+
+        controller = ConfigActionsController(
+            policy=self.config_actions_policy,
+            backend=selected_backend,
+        )
+        return controller.execute(actions, context)
```
allenrobel, line 106: The controller's skip paths (`no_fabrics`, `ineligible`, `actions_disabled`, `no_targets`) return `status: "skipped"` but nothing surfaces that to the user — no `rest_send.warn()`, and this facade returns the result unchanged. This is the same visibility gap fixed in this file in #491, where the switchless-fabric skip in `execute_config_actions()` was switched from `logger.info()` (invisible unless `ND_LOGGING_CONFIG` is set) to `self.rest_send.warn(...)`.

The controller is transport-agnostic so it can't warn itself, which makes this facade the natural place: after `controller.execute()`, translate skipped steps into `self.rest_send.warn(...)` so integrations get the surfaced-skip behavior by default instead of each caller having to remember to do it.

plugins/module_utils/nd_argument_specs.py
```diff
@@ -40,55 +45,46 @@ def nd_argument_spec() -> dict[str, Any]:
...
-    Return the shared `config_actions` argument spec fragment, filtered to the allowlist `include`.
-
-    The full option set is `save`, `deploy`, and `type`. Modules that expose only a subset pass the option names they support, e.g.
-    `config_actions_spec(include=("deploy",))` for the `nd_interface_*` modules.
+    Return the shared `config_actions` argument spec fragment.
 
-    `type` accepts `resource`, `switch`, and `global`, per the contract in issue #368. The companion per-resource `deploy` key described in that
-    issue is not part of this fragment yet: it lives in each module's `config` suboptions and its interaction with `config_actions` (mutually
-    exclusive, or gated on `type == "resource"`) is still under discussion on #368. It will be added as a separate fragment once settled.
+    This compatibility wrapper delegates to the policy-based implementation in `config_actions.argument_spec`.
```
allenrobel, line 71: Two pieces of load-bearing docstring content are dropped here without being re-homed in `config_actions/argument_spec.py`:

1. The allowlist rationale on `_select_options()` ("An allowlist (rather than an exclude list) is deliberate: options added to a shared fragment in the future must never silently appear in modules that composed the fragment before the addition") — the mechanics survive, but the reason it must stay an allowlist is now undocumented in both the wrapper and the new canonical implementation.
2. The note that the per-resource `deploy` key's interaction with `config_actions` ("mutually exclusive, or gated on `type == 'resource'`") was still under discussion on #368. This PR implements one specific resolution (`RESOURCE_CONFIG_ACTIONS.resource_interaction = "type_resource_gated"`) — #368 is still open, so it would be good to note the decision on that issue and carry the rationale into `policies.py`/`argument_spec.py` rather than silently settling it.

allenrobel (commented):
### Code review

Review of the common config-actions framework. Findings posted inline; the switch-module wiring on top of this framework was reviewed separately on #499.

plugins/module_utils/config_actions/backend.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def save(self, context: ConfigActionsContext, fabric_name: str) -> Any:
+        """
+        # Summary
+
+        Save or recalculate configuration for `fabric_name`.
+
+        ## Raises
+
+        Exception
```
allenrobel: All four Protocol methods (`save` here, `deploy_global` L48, `deploy_switches` L59, `deploy_resources` L70) list a bare `Exception` under `## Raises`. CLAUDE.md: "use `### ExceptionType` subheadings listing conditions as a Markdown list".

plugins/module_utils/config_actions/controller.py
```diff
@@ -0,0 +1,251 @@
...
+
+    def execute(self, actions: ConfigActions, context: ConfigActionsContext) -> ConfigActionsResult:
+        """
+        # Summary
+
+        Execute or plan config actions using the supplied backend.
+
+        ## Raises
+
+        Exception
```
allenrobel: `execute()` (here) and `_deploy()` (L127) list a bare `Exception` under `## Raises` — same format issue as `backend.py`. CLAUDE.md: "use `### ExceptionType` subheadings listing conditions as a Markdown list".

plugins/module_utils/config_actions/policies.py
```diff
@@ -0,0 +1,70 @@
...
+    defaults=ConfigActionsDefaults(deploy=True, type="switch"),
+    allowed_types=frozenset({"switch", "resource"}),
+    deploy_requires_save=False,
+    resource_interaction="type_resource_gated",
+)
+
+LEGACY_CONFIG_ACTIONS = ConfigActionsPolicy(
+    name="legacy",
+    supported=frozenset({"save", "deploy", "type"}),
+    defaults=ConfigActionsDefaults(save=True, deploy=True, type="switch"),
```
gmicol: issue: Preserve the canonical opt-in deploy default - High

Could the legacy and interface-facing policies preserve the `deploy=False` contract now on upstream/develop?

Commit 97cc6a5 changed `config_actions_spec(include=("deploy",))` to opt-in deployment. `nd_interface_ethernet_access` and `nd_interface_loopback` now consume that helper and document that only `deploy: true` pushes changes.

Retaining `deploy=True` while resolving the current conflict would make an explicit empty block inject `deploy=true` and unexpectedly push staged interface changes.

suggested fix:

Set `deploy=False` for `LEGACY_CONFIG_ACTIONS` and `INTERFACE_CONFIG_ACTIONS`, then update the policy and compatibility tests. Preserve `True` only for families whose established contract still requires it.

plugins/module_utils/orchestrators/config_actions_mixin.py
```diff
@@ -64,6 +73,38 @@ class ConfigActionsMixin:
           ``{"switchIds": [...]}``.
     """
 
+    config_actions_policy: ConfigActionsPolicy = FABRIC_CONFIG_ACTIONS
+    config_actions_backend_class: type[ConfigActionsBackend] | None = None
```
gmicol: issue: Keep mixin configuration out of Pydantic model fields - Medium

Could these hooks use `ClassVar`? Pydantic collects both annotations as instance fields. A normal class-level backend override raises `PydanticUserError`, while supplying the backend class as a field value raises `TypeError` because `ConfigActionsBackend` is not runtime-checkable.

suggested fix:

Annotate both hooks with `ClassVar`, or define a typed `ClassVar` backend factory if construction requires the owning orchestrator. Cover the configured-backend path with a concrete `NDBaseOrchestrator` subclass.

plugins/module_utils/config_actions/controller.py
```diff
@@ -0,0 +1,255 @@
...
+            )
+
+        if not actions.save and not actions.deploy:
+            return ConfigActionsResult(
+                requested=actions,
+                effective=actions,
+                status="skipped",
+                reason="actions_disabled",
+                targets=targets,
+            )
```
gmicol: issue: Reconcile top-level and per-resource deploy decisions - Medium

Could this early return account for explicit resource overrides? 
`resource_deploy_enabled()` says an item override takes precedence over the top-level value. For top-level `deploy=false` with `config[0].deploy=true`, parsing reports that resource as enabled, but the controller returns `actions_disabled`.

example:
```YAML
config_actions:
  deploy: false
  type: resource
config:
  - name: EXAMPLE_NAME
    deploy: true
```

plugins/module_utils/config_actions/controller.py
```diff
@@ -0,0 +1,255 @@
...
+
+        steps: list[ConfigActionStepResult] = []
+        for fabric_name in normalized_context.fabric_names:
+            if actions.save:
+                try:
+                    response = self.backend.save(normalized_context, fabric_name)
+                    steps.append(ConfigActionStepResult(action="save", status="completed", target=fabric_name, response=response))
+                except Exception as exc:  # pylint: disable=broad-exception-caught
+                    steps.append(ConfigActionStepResult(action="save", status="failed", target=fabric_name, error=str(exc)))
+                    return self._result(actions, targets, steps)
```
gmicol, line 114: issue: Preserve structured backend failure details - Medium

Could the controller retain structured exception data instead of reducing every failure to `str(exc)`?

`NDModuleError` can carry the HTTP status, request payload, response payload, and raw body. Both save and deploy catches discard those fields, preventing faithful 4xx/5xx or partial-response reporting and evidence-based retry decisions.

suggested fix:

Add structured error fields to `ConfigActionStepResult`, or raise a typed exception carrying the partial `ConfigActionsResult` while preserving the original exception.

example:
```python
raise NDModuleError(
    "deploy rejected",
    status=409,
    response_payload={"errors": [...]},
)
```

plugins/module_utils/config_actions/controller.py
```diff
@@ -0,0 +1,255 @@
...
+
+        - Raised by backend deploy operations when the controller requests a supported deploy scope.
+        """
+        if actions.type == "global":
+            response = self.backend.deploy_global(context, fabric_name)
+            return ConfigActionStepResult(action="deploy", scope="global", status="completed", target=fabric_name, response=response)
+        if actions.type == "switch":
+            if not context.switch_ids:
+                return ConfigActionStepResult(action="deploy", scope="switch", status="skipped", target=fabric_name, error="no_targets")
+            response = self.backend.deploy_switches(context, fabric_name, context.switch_ids)
```
gmicol, line 143: issue: Scope deploy targets to their fabric - Medium

Could scoped targets be associated with each fabric, or could ambiguous multi-fabric input be rejected? 
`execute()` loops over `fabric_names`, but this line sends the same flat `switch_ids` tuple to every fabric; resource deploy repeats the pattern. Targets from two fabrics are therefore submitted to both fabric-specific endpoints.

suggested fix:

Key targets by fabric, execute one context per fabric, or reject multiple fabrics whenever scoped targets are supplied. Add a two-fabric ordering and partial-failure test.

example:
```python
ConfigActionsContext(
    fabric_names=("FAB1", "FAB2"),
    switch_ids=("FAB1-S1", "FAB2-S1"),
)
```

gmicol (changes requested):
Just a few comments - The canonical-default regression and read-only write path (this was mentioned in a previous comment) are both potentially unsafe mutations. The Pydantic hook failure and remaining controller-contract defects should also be resolved.

Also, the branch conflicts with current `develop`- Could you rebase your local branch and fix the conflicts?

plugins/module_utils/config_actions/backend.py
AKDRG: Fixed.

plugins/module_utils/config_actions/controller.py
AKDRG: Fixed.

plugins/module_utils/nd_argument_specs.py
AKDRG: Fixed.

plugins/module_utils/orchestrators/config_actions_mixin.py
AKDRG: Fixed.

plugins/module_utils/config_actions/parser.py
AKDRG: Fixed.

plugins/module_utils/config_actions/policies.py
AKDRG: Fixed!

plugins/module_utils/orchestrators/config_actions_mixin.py
AKDRG: Fixed!

plugins/module_utils/config_actions/controller.py
AKDRG: Fixed!

plugins/module_utils/config_actions/controller.py
AKDRG: Fixed!

AKDRG: The controller no longer reduces backend failures to only `str(exc)`. Failed save/deploy steps now preserve structured exception details in the config action result:

- `error`
- `error_type`
- `http_status`
- `request_payload`
- `response_payload`
- `raw`

This keeps `NDModuleError` details visible to the module result, including HTTP status, request body, response payload, and raw controller body when available.

plugins/module_utils/config_actions/controller.py
AKDRG: Config action scoped targets can now be associated with their owning fabric through:

- `switch_ids_by_fabric`
- `resources_by_fabric`

The controller now executes one single-fabric context per fabric and passes only that fabric’s scoped targets to the backend.

For ambiguous multi-fabric input using flat scoped targets, the controller now fails closed:

- multi-fabric `type: switch` with flat `switch_ids` is rejected
- multi-fabric `type: resource` with flat `resources` is rejected

This prevents targets from one fabric being submitted to another fabric’s endpoint.

plugins/module_utils/config_actions/controller.py
AKDRG: The controller now uses `deploy_requested()` instead of only checking top-level `deploy`.

This means deploy is active when either:

- `config_actions.deploy: true`
- any resource item has `deploy: true`

So `config_actions.deploy: false` with `config[].deploy: true` no longer returns `actions_disabled`.

The reverse is also supported: `config_actions.deploy: true` with `config[].deploy: false` lets the module exclude that resource from deploy target collection.

allenrobel (approved):
Re-reviewed the changes since my 2026-08-10 review (the commits from "Surface config action skip warnings" through "Scope config action deploy targets by fabric", plus the docstring fixes). All five of my earlier threads are addressed:

- Read-only states now force `save`/`deploy` off after validation regardless of `provided`; the type-only `gathered` case and item-level `deploy` rejection are covered by tests.
- Skipped controller results and skipped action steps are surfaced through `rest_send.warn()` in the mixin facade.
- The allowlist rationale and the #368 resolution are now documented in `argument_spec.py` / `policies.py`.
- `## Raises` sections use the `### ExceptionType` format.

Also read through the changes for gmicol's threads: `ClassVar` hooks, opt-in `deploy` defaults for the legacy/interface policies, `deploy_requested()` reconciling top-level and per-resource deploy, structured failure details on `ConfigActionStepResult`, and per-fabric target scoping that fails closed on flat multi-fabric targets. The new tests cover each of those paths.

Unit tests for `config_actions/`, the mixin, and `nd_argument_specs` pass locally (76 passed) and CI is green. Leaving gmicol's two open threads to gmicol.

mikewiebe (approved):

PR #499, opened 2026-08-10
Config Actions: Wire nd_manage_switches to the Common Save and Deploy Framework

AKDRG:
## Related Issue(s)

Related to #368

Merge dependency on common framework #498

## Proposed Changes

- Wire `nd_manage_switches` into the shared `config_actions` framework.
- Parse and validate switch module `config_actions` through the common policy-driven parser.
- Add a switch-specific config actions backend for:
  - fabric config save
  - switch-scoped deploy
  - global fabric deploy
- Preserve existing switch module defaults:
  - `config_actions.save: true`
  - `config_actions.deploy: true`
  - `config_actions.type: switch`
- Enforce common validation behavior:
  - reject explicit empty `config_actions: {}`
  - reject `deploy: true` with `save: false`
  - reject unsupported deploy types for switch workflows
- Route switch finalize flows through the common controller while preserving existing add, migrate, re-add, POAP, preprovision, RMA, and idempotent deploy behavior.
- Add result registration for config save/deploy actions so config action execution is visible in module output.
- Keep idempotent out-of-sync switches deployable on a later run when `config_actions.deploy` is enabled.
- Add focused unit coverage for switch config action finalize behavior and common framework integration.

## Test Notes

- Ran formatting:
  - `black --check -l 159 plugins tests`
  - Result: passed

- Ran targeted sanity checks:
  - `ansible-test sanity --test pep8 --color --truncate 0`
  - `ansible-test sanity --test pylint --color --truncate 0`
  - Result: passed

- Ran full sanity from a materialized collection tree:
  - `ansible-test sanity --color --truncate 0`
  - Result: passed
  - Note: Python 3.10 and 3.12 compile/import checks were skipped locally because those interpreters are not installed.

- Live tested switch config actions on `AK-VXLAN`:
  - 8-switch import with config save and switch-level deploy enabled.
  - Switch-level deploy flow completed.
  - Global deploy flow was exercised.
  - Subsequent deploy run after a staged/non-deployed run was validated.
  - Cleanup/removal workflow was tested.

Cisco Nexus Dashboard 4.2.1

- [ ] analyze
- [ ] infa
- [x] manage
- [ ] onemanage
- [x] other

## Checklist

- [ ] Latest commit is rebased from develop with merge conflicts resolved


plugins/module_utils/manage_switches/nd_switch_resources.py
```diff
@@ -1358,33 +1389,78 @@ def bulk_update_roles(
...
-            self.ctx.log.info("Saving fabric configuration")
-            self.fabric_utils.save_config()
+        context = ConfigActionsContext(
+            fabric_names=(self.ctx.fabric,),
+            state=None,
+            check_mode=self.ctx.nd.module.check_mode,
+            switch_ids=tuple(serial_numbers or ()),
+        )
+        controller = ConfigActionsController(SWITCH_CONFIG_ACTIONS, SwitchConfigActionsBackend(self.fabric_utils))
+        result = controller.execute(actions, context)
```
allenrobel, line 1420: When `deploy_type` is `switch` and `serial_numbers` is empty, the controller returns `status: "skipped", error: "no_targets"` for the deploy step and the module still reports success. The code this replaces fell back to a global deploy with a warning in exactly that case (`"Switch-level deploy requested but no serial numbers provided — falling back to global deploy"`, removed in this hunk).

The idempotent config-sync paths call `finalize(serial_numbers=sync_serials)` (three sites, ~L3241/L3389/L3528) where `sync_serials` is built from a filtered comprehension (`switch_id` must be truthy) and can legitimately be empty while a deploy is still required — in that case pending config-sync now goes undeployed, with the skip visible only in the nested `DATA.actions[].error` field.

Suggest either restoring the global-deploy fallback for this path, or at minimum warning/failing when a requested deploy is skipped for lack of targets, so the pending config doesn't silently stay undeployed.

plugins/module_utils/manage_switches/config_actions_backend.py
```diff
@@ -0,0 +1,91 @@
...
+        """
+        # Summary
+
+        Deploy all pending fabric configuration.
+
+        ## Raises
+
+        Exception
+        """
+        return self.fabric_utils.deploy_config()
```
allenrobel, line 69: `save()` (here), `deploy_global()` (L77), and `deploy_switches()` (L89) list a bare `Exception` under `## Raises`. CLAUDE.md: "use `### ExceptionType` subheadings listing conditions as a Markdown list". `deploy_resources()` in this same file shows the expected format (`### NotImplementedError` + bullet list).

allenrobel (commented):
### Code review

Reviewed the switch-module wiring into the common config-actions framework (findings in the `config_actions/` package itself are out of scope here since they land via #498). Two findings, posted inline.

plugins/module_utils/manage_switches/config_actions_backend.py
AKDRG: Fixed

plugins/module_utils/manage_switches/nd_switch_resources.py
AKDRG: Great catch. 
Since we are specfically targeting "switch" deploy_type, doing a global deploy should not be done. So, I'm throwing up an error now.

allenrobel (approved):
Verified `8327cd91` addresses both threads: `finalize()` now fails on switch-scoped deploy with no targets, and `_idempotent_sync_serials()` fails early naming any out-of-sync switch without a resolved switchId. Unit tests pass locally (39 passed). LGTM.

nikhilsrikrishna:
### Check mode reports no change although the same task performs save/deploy in a normal run

Consider a switch that already exists in the fabric and matches the requested inventory configuration, but NDFC reports it as out of config sync.

For this input, a normal module run recognizes that pending configuration must be saved/deployed and calls `finalize()` for the switch. Check mode, however, reports `changed: false` because there is no switch to add, update, or delete.

The two executions therefore disagree about whether the task has work to perform:

```yaml
# check mode
changed: false
save_deploy_required: true

# normal execution
save/deploy is performed
```

The state handler correctly detects the pending configuration through the switch config-sync status:

```python
sync_serials = self._idempotent_sync_serials(plan, existing_by_ip)
idempotent_save_req = bool(sync_serials)
```

It includes this condition when deciding whether work exists and exposes it in the check-mode details as `save_deploy_required`.

The normal execution path then acts on the condition:

```python
if not switch_actions and idempotent_save_req:
    self.fabric_ops.finalize(serial_numbers=sync_serials)
```

But the final check-mode result is calculated only from the inventory add/update/delete list:

```python
changed = bool(diff_list)
```

Because an out-of-sync switch can require save/deploy without requiring an inventory change, `diff_list` is empty and `changed` becomes false. This is inconsistent with both `save_deploy_required: true` and the normal execution behavior.

AKDRG:
> ### Check mode reports no change although the same task performs save/deploy in a normal run
> Consider a switch that already exists in the fabric and matches the requested inventory configuration, but NDFC reports it as out of config sync.
> 
> For this input, a normal module run recognizes that pending configuration must be saved/deployed and calls `finalize()` for the switch. Check mode, however, reports `changed: false` because there is no switch to add, update, or delete.
> 
> The two executions therefore disagree about whether the task has work to perform:
> 
> ```yaml
> # check mode
> changed: false
> save_deploy_required: true
> 
> # normal execution
> save/deploy is performed
> ```
> 
> The state handler correctly detects the pending configuration through the switch config-sync status:
> 
> ```python
> sync_serials = self._idempotent_sync_serials(plan, existing_by_ip)
> idempotent_save_req = bool(sync_serials)
> ```
> 
> It includes this condition when deciding whether work exists and exposes it in the check-mode details as `save_deploy_required`.
> 
> The normal execution path then acts on the condition:
> 
> ```python
> if not switch_actions and idempotent_save_req:
>     self.fabric_ops.finalize(serial_numbers=sync_serials)
> ```
> 
> But the final check-mode result is calculated only from the inventory add/update/delete list:
> 
> ```python
> changed = bool(diff_list)
> ```
> 
> Because an out-of-sync switch can require save/deploy without requiring an inventory change, `diff_list` is empty and `changed` becomes false. This is inconsistent with both `save_deploy_required: true` and the normal execution behavior.

Implemented the check-mode fix on config_actions_switches_integration.
- Check mode now treats pending config save/deploy for idempotent out-of-sync switches as planned work.
- The final result now reports changed: true when normal mode would call finalize().
- The synthetic check-mode diff includes a clear config_actions entry with save, deploy, deploy_type, and serial_numbers.

Issue #489, opened 2026-08-06
Enhance nd_manage_networks support for empty attachments and PVLAN networks


AKDRG:
### New module/feature description:

```shell
We need to enhance `nd_manage_networks` to support Network attachments without explicit interface bindings and add playbook-facing support for PVLAN primary and secondary community Networks.

The module should allow `attach[].interfaces` to be omitted or set to an empty list. This enables attaching a Network to a switch without selecting specific interfaces.

The module should also expose PVLAN Network fields:


vlan_network_type:
  description:
    - VLAN network role for normal and PVLAN networks.
    - Defaults to normal when omitted.
    - private_secondary_community requires primary_network_id.
  type: str
  choices:
    - normal
    - private_primary
    - private_secondary_community

primary_network_id:
  description:
    - Primary Network ID associated with a PVLAN secondary Network.
  type: int


Notes:

- `attach[].interfaces` should default to an empty list when omitted.
- Empty interface lists should be preserved in the attachment payload.
- `private_primary` Networks should be supported as L2-only PVLAN primary Networks.
- `private_secondary_community` Networks should require `primary_network_id`.
- PVLAN secondary Networks should be treated as L2-only and should not emit `l3Data`.
- `primaryNetworkId` should be preserved in standalone and MCFG oneManage Network payload flows.
- Unsupported PVLAN roles should not be exposed in the public playbook contract.
```

### Which resource area of Cisco Nexus Dashboard does this new feature belong to?

manage

### New or Affected Documentation for New Feature

```shell
`nd_manage_networks`
```

### Potential Ansible Task Config

```yaml
- name: Attach Network to switch without per-interface bindings
  cisco.nd.nd_manage_networks:
    fabric_name: fab1
    state: merged
    config:
      - network_name: EMPTY_ATTACH_NET
        is_l2only: true
        network_id: 50110
        vlan_id: 2110
        attach:
          - ip_address: 10.1.1.11
            interfaces: []



- name: Create a private primary Network
  cisco.nd.nd_manage_networks:
    fabric_name: fab1
    state: merged
    config:
      - network_name: PVLAN_PRIMARY
        vlan_network_type: private_primary
        is_l2only: true
        network_id: 50100
        vlan_id: 2100
        deploy: false



- name: Create a private secondary community Network
  cisco.nd.nd_manage_networks:
    fabric_name: fab1
    state: merged
    config:
      - network_name: PVLAN_SECONDARY_COMMUNITY
[...]

PR #458, opened 2026-07-30
ND_Manage_Networks: Support Empty Network Attachment Interfaces and PVLAN Secondary Networks

AKDRG:
## Related Issue(s)

https://github.com/CiscoDevNet/ansible-nd/issues/489

## Proposed Changes

- Allow `nd_manage_networks` attachments to omit `attach[].interfaces`; missing interfaces now default to an empty list.
- Preserve empty interface payloads so Network attachments can be applied without explicit interface selection.
- Add playbook-facing PVLAN Network support for:
  - `vlan_network_type: private_primary`
  - `vlan_network_type: private_secondary_community`
  - `primary_network_id` for secondary community Networks
- Limit the public PVLAN contract to tested/supported values: `normal`, `private_primary`, and `private_secondary_community`.
- Ensure PVLAN secondary Network payloads are L2-only and do not emit `l3Data`.
- Preserve `primaryNetworkId` in standalone and MCFG oneManage Network payload flows.
- Avoid injecting contextual L3 defaults from argspec into sparse/PVLAN Network configs.

## Test Notes

- Ran focused unit tests:
  - `PYENV_VERSION=ndfclab PYTHONPATH=$PWD/collections python -m pytest -q tests/unit/module_utils/orchestrators/test_networks.py`
  - Result: `84 passed`
- Ran formatting:
  - `black -l 159`
- Live tested on `Fabric`:
  - `private_primary` Network create succeeded.
  - `private_secondary_community` Network create succeeded with `primary_network_id`.
  - Secondary community Network merged update, replaced, explicit delete, and cleanup succeeded.

## Cisco Nexus Dashboard Version

Cisco Nexus Dashboard 4.2.1

## Related ND API Resource Category

- [ ] analyze
- [ ] infa
- [x] manage
- [x] onemanage
- [ ] other

## Checklist

- [x] Latest commit is rebased from develop with merge conflicts resolved
- [x] New or updates to documentation has been made accordingly
- [ ] Assigned the proper reviewers

PR #500, opened 2026-08-10
ND Manage VRFs, Networks, and Switches: Align ND 4.2 Fields, Attachments, and Delete Readiness

AKDRG:
## Related Issue(s)

Related to CiscoDevNet/ansible-nd#458, https://github.com/CiscoDevNet/ansible-nd/issues/489

## Proposed Changes

This PR updates the user-facing module fields and generated payloads to match the ND 4.2 Manage API behavior validated across Networks, VRFs, and Switch bootstrap workflows.

### Networks

- Removed legacy/DCNM-style compatibility aliases from `nd_manage_networks`.
- Removed unsupported user-facing options such as `query` state, `network_type`, `tenant_name`, `is_l2only`, `rt_auto`, `enable_ir`, `l2_fabric_data`, and older shorthand field aliases.
- Kept ND-native canonical fields such as `arp_suppression`, `mtu`, `route_target_both`, `multicast_group_address`, and `ds_vni`.
- Added user-facing PVLAN support with simple values:
  - `normal`
  - `primary`
  - `community`
  - `isolated`
- Internally maps PVLAN values to the ND API values:
  - `primary` -> `privatePrimary`
  - `community` -> `privateSecondaryCommunity`
  - `isolated` -> `privateSecondaryIsolated`
- Added validation for PVLAN secondary networks:
  - requires `primary_network_id`
  - rejects L3-only fields
  - places secondary `vlan_id` in `networkTemplateConfig.vlanId`
  - omits top-level `vlanId` for secondary PVLAN payloads
- Allows network attachments with omitted `interfaces`, defaulting them to an empty list.
- Added structured `attachment_options` support using snake_case fields.
- Replaced attachment `extra_config` with `freeform_config`.

### VRFs

- Removed legacy attachment/config aliases to keep VRF input ND-native and consistent with Networks.
- Updated VRF docs/templates/tests to reflect canonical field names.

### VRFs and Networks
- Improved VRF and Network delete-readiness checks after detach.
- Added shared delete-readiness handling to avoid incomplete delete decisions from stale attachment rows.
- Optimized targeted controller queries for VRF and Network workflows to reduce unnecessary full-fabric reads.
- Added/updated unit and integration coverage for the changed Network, VRF, attachment, and delete-readiness behavior.

### Switches

- Updated bootstrap import payloads for ND 4.2 fields:
  - uses `softwareImage`
  - removes legacy `imagePolicy`
  - supports `dhcpBootstrapIp`
  - supports `seedSwitch`
- Removed `software_image` from pre-provision input, since pre-provision uses software version rather than software image.
- Updated switch module documentation and examples accordingly.

### Tests

- Updated Network unit tests for:
  - PVLAN type mapping
  - secondary PVLAN payload generation
  - attachment interface defaults
  - attachment option validation
  - canonical argspec behavior
  - removed legacy aliases
- Updated VRF and Switch unit/integration coverage impacted by the field cleanup.
- Updated Network integration playbooks for standalone, MSD, and MCFG flows.

## Test Notes

Validated locally:

- `pytest` focused Network tests:
  - `104 passed`
- `black -l 159 --check`
- `git diff --check`
- `ansible-test sanity --test validate-modules --test import --test pep8 --test pylint --python 3.11`

Live validation was also performed against ND 4.2 Manage workflows for Network field behavior, PVLAN payload handling, and switch bootstrap field handling.

Developed and validated against Cisco Nexus Dashboard 4.2.x Manage APIs.

- [ ] analyze
- [ ] infra
- [x] manage
- [ ] onemanage
- [ ] other

## Checklist

- [x] New or updated documentation has been made accordingly


plugins/module_utils/models/manage_networks/network_data_models.py
```diff
@@ -368,15 +369,26 @@ class NetworkBaseModel(NetworkCommonModel):
...
     def from_response(cls, response: dict[str, Any], **kwargs) -> "NetworkBaseModel":
         normalized = dict(response)
         if "layer" not in normalized and "networkMode" in normalized:
             normalized["layer"] = normalized["networkMode"]
         l2_data = normalized.get("l2Data")
-        if isinstance(l2_data, dict) and "rtAuto" not in l2_data and "disableRtAuto" in l2_data:
+        if isinstance(l2_data, dict):
             l2_data = dict(l2_data)
-            l2_data["rtAuto"] = not l2_data["disableRtAuto"]
+            l2_data.pop("disableRtAuto", None)
```
allenrobel, line 385: `from_response()` previously translated the GET response's `l2Data.disableRtAuto` into `rtAuto` (`rtAuto = not disableRtAuto`, in place since the module's original commit); this now pops the key without ever populating `rtAuto`, so `rt_auto` is never read back from the controller even though the field still exists on `DefaultL2DataModel` with `alias="rtAuto"`.

Lab-verified on ND 4.2.1 (2026-08-31): `GET /api/v1/manage/fabrics/{fabric}/networks/{name}` still returns `l2Data.disableRtAuto` (and no `rtAuto`). Running this PR's head twice with identical config (`state: merged`, L2 network with `route_target_both: true`) reports `changed: true` on the second run — `before[]` has no `l2_data.rt_auto` while `after[]` has `rt_auto: true`, and that is the only delta. So any playbook managing `route_target_both` on an L2 network gets a perpetual false diff.

Suggest restoring the inversion (or populating `rt_auto` from `disableRtAuto` wherever the read side now normalizes) so `route_target_both` stays idempotent. The renamed test `test_network_response_omits_disable_rt_auto_without_normalizing_rt_auto` would flip back to asserting the translated value.

allenrobel (commented):
### Code review

One finding, lab-verified against ND 4.2.1 — see the inline comment on `network_data_models.py`.

plugins/module_utils/models/manage_networks/network_data_models.py
allenrobel: Follow-up: this deviation is now recorded in the maintainers' bug-tracker vault, so the restored workaround can carry a marker per the repo convention. When you re-add the inversion here, please put this directly above it:

```python
# TODO(4.2.1) network-rt-auto-inverted-echo
# ND's network GET echoes l2Data.disableRtAuto (inverted polarity) and never rtAuto, even though the
# write schema uses rtAuto (POST with rtAuto is accepted). Translate on read so route_target_both
# stays idempotent: rtAuto = not disableRtAuto. Lab-verified on ND 4.2.1.
```

`4.2.1` is the ND version the discrepancy was verified against, and `network-rt-auto-inverted-echo` is the stable vault id — using it verbatim lets the marker resolve to the vault note. The follow-on comment lines are yours to adjust; only the `TODO(4.2.1) network-rt-auto-inverted-echo` line needs to stay exact.

AKDRG: Hey Allen,
Thanks for the comment. 
This was valid for the earlier revision, but the current branch resolves it by removing this field from the managed Network module surface.

route_target_both / rt_auto is no longer exposed as a user-facing argspec field, and the module no longer maps it into the Network payload or read-side managed model. Because of that, `2Data.disableRtAuto is intentionally ignored instead of being inverted back into rtAuto.

The reason we chose removal instead of restoring the inversion is that ND 4.2 does not consistently honor this as a stable user-managed field for the Network module and in the GUI . Keeping it caused false diffs/idempotency drift.

Current behavior after the fix:

- route_target_both is rejected by argspec as unsupported.
- rtAuto is not generated in create/update payloads.
- disableRtAuto from GET is not converted into module intent.
- Repeated runs no longer drift on this field because it is no longer part of the desired/current comparison.

allenrobel (commented):
### Code review

Re-review of head 79360f9c. The earlier `disableRtAuto` finding is resolved by the field removal (verified: no `rt_auto` / `route_target_both` residue in the argspec, payload builder, `DefaultL2DataModel`, docs, or tests, so the `pop` is now inert); I resolved that thread. Two items inline: one code-traced idempotency issue on secondary PVLAN networks under `replaced` / `overridden`, and one optional note on attachment `instanceValues`. One body-level note:

1. The PR body links #458 / #489 as "Related to", so `closingIssuesReferences` is empty. Team convention is a closing keyword in every PR body (CLAUDE.md PR template: "Link issues with `Fixes #xxx`, `Closes #xxx`, or `Resolves #xxx`").

plugins/module_utils/orchestrators/networks.py
```diff
@@ -321,13 +340,76 @@ def _transform_config_to_payload_model_data(self, config: dict[str, Any], fabric
...
+    ) -> dict[str, Any]:
+        template_config = {
+            "isLayer2Only": "true",
+            "networkMode": NetworkLayer.LAYER2.value,
+            "networkName": self._value(config, "network_name", "networkName"),
+            "networkType": NetworkType.USER_DEFINED.value,
+            "nveId": "1",
+            "segmentId": self._template_config_string(self._value(config, "network_id", "networkId")),
+            "type": _PRIVATE_SECONDARY_TEMPLATE_BY_TYPE[vlan_network_type],
+            "vlanId": self._template_config_string(self._value(config, "vlan_id", "vlanId")),
```
allenrobel, line 372: `_private_secondary_payload_model_data` returns no top-level `vlan_id` (the VLAN lives only in `network_template_config["vlanId"]`), but the read side promotes `networkTemplateConfig.vlanId` to a top-level `vlanId` on every queried item (`_schema_fields_from_top_down_template`, called from `_normalize_query_network_item`). Under `replaced` / `overridden`, `get_diff` runs with `exclude_unset=False` and the reverse `has_removals` pass (`models/base.py`, `utils.py`) sees a non-empty `vlanId` on the existing model and no key on the proposed one, so a `community` / `isolated` network classifies as changed on every run. `merged` is unaffected (`exclude_unset=True`).

Traced in code only (no PVLAN on my lab), so please confirm with a second `state: replaced` run. If it reproduces, either keep `vlan_id` populated on the proposed model and drop it from the wire payload in `to_payload`, or add `vlanId` to `reverse_diff_exclude` for the secondary types.

plugins/module_utils/orchestrators/network_attachment_manager.py
```diff
@@ -513,8 +519,20 @@ def attachment_matches(existing: dict[str, Any], desired: dict[str, Any]) -> boo
...
+            "dpu_affinity": attachment_options.get("dpu_affinity"),
+            "svi_enabled": attachment_options.get("svi_enabled"),
+            "switch_route_target_import": attachment_options.get("switch_route_target_import"),
+            "switch_route_target_export": attachment_options.get("switch_route_target_export"),
+            "is_active": attachment_options.get("is_active"),
+        }
+        raw = {key: value for key, value in raw.items() if value is not None}
+        if not raw:
+            return {}
+        return NetworkAttachmentInstanceValuesModel(**raw).to_payload()
```
allenrobel, line 535: `NetworkAttachmentInstanceValuesModel.dpu_secure` defaults to `False` (not `None`), and `to_payload()` only excludes `None`, so any set `attachment_options` key makes this emit `"dpuSecure": false` that the user never specified. I probed this on ND 4.2.1.10: a POST with `{"sviEnabled": true, "dpuSecure": false}` is echoed back verbatim by `networkAttachments/query`, so runs stay idempotent. The residual effect is a silent write of `dpuSecure: false` (spec: "Only valid for smart switches") on a smart switch that had it set elsewhere, when a task names only, say, `svi_enabled`. Same class as the `to_config(exclude_unset=True)` change to `_parse_config` in this PR; `to_payload(exclude_unset=True)` here would match it. Your call whether to change it in this PR.

plugins/module_utils/models/manage_networks/network_data_models.py
allenrobel: Verified on 79360f9c: `route_target_both` / `rt_auto` is gone from the argspec, payload builder, `DefaultL2DataModel`, docs, and tests, so the `disableRtAuto` pop is inert. Resolving.

plugins/module_utils/orchestrators/network_attachment_manager.py
AKDRG: `dpu_secure` now defaults to `None` instead of `False` in both Network and VRF attachment instance-value models. This keeps omitted user intent omitted in the generated payload, while still preserving an explicit `dpu_secure: false`.

I chose the model-level fix instead of switching the whole attachment POST to `exclude_unset=True`, because the broader POST payload has other defaulted fields, such as interface defaults, that should continue to be emitted consistently.

plugins/module_utils/orchestrators/networks.py
AKDRG: The issue was reproduced and addressed for `community` / `isolated` secondary PVLAN networks. The proposed model now retains the user-facing `vlan_id` for diff/idempotence purposes, while the payload builder places the VLAN under `networkTemplateConfig.vlanId`, which is where ND persists it for private secondary networks.

This avoids the repeated `replaced` / `overridden` drift caused by the read side promoting `networkTemplateConfig.vlanId` back to top-level `vlanId`.

allenrobel:
@AKDRG thanks for the quick turnaround on both round-2 threads. I verified "Fix PVLAN network idempotence" and "Fix dpuSecure attachment defaults" on the current head (`ndpytest tests/unit/` passes, 4186 tests) and resolved both threads.

One remaining housekeeping item: the PR body links #458 / #489 with "Related to", so GitHub's closing-issue reference is still empty. Per the team convention that every PR carries a closing keyword, could you add a `Closes #<N>` line (or `Fixes` / `Resolves`) for the issue this PR completes? That is the only thing I have left on this one.

allenrobel (approved):
Both round-2 findings (PVLAN secondary replaced/overridden false diff, dpuSecure default injection) are fixed on head 26fa2a9a and verified locally; full unit suite passes. Approving. See the conversation comment for the one remaining body-text item (Closes #N).

plugins/module_utils/models/manage_networks/config_models.py
```diff
@@ -137,6 +210,46 @@ def _validate_multicast_group(cls, v: str | None) -> str | None:
...
+        monitor_fields = {
+            "vlan_netflow_monitor": vlan_netflow_monitor,
+            "interface_netflow_monitor": interface_netflow_monitor,
+        }
+        configured = [field for field, value in monitor_fields.items() if value is not None]
+        if configured and netflow_enable is not True:
+            raise ValueError("netflow monitor fields require netflow_enable=true: " + ", ".join(configured))
+
 
 class NetworkConfigModel(NDBaseModel):
```
mikewiebe, line 255: ### Medium: Omitting layer passes validation but becomes an invalid layer-3 Network

**Issue**

The public model permits layer to be omitted and only requires vrf_name when
layer is explicitly layer3. Later, the runtime silently converts an omitted
layer to layer3. This creates a configuration that passed local and check-mode
validation but lacks the vrfName required by both active Manage API contracts.

**Evidence**

- [`config_models.py` lines 254-268](https://github.com/CiscoDevNet/ansible-nd/pull/500/files#diff-c6e6119ccf549ab83e20a0ec55da3330680de73b36b9e9e2faa8e21bd92d90f8R254)
  defines layer with a None default.
- [The cross-field validator at lines 413-428](https://github.com/CiscoDevNet/ansible-nd/pull/500/files#diff-c6e6119ccf549ab83e20a0ec55da3330680de73b36b9e9e2faa8e21bd92d90f8R413)
  requires vrf_name only when
  self.layer equals layer3, so None bypasses it.
- [`network_argument_specs.py` lines 107-124](https://github.com/CiscoDevNet/ansible-nd/pull/500/files#diff-40ef579487d1e9cf5406f3972ee1a4ea97e29d6c7fb1f70428487e713f8f3f97R107)
  does not make layer required or assign a default.
- [`network_workflow_coordinator.py` lines 242-273](https://github.com/CiscoDevNet/ansible-nd/pull/500/files#diff-631a8f0f55f6b9f2be39226883a2307ad06a2a63d10a59259db38a3956d8c8dbR242)
  preserves sparse input with exclude_unset.
- [`networks.py` lines 166-170](https://github.com/CiscoDevNet/ansible-nd/pull/500/files#diff-4b5d1318602e47f14efc7a0d6ea27b9517e6d5da198b9479ffa770bbde72b842R166)
  then maps a
  missing layer to layer3. Lines 330-342 synthesize vrf_name=NA only for
  layer2, leaving implicit layer3 without a VRF.
- An exact-head model/runtime probe reproduced the behavior shown below.

**Practical example**

The user omits both `layer` and `vrf_name`:

```yaml
- name: Create a Network
  cisco.nd.nd_manage_networks:
    fabric_name: fab1
    state: merged
    config:
      - network_name: net_no_vrf
        network_id: 30001
        vlan_id: 101
```

Local validation effectively sees:

```yaml
layer: null
vrf_name: null
```

Because `layer` is not explicitly `layer3`, the VRF validation does not run
and the task is accepted. Runtime later produces the equivalent of:

```yaml
networkName: net_no_vrf
layer: layer3
networkId: 30001
vlanId: 101
# vrfName is missing
```

Expected result: the module should either reject the input locally because a
Layer 3 Network requires `vrf_name`, or require the user to specify `layer`
explicitly.

Actual result: check mode can report a valid plan, but normal execution sends
an incomplete Layer 3 Network definition to the controller.

**ND/OpenAPI applicability**

| ND release | Manage API | Result | Evidence |
| --- | --- | --- | --- |
| ND 4.2.1 | 1.1.411 | does not conform | `networkCommon` requires `vrfName`; the implicit layer-3 payload can omit it. |
| ND 4.3.1 | 1.2.224 | does not conform | `networkCommon` still requires `vrfName`; the same payload is invalid. |

Finding scope: all active ND releases.
Target-version evidence: confirmed — both releases are active default baselines for this collection and PR #500 explicitly targets ND 4.2 field alignment.

**Existing PR overlap**

No matching open issue found. [#489](https://github.com/CiscoDevNet/ansible-nd/issues/489)
was inspected and excluded as an exact owner because it covers broader Network
field/PVLAN alignment but does not record this omitted-layer-to-invalid-L3
path. The current-PR delta is actionable.

**Impact**

Check mode can report a valid plan even though normal execution sends an
invalid or ambiguous layer-3 definition. Failure is deferred to the controller,
and behavior can vary by controller version or topology instead of failing
locally with a clear message.

**Suggested fix**

Make the layer default explicit before cross-field validation and require
vrf_name when that effective value is layer3, or require layer whenever a
Network definition is supplied. Add omitted-layer tests for standalone and
parent workflows in both check and normal mode, asserting the exact payload.

---

plugins/module_utils/models/manage_networks/config_models.py
```diff
@@ -28,21 +32,75 @@
...
+        NetworkAttachmentMode.TRUNK.value,
+    }
+)
+_PRIVATE_PRIMARY_INTERFACE_MODES = frozenset(
+    {
+        NetworkAttachmentMode.PROMISCUOUS.value,
+        NetworkAttachmentMode.TRUNK_PROMISCUOUS.value,
+    }
+)
+_PRIVATE_SECONDARY_INTERFACE_MODES = frozenset(
```
mikewiebe, line 64: ### Medium: The module rejects the PVLAN host mode used by ND 4.3.1

**Issue**

ND 4.3.1 uses `pvlanHost` for attaching an interface as a PVLAN host and marks
the older `host` value as deprecated for this purpose. The module accepts only
`host`, so it rejects the ND 4.3.1 value before contacting the controller.

**Evidence**

- [`config_models.py` lines 63-68](https://github.com/CiscoDevNet/ansible-nd/pull/500/files#diff-c6e6119ccf549ab83e20a0ec55da3330680de73b36b9e9e2faa8e21bd92d90f8R63)
  accepts only `host` and `trunkSecondary`.
- [`enums.py` lines 144-157](https://github.com/CiscoDevNet/ansible-nd/pull/500/files#diff-b544691764257ac253ca7d756444c361423dfc98f287671aca264ae6583a3a9fR144)
  omits `pvlanHost` from `NetworkAttachmentMode`.
- [`config_models.py` lines 91-96](https://github.com/CiscoDevNet/ansible-nd/pull/500/files#diff-c6e6119ccf549ab83e20a0ec55da3330680de73b36b9e9e2faa8e21bd92d90f8R91)
  rejects values outside that enum.
- [`nd_manage_networks.py` lines 160-168](https://github.com/CiscoDevNet/ansible-nd/pull/500/files#diff-28c2ca22beff4f6a57eb86c269dffc3e68f3a06ef845597d91e11b9e8427f89eR160)
  documents the same older
  choices.

**Practical example**

A user running ND 4.3.1 submits:

```yaml
- name: Attach a community Network as a PVLAN host
  cisco.nd.nd_manage_networks:
    fabric_name: fab1
    state: merged
    config:
      - network_name: community_101
        vlan_network_type: community
        network_id: 50101
        primary_network_id: 50100
        vlan_id: 101
        attach:
          - ip_address: 192.0.2.11
            interfaces:
              - interface_range: Ethernet1/10
                mode: pvlanHost
```

Expected result: the module accepts the ND 4.3.1 value and sends:

```yaml
mode: pvlanHost
```

Actual result: the module rejects `pvlanHost` during local validation and
never sends a request to ND. Changing the input to the following avoids the
local error:

```yaml
mode: host
```

However, `host` is the older ND 4.2.1 value and is deprecated for this use in
ND 4.3.1.

**ND/OpenAPI applicability**

| ND release | Manage API | Result | Evidence |
| --- | --- | --- | --- |
| ND 4.2.1 | 1.1.411 | conforms | The schema and PVLAN-host POST example use `mode: host`, which the module accepts. |
| ND 4.3.1 | 1.2.224 | contract differs | The enum adds `pvlanHost`, deprecates `host` for this use, and the module rejects `pvlanHost`. |

Finding scope: ND 4.3.1 only.
Target-version evidence: confirmed — ND 4.3.1 is an active default baseline in the collection's version registry.

**Existing PR overlap**

No matching open issue found. [#489](https://github.com/CiscoDevNet/ansible-nd/issues/489)
was inspected and excluded as an exact owner because it covers PVLAN support
generally but does not identify this difference between ND 4.2.1 and ND 4.3.1.

**Impact**

Users cannot use the value documented for ND 4.3.1. If a future ND release
stops accepting the older `host` value, PVLAN host attachments made through
this module will fail completely.

**Suggested fix**

Accept both values and send the correct one for the detected ND version:

- ND 4.2.1: `host`
- ND 4.3.1: `pvlanHost`

Add tests that verify the exact request sent for each ND version.

---

plugins/module_utils/orchestrators/network_argument_specs.py
```diff
@@ -29,110 +36,86 @@ def _network_interface_spec():
...
         ipv6_trm=dict(type="bool"),
         netflow_enable=dict(type="bool"),
+        vlan_netflow_monitor=dict(type="str"),
+        interface_netflow_monitor=dict(type="str"),
         gateway_on_border=dict(type="bool"),
-        l3gw_on_border=dict(type="bool"),
     )
 
 
 def network_base_argument_spec():
```
mikewiebe, line 107: ### Medium: Examples and tests still use the removed `is_l2only` setting

**Issue**

This PR replaces `is_l2only` with `layer`, but a runnable module example and
the staged integration tests still use `is_l2only`. Because the module no
longer accepts that setting, Ansible stops during input validation before
contacting ND.

**Evidence**

- [`network_argument_specs.py` lines 107-124](https://github.com/CiscoDevNet/ansible-nd/pull/500/files#diff-40ef579487d1e9cf5406f3972ee1a4ea97e29d6c7fb1f70428487e713f8f3f97R107)
  exposes layer and has no is_l2only
  option.
- [`nd_manage_networks.py` lines 548-567](https://github.com/CiscoDevNet/ansible-nd/pull/500/files#diff-28c2ca22beff4f6a57eb86c269dffc3e68f3a06ef845597d91e11b9e8427f89eR548)
  still uses is_l2only in runnable examples.
- There are 45 stale occurrences across that EXAMPLES block and:
  - [`tests/integration/targets/nd_manage_networks/tests/standalone/staged.yaml`](https://github.com/CiscoDevNet/ansible-nd/pull/500/files#diff-42d0af2f59314d8bcef8db573089ff66b0c1c5e184d479f433bfa1be7a3a2f0aR1)
  - tests/integration/targets/nd_manage_networks/tests/msd/staged.yaml
  - tests/integration/targets/nd_manage_networks/tests/mcfg/staged.yaml
- The review helper reports undocumented `config[].is_l2only` properties.
- A real module-input validation probe returned:
  `config.is_l2only. Supported parameters include ... layer ...`

**Practical example**

The shipped overridden-state example contains both the replacement setting
and the removed setting:

```yaml
- name: Override Network configuration
  cisco.nd.nd_manage_networks:
    fabric_name: fab1
    state: overridden
    config:
      - network_name: Network_BLUE
        layer: layer2
        network_id: 50010
        vlan_id: 2001
        vlan_name: Network_BLUE_REPLACED
        is_l2only: true
```

A user copying this example receives an error similar to:

```text
Unsupported parameter: config.is_l2only
Supported parameters include: layer
```

No request is sent to ND. The corrected task should remove `is_l2only`:

```yaml
- name: Override Network configuration
  cisco.nd.nd_manage_networks:
    fabric_name: fab1
    state: overridden
    config:
      - network_name: Network_BLUE
        layer: layer2
        network_id: 50010
        vlan_id: 2001
        vlan_name: Network_BLUE_REPLACED
```

Expected result: every example and integration test uses `layer: layer2`.

Actual result: the remaining `is_l2only` setting causes input validation to
fail before the Network can be checked, staged, or changed.

**ND/OpenAPI applicability**

**Existing PR overlap**

No matching open issue found. [#489](https://github.com/CiscoDevNet/ansible-nd/issues/489)
was inspected and excluded as an exact owner because it covers the wider
Network-field change but does not specifically track these broken examples and
integration tests.

**Impact**

- Users copying the documentation receive an immediate error.
- The staged integration tests cannot test their intended behavior.
- Current CI does not expose the problem because the integration-test job is
  skipped.

**Suggested fix**

Replace every remaining `is_l2only` occurrence with the correct `layer` value,
normally `layer: layer2`. Then run the example checker and the standalone,
MSD, and MCFG staged integration tests.

---

plugins/modules/nd_manage_networks.py
```diff
@@ -593,9 +658,12 @@
   elements: dict
```
mikewiebe, line 658: ### Low: Module documentation does not explain all returned result fields

**Issue**

The three modules changed by this PR do not fully document the information they
can return. `nd_manage_switches` has an empty `RETURN` section, while
`nd_manage_networks` and `nd_manage_vrfs` omit fields that their code can
return, including `output_level`, `proposed`, `logs`, `gathered`, and failure
`msg`.

**Evidence**

- [`nd_manage_networks.py` lines 577-659](https://github.com/CiscoDevNet/ansible-nd/pull/500/files#diff-28c2ca22beff4f6a57eb86c269dffc3e68f3a06ef845597d91e11b9e8427f89eR658)
  does not document every supported result field.
- [`nd_manage_switches.py` lines 415-416](https://github.com/CiscoDevNet/ansible-nd/blob/3f1c6994b99d07af064ca5b720b011e7d2f6fec3/plugins/modules/nd_manage_switches.py#L415-L416)
  has an empty `RETURN` block.
- [`nd_manage_vrfs.py` lines 647-729](https://github.com/CiscoDevNet/ansible-nd/blob/3f1c6994b99d07af064ca5b720b011e7d2f6fec3/plugins/modules/nd_manage_vrfs.py#L647-L729)
  also omits supported result fields.
- [The switch runtime at `nd_switch_resources.py` lines 3035-3044](https://github.com/CiscoDevNet/ansible-nd/blob/3f1c6994b99d07af064ca5b720b011e7d2f6fec3/plugins/module_utils/manage_switches/nd_switch_resources.py#L3035-L3044)
  returns information that has no public `ansible-doc` description.

**Practical example**

A user gathers the switches and registers the result:

```yaml
- name: Gather the switches managed by the fabric
  cisco.nd.nd_manage_switches:
    fabric_name: fab1
    state: gathered
  register: switch_result

- name: Display the gathered switches
  ansible.builtin.debug:
    var: switch_result.gathered
```

The user then runs:

```bash
ansible-doc cisco.nd.nd_manage_switches
```

The module's `RETURN` section is empty. It does not explain whether `gathered`
is valid, when it is returned, whether it contains a list or dictionary, or
what each item contains.

A similar problem occurs with a Network check-mode task:

```yaml
- name: Preview a Network change
  cisco.nd.nd_manage_networks:
    fabric_name: fab1
    state: merged
    config:
      - network_name: Network_BLUE
        layer: layer2
        network_id: 50010
        vlan_id: 2001
  check_mode: true
  register: network_result

- name: Display the proposed change
  ansible.builtin.debug:
    var: network_result.proposed
```

The runtime can return `proposed`, but the module documentation does not
explain its type, contents, or when it is available.

Expected result: `ansible-doc` describes each supported field. For example:

```yaml
gathered:
  description: Switches returned when state is gathered.
  returned: When state is gathered
  type: list
  elements: dict
```

Actual result: the fields may exist at runtime, but users must inspect the
implementation or experiment with the module to learn how to use them.

**ND/OpenAPI applicability**

**Existing PR overlap**

No matching open issue found. [#304](https://github.com/CiscoDevNet/ansible-nd/issues/304)
and [#530](https://github.com/CiscoDevNet/ansible-nd/issues/530) were inspected
and excluded as exact owners because they cover wider result-format consistency
but not the module-specific documentation required here.

**Impact**

Users cannot confidently use registered results in playbooks, and
`ansible-doc` cannot explain gathered, check-mode, verbose, or failure output.

**Suggested fix**

Document `changed`, `output_level`, `before`, `after`, `diff`, `proposed`,
`logs`, `gathered`, and failure `msg` as applicable. For every field, include
its type, when it is returned, and a representative sample. Use
`plugins/modules/nd_interface_port_channel_access.py` as the current reference
format.

plugins/module_utils/models/manage_networks/config_models.py
AKDRG: Fixed. 
The fix makes the Network layer default explicit during model validation instead of letting it stay None and later silently become layer3 in payload construction.
Now, if a user omits layer, the module treats it as the effective default layer and applies the same cross-field validation immediately. That means a Network that resolves to Layer 3 must provide vrf_name locally, including in check mode, before any controller request is attempted.

plugins/module_utils/models/manage_networks/config_models.py
AKDRG: Handled it for 4.3 and lesser.
Added handling to make the payloads depending on the controller version.
- ND 4.2.x sends host
- ND 4.3.x and newer sends pvlanHost
Also, the fix separates the user-facing interface mode from the controller API value using snake case value for pvlan_host. This avoids exposing version-specific API spelling directly in the playbook. Idempotency is handled by converting the readback.

plugins/module_utils/orchestrators/network_argument_specs.py
AKDRG: Fixed!

plugins/modules/nd_manage_networks.py
AKDRG: Fixed!

Issue #457, opened 2026-07-30
Discuss: should RestSend retry mutations (POST/PUT/DELETE) that fail with non-success codes?


allenrobel:
## Context

PR #398 (fixes #295) added a terminal-vs-retryable classification to the shared REST layer, prompted by [mikewiebe's finding](https://github.com/CiscoDevNet/ansible-nd/pull/398#discussion_r3679248511) that `RestSend` replayed a non-idempotent batch mutation ~60 times over its 300-second budget when the failure was a deterministic per-item application rejection (live-verified on ND 4.2.1: replays turned an already-succeeded ACL create into "already exists" failures).

The fix classifies a failure as **terminal** (`retryable=False`, one submission) when the RETURN_CODE is a success code but the body carries an embedded error — the application definitively rejected the request. Failures with **non-success codes** (5xx, timeouts) deliberately kept the historical retry behavior, scoped out of #398 to limit blast radius.

## The open question

For POST/PUT/DELETE, a 5xx does not guarantee the mutation was **not** applied server-side. Replaying it carries the same duplicate-side-effect risk the terminal case demonstrated — just triggered by a transient-looking failure instead of a deterministic one. Today `ResponseHandler` marks every non-success-code mutation failure `retryable=True` and `RestSend._commit_normal_mode()` resubmits the identical payload until timeout (default 300 s / 5 s interval).

GET retries are not in question — they are idempotent and serve eventual-consistency polling.

## Possible directions

1. **Never retry mutations at the transport layer** — push retries to orchestrators, which have the domain knowledge to decide (e.g. L3Out's bounded "not found" attach retry). Simplest rule, but removes transient-recovery behavior existing modules may rely on.
2. **Classify by failure mode** — retry only failures where the request provably never reached the application (connection errors; possibly 502/503/504), treat HTTP 500 like the terminal case. Closer to correct, needs agreement on which codes are "not processed".
3. **Keep as-is and document** — accept the replay risk for 5xx as the availability trade-off.

The seam from #398 makes any of these a one-line policy change in `ResponseHandler._handle_post_put_delete_response()` (the `retryable` classification); `RestSend` needs no further changes.

## References

- `plugins/module_utils/rest/response_handler_nd.py` — `retryable` classification
- `plugins/module_utils/rest/rest_send.py` — `_commit_normal_mode()` retry loop and terminal break
- PR #398 discussion threads; issue #295; #397 (detector consolidation, related but distinct)

[...]

Issue #188, opened 2026-03-04
CR: Should ResponseHandler convert response to Pydantic model? (DCNE-690)


allenrobel:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Look into whether ResponseHandler (response_handler_nd.py) should convert the response to a Pydantic model, or if this should be handled by an injected dependency.

### New or Affected Module(s):

* Infrastructure related. No new modules.

### ND version and ND Platform

* V 4.2

### Collection versions

* Future work

### References

* https://github.com/CiscoDevNet/ansible-nd/pull/185#discussion_r2884912465)

PR #502, opened 2026-08-11
Classify 4xx (except 429) as terminal in RestSend retry policy

allenrobel:
## Related Issue(s)

Addresses the 4xx half of #457 (deliberately does NOT close it — the 5xx retry question remains open for discussion there).

## Proposed Changes

Implements the independently-shippable piece proposed in [#457's field-report comment](https://github.com/CiscoDevNet/ansible-nd/issues/457#issuecomment-5220049760): a 4xx response proves the request reached the application and was rejected, so replaying the identical request cannot succeed. Previously a deterministic 400 (e.g. an immediately-rejected fabric mutation) was resubmitted every 5 seconds for the full 300-second `RestSend` budget, delaying module failure by 5 minutes with no user-facing knob to shorten it.

- `NdV1Strategy.is_terminal_client_error(return_code, verb)` — the review round moved the classification from the handler onto the injected `ResponseValidationStrategy`, the handler's documented extension point for version-specific status-code policy. A 4xx is terminal unless transient: 408/421/425/429 for every verb (RFC 9110 §15.5.9/§15.5.20, RFC 8470 §5.2, rate limiting; ND 4.2.1 documents none of 408/421/425, so retaining them preserves historical behavior), plus 409 for GET only — the ND 4.2.1 OpenAPI documents 409 on safe GETs (manage v1.1.411: `/links`, `/links/{linkId}`, `/logicalLinks`, `/remoteFabrics`, `/anomalyRules/postProcessingRules`; onemanage: three more) where the conflict is with the resource's current state and can clear on its own. Mutation 409 stays terminal (deterministic create/update conflict). The dcnm-era retry-on-400 cases do not carry over.
- **Mutations (POST/PUT/DELETE):** the `retryable` classification introduced by #398 now also excludes terminal client errors. 5xx keeps the historical retry behavior — that half of #457 is untouched.
- **GET:** results now carry `retryable`, applying the same 4xx rule (with the GET-only 409 carve-out above) per the issue's scope note — a deterministic 400 on a GET burned the same budget. 5xx GETs remain retryable for eventual-consistency polling, and the 404 not-found-is-success contract is unchanged.
- `RestSend` needs no changes — the terminal break from #398 already consumes `retryable` verb-agnostically.
- **Strategy-injection compatibility (review round 2):** `is_terminal_client_error()` lives on a separate `@runtime_checkable` `TerminalClientErrorPolicy` capability protocol rather than on `ResponseValidationStrategy`, so the assignment-time `isinstance` check in `ResponseHandler.validation_strategy` is back to the original contract and a strategy written against the pre-PR protocol is still accepted. The handler feature-detects the capability (`_is_terminal_client_error()` via `getattr`) and falls back to the historical retryable=True for every non-success 4xx when the injected strategy lacks it. `NdV1Strategy` satisfies both protocols and is unchanged.

Test changes: five new ResponseHandler tests (01500–01540: POST 400 terminal, POST 429 retryable, GET 400 terminal, GET 404 contract, DELETE 404 terminal) and a deliberate rewrite of 01440, which previously pinned the "GET results carry no retryable key" behavior this PR removes. The review round added handler tests 01550–01590 (GET 409 retryable, POST 409 terminal, POST 408 / GET 425 / GET 421 retryable) and RestSend test 01120 (GET 409 retries and consumes the recovery 200). Round 2 added a shared `LegacyStrategy` test helper (pre-PR protocol members only), handler tests 01700–01730 (legacy strategy accepted at assignment, POST 400 / GET 400 keep retryable=True, protocol split) and RestSend test 01130 (legacy strategy injected end-to-end: POST 400 retries and consumes the recovery 200).

## Test Notes

- Full unit suite passes in the nd-dev container machine: 4295 passed (`ndpytest tests/unit/`)
- `pylint` 10.00/10 on the changed modules; `black`/`isort` clean; `mypy` reports only one error pre-existing on develop (the review round incidentally fixed the other pre-existing one)

4.2.1

* [ ] analyze
* [ ] infra
* [ ] manage
* [ ] onemanage
* [x] other

## Checklist

https://claude.ai/code/session_019XZiWNrVHYXbeNXRM8kUrz

### DIFF
--- a/plugins/module_utils/rest/protocols/response_validation.py
+++ b/plugins/module_utils/rest/protocols/response_validation.py
@@ -14,6 +14,9 @@ including status code validation and error message extraction.
 
 When ND API v2 is released with different status codes or response formats,
 implementing a new strategy class allows clean separation of v1 and v2 logic.
+
+`TerminalClientErrorPolicy` is an optional, separately-checked capability: a strategy that also implements it opts in to
+classifying 4xx client errors as terminal (not retryable). Strategies that omit it remain valid and keep every 4xx retryable.
 """
 
 # isort: off
@@ -145,34 +148,6 @@ class ResponseValidationStrategy(Protocol):
         """
         ...
 
-    def is_terminal_client_error(self, return_code: int, verb: HttpVerbEnum) -> bool:
-        """
-        # Summary
-
-        Check whether `return_code` is a 4xx client error that is terminal (not retryable) for `verb`.
-
-        ## Description
-
-        A 4xx response proves the request reached the application and was rejected, so an identical replay is deterministic and retrying wastes the
-        retry budget. Implementations decide which 4xx codes are nevertheless transient (and how that depends on the verb) for their API version —
-        e.g. rate limiting, or conflict codes the API documents on safe reads. Non-4xx codes must return False: this method classifies only client
-        errors, leaving success and 5xx policy to the caller.
-
-        ## Parameters
-
-        - return_code: HTTP status code to check
-        - verb: The HTTP verb of the request the response answers
-
-        ## Returns
-
-        - True if the code is a 4xx client error that is terminal for `verb`, False otherwise
-
-        ## Raises
-
-        None
-        """
-        ...
-
     def is_changed(self, response: dict) -> bool:
         """
         # Summary
@@ -249,5 +224,54 @@ class ResponseValidationStrategy(Protocol):
         ## Raises
 
         None - should return None gracefully if error message cannot be extracted
+        """
+        ...
+
+
+@runtime_checkable
+class TerminalClientErrorPolicy(Protocol):  # pylint: disable=too-few-public-methods
+    """
+    # Summary
+
+    Optional capability protocol: classify 4xx client errors as terminal (not retryable) or transient for a given verb.
+
+    ## Description
+
+    This is deliberately separate from `ResponseValidationStrategy`, which is the contract `ResponseHandler.validation_strategy` enforces at
+    assignment. A strategy written against the original protocol (every member of `ResponseValidationStrategy`, nothing more) remains valid and
+    keeps the historical behavior — every non-success 4xx stays retryable. A strategy that additionally satisfies this protocol opts in to
+    terminal-4xx classification: `ResponseHandler` feature-detects `is_terminal_client_error` and consults it before deciding `retryable`
+    (issue #457). `NdV1Strategy` implements both protocols.
+
+    ## Raises
+
+    None - implementations may raise exceptions per their logic
+    """
+
+    def is_terminal_client_error(self, return_code: int, verb: HttpVerbEnum) -> bool:
+        """
+        # Summary
+
+        Check whether `return_code` is a 4xx client error that is terminal (not retryable) for `verb`.
+
+        ## Description
+
+        A 4xx response proves the request reached the application and was rejected, so an identical replay is deterministic and retrying wastes the
+        retry budget. Implementations decide which 4xx codes are nevertheless transient (and how that depends on the verb) for their API version —
+        e.g. rate limiting, or conflict codes the API documents on safe reads. Non-4xx codes must return False: this method classifies only client
+        errors, leaving success and 5xx policy to the caller.
+
+        ## Parameters
+
+        - return_code: HTTP status code to check
+        - verb: The HTTP verb of the request the response answers
+
+        ## Returns
+
+        - True if the code is a 4xx client error that is terminal for `verb`, False otherwise
+
+        ## Raises
+
+        None
         """
         ...
--- a/plugins/module_utils/rest/protocols/response_validation.py
+++ b/plugins/module_utils/rest/protocols/response_validation.py
@@ -43,6 +43,8 @@ except ImportError:
 
 from typing import Optional
 
+from ansible_collections.cisco.nd.plugins.module_utils.enums import HttpVerbEnum
+
 # pylint: disable=unnecessary-ellipsis
 
 
@@ -136,6 +138,34 @@ class ResponseValidationStrategy(Protocol):
         ## Returns
 
         - True if code matches not_found_code, False otherwise
+
+        ## Raises
+
+        None
+        """
+        ...
+
+    def is_terminal_client_error(self, return_code: int, verb: HttpVerbEnum) -> bool:
+        """
+        # Summary
+
+        Check whether `return_code` is a 4xx client error that is terminal (not retryable) for `verb`.
+
+        ## Description
+
+        A 4xx response proves the request reached the application and was rejected, so an identical replay is deterministic and retrying wastes the
+        retry budget. Implementations decide which 4xx codes are nevertheless transient (and how that depends on the verb) for their API version —
+        e.g. rate limiting, or conflict codes the API documents on safe reads. Non-4xx codes must return False: this method classifies only client
+        errors, leaving success and 5xx policy to the caller.
+
+        ## Parameters
+
+        - return_code: HTTP status code to check
+        - verb: The HTTP verb of the request the response answers
+
+        ## Returns
+
+        - True if the code is a 4xx client error that is terminal for `verb`, False otherwise
 
         ## Raises
 
--- a/plugins/module_utils/rest/response_handler_nd.py
+++ b/plugins/module_utils/rest/response_handler_nd.py
@@ -170,6 +170,25 @@ class ResponseHandler:
         else:
             self._handle_post_put_delete_response()
 
+    def _is_terminal_client_error(self, return_code) -> bool:
+        """
+        # Summary
+
+        Return True when `return_code` is a 4xx client error other than 429 (Too Many Requests).
+
+        ## Description
+
+        A 4xx response proves the request reached the application and was rejected: replaying the identical request
+        cannot succeed, so the failure is terminal for every verb. 429 is the one transient 4xx (rate limiting) and
+        stays retryable. No retryable 4xx is documented for any ND 4.2.1 endpoint (the dcnm-era retry-on-400 cases
+        do not carry over). See issue #457.
+
+        ## Raises
+
+        None
+        """
+        return isinstance(return_code, int) and 400 <= return_code <= 499 and return_code != 429
+
     def _handle_get_response(self) -> None:
         """
         # Summary
@@ -183,6 +202,11 @@ class ResponseHandler:
             -   success:
                     -   True if RETURN_CODE in (200, 201, 202, 204, 207, 404)
                     -   False otherwise (error status codes)
+            -   retryable:
+                    -   False when the request succeeded, or when it failed with a 4xx code other than 429 (the
+                        request reached the application and was rejected; an identical replay cannot succeed)
+                    -   True when it failed with any other code (e.g. 5xx — potentially transient, and GET retries
+                        also serve eventual-consistency polling)
         """
         result = {}
         return_code = self.response.get("RETURN_CODE")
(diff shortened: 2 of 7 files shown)


akinross (approved):

plugins/module_utils/rest/response_handler_nd.py
```diff
@@ -170,6 +170,25 @@ def _handle_response(self) -> None:
...
+        A 4xx response proves the request reached the application and was rejected: replaying the identical request
+        cannot succeed, so the failure is terminal for every verb. 429 is the one transient 4xx (rate limiting) and
+        stays retryable. No retryable 4xx is documented for any ND 4.2.1 endpoint (the dcnm-era retry-on-400 cases
+        do not carry over). See issue #457.
+
+        ## Raises
+
+        None
+        """
+        return isinstance(return_code, int) and 400 <= return_code <= 499 and return_code != 429
```
mikewiebe: ### Medium: Preserve retryable 4xx responses instead of terminalizing the entire class

**Issue**

`_is_terminal_client_error()` treats every 4xx except 429 as a definitive application rejection for every verb. That is too broad: HTTP 408, 421, and 425 explicitly support retry, and the bundled ND OpenAPI documents 409 on safe GET operations whose conflict is tied to the resource's current state.

**Evidence**

- [`plugins/module_utils/rest/response_handler_nd.py` line 190](https://github.com/CiscoDevNet/ansible-nd/pull/502/files#diff-afe56188371bcdae64893568b8f16526417bdc5fb02be25c5b8c35eed1bf7c8eR190)
  classifies the full `400..499` range as terminal with only 429 excluded.
- [`plugins/module_utils/rest/response_handler_nd.py` lines 225-228](https://github.com/CiscoDevNet/ansible-nd/pull/502/files#diff-afe56188371bcdae64893568b8f16526417bdc5fb02be25c5b8c35eed1bf7c8eR228)
  applies that same predicate to GET, even though GET is safe/idempotent and its existing retries also serve polling and eventual-consistency workflows.
- [`plugins/module_utils/rest/rest_send.py` lines 377-384](https://github.com/CiscoDevNet/ansible-nd/blob/5292577365d6d0ab30186935844ea832c7dd2167/plugins/module_utils/rest/rest_send.py#L377-L384)
  immediately exits the retry loop when the handler returns `retryable=False`.
- The bundled Manage OpenAPI v1.1.411 spec documents 409 on 23 operations, including five GETs: `GET /anomalyRules/postProcessingRules`, `GET /links`, `GET /links/{linkId}`, `GET /logicalLinks`, and `GET /remoteFabrics`. Its shared response says the conflict is with the resource's current state; the spec contains no retryability metadata that supports blanket terminalization.
- [RFC 9110 sections 15.5.9 and 15.5.20](https://www.rfc-editor.org/rfc/rfc9110.html#section-15.5.9) permit retrying 408 and 421 respectively; 421 can be retried even for a non-idempotent method when a different connection is used. [RFC 8470 section 5.2](https://www.rfc-editor.org/rfc/rfc8470.html#section-5.2) expects an automatic retry after 425 outside early data.
- A local check against the exact PR head returned `retryable=False` for GET 408, 409, and 425, while only GET 429 remained retryable.

**Existing PR overlap**

Related open issue: [#457](https://github.com/CiscoDevNet/ansible-nd/issues/457) is the design source for this PR and proposes the same blanket 4xx-except-429 rule. It does not track the protocol-level 408/421/425 exceptions or the OpenAPI-documented GET 409 counterexample, so this finding narrows and corrects that policy rather than duplicating tracked work.

**Impact**

A concrete ND scenario is topology reconciliation after a switch is added or updated. An Ansible module requests `GET /links` while ND is still reconciling the fabric and receives the OpenAPI-documented `409 Conflict` because the link data is temporarily unavailable in its current state. A retry a few seconds later can return `200 OK` with the completed link data.

```text
develop: GET /links -> 409 Conflict -> retry -> 200 OK -> task succeeds
PR #502: GET /links -> 409 Conflict -> stop retrying -> task fails
```

The operator therefore sees a failed automation job even though nothing is permanently wrong, and running the same playbook again moments later can succeed. This is a representative scenario inferred from ND's OpenAPI contract, not a live-testbed reproduction. The same blanket classification also suppresses standards-defined retry handling for 408, 421, and 425, and because `ResponseHandler` is shared, the regression affects every module using RestSend.

**Suggested fix**

Split safe GET policy from mutation policy and replace the blanket range check with explicit, evidence-backed classifications. At minimum, keep 408 and 425 retryable; preserve retry/polling for safe GET 409; and handle 421 by reopening or changing the connection before retrying. Put the version-specific decision on the injected `ResponseValidationStrategy` (or another policy object), matching this file's documented extension point, and add handler plus RestSend tests for the retained retry cases.

---

allenrobel: Addressed in commit "Keep RFC-transient 4xx and safe-GET 409 retryable; move the policy to NdV1Strategy":

- 408, 421, and 425 join 429 in the transient set for every verb. Verified against the bundled 4.2.1 specs that ND documents none of the three on any endpoint or section, so this simply preserves the pre-#502 retry behavior for codes ND is not expected to send.
- GET 409 is retryable again. I reproduced your spec query locally (manage v1.1.411: 23 operations documenting 409, the same five GETs; onemanage documents three more), and the shared 409 response ties the conflict to the resource's current state — so GET retries/polls through it.
- The classification now lives on `NdV1Strategy.is_terminal_client_error(return_code, verb)` with the protocol updated, per the handler's documented extension point for version-specific status-code policy.

Two deliberate deviations from the suggested fix:

- Mutation 409 stays terminal. The 18 mutation operations documenting 409 in the manage spec are create/update conflicts (resource already exists), which an identical replay cannot resolve — retrying those would reintroduce the full-retry-budget stall this PR removes. I read that as consistent with your suggested fix, which scopes the 409 carve-out to safe GETs.
- No connection handling for 421. RFC 9110 wants the 421 retry on a different connection, but connection lifecycle belongs to Ansible's persistent httpapi transport, which `RestSend` cannot reach (that layer re-establishes a dropped connection transparently, so a retry rides whatever connection it provides). 421 is classified transient and the caveat is documented at the constant.

Coverage: handler tests 01550–01590 (GET 409 retryable, POST 409 terminal, POST 408 / GET 425 / GET 421 retryable) and RestSend test 01120, which is your Impact scenario — GET /links returns 409, the retry consumes the recovery 200.

plugins/module_utils/rest/protocols/response_validation.py
```diff
@@ -143,6 +145,34 @@ def is_not_found(self, return_code: int) -> bool:
         """
         ...
 
+    def is_terminal_client_error(self, return_code: int, verb: HttpVerbEnum) -> bool:
```
mikewiebe: ### Medium: Preserve compatibility for previously valid injected response strategies

**Issue**

Adding `is_terminal_client_error()` to the runtime-checkable `ResponseValidationStrategy` makes it mandatory for every injected strategy. A strategy that satisfied the complete pre-PR protocol and was accepted by `ResponseHandler.validation_strategy` now raises `TypeError` at assignment.

**Evidence**

- [`plugins/module_utils/rest/protocols/response_validation.py` line 148](https://github.com/CiscoDevNet/ansible-nd/pull/502/files#diff-899320f690a986ee9404ac46c2de2b1bf99029cbea5b73fea4c54a6d47f48a2fR148)
  adds the classifier as a required member of the `@runtime_checkable` protocol.
- [`plugins/module_utils/rest/response_handler_nd.py` lines 414-431](https://github.com/CiscoDevNet/ansible-nd/blob/dc7b723a7e065d0b3d6e1ff3111c08aab52643fb/plugins/module_utils/rest/response_handler_nd.py#L414-L431)
  rejects injected objects that no longer satisfy the widened protocol.
- [`plugins/module_utils/rest/response_handler_nd.py` lines 23-24 and 47-51](https://github.com/CiscoDevNet/ansible-nd/blob/dc7b723a7e065d0b3d6e1ff3111c08aab52643fb/plugins/module_utils/rest/response_handler_nd.py#L23-L51)
  documents strategy injection as the supported extension point for another API version.
- A local compatibility reproducer assigned a structural strategy implementing every pre-PR protocol member. Exact base accepted it and printed `accepted`; current head raised `TypeError: ResponseHandler.validation_strategy: Expected ResponseValidationStrategy`.

**Existing PR overlap**

No matching open issue found. Open issue [#188](https://github.com/CiscoDevNet/ansible-nd/issues/188) concerns converting responses to Pydantic models, and [#457](https://github.com/CiscoDevNet/ansible-nd/issues/457) concerns retry classification; neither tracks compatibility for injected strategies.

**Impact**

An existing out-of-tree or downstream API-version strategy that was valid before this PR can no longer be assigned, even though its prior response behavior remains implementable. This turns an additive retry policy into a breaking change at the handler's documented extension seam.

For example, a downstream team supporting an upcoming ND API version can inject an `NdV2Strategy` that implements every member of the pre-PR protocol:

```python
class NdV2Strategy:
    @property
    def success_codes(self):
        return {200, 201, 202, 204, 207}

    @property
    def not_found_code(self):
        return 404

    def is_success(self, response):
        return response["RETURN_CODE"] in self.success_codes

    def is_not_found(self, return_code):
        return return_code == self.not_found_code

    def is_changed(self, response):
        return response.get("modified") != "false"

    def is_changed_on_failure(self, response):
        return False

    def extract_error_message(self, response):
        return response.get("MESSAGE")


handler = ResponseHandler()
handler.validation_strategy = NdV2Strategy()
```

The exact base accepts this strategy. With PR #502, assignment fails before the first API request:

```text
TypeError: ResponseHandler.validation_strategy:
Expected ResponseValidationStrategy. Got <class 'NdV2Strategy'>.
```

Ordinary playbook users are unlikely to hit this directly, but downstream developers using the documented strategy-injection extension point must update in lockstep with the collection.

**Suggested fix**

Keep terminal-client-error classification as an optional capability: split it into a separate capability protocol or feature-detect it with `getattr()`, using the historical `retryable=True` behavior when an older strategy lacks the method. Preserve assignment compatibility for the pre-change protocol and add a regression test that injects and executes a legacy structural strategy.

---

allenrobel: Fixed in commit "Make terminal-4xx classification an optional strategy capability", taking the split-protocol variant of the suggested fix:

- `is_terminal_client_error()` moved off `ResponseValidationStrategy` onto a separate `@runtime_checkable` `TerminalClientErrorPolicy` capability protocol, so the assignment-time `isinstance` check in `ResponseHandler.validation_strategy` is back to the original contract. Your `NdV2Strategy` reproducer is accepted again.
- `ResponseHandler._is_terminal_client_error()` feature-detects the capability with `getattr()`; when the injected strategy lacks it the handler returns "not terminal", so every non-success 4xx keeps the historical retryable=True rather than the handler terminalizing on the strategy's behalf. Both call sites (GET and mutation paths) go through the helper.
- `NdV1Strategy` is unchanged and satisfies both protocols.

Regression coverage: a shared `tests/unit/module_utils/legacy_response_strategy.py` (`LegacyStrategy`, pre-PR members only — your reproducer, essentially), handler tests 01700–01730 (assignment accepted, POST 400 and GET 400 stay retryable, `isinstance` split holds for both strategies), and RestSend test 01130 which injects the legacy strategy end-to-end: POST 400 retries and consumes the recovery 200.

Suite 4295 green; lint/mypy unchanged apart from the pre-existing residuals. PR body updated.

mikewiebe (approved):
Re-reviewed at cd26726f. The optional terminal-client-error policy now preserves legacy injected strategies while retaining the intended transient-status and GET 409 behavior. Focused tests and CI are green. Approving with the remaining RestSend-level 400/429 test hardening treated as nonblocking.

plugins/module_utils/rest/protocols/response_validation.py
allenrobel: Resolving to unblock merge

PR #503, opened 2026-08-12
Staged preview refinement for vpc_pair with additional tests

sivakasi-cisco:
## What this PR does

The vPC pair module supports a "stage now, apply later" flow — you can save a
change without deploying it, then re-run later to push it. A few bugs made that
"apply later" step quietly do nothing. This PR fixes those and adds tests.

## Problems fixed

1. **Staged delete never completed.** When a pair was deleted with `save: false`
   and `deploy: false`, the controller removed it from `/vpcPairs` right away but
   the switches stayed pending. Re-running the same delete with `save: true` /
   `deploy: true` saw the pair was already gone and reported `changed: false`
   without calling `configSave` or `switchActions/deploy`, so the switches stayed
   stuck.

2. **Staged delete broke with more than one pair.** If pair A/B was being deleted
   while an unrelated pair C/D still existed, the delete query returned early on
   the non-empty list (C/D) and never reached the recovery path for A/B, so A/B
   was reported as `no_change`.

3. **Repeat save-only reported changed.** Running the same save-only task again
   always reported `changed: true`, even when nothing was different and the
   pending config was identical.

4. **Optimistic status.** The module trusted the controller's summary and could
   call a pair "done" before the switches were actually in sync.

5. **Partial deploy looked successful.** A `switchActions/deploy` 207 where one
   peer failed and the other succeeded surfaced as a top-level `changed: true`
   success, because only the outer status was checked.

## What changed

- **Delete uses the requested config as the source of truth** (`query.py`). New
  helpers filter `/vpcPairs` down to the requested pairs, rebuild a requested pair
  that is already gone, and drop unrelated pairs from delete reconciliation
  (`_filter_vpc_pairs_by_requested_config`, `_reconstruct_requested_delete_pairs`,
  `_build_delete_existing_pairs`). This fixes both the single-pair and multi-pair
  staged-delete cases.
- **Pending work is flushed on save, not just deploy** (`actions.py`). When a
  later save/deploy delete finds the pair already unpaired, still out-of-sync
  member switches are flagged so `configSave` + `switchActions/deploy` run. Only
  switches actually reported out-of-sync are flagged, so it stays idempotent.
- **Real per-switch deploy status is checked** (`deploy.py`). A 207 body is read
  per switch (`DATA.switchIds[]`) and the run fails if any switch status is a
  failure, instead of trusting the outer 2xx.
- **Change flag reflects real work** (`resources.py`). A deploy that pushed config
  (or a save with a real declarative diff) sets module-level `changed: true`; a
  bare `configSave` with no diff stays `changed: false`, keeping repeat save-only
  runs idempotent.
- **`deploy` now defaults to `false`** (`nd_argument_specs.py`). `save` still
  defaults to `true`, so by default the module saves but does not deploy unless
  asked. (This addresses the defaults question raised in review.)
- **Small cleanup** — removed a dead `else` branch in the sync-state check and
  fixed the `OperationType.UPDATE` call in `utils.py`.

## Tests

- Unit tests for the delete query helpers (`test_manage_vpc_pair_query.py`).
- Unit tests for the per-switch deploy failure handling (`test_manage_vpc_pair_deploy.py`).
- Updated argument-spec test for the new `deploy` default (`test_nd_argument_specs.py`).

## How it was verified

Reproduced each scenario against a live fabric (create → stage delete → re-run
with save/deploy; two-pair delete; repeat save-only) and confirmed the switches
end in sync and the change status is accurate.


plugins/module_utils/manage_vpc_pair/query.py
```diff
@@ -1203,14 +1205,22 @@ def _set_lightweight_context(
...
-                if sync_state is True:
-                    pair_not_in_sync = False
+                if config_sync_state is False:
+                    pair_not_in_sync = True
                 elif sync_state is False:
                     pair_not_in_sync = True
+                elif sync_state is True:
+                    pair_not_in_sync = False
                 else:
                     pair_not_in_sync = config_sync_state is False
```
allenrobel: The last precedence bullet ("overview=None falls back to explicit switch out-of-sync") describes this final `else`, but the branch is unreachable as written: `config_sync_state is False` is already consumed by the first `if` in the chain, so by the time the `else` runs, `config_sync_state` can only be `True` or `None` and `config_sync_state is False` always evaluates to `False`. Runtime behavior is still correct — the first branch is what provides the fallback — but consider simplifying the `else` to `pair_not_in_sync = False` and reworking that bullet so a future reader doesn't treat the dead branch as live logic.

plugins/module_utils/manage_vpc_pair/resources.py
```diff
@@ -588,5 +588,9 @@ def execute(self, fabric_name: str) -> dict[str, Any]:
                 "deployment_needed",
                 self.needs_deployment_handler(result, nd_manage_vpc_pair),
             )
+            # A save/deploy that actually ran (e.g. saving an out-of-sync pair
+            # with no declarative diff) must surface as a module-level change.
+            if deploy_result.get("changed"):
+                result["changed"] = True
```
allenrobel, line 594: This newly promotes `deploy_result["changed"]` to the module-level `changed` flag. The promotion itself follows the pattern the network/VRF coordinators use, but the switch-scope branch of `custom_vpc_deploy` in `deploy.py` still registers `success=True, changed=True` on any 2xx without inspecting the HTTP 207 per-switch `status` array (`DATA.switchIds[]`), so a partial per-switch failure (one vPC peer fails, its peer succeeds) now surfaces as a successful `changed=true` run at the top level. This was raised in the PR #338 review; that thread was resolved on the understanding that the switch-scope deploy path had been removed, but it is still present on develop. Could `deploy.py` check the per-switch statuses before registering (or a follow-up issue be filed) so this top-level flag isn't fed by an unverified signal?

allenrobel (commented):
### Code review

Two observations on the sync-state refinement — one comment/dead-code cleanup in `query.py`, one question about the `changed` promotion in `resources.py`. The `utils.py` `OperationType.UPDATE` fix looks correct (verified `register_action_api_call` has a single caller, and the per-call `Results` scoping means no blast radius).

allenrobel:
Just a comment rather than a review comment, but it would be helpful to reviewers if you describe what this PR does in the PR description, along with a description of the changes and why they are needed.

sivakasi-cisco:
> Just a comment rather than a review comment, but it would be helpful to reviewers if you describe what this PR does in the PR description, along with a description of the changes and why they are needed.

Hi Allen, True. I moved it from draft to open and had missed it

plugins/module_utils/manage_vpc_pair/query.py
sivakasi-cisco: Yes. That line was effectively dead. The False case is already handled at the top, so in the else, config_sync_state is only True or None, making config_sync_state is False would always be False. I have simplified it to "pair_not_in_sync = False"

plugins/module_utils/manage_vpc_pair/resources.py
sivakasi-cisco: I have resolved it here with a UT. Yes, a 207 can be a mixed result: outer call is 2xx but one peer can fail while the other succeeds. The old code only checked the outer status, so it reported - changed=true on a partial failure.

 Now I read switchIds[] and raise if any switch status is failed/failure/error/success/notExecuted and checking if each still passes and corrected them here

nikhilsrikrishna:
Could you please confirm whether this is the intended behavior for `config_actions.type: switch`? Currently, the module appears to deploy every switch in the fabric that is not explicitly in-sync, rather than only the vPC peer switches managed by the current invocation ([code](https://github.com/CiscoDevNet/ansible-nd/blob/4bfefe19c5e4c17c864d55a621396ef4e80f730d/plugins/module_utils/manage_vpc_pair/deploy.py#L121-L141)). Would it be more appropriate to limit the deployment to the affected vPC peer switches?

nikhilsrikrishna:
### Problem scenario

A vPC pair deletion can be staged using `save: false` and `deploy: false`, with the intention of saving and deploying it in a later module invocation.

During the staged deletion, the controller removes the pair from the vPC pair inventory before the switch-side removal is saved or deployed. When the module is subsequently invoked with `save: true` and `deploy: true`, it sees that the pair is absent and treats the operation as complete.

The later invocation therefore returns `changed: false` and does not call either `configSave` or `switchActions/deploy`, even though the removal is still pending for the switches.

### Steps to reproduce

1. Create and deploy a vPC pair:

   ```yaml
   - cisco.nd.nd_manage_vpc_pair:
       fabric_name: "<fabric>"
       state: merged
       config:
         - peer1_switch_id: "<peer1>"
           peer2_switch_id: "<peer2>"
           use_virtual_peer_link: false
       config_actions:
         save: true
         deploy: true
         type: switch
   ```

2. Stage its deletion without saving or deploying:

   ```yaml
   - cisco.nd.nd_manage_vpc_pair:
       fabric_name: "<fabric>"
       state: deleted
       config:
         - peer1_switch_id: "<peer1>"
           peer2_switch_id: "<peer2>"
       config_actions:
         save: false
         deploy: false
         type: switch
   ```

3. After this task:

   - The module reports `changed: true`.
   - The intended-pair inventory no longer contains the pair.
   - The direct `vpcPair` endpoints for both peers return `{}`.
   - Both switches report `configSyncStatus: pending`.
   - The vPC configuration has not yet been removed.
   - `pendingConfig` is empty because `configSave` has not been called.

4. Invoke the same deletion again with save and deploy enabled:

   ```yaml
   - cisco.nd.nd_manage_vpc_pair:
       fabric_name: "<fabric>"
       state: deleted
       config:
         - peer1_switch_id: "<peer1>"
           peer2_switch_id: "<peer2>"
       config_actions:
         save: true
         deploy: true
         type: switch
   ```

5. The second invocation returns:

   ```text
   changed: false
   VPC pair <peer1>-<peer2> is already unpaired on the controller.
   No deletion required.
   ```

   The module trace contains no calls to:

   ```text
   POST /actions/configSave
   POST /switchActions/deploy
   ```

### Controller intent verification

After the second module invocation returned `changed: false`, I called `POST /actions/configSave` directly.

Subsequent `pendingConfig` responses for both peers contained the staged removal commands:

```yaml
pendingConfigs:
  - no interface port-channel500
  - no interface Vlan3600
  - no vlan 3600
  - no vpc domain 1
  - no feature vpc
  # plus peer-link member cleanup and interface restoration commands
```

This confirms that the staged removal intent was still available to the controller even though the pair was no longer present in the pair inventory.

Calling `POST /switchActions/deploy` directly for the same peers completed the removal. The endpoint returned HTTP 207 with `status: success` and `message: Deployed Successfully` for both switches. Final verification showed both peers `inSync`, empty `pendingConfig` responses, and empty direct `vpcPair` responses.

nikhilsrikrishna:
I noticed that `config_actions.save` and `config_actions.deploy` default to `true` when they are not explicitly provided. Could you please confirm whether this is the expected behavior?

nikhilsrikrishna:
**Impact: Low**

### Problem scenario

A vPC pair can be saved without being deployed using `save: true` and `deploy: false`. Because deployment is disabled, the switches remain `pending/outOfSync`.

Repeating the identical save-only operation performs another `configSave` and reports `changed: true`, even when there is no declarative pair difference and the pending configuration is unchanged.

### Steps to reproduce

1. Create a controller-only pair:

```yaml
- cisco.nd.nd_manage_vpc_pair:
    fabric_name: "<fabric>"
    state: merged
    config:
      - peer1_switch_id: "<peer1>"
        peer2_switch_id: "<peer2>"
        use_virtual_peer_link: false
    config_actions:
      save: false
      deploy: false
      type: switch
```

2. Save the pending pair without deploying:

```yaml
- cisco.nd.nd_manage_vpc_pair:
    fabric_name: "<fabric>"
    state: merged
    config:
      - peer1_switch_id: "<peer1>"
        peer2_switch_id: "<peer2>"
        use_virtual_peer_link: false
    config_actions:
      save: true
      deploy: false
      type: switch
```

3. Repeat the exact task from step 2.

### Observed behavior

The repeated save-only invocation returns:

```yaml
changed: true
class_diff:
  created: []
  updated: []
  deleted: []
deployment_changed: true
deployment_needed: true
deployment_paths:
  - /api/v1/manage/fabrics/<fabric>/actions/configSave
```

The API trace for the repeated invocation contains:

```text
POST /api/v1/manage/fabrics/<fabric>/actions/configSave
{"type":"switch"}

HTTP 200
{"status":"Config save is completed"}
```

It contains no vPC-pair `PUT` and no `switchActions/deploy` request.

Direct `pendingConfig` reads for both peers returned the same command lists before and after the repeated operation. The pair identity and configuration were also unchanged.

The repeated task is therefore reported as `changed: true` solely because the still-undeployed pair triggers another `configSave`, whose successful result is promoted to the module-level `changed` value. This makes repeated `save: true, deploy: false` operations non-idempotent.

sivakasi-cisco:
> Could you please confirm whether this is the intended behavior for `config_actions.type: switch`? Currently, the module appears to deploy every switch in the fabric that is not explicitly in-sync, rather than only the vPC peer switches managed by the current invocation ([code](https://github.com/CiscoDevNet/ansible-nd/blob/4bfefe19c5e4c17c864d55a621396ef4e80f730d/plugins/module_utils/manage_vpc_pair/deploy.py#L121-L141)). Would it be more appropriate to limit the deployment to the affected vPC peer switches?

Hi Nikhil, thank you for the comments. There is one more PR where the config_actions are explictly handled for vpc pair here. Just to avoid repetitive changes, I have handled the default deploy to false in that.

https://github.com/CiscoDevNet/ansible-nd/pull/513/changes

sivakasi-cisco:
@nikhilsrikrishna , Thank you for the comments. 

 The vPC‑pair module lets to stage a change now and apply it later, but the few bugs made that "do it later" step quietly do nothing - thank you for pointing.

"Deploy later" now actually deploys. Before, if we staged a pair (or a deletion) and re‑ran later to push it, the module looked at the controller, decided "nothing to do," reported no change, and left the switches stuck half‑done. Now it notices the still‑pending work and finishes the job.

The module now trusts the switches' real status, so it doesn't call something "done" too early. Saving twice no longer lies about changing things. Before, running a save‑only step again and again always said "changed," even when nothing was actually different. Now a repeat save‑only correctly reports no change — which is what automation expect in case of idempotency.

Now, the "stage now, apply later" flow for vPC pairs behaves right, it completes pending work, reports change status accurately, and is safe to re‑run

sivakasi-cisco (commented):
Comments addressed and tested

nikhilsrikrishna:
Thanks, Sivakami, for addressing the earlier comments and explaining the changes. The latest changes address the staged-deletion issue we discussed. However, while reviewing the implementation, I noticed a possible additional edge case when the fabric contains multiple independent vPC pairs.

Consider two pairs:

- A/B — the pair being deleted.
- C/D — an unrelated pair that remains configured.

First, A/B is deleted with `save: false` and `deploy: false`. As observed in the original reproduction, A/B disappears from `/vpcPairs`, while switches A and B remain pending/out-of-sync because the removal has not been saved or deployed. C/D remains in `/vpcPairs`.

The deletion of A/B is then invoked again with `save: true` and `deploy: true`.

On this later invocation, `/vpcPairs` returns C/D. Because `have` is non-empty, the deleted-state query returns immediately ([query.py](https://github.com/CiscoDevNet/ansible-nd/blob/1bdd7321df181da3a39491b82998845e8bfd2562/plugins/module_utils/manage_vpc_pair/query.py#L851-L874)):

```python
elif have:
    return _set_lightweight_context(have)
```

This happens without checking whether C/D matches the requested A/B pair and prevents the fallback that reconstructs an already-absent requested pair from being reached.

The state machine consequently receives C/D as the existing pair but is asked to delete A/B. Since A/B cannot be found, the operation is recorded as `no_change` ([resources.py](https://github.com/CiscoDevNet/ansible-nd/blob/1bdd7321df181da3a39491b82998845e8bfd2562/plugins/module_utils/manage_vpc_pair/resources.py#L499-L506)).

As a result, `custom_vpc_delete()` and the pending-member recovery helper are not called. No pending or out-of-sync signal is recorded, so `configSave` and `switchActions/deploy` are skipped despite `deploy: true`.

One possible approach would be for the deleted-state query to use the requested pair identifiers as its source of truth:

- Retain requested pairs that are still present in `/vpcPairs`.
- Reconstruct requested pairs that are already absent so the recovery check can run.
- Exclude unrelated pairs such as C/D from deletion reconciliation.

The existing `_filter_vpc_pairs_by_requested_config()` helper ([query.py](https://github.com/CiscoDevNet/ansible-nd/blob/1bdd7321df181da3a39491b82998845e8bfd2562/plugins/module_utils/manage_vpc_pair/query.py#L543-L585)) may be reusable or generalizable for this.

Could you please check whether this multi-pair staged-deletion scenario needs to be handled?

sivakasi-cisco:
> Thanks, Sivakami, for addressing the earlier comments and explaining the changes. The latest changes address the staged-deletion issue we discussed. However, while reviewing the implementation, I noticed a possible additional edge case when the fabric contains multiple independent vPC pairs.
> 
> Consider two pairs:
> 
> * A/B — the pair being deleted.
> * C/D — an unrelated pair that remains configured.
> 
> First, A/B is deleted with `save: false` and `deploy: false`. As observed in the original reproduction, A/B disappears from `/vpcPairs`, while switches A and B remain pending/out-of-sync because the removal has not been saved or deployed. C/D remains in `/vpcPairs`.
> 
> The deletion of A/B is then invoked again with `save: true` and `deploy: true`.
> 
> On this later invocation, `/vpcPairs` returns C/D. Because `have` is non-empty, the deleted-state query returns immediately ([query.py](https://github.com/CiscoDevNet/ansible-nd/blob/1bdd7321df181da3a39491b82998845e8bfd2562/plugins/module_utils/manage_vpc_pair/query.py#L851-L874)):
> 
> ```python
> elif have:
>     return _set_lightweight_context(have)
> ```
> 
> This happens without checking whether C/D matches the requested A/B pair and prevents the fallback that reconstructs an already-absent requested pair from being reached.
> 
> The state machine consequently receives C/D as the existing pair but is asked to delete A/B. Since A/B cannot be found, the operation is recorded as `no_change` ([resources.py](https://github.com/CiscoDevNet/ansible-nd/blob/1bdd7321df181da3a39491b82998845e8bfd2562/plugins/module_utils/manage_vpc_pair/resources.py#L499-L506)).
> 
> As a result, `custom_vpc_delete()` and the pending-member recovery helper are not called. No pending or out-of-sync signal is recorded, so `configSave` and `switchActions/deploy` are skipped despite `deploy: true`.
> 
> One possible approach would be for the deleted-state query to use the requested pair identifiers as its source of truth:
> 
> * Retain requested pairs that are still present in `/vpcPairs`.
> * Reconstruct requested pairs that are already absent so the recovery check can run.
> * Exclude unrelated pairs such as C/D from deletion reconciliation.
> 
> The existing `_filter_vpc_pairs_by_requested_config()` helper ([query.py](https://github.com/CiscoDevNet/ansible-nd/blob/1bdd7321df181da3a39491b82998845e8bfd2562/plugins/module_utils/manage_vpc_pair/query.py#L543-L585)) may be reusable or generalizable for this.
> 
> Could you please check whether this multi-pair staged-deletion scenario needs to be handled?

Thanks for pointing.  I reviewed the multi-pair staged-delete path and addressed it by making the requested delete configuration.

For a request to delete A/B when /vpcPairs only returns unrelated C/D, the delete query now does the below

- Filters the list response to retain only requested pairs.
- Reconstructs A/B from the requested configuration when it is already absent from /vpcPairs.
- Excludes unrelated pairs such as C/D from delete reconciliation.

This allows the delete state machine to process A/B and run the pending-member recovery path, so a later save: true, deploy: true invocation can save and deploy the previously staged removal.

I also added unit coverage for the below.

- absent requested A/B with unrelated C/D present;
- requested pairs that are still present;
- a mixture of present and reconstructed requested pairs.
The query tests pass locally.

sivakasi-cisco:
> **Impact: Low**
> 
> ### Problem scenario
> A vPC pair can be saved without being deployed using `save: true` and `deploy: false`. Because deployment is disabled, the switches remain `pending/outOfSync`.
> 
> Repeating the identical save-only operation performs another `configSave` and reports `changed: true`, even when there is no declarative pair difference and the pending configuration is unchanged.
> 
> ### Steps to reproduce
> 1. Create a controller-only pair:
> 
> ```yaml
> - cisco.nd.nd_manage_vpc_pair:
>     fabric_name: "<fabric>"
>     state: merged
>     config:
>       - peer1_switch_id: "<peer1>"
>         peer2_switch_id: "<peer2>"
>         use_virtual_peer_link: false
>     config_actions:
>       save: false
>       deploy: false
>       type: switch
> ```
> 
> 2. Save the pending pair without deploying:
> 
> ```yaml
> - cisco.nd.nd_manage_vpc_pair:
>     fabric_name: "<fabric>"
>     state: merged
>     config:
>       - peer1_switch_id: "<peer1>"
>         peer2_switch_id: "<peer2>"
>         use_virtual_peer_link: false
>     config_actions:
>       save: true
>       deploy: false
>       type: switch
> ```
> 
> 3. Repeat the exact task from step 2.
> 
> ### Observed behavior
> The repeated save-only invocation returns:
> 
> ```yaml
> changed: true
> class_diff:
>   created: []
>   updated: []
>   deleted: []
> deployment_changed: true
> deployment_needed: true
> deployment_paths:
>   - /api/v1/manage/fabrics/<fabric>/actions/configSave
> ```
> 
> The API trace for the repeated invocation contains:
> 
> ```
> POST /api/v1/manage/fabrics/<fabric>/actions/configSave
> {"type":"switch"}
> 
> HTTP 200
> {"status":"Config save is completed"}
> ```
> 
> It contains no vPC-pair `PUT` and no `switchActions/deploy` request.
> 
> Direct `pendingConfig` reads for both peers returned the same command lists before and after the repeated operation. The pair identity and configuration were also unchanged.
> 
> The repeated task is therefore reported as `changed: true` solely because the still-undeployed pair triggers another `configSave`, whose successful result is promoted to the module-level `changed` value. This makes repeated `save: true, deploy: false` operations non-idempotent.

It was a real bug, Nikhil. Thanks for pointing.

The problem was how we counted a save as a change. A bare configSave (Recalculate & Save) with no actual pair diff doesn't push anything to the switches, it just recomputes the same config and leaves them pending. But we were still treating that successful save as a module-level change, so every repeat save=true, deploy=false run kept reporting changed: true even though nothing was different.

I fixed the change detection so changed is only driven by a real declarative diff (create/update/delete) or an actual deploy. A bare configSave with no diff is now treated as idempotent, so a second identical save-only run reports changed: false.

One thing to be clear on: the pair is still genuinely out-of-sync until it's deployed, so the save step can still run on a repeat, but it no longer flips the run to changed: true. The change flag now reflects real config deltas, not the save call itself.

Verified with the save-only repro (issue467_saveonly_idempotency_repro) and a unit test.

sivakasi-cisco:
> ### Problem scenario
> A vPC pair deletion can be staged using `save: false` and `deploy: false`, with the intention of saving and deploying it in a later module invocation.
> 
> During the staged deletion, the controller removes the pair from the vPC pair inventory before the switch-side removal is saved or deployed. When the module is subsequently invoked with `save: true` and `deploy: true`, it sees that the pair is absent and treats the operation as complete.
> 
> The later invocation therefore returns `changed: false` and does not call either `configSave` or `switchActions/deploy`, even though the removal is still pending for the switches.
> 
> ### Steps to reproduce
> 1. Create and deploy a vPC pair:
>    ```yaml
>    - cisco.nd.nd_manage_vpc_pair:
>        fabric_name: "<fabric>"
>        state: merged
>        config:
>          - peer1_switch_id: "<peer1>"
>            peer2_switch_id: "<peer2>"
>            use_virtual_peer_link: false
>        config_actions:
>          save: true
>          deploy: true
>          type: switch
>    ```
> 2. Stage its deletion without saving or deploying:
>    ```yaml
>    - cisco.nd.nd_manage_vpc_pair:
>        fabric_name: "<fabric>"
>        state: deleted
>        config:
>          - peer1_switch_id: "<peer1>"
>            peer2_switch_id: "<peer2>"
>        config_actions:
>          save: false
>          deploy: false
>          type: switch
>    ```
> 3. After this task:
>    
>    * The module reports `changed: true`.
>    * The intended-pair inventory no longer contains the pair.
>    * The direct `vpcPair` endpoints for both peers return `{}`.
>    * Both switches report `configSyncStatus: pending`.
>    * The vPC configuration has not yet been removed.
>    * `pendingConfig` is empty because `configSave` has not been called.
> 4. Invoke the same deletion again with save and deploy enabled:
>    ```yaml
>    - cisco.nd.nd_manage_vpc_pair:
>        fabric_name: "<fabric>"
>        state: deleted
>        config:
>          - peer1_switch_id: "<peer1>"
>            peer2_switch_id: "<peer2>"
>        config_actions:
>          save: true
>          deploy: true
>          type: switch
>    ```
> 5. The second invocation returns:
>    ```
>    changed: false
>    VPC pair <peer1>-<peer2> is already unpaired on the controller.
>    No deletion required.
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    The module trace contains no calls to:
>    ```
>    POST /actions/configSave
>    POST /switchActions/deploy
>    ```
> 
> ### Controller intent verification
> After the second module invocation returned `changed: false`, I called `POST /actions/configSave` directly.
> 
> Subsequent `pendingConfig` responses for both peers contained the staged removal commands:
> 
> ```yaml
> pendingConfigs:
>   - no interface port-channel500
>   - no interface Vlan3600
>   - no vlan 3600
>   - no vpc domain 1
>   - no feature vpc
>   # plus peer-link member cleanup and interface restoration commands
> ```
> 
> This confirms that the staged removal intent was still available to the controller even though the pair was no longer present in the pair inventory.
> 
> Calling `POST /switchActions/deploy` directly for the same peers completed the removal. The endpoint returned HTTP 207 with `status: success` and `message: Deployed Successfully` for both switches. Final verification showed both peers `inSync`, empty `pendingConfig` responses, and empty direct `vpcPair` responses.

Thanks Nikhil. Earlier when a delete was stagedwith save: false / deploy: false, the controller removed the pair from /vpcPairs, so the next run saw it as absent and reported changed: false without calling configsave or deploy.

 Now the delete path rebuilds the requested pair from config when it is already gone, sees the switches are still pending, and runs configSave + switchActions/deploy. So the second run with save: true / deploy: true actually
finishes the removal. Covered by issue467_staged_delete_repro and a UT.

sivakasi-cisco:
> **Impact: Low**
> 
> ### Problem scenario
> A vPC pair can be saved without being deployed using `save: true` and `deploy: false`. Because deployment is disabled, the switches remain `pending/outOfSync`.
> 
> Repeating the identical save-only operation performs another `configSave` and reports `changed: true`, even when there is no declarative pair difference and the pending configuration is unchanged.
> 
> ### Steps to reproduce
> 1. Create a controller-only pair:
> 
> ```yaml
> - cisco.nd.nd_manage_vpc_pair:
>     fabric_name: "<fabric>"
>     state: merged
>     config:
>       - peer1_switch_id: "<peer1>"
>         peer2_switch_id: "<peer2>"
>         use_virtual_peer_link: false
>     config_actions:
>       save: false
>       deploy: false
>       type: switch
> ```
> 
> 2. Save the pending pair without deploying:
> 
> ```yaml
> - cisco.nd.nd_manage_vpc_pair:
>     fabric_name: "<fabric>"
>     state: merged
>     config:
>       - peer1_switch_id: "<peer1>"
>         peer2_switch_id: "<peer2>"
>         use_virtual_peer_link: false
>     config_actions:
>       save: true
>       deploy: false
>       type: switch
> ```
> 
> 3. Repeat the exact task from step 2.
> 
> ### Observed behavior
> The repeated save-only invocation returns:
> 
> ```yaml
> changed: true
> class_diff:
>   created: []
>   updated: []
>   deleted: []
> deployment_changed: true
> deployment_needed: true
> deployment_paths:
>   - /api/v1/manage/fabrics/<fabric>/actions/configSave
> ```
> 
> The API trace for the repeated invocation contains:
> 
> ```
> POST /api/v1/manage/fabrics/<fabric>/actions/configSave
> {"type":"switch"}
> 
> HTTP 200
> {"status":"Config save is completed"}
> ```
> 
> It contains no vPC-pair `PUT` and no `switchActions/deploy` request.
> 
> Direct `pendingConfig` reads for both peers returned the same command lists before and after the repeated operation. The pair identity and configuration were also unchanged.
> 
> The repeated task is therefore reported as `changed: true` solely because the still-undeployed pair triggers another `configSave`, whose successful result is promoted to the module-level `changed` value. This makes repeated `save: true, deploy: false` operations non-idempotent.

Correct.  A repeat save-only was always reporting changed: true even when
nothing was different. I fixed the change detection so a second identical save-only
run reports changed: false and does not call configSave when there is no real pair
difference and the pending config is the same. Verified with the save-only repro.

sivakasi-cisco:
> I noticed that `config_actions.save` and `config_actions.deploy` default to `true` when they are not explicitly provided. Could you please confirm whether this is the expected behavior?

Thanks Nikhil. I have modified changed
deploy to default to false. save still defaults to true. So by default it saves the
config but does not deploy unless you explicitly set deploy: true.

sivakasi-cisco (commented):
Changes made to address the review comments.

Also made it to go hand-in-hand with [PR513. ](https://github.com/CiscoDevNet/ansible-nd/pull/513). Integ test passed successfully with combined changes

nikhilsrikrishna (commented):
I noticed that `config_actions.deploy` is  defaulting to `true` in a few places:

- `nd_manage_vpc_pair.py` documentation
- `ConfigActionsModel.deploy`
- `VpcPairPlaybookConfigModel.get_argument_spec()`
- `get_config_actions()` in `common.py`
- `custom_vpc_deploy()` in `deploy.py`

Could you please check whether these should be updated to `false` for consistency?

plugins/module_utils/manage_vpc_pair/deploy.py
```diff
@@ -302,6 +349,24 @@ def custom_vpc_deploy(nrm: Any, fabric_name: str, result: dict[str, Any]) -> dic
                 deploy_payload = {"switchIds": switch_ids}
                 if switch_ids:
                     response = fabric_utils.deploy_switches(switch_ids)
+                    # A 2xx (including 207 Multi-Status) can still carry per-switch
+                    # failures; do not report a clean changed=true when one vPC peer
+                    # deployed and the other failed.
+                    switch_failures = _switch_deploy_failures(response.get("response_data"))
```
allenrobel: Follow-up on the resolved 207 thread: this parser cannot see a failure body in production. `fabric_utils.deploy_switches()` goes through `NDModuleV2.request()`, which raises `NDModuleError` whenever the RestSend result is not success, and `NdV1Strategy.is_success()` already scans `DATA.switchIds[]`: on a 207 any item whose `status` is not exactly `success` fails the request (plain-200 bodies fail on `failed`/`failure`/`error`). So a partial failure lands in the existing `except NDModuleError` below, with the per-item detail already formatted by the strategy's `error_message`, and `switch_failures` is always empty here.

The new tests pass only because `_FakeNDModuleV2.request()` returns the raw dict and bypasses RestSend. Two consequences: `test_manage_vpc_pair_deploy_00110` exercises a path RestSend never allows through, and `test_manage_vpc_pair_deploy_00120` asserts that `notExecuted` completes normally, which is the opposite of what the stack does on a real 207 (the request fails).

Suggest dropping `_switch_deploy_failures` / `_SWITCH_DEPLOY_FAILURE_STATUSES` and the three tests. If `notExecuted` needs to be tolerated for `switchActions/deploy`, that is a strategy-level decision (issue #397, vault note `multi-status-207-status-field-inconsistent`), not a module-level denylist.

plugins/module_utils/manage_vpc_pair/actions.py
```diff
@@ -518,6 +522,53 @@ def custom_vpc_update(nrm: Any) -> dict[str, Any] | None:
...
+
+    A prior save=false delete unpairs the pair on the controller but leaves the
+    member switches out-of-sync. When a later save/deploy delete finds the pair
+    already gone, record any still out-of-sync member switch in
+    ``_not_in_sync_pairs`` so the deploy step runs configSave + switchActions/deploy.
+    Only switches explicitly reported out-of-sync are flagged, so the run stays
+    idempotent once the switches are back in sync.
+    """
+    try:
+        switches = _validate_fabric_switches(nd_v2, fabric_name)
```
allenrobel: `_validate_fabric_switches` is a GET of the fabric switch inventory, and `custom_vpc_delete` runs once per pair from `_manage_delete_state`, so a `state: deleted` re-run listing N already-unpaired pairs (the multi-pair case this PR adds) issues N identical inventory GETs, and then `_get_switches_needing_deploy` fetches the same inventory again. CLAUDE.md Performance: "Fetch each controller resource at most once per module run. If two code paths need the same data, cache it on the orchestrator or pass it through."

Suggest fetching once (e.g. cache the switch map in `nrm.module.params` on first use and reuse it from deploy.py) and gating this helper on `config_actions.save or config_actions.deploy`. With both false nothing is flushed, so the GET and the "Flushing the pending removal via save/deploy" warning are both wasted.

plugins/module_utils/nd_argument_specs.py
```diff
@@ -85,7 +85,7 @@ def config_actions_spec(include: Iterable[str] | None = None) -> dict[str, Any]:
     options: dict[str, Any] = {
         "save": {"type": "bool", "default": True},
         "deploy": {"type": "bool", "default": False},
-        "type": {"type": "str", "default": "switch", "choices": ["resource", "switch", "global"]},
+        "type": {"type": "str", "choices": ["switch", "global"]},
```
allenrobel, line 88: This fragment is not consumed by `nd_manage_vpc_pair` (the module builds `config_actions` from `VpcPairPlaybookConfigModel.get_argument_spec()`), so this hunk has no effect on the behavior this PR fixes. What it does change is the shared `config_actions_spec()` used by the `nd_interface_*` modules (adopted in #524): it removes the `resource` choice and the `switch` default, and replaces the #368 contract note with a scope claim (`switch` deploys "only the switches left out-of-sync by the change") that the vPC deploy path does not implement today (`_get_switches_needing_deploy` sweeps every out-of-sync switch in the fabric).

The identical hunk, and the matching `test_nd_argument_specs` edit, are already in #513. Suggest dropping `nd_argument_specs.py` and `test_nd_argument_specs.py` from this PR and letting #513 carry the contract change where it can be reviewed with the rest of the config_actions work.

allenrobel (commented):
### Code review (re-review)

Re-reviewed the current head after the 2026-08-13 comments. The staged-delete reconstruction and the save-only idempotency changes look right: IP inputs are normalized to serials before the new delete filter runs, unknown serials are never flagged as pending, and the partial-deploy raise propagates past the `except NDModuleError`. Three follow-ups inline.

**Sequencing note for #513.** A trial merge of the current heads of #503 (`1d45d86a`) and #513 (`cb3fe6d9`) conflicts in `plugins/module_utils/manage_vpc_pair/deploy.py`, `plugins/module_utils/manage_vpc_pair/resources.py`, `plugins/module_utils/models/manage_vpc_pair/vpc_pair_model.py`, and `tests/unit/module_utils/test_manage_vpc_pair_deploy.py` (the byte-identical hunks in `nd_argument_specs.py`, `utils.py`, and the module doc auto-merge cleanly). Whichever PR lands first, the other will need a rebase before it can merge.

sivakasi-cisco:
> I noticed that `config_actions.deploy` is defaulting to `true` in a few places:
> 
> * `nd_manage_vpc_pair.py` documentation
> * `ConfigActionsModel.deploy`
> * `VpcPairPlaybookConfigModel.get_argument_spec()`
> * `get_config_actions()` in `common.py`
> * `custom_vpc_deploy()` in `deploy.py`
> 
> Could you please check whether these should be updated to `false` for consistency?

Hi Nikhil, I have another PR - https://github.com/CiscoDevNet/ansible-nd/pull/513 in which the changes for deploy to false is handled. I had tested both by combining the changes. I remember you had highlighted this earlier as well. As I had raised the other review and obtained some comments on it, I had sticked to that. Can you pls have a look if that is okay?

nikhilsrikrishna:
> > I noticed that `config_actions.deploy` is defaulting to `true` in a few places:
> > 
> > * `nd_manage_vpc_pair.py` documentation
> > * `ConfigActionsModel.deploy`
> > * `VpcPairPlaybookConfigModel.get_argument_spec()`
> > * `get_config_actions()` in `common.py`
> > * `custom_vpc_deploy()` in `deploy.py`
> > 
> > Could you please check whether these should be updated to `false` for consistency?
> 
> Hi Nikhil, I have another PR - #513 in which the changes for deploy to false is handled. I had tested both by combining the changes. I remember you had highlighted this earlier as well. As I had raised the other review and obtained some comments on it, I had sticked to that. Can you pls have a look if that is okay?

Thanks, Sivakami. I checked the latest changes, and config_actions.deploy now defaults to false consistently across. This addresses my comment. Thank you.

sivakasi-cisco:
> > > I noticed that `config_actions.deploy` is defaulting to `true` in a few places:
> > > 
> > > * `nd_manage_vpc_pair.py` documentation
> > > * `ConfigActionsModel.deploy`
> > > * `VpcPairPlaybookConfigModel.get_argument_spec()`
> > > * `get_config_actions()` in `common.py`
> > > * `custom_vpc_deploy()` in `deploy.py`
> > > 
> > > Could you please check whether these should be updated to `false` for consistency?
> > 
> > 
> > Hi Nikhil, I have another PR - #513 in which the changes for deploy to false is handled. I had tested both by combining the changes. I remember you had highlighted this earlier as well. As I had raised the other review and obtained some comments on it, I had sticked to that. Can you pls have a look if that is okay?
> 
> Thanks, Sivakami. I checked the latest changes, and config_actions.deploy now defaults to false consistently across. This addresses my comment. Thank you.

Thank you, Nikhil

plugins/module_utils/manage_vpc_pair/deploy.py
sivakasi-cisco: Hi Allen, 

This parser cannot see a failure body in production. deploy_switches() goes through NDModuleV2.request(), and NdV1Strategy.is_success() already fails the request on any switchIds[] row that is not a "success", so a partial 207 is raised as NDModuleError before my parser ever runs. That means switch_failures was always empty, and the three new tests only passed because the fake _FakeNDModuleV2.request() returns the raw dict and bypasses RestSend. They were exercising a path the real stack never allows, and one of them even asserted that notExecuted "completes normally", which is the opposite of what really happens.


I have dropped the initially kept (_is_non_fatal_switch_deploy_207 + its test), since that is also a module-level decision about tolerating notExecuted.As you mentioned, notExecuted handling is now deferred entirely to the strategy (#397). Also, removed _switch_deploy_failures and _SWITCH_DEPLOY_FAILURE_STATUSES and the three tests. Thanks for the thorough explanation.

plugins/module_utils/nd_argument_specs.py
sivakasi-cisco: True. I merged them (PR 503 adn PR 513)together to run the tests and then splitted few changes. I reverted them from 503, here.

plugins/module_utils/manage_vpc_pair/actions.py
sivakasi-cisco: Thanks, Allen. I have fixed Fixed both:

-> Fetch once (delete): added _fabric_switches_cached(), which caches inventory on nrm.module.params keyed by fabric, so that the state "deleted" reruned over N unpaired pairs will now does 1 GET instead of N.
-> Gate: _flag_pending_member_switches_for_deploy will return early when save and deploy are both false. It will skip the GET and the "Flushing the pending removal" warning.

-> On "reuse it from deploy.py", I had left it for the following reasons. 
Here, the delete-phase read happens before configSave, but the deploy phase has to read after configSave to see which switches are still out-of-sync and actually need deploying.

 If it reused the pre-save snapshot it'd be deciding off stale sync state. deploy.py also relies on force_deploy_serials to cover a transient right after configSave where a peer briefly reports in-sync, which a shared snapshot would defeat. So it's down to 2 GETs total (one per phase) — the fewest I can do without crossing the configSave boundary. I have mentioned it in the _fabric_switches_cached docstring as well

nikhilsrikrishna:
I noticed a mismatched delete scenario. If the requested deletion is `A/B`, but the controller reports that `A` is currently paired with `X`, the module should reject the request because `A/B` is not the active pair.

Instead, the current flow continues as though `A/B` exists and can send an `unPair` request for `A/B`.

This originates in `_build_delete_existing_pairs()`: exact-match filtering removes `A/X` from the list passed to the delete state machine as the existing pairs. Because `A/B` is then considered missing, `_reconstruct_requested_delete_pairs()` creates a synthetic `A/B` in-memory object and adds it to that existing-pairs list. No `A/B` object is created on the controller, but the deletion path treats the in-memory object as existing controller state. The normal deletion flow is then unnecessarily carried out using this synthetic entry, resulting in an `unPair` request for a pair the controller did not report.

Could we check for member conflicts against the unfiltered controller pairs before reconstruction and reject the operation when either switch is paired with a different peer?

nikhilsrikrishna (commented):
`changed` is misreported in the following scenarios: if creating A/B succeeds but `configSave` fails, or if deployment succeeds for A but fails for B, the final failure reports `changed: false` despite the partial mutation. Could you please preserve the state-machine and partial-deployment change status?

nikhilsrikrishna:
Following up on [the synthetic-pair issue](https://github.com/CiscoDevNet/ansible-nd/pull/503#issuecomment-5711014983), the same reconstruction also affects the reported state after a successful deletion.

After `A/B` is unpaired, the module's in-memory existing state is correctly empty. However, `_refresh_after_state()` runs the deleted-state query again. Although the controller correctly returns no pairs, the query reconstructs `A/B` from the original request and replaces the empty state with that synthetic object.

The result can therefore report `A/B` under both `deleted` and `after/current`, even though it was successfully removed from the controller.

Could we limit reconstruction to the initial staged-delete recovery and use an authoritative query without reconstruction for the post-write refresh?

allenrobel (approved):
Verified the 2026-09-15 follow-ups against head c30dc0c8: the module-level 207 parser and its tests are gone (notExecuted deferred to #397), the delete phase now reads the fabric inventory once per run and skips the read entirely when save and deploy are both false, and the `nd_argument_specs.py` hunk is back in #513 where it belongs. Keeping the second inventory read in the deploy phase is the right call since it has to see post-configSave sync state. CI is green. Approving from my side; Nikhil's open note on `changed` under partial save/deploy failure is still outstanding.

https://claude.ai/code/session_013U847htGHYgohJPB2mRF7F

sivakasi-cisco:
> I noticed a mismatched delete scenario. If the requested deletion is `A/B`, but the controller reports that `A` is currently paired with `X`, the module should reject the request because `A/B` is not the active pair.
> 
> Instead, the current flow continues as though `A/B` exists and can send an `unPair` request for `A/B`.
> 
> This originates in `_build_delete_existing_pairs()`: exact-match filtering removes `A/X` from the list passed to the delete state machine as the existing pairs. Because `A/B` is then considered missing, `_reconstruct_requested_delete_pairs()` creates a synthetic `A/B` in-memory object and adds it to that existing-pairs list. No `A/B` object is created on the controller, but the deletion path treats the in-memory object as existing controller state. The normal deletion flow is then unnecessarily carried out using this synthetic entry, resulting in an `unPair` request for a pair the controller did not report.
> 
> Could we check for member conflicts against the unfiltered controller pairs before reconstruction and reject the operation when either switch is paired with a different peer?

Hi Nihil, I have updated the delete flow to check the complete pair list returned by the controller before filtering or rebuilding the requested pair. Now, if the request is to delete A/B but A is already paired with X, the module stops with a clear error and does not send an unPair request for A/B. I have added a unit test for this case as well.

sivakasi-cisco:
> `changed` is misreported in the following scenarios: if creating A/B succeeds but `configSave` fails, or if deployment succeeds for A but fails for B, the final failure reports `changed: false` despite the partial mutation. Could you please preserve the state-machine and partial-deployment change status?

I have updated the failure handling so that it does not lose a change that has already happened. The task will still fail if configSave fails or one of the switches fails during deployment, but changed will remain true if the pair was already created or deleted, or if part of the deployment had already changed the state. I have added tests for both the configSave failure and partial deployment cases. We also saw this in the combined live runs - the pair deletion completed, one switch deployment failed and the results are correctly preserved with changed: true.

For a mixed HTTP 207 response, notExecuted by itself is treated as non-fatal, but the task still fails if any switch reports an actual deployment failure.(notExecuted deferred to https://github.com/CiscoDevNet/ansible-nd/issues/397),

sivakasi-cisco:
> Following up on [the synthetic-pair issue](https://github.com/CiscoDevNet/ansible-nd/pull/503#issuecomment-5711014983), the same reconstruction also affects the reported state after a successful deletion.
> 
> After `A/B` is unpaired, the module's in-memory existing state is correctly empty. However, `_refresh_after_state()` runs the deleted-state query again. Although the controller correctly returns no pairs, the query reconstructs `A/B` from the original request and replaces the empty state with that synthetic object.
> 
> The result can therefore report `A/B` under both `deleted` and `after/current`, even though it was successfully removed from the controller.
> 
> Could we limit reconstruction to the initial staged-delete recovery and use an authoritative query without reconstruction for the post-write refresh?

Yes. I have now limited the pair reconstruction to the initial staged-delete check, where it is needed to recover an earlier pending deletion. For the refresh after the delete, the module now uses only the actual response from the controller. So, once A/B is deleted, after/current remains empty and the temporary A/B entry is not added back. I have added a unit test for this scenario.

PR #505, opened 2026-08-12
ND Manage VRFs & Networks: Fix Replaced First-Create Attachment Flow

### LINKED ISSUE #506: Replaced state fails on first-create VRF/Network attachment workflows
## Summary

For `nd_manage_vrfs` and `nd_manage_networks`, `state: replaced` fails when the target VRF/Network does not already exist and the desired config includes `attach`.

The same playbook succeeds after the VRF/Network exists, but the first run errors before the CRUD create can complete. This makes `replaced` unusable as an initial declarative create state for attached VRFs/Networks.

The issue affects both modules because they use the same workflow shape:

- `vrf_state_machine` + `vrf_attachment_manager`
- `network_state_machine` + `network_attachment_manager`

## Root Cause

The `replaced` flow performs attachment pre-processing before the generic CRUD state machine has created the VRF/Network.

For first-create scenarios:

1. Desired config includes a new VRF/Network with `attach`.
2. The resource does not exist yet on the controller.
3. The attachment pre-phase attempts to query current attachments for the desired resource.
4. ND returns an error because the VRF/Network is missing.
5. The module fails before the state machine can create the resource.

This differs from the desired workflow, where `replaced` should be able to create the missing resource first and then apply attachments.

## Reproduction

Example VRF case:

```yaml
- name: Replace/create VRF with attachment
  cisco.nd.nd_manage_vrfs:
    fabric_name: AK-VXLAN
    state: replaced
    config:
      - vrf_name: BLUE
        vrf_id: 50001
        attach:
          - ip_address: 192.0.2.10
```

Example Network case:

```yaml
- name: Replace/create Network with attachment
  cisco.nd.nd_manage_networks:
    fabric_name: AK-VXLAN
    state: replaced
    config:
      - network_name: BLUE_NET
        network_id: 30001
        vrf_name: BLUE
        attach:
          - ip_address: 192.0.2.10
            interfaces:
              - interface_range: Ethernet1/1
                mode: trunk
```

Expected:

- If the VRF/Network does not exist, `replaced` creates it.
- Attachment is applied after the resource exists.
- Module reports `changed: true`.

Actual:

- First run fails with a controller-side attachment query/update error.
- Re-running after the resource exists succeeds.

## Notes for the Fix

The fix should make `replaced` use the same state-machine-aware attachment sequencing pattern as `overridden`.

The corrected flow should be:

1. Initialize the generic state machine and derive current VRF/Network names from `sm.existing`.
2. Query current attachments only for desired resources that already exist.
3. Skip missing-resource attachment pre-query on first-create.
4. Run the generic CRUD state machine.
5. Apply post-attach after the VRF/Network exists.
6. Deploy pending attachment changes.

For `overridden`, omitted-resource deletion handling should remain intact because that state still needs to inspect all existing resources and detach/delete resources omitted from desired config.

## Affected Code

- `plugins/module_utils/orchestrators/vrf_state_machine.py`
- `plugins/module_utils/orchestrators/network_state_machine.py`
- `plugins/module_utils/orchestrators/vrf_attachment_manager.py`
- `plugins/module_utils/orchestrators/network_attachment_manager.py`

## Environment

- Collection: `cisco.nd` 2.0.0 / develop
- ND version: observed against ND 4.2.x workflows
- Affected modules:
  - `cisco.nd.nd_manage_vrfs`
  - `cisco.nd.nd_manage_networks`

AKDRG:
## Related Issue(s)

Resolves #506 

## Proposed Changes

- Fix `state: replaced` first-create workflows for `nd_manage_vrfs` and `nd_manage_networks` when the desired config includes attachments.
- Align `replaced` with the state-machine-aware attachment sequencing used by `overridden`.
- Avoid querying attachment endpoints for VRFs/Networks that do not exist yet.
- Preserve the correct workflow order:
  - derive current VRF/Network names from the generic state machine
  - pre-detach only existing desired resources
  - create/update the VRF or Network through CRUD
  - post-attach after the resource exists
  - deploy pending attachment changes
- Keep `overridden` behavior intact, including omitted-resource pre-delete handling.
- Add focused unit coverage for:
  - VRF `state: replaced` first-create with attachment
  - Network `state: replaced` first-create with attachment
  - skipping missing-resource attachment pre-query before CRUD create

## Test Notes

- Added unit tests covering the first-create replaced attachment workflow for both VRFs and Networks.
- Verified the state machine runs CRUD before post-attachment processing when the resource is absent.
- Verified pre-attachment processing receives an empty current attachment map instead of querying missing resources.

Focused tests to run:

```bash
PYENV_VERSION=ndfclab PYTHONPATH=$PWD/collections python -m pytest -q \
  tests/unit/module_utils/orchestrators/test_vrf_workflow_coordinator.py \
  tests/unit/module_utils/orchestrators/test_networks.py
```

N/A - workflow sequencing/unit-test fix. No new API surface introduced.

- [ ] analyze
- [ ] infa
- [x] manage
- [ ] onemanage
- [ ] other

## Checklist

### DIFF
--- a/plugins/module_utils/orchestrators/network_attachment_manager.py
+++ b/plugins/module_utils/orchestrators/network_attachment_manager.py
@@ -94,6 +94,7 @@ class NetworkAttachmentManager:
         desired: dict[tuple[str, str], dict[str, Any]] | None = None,
         current_network_names: list[str] | None = None,
         current: dict[tuple[str, str], dict[str, Any]] | None = None,
+        attachment_details: list[dict[str, Any]] | None = None,
     ) -> dict[str, Any]:
         state = module_args.get("state", "merged")
         config = module_args.get("config") or []
@@ -119,9 +120,25 @@ class NetworkAttachmentManager:
         query_names = current_network_names
         if query_names is None:
             query_names = None if state == "overridden" else configured_network_names(config)
+        raw_attachment_details = attachment_details
         if current is None:
-            current = self.current_attachment_map(module_args, strategy, query_names)
-            self._trace("network_attachment_current_loaded", phase=phase, queried_network_names=query_names, current_count=len(current))
+            if raw_attachment_details is None:
+                raw_attachment_details = self.current_attachment_details(module_args, strategy, query_names)
+            current = self.attachment_map_from_details(raw_attachment_details, query_names)
+            self._trace(
+                "network_attachment_current_loaded",
+                phase=phase,
+                queried_network_names=query_names,
+                attachment_count=len(raw_attachment_details or []),
+                current_count=len(current),
+            )
+
+        if desired:
+            desired = self.expand_desired_attachments_with_vpc_peers(
+                desired,
+                raw_attachment_details if raw_attachment_details is not None else current.values(),
+            )
+            self._trace("network_attachment_desired_expanded", phase=phase, desired_count=len(desired))
 
         payloads = self.planned_detach_payloads(state, config, current, desired) if phase == "pre" else self.planned_attach_payloads(current, desired)
         if not payloads:
@@ -140,6 +157,7 @@ class NetworkAttachmentManager:
         )
         self._trace("network_attachment_phase_end", phase=phase, payload_count=len(payloads), deploy_target_count=len(deploy_targets))
         trace["current"] = current
+        trace["desired"] = desired
         trace["payloads"] = payloads
         return trace
 
@@ -368,6 +386,43 @@ class NetworkAttachmentManager:
         network_names: list[str] | None = None,
     ) -> dict[tuple[str, str], dict[str, Any]]:
         return self.attachment_map_from_details(self.current_attachment_details(module_args, strategy, network_names), network_names)
+
+    @staticmethod
+    def expand_desired_attachments_with_vpc_peers(
+        desired: dict[tuple[str, str], dict[str, Any]],
+        attachment_details: Any,
+    ) -> dict[tuple[str, str], dict[str, Any]]:
+        """Add vPC peer attachments from attachment-query rows."""
+        if not desired or not attachment_details:
+            return desired
+
+        detail_by_key: dict[tuple[str, str], dict[str, Any]] = {}
+        for attachment in attachment_details:
+            if not isinstance(attachment, dict):
+                continue
+            network_name = attachment.get("networkName")
+            switch_id = attachment.get("switchId")
+            if network_name and switch_id:
+                detail_by_key[(network_name, switch_id)] = attachment
+
+        expanded = dict(desired)
+        for key, payload in list(desired.items()):
+            detail = detail_by_key.get(key)
+            peer_switch_id = detail.get("peerSwitchId") if detail else None
+            if not peer_switch_id:
+                continue
+
+            peer_key = (key[0], peer_switch_id)
+            if peer_key in expanded:
+                continue
+
+            peer_payload = dict(payload)
+            peer_payload["switchId"] = peer_switch_id
+            if "interfaces" in peer_payload:
+                peer_payload["interfaces"] = []
+            expanded[peer_key] = peer_payload
+
+        return expanded
 
     @staticmethod
     def attachment_map_from_details(
--- a/plugins/module_utils/orchestrators/network_attachment_manager.py
+++ b/plugins/module_utils/orchestrators/network_attachment_manager.py
@@ -38,6 +38,9 @@ from ansible_collections.cisco.nd.plugins.module_utils.models.manage_networks.ne
     NetworkAttachmentValidateInterfaceModel,
     NetworkAttachmentValidateInterfacesPayloadModel,
 )
+from ansible_collections.cisco.nd.plugins.module_utils.orchestrators.attachment_vpc_peer_expander import (
+    expand_desired_attachments_with_vpc_peers,
+)
 from ansible_collections.cisco.nd.plugins.module_utils.orchestrators.network_config_utils import (
     configured_network_names,
     deploy_enabled_by_network,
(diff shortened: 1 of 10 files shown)


plugins/module_utils/orchestrators/vrf_state_machine.py
```diff
@@ -154,29 +165,33 @@ def run_overridden(
...
                     module_args,
                     strategy,
-                    current_vrf_names,
+                    attachment_query_vrf_names,
                 )
-                if current_vrf_names
+                if attachment_query_vrf_names
                 else []
             )
             current_desired_attachments = self.coordinator._attachment_map_from_details(
```
mikewiebe, line 179: ### High: `replaced` regresses existing VRF vPC attachment and leaves first-create incomplete

**Issue**

The new `replaced` helper converts raw attachment-query rows into an attached-only map before invoking the VRF attachment manager. That discards `peerSwitchId` metadata carried by `attach: false` vPC rows and suppresses the post-CRUD query that a newly created VRF needs to discover its peer.

This has two distinct compatibility effects. For an existing but unattached VRF, it is a regression: the base implementation queried the raw rows and expanded one selected vPC member to both peers, while this PR no longer does. For first-create, the old flow failed before CRUD; this PR creates the VRF first but can still fail during vPC attachment, so that case is an incomplete fix rather than a strict regression.

**Evidence**

- [`vrf_state_machine.py` line 179](https://github.com/CiscoDevNet/ansible-nd/pull/505/files#diff-877c7cec30271df39312f708718668ef3dcf85c29adf94856ef5c887a01943c1R179) collapses the raw rows with `_attachment_map_from_details()`, then passes that non-`None` cache into both the pre- and post-attachment phases at lines 195-202 and 221-228.
- [The base `replaced` flow](https://github.com/CiscoDevNet/ansible-nd/blob/4b0f681f4b897dde3b4d53f74088ac562982b6db/plugins/module_utils/orchestrators/vrf_state_machine.py#L90-L125) passed no cached `current` value into the pre-phase. For an existing VRF, the real attachment manager therefore queried and retained the raw vPC rows before expanding the desired attachment.
- [`vrf_attachment_manager.py` lines 120-134](https://github.com/CiscoDevNet/ansible-nd/blob/de77240b6d14127e25081efd1f2473e44f63e075/plugins/module_utils/orchestrators/vrf_attachment_manager.py#L120-L134) queries raw details only when `current is None`; [`attachment_map_from_details()` lines 436-452](https://github.com/CiscoDevNet/ansible-nd/blob/de77240b6d14127e25081efd1f2473e44f63e075/plugins/module_utils/orchestrators/vrf_attachment_manager.py#L436-L452) drops every row whose `attach` value is not `True`.
- [`expand_desired_attachments_with_vpc_peers()` lines 301-334](https://github.com/CiscoDevNet/ansible-nd/blob/de77240b6d14127e25081efd1f2473e44f63e075/plugins/module_utils/orchestrators/vrf_attachment_manager.py#L301-L334) explicitly reads `peerSwitchId` from those raw rows. The existing regression fixture at [`test_vrf_workflow_coordinator.py` lines 2082-2125](https://github.com/CiscoDevNet/ansible-nd/blob/de77240b6d14127e25081efd1f2473e44f63e075/tests/unit/module_utils/orchestrators/test_vrf_workflow_coordinator.py#L2082-L2125) models both vPC rows as `attach: false` and expects one desired switch to expand to both peers.
- The added first-create test stubs the real attachment manager and manufactures one-switch output at [`test_vrf_workflow_coordinator.py` lines 252-264](https://github.com/CiscoDevNet/ansible-nd/pull/505/files#diff-63af70e8a067940026cd9ddeb6d5a865f9069e19bab81131a5d318e4e0812dfbR252), so it cannot detect the lost peer.
- A direct probe on the exact head produced: `raw_rows=2`, `attached_map={}`, raw-row expansion `[(BLUE, SERIAL1), (BLUE, SERIAL2)]`, and new cached-map expansion `[(BLUE, SERIAL1)]`.
- The bundled Nexus Dashboard Manage 1.1.411 OpenAPI defines `peerSwitchId` on `vrfAttachmentDetail`, while `POST /fabrics/{fabricName}/vrfAttachments` accepts explicit per-switch attachment rows. The behavior therefore depends on preserving query metadata before building the write payload.

Practical example: assume `10.122.84.181` (`SERIAL1`) and `10.122.84.182` (`SERIAL2`) form a vPC pair. A user intentionally names only one member because the module's existing peer-expansion behavior is supposed to add the other:

```yaml
- name: Replace/create and attach BLUE
  cisco.nd.nd_manage_vrfs:
    fabric_name: test-fabric
    state: replaced
    config:
      - vrf_name: BLUE
        vrf_id: 50001
        attach:
          - ip_address: 10.122.84.181
```

ND exposes the pair through currently unattached query rows:

```json
[
  {"vrfName": "BLUE", "switchId": "SERIAL1", "peerSwitchId": "SERIAL2", "attach": false},
  {"vrfName": "BLUE", "switchId": "SERIAL2", "peerSwitchId": "SERIAL1", "attach": false}
]
```

The expected write contains both `SERIAL1` and `SERIAL2`. This PR filters both discovery rows out and submits only `SERIAL1`, so ND may return `VPC details not found`.

**Existing PR overlap**

No matching existing PR comment found. PR #505 currently has no issue comments, reviews, or review threads.

Related open issue [#506](https://github.com/CiscoDevNet/ansible-nd/issues/506) tracks the first-create sequencing failure this PR intends to fix, but it does not record loss of vPC peer metadata or the same regression for an existing-but-unattached VRF. No exact open issue was found for this vPC regression.

**Impact**

For an existing but unattached VRF, a previously working `replaced` workflow now targets only one vPC member and can fail, making this a concrete regression. For first-create, the PR moves failure from before CRUD to after CRUD: the VRF can be created, but the attachment fails and leaves partial controller state. Re-runs remain unable to converge because the `attach: false` discovery rows are discarded again, and check mode previews the same incorrect one-member plan.

**Suggested fix**

Preserve raw attachment details through vPC peer expansion for existing VRFs. After successful CRUD, query attachment details for desired VRFs that did not exist before the run, then expand their peers before post-attach; mixed existing/new configurations need both sources combined. Add real `VrfAttachmentManager` orchestration tests for first-create and existing-but-unattached vPC flows, plus mixed input, check mode, and 207 failure behavior, asserting both peer attachment and deploy targets.

AKDRG: Great Catch! Fixed it along with Networks.

plugins/module_utils/orchestrators/network_attachment_manager.py
```diff
@@ -369,6 +387,43 @@ def current_attachment_map(
     ) -> dict[tuple[str, str], dict[str, Any]]:
         return self.attachment_map_from_details(self.current_attachment_details(module_args, strategy, network_names), network_names)
 
+    @staticmethod
+    def expand_desired_attachments_with_vpc_peers(
```
mtarking, line 394: expand_desired_attachments_with_vpc_peers appears to be a copy-paste into the Network manager and essentially verbatim from vrf_attachment_manager.py. Two independent copies may naturally drift. Consider a shared helper/mixin via an enhancement issue so a future peer-expansion fix doesn't have to be applied twice (and risk being applied to only one). In the mean time, maybe a comment in or above both the network and vrf usage to inform the user to review the other if a change is made in one.

mtarking (changes requested):

plugins/module_utils/orchestrators/network_attachment_manager.py
AKDRG: Made these into wrapper methods which use the shared utility function.

mtarking (approved):

mikewiebe (approved):

PR #507, opened 2026-08-12
Add gathered state for ethernet access, trunk-host, and fabric update group modules

deekpand-cisco:
## Related Issue(s)

Depends on PR #391 (gathered state framework + loopback/local_user filtering).
This PR is stacked on top of PR #391 

## Proposed Changes

Extends gathered state and Lucene filtering to additional modules:
- `nd_interface_ethernet_trunk_host`
- `nd_fabric_update_group`

Changes include:
- Model `gathered_spec` and `gathered_transform` definitions
- Orchestrator gathered workflow with pagination in base_interface, ethernet_base, and fabric_update_group
- `gathered` added to state choices with documentation and examples
- Unit tests for gathered filtering round-trip
- Integration test tasks for gathered state

## Test Notes

- All 4093 unit tests passing (full repo-wide suite)
- Integration tested against live ND (ethernet_access, ethernet_trunk_host, fabric_update_group gathered states)

4.2.1.10

* [ ] analyze
* [ ] infa
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

### DIFF
--- a/plugins/module_utils/gathered_filter.py
+++ b/plugins/module_utils/gathered_filter.py
@@ -71,6 +71,7 @@ def build_lucene_expressions(
     can union and deduplicate their responses to preserve gathered-state OR
     semantics without relying on endpoint-specific Lucene OR support.
     """
+    # TODO(4.2.1) interface-lucene-or-silently-empty
     filter_items = filters or [{}]
     expressions: list[str] = []
 
--- a/plugins/module_utils/gathered_filter.py
+++ b/plugins/module_utils/gathered_filter.py
@@ -203,13 +203,15 @@ def filter_gathered_response(
     filters: list[dict[str, Any]],
     model_class: type[NDBaseModel],
     normalize_filter: Callable[[dict[str, Any]], dict[str, Any]] | None = None,
-) -> list[dict[str, Any]]:
+) -> list[NDBaseModel]:
     """
-    Apply gathered filters to raw API objects.
+    Apply gathered filters to raw API objects and return validated models.
 
     Criteria inside one filter item use AND semantics.
     Multiple filter items use OR semantics.
-    Returned objects remain in API shape for NDConfigCollection.from_api_response().
+
+    Returns validated model instances directly so callers can construct an
+    NDConfigCollection without redundant from_response() calls.
     """
     # Validation duplicates validate_gathered_filters() intentionally so this
     # function remains safe to call standalone (e.g., from unit tests).
@@ -227,7 +229,7 @@ def filter_gathered_response(
 
         active_filters.append(normalized)
 
-    filtered: list[dict[str, Any]] = []
+    filtered_models: list[NDBaseModel] = []
     seen_identifiers = set()
 
     for response_item in response_data:
@@ -243,9 +245,9 @@ def filter_gathered_response(
             continue
 
         seen_identifiers.add(identifier)
-        filtered.append(response_item)
+        filtered_models.append(model)
 
-    return filtered
+    return filtered_models
 
 
 def validate_gathered_filters(
--- a/plugins/module_utils/gathered_filter.py
+++ b/plugins/module_utils/gathered_filter.py
@@ -108,6 +108,102 @@ def _contains_active_value(value: Any) -> bool:
     # valid loopback fields and are treated as absent criteria.
     return value not in (None, "")
 
+def _extract_active_leaf_paths(data: dict[str, Any], prefix: str = "") -> set[str]:
+    """Return dot-paths for every non-None, non-empty leaf value in a nested dict.
+
+    Given a gathered filter dict (which may be nested due to Ansible's argument
+    spec structure), this walks the tree and identifies which properties the user
+    actually supplied a value for.
+
+    Ansible pads all omitted suboptions with None after argument validation.
+    Those Nones are not user-supplied criteria and are correctly skipped.
+
+    Args:
+        data: A (possibly nested) filter dict from the user's config item.
+        prefix: Internal recursion prefix — callers should not pass this.
+
+    Returns:
+        A set of dot-path strings representing active (non-None) leaf values.
+
+    Examples:
+        >>> _extract_active_leaf_paths({"switch_ip": "10.1.1.1", "interface_name": None})
+        {"switch_ip"}
+
+        >>> _extract_active_leaf_paths({
+        ...     "config_data": {"network_os": {"policy": {"vrf": "mgmt", "ip": None}}}
+        ... })
+        {"config_data.network_os.policy.vrf"}
+
+        >>> _extract_active_leaf_paths({"login_id": "admin"})
+        {"login_id"}
+
+        >>> _extract_active_leaf_paths({"config_data": None})
+        set()  # None at any level is skipped
+    """
+    paths: set[str] = set()
+    for key, value in data.items():
+        full_path = f"{prefix}.{key}" if prefix else key
+        if isinstance(value, dict):
+            paths |= _extract_active_leaf_paths(value, full_path)
+        elif value not in (None, "", []):
+            # False and 0 ARE active values (e.g., admin_state: false)
+            paths.add(full_path)
+    return paths
+
+
+def _reject_unsupported_filter_properties(
+    filter_item: dict[str, Any],
+    supported_properties: tuple[str, ...],
+    index: int,
+) -> None:
+    """Raise ValueError if a filter item contains properties not in the supported set.
+
+    This is the enforcement function for the model's gathered_filter_properties
+    declaration. Called exactly once per filter item during pre-flight validation,
+    before any API request or normalization.
+
+    Args:
+        filter_item: One raw filter dict from the user's config list.
+        supported_properties: The model's declared tuple of allowed dot-paths.
+        index: The position of this filter item in the config list (for error messages).
+
+    Raises:
+        ValueError: With a message listing unsupported properties and the full
+            set of supported properties. Formatted for direct use in Ansible's
+            module.fail_json(msg=...).
+
+    Examples:
+        # Passes (switch_ip is supported):
+        >>> _reject_unsupported_filter_properties(
+        ...     {"switch_ip": "10.1.1.1"},
+        ...     ("switch_ip", "interface_name"),
+        ...     index=0,
+        ... )  # No error
+
+        # Fails (extra_config is not supported):
+        >>> _reject_unsupported_filter_properties(
+        ...     {"config_data": {"network_os": {"policy": {"extra_config": "cli"}}}},
+        ...     ("switch_ip", "interface_name", "config_data.network_os.policy.vrf"),
+        ...     index=0,
+        ... )
+        # ValueError: "Gathered filter at index 0 contains unsupported properties: 
+        #   'config_data.network_os.policy.extra_config'. 
+        #   Supported gathered filter properties: 'switch_ip', 'interface_name', ..."
+    """
+    supplied_paths = _extract_active_leaf_paths(filter_item)
+    supported_set = set(supported_properties)
+    unsupported = sorted(supplied_paths - supported_set)
+    if unsupported:
+        raise ValueError(
+            "Gathered filter at index {0} contains unsupported properties: {1}. "
+            "Supported gathered filter properties: {2}".format(
+                index,
+                ", ".join("'{0}'".format(p) for p in unsupported),
+                ", ".join("'{0}'".format(p) for p in supported_properties),
+            )
+        )
+
+
 
 def filter_gathered_response(
     response_data: list[dict[str, Any]],
@@ -160,24 +256,54 @@ def filter_gathered_response(
 def validate_gathered_filters(
         filters: list[Any],
         normalize_filter: Callable[[dict[str, Any]], dict[str, Any]] | None = None,
+        supported_properties: tuple[str, ...] = (),
 ) -> None:
     """ 
     Pre-flight validation for gathered filter items.
 
-    Raises ValueError if any filter item is malformed or contains no actionable 
-    criteria. Call this before making API requests so that invalid input is r
-    rejected without wasted network I/O.
+    Validates each filter item through a strict pipeline:
+      1. Type check — must be a dict
+      2. Property check — all active properties must be in the supported set
+      3. Normalize — transform values to canonical form (e.g., lowercase)
+      4. Active check — at least one non-None criterion must exist
+
+    Property validation runs BEFORE normalization because:
+      - It checks KEYS (property paths), normalization transforms VALUES.
+      - Unsupported properties should be rejected without wasting normalization effort.
+      - Error messages should show what the user wrote, not the normalized form.
+
+    Args:
+        filters: The raw config list from module.params["config"].
+            Example: [{"switch_ip": "10.1.1.1", "interface_name": "loopback0"}]
+
+        normalize_filter: Model-specific value transformer (e.g., lowercase interface_name).
+            Signature: (dict) -> dict. May be None if no normalization needed.
+
+        supported_properties: The model's gathered_filter_properties tuple.
+            Example: ("switch_ip", "interface_name", "config_data.network_os.policy.vrf")
+            Empty tuple means property validation is skipped (backward-compatible).
+
+    Raises:
+        ValueError: If any filter item fails any validation step.
     """
     for idx, filter_item in enumerate(filters):
+        # Step 1: Type check
         if not isinstance(filter_item, dict):
             raise ValueError(
                 f"Each gathered filter item must be a dictionary, "
                 f"got {type(filter_item).__name__} at index {idx}."
             )
+        
+        # Step 2: Property validation (before normalize - checks keys, not values)
+        if supported_properties:
+            _reject_unsupported_filter_properties(filter_item, supported_properties, idx)
+
+        # Step 3: Normalize values to canonical form
         normalized = deepcopy(filter_item)
         if normalize_filter is not None:
             normalized = normalize_filter(normalized)
         
+        #Step 4: Active value check
         if not _contains_active_value(normalized):
             raise ValueError(
                 f"Gathered filter item at index {idx} must contain "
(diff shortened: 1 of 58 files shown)


plugins/module_utils/orchestrators/ethernet_base.py
```diff
@@ -617,7 +642,31 @@ def query_one(self, model_instance: ModelType, **kwargs) -> ResponseType:
...
+        ### RuntimeError
+
+        - Via `FabricContext.switch_map` if the switches API query fails.
+        """
+        switch_map = self.fabric_context.switch_map
+        if self.rest_send.params.get("state") in ("overridden", "gathered"):
+            return switch_map
+        config_items = self.rest_send.params.get("config") or []
+        config_ips = {item.get("switch_ip") for item in config_items if item.get("switch_ip")}
+        return {ip: sid for ip, sid in switch_map.items() if ip in config_ips}
```
allenrobel: This override reintroduces the `_switches_to_query` duplication that PR #390 deliberately hoisted into `NDBaseInterfaceOrchestrator` — it is near-identical to the base-class version except for adding `"gathered"` to the state check. The `gathered` branch also appears unreachable: both ethernet orchestrators set `gathered_lucene_spec`, so `query_all` short-circuits to `_query_all_for_gathered()` (which reads `fabric_context.switch_map` directly) before `_switches_to_query` is ever consulted for that state. If the `gathered` condition is genuinely needed, it belongs in the base-class method; otherwise this override can be dropped. If kept, the docstring still says only `state: overridden` is fabric-wide.

plugins/module_utils/models/base.py
```diff
@@ -117,6 +154,43 @@ def _build_payload_nested(self, data: Dict[str, Any]) -> Dict[str, Any]:
...
+        """
+        keys: Set[str] = set()
+        for field_name, field_info in cls.model_fields.items():
+            extra = field_info.json_schema_extra
+            if isinstance(extra, dict) and extra.get("secret"):
+                keys.add((field_info.alias or field_name) if by_alias else field_name)
+        return keys
+
+    @classmethod
+    def collect_secret_values(cls, config_item: Dict[str, Any]) -> Set[str]:
```
allenrobel: New code should use modern (PEP 585/604) annotations per team convention: `dict[str, Any]` / `set[str]` rather than `Dict` / `Set`.

plugins/module_utils/nd_output.py
```diff
@@ -6,31 +6,63 @@
...
         self._changed: bool = False
         self._before: Union[NDConfigCollection, List] = []
         self._after: Union[NDConfigCollection, List] = []
         self._diff: Union[NDConfigCollection, List] = []
         self._proposed: Union[NDConfigCollection, List] = []
         self._logs: List = []
         self._extra: Dict[str, Any] = {}
+        # Argument-spec ``config.options`` mapping used to prune gathered output
+        # down to valid module arguments so it round-trips as ``config``.
+        self._gathered_spec: Dict[str, Any] = {}
```
allenrobel: Modern annotations for new code: `dict[str, Any]` here, and `dict[str, Any] | None` for the new `gathered_spec` parameter in `assign()` (L136), rather than `Dict` / `Optional[Dict[...]]`.

tests/unit/module_utils/models/test_base_secrets.py
```diff
@@ -0,0 +1,46 @@
...
+
+"""
+Unit tests for the generic (base-model) secret handling: a single
+``json_schema_extra={"secret": True}`` tag drives both no_log value collection
+and output/diff stripping, while the value is kept in the controller payload.
+"""
+
+from __future__ import annotations
+
+from typing import ClassVar, List, Optional
```
allenrobel: New file: prefer modern annotations — `str | None`, `ClassVar[list[str] | None]` — and drop the `List` / `Optional` imports (the file already has `from __future__ import annotations`).

tests/unit/module_utils/test_utils.py
```diff
@@ -0,0 +1,110 @@
...
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+"""
+Unit tests for module_utils/utils.py helpers.
+"""
+
+from __future__ import absolute_import, annotations, division, print_function
+
+__metaclass__ = type  # pylint: disable=invalid-name
```
allenrobel: New file: only `from __future__ import annotations` is needed — `absolute_import` / `division` / `print_function` and `__metaclass__ = type` are Python 2 boilerplate we no longer add in new code.

plugins/module_utils/gathered_filter.py
```diff
@@ -0,0 +1,302 @@
...
+def build_lucene_expressions(
+    filters: list[dict[str, Any]],
+    spec: GatheredLuceneSpec,
+) -> list[str]:
+    """Build one server-side AND expression per gathered config item.
+
+    Separate config items intentionally remain separate expressions. Callers
+    can union and deduplicate their responses to preserve gathered-state OR
+    semantics without relying on endpoint-specific Lucene OR support.
+    """
```
allenrobel, line 73: The one-expression-per-config-item design here is the workaround for the interface list endpoint's Lucene OR bug (OR returns empty instead of erroring). Per the team's workaround-marker convention it should carry `# TODO(4.2.1) interface-lucene-or-silently-empty` (the bug-tracker note for this discrepancy) so the workaround can be audited once ND fixes the endpoint. Each workaround site gets its own marker — see also the sibling sites in `base_interface.py` and `loopback_interface.py`.

plugins/module_utils/orchestrators/base_interface.py
```diff
@@ -328,6 +337,136 @@ def _configure_endpoint(self, api_endpoint, switch_sn: str):
...
 
+    def _build_gathered_query_plan(self, gathered_filters: list[dict]) -> dict[str, tuple[str, set[str]]]:
+        """
+        # Summary
+
+        Build one or more Lucene expressions for each target switch from gathered filters.
+
+        Uses the orchestrator's ``gathered_lucene_spec`` ClassVar to map Ansible filter fields to Lucene
+        fields names. Each filter item produces a separate expression (the interface endpoint does not
+        reliably support Lucene OR). If a filter item includes ``switch_ip``, only that switch is targeted;
```
allenrobel, line 348: Second Lucene-OR workaround site — should also carry `# TODO(4.2.1) interface-lucene-or-silently-empty` per the workaround-marker convention.

plugins/module_utils/orchestrators/loopback_interface.py
```diff
@@ -224,15 +249,153 @@ def query_one(self, model_instance: LoopbackInterfaceModel, **kwargs) -> Respons
...
+        policy_type = interface.get("configData", {}).get("networkOS", {}).get("policy", {}).get("policyType")
+
+        return policy_type == "loopback"
+
+    def _build_gathered_query_plan(self, gathered_filters: list[dict]) -> dict[str, tuple[str, set[str]]]:
+        """
+        Build one or more Lucene expressions for each target switch.
+
+        Multiple filter items remain separate expressions because the
+        interface endpoint does not reliably support Lucene OR.
```
allenrobel: Third Lucene-OR workaround site — should also carry `# TODO(4.2.1) interface-lucene-or-silently-empty` per the workaround-marker convention (moot if this copy is consolidated into the base class as suggested below).

plugins/module_utils/orchestrators/loopback_interface.py
```diff
@@ -224,15 +249,153 @@ def query_one(self, model_instance: LoopbackInterfaceModel, **kwargs) -> Respons
...
+        policy. System-managed underlay loopbacks are intentionally excluded.
+        """
+        if interface.get("interfaceType") != "loopback":
+            return False
+
+        policy_type = interface.get("configData", {}).get("networkOS", {}).get("policy", {}).get("policyType")
+
+        return policy_type == "loopback"
+
+    def _build_gathered_query_plan(self, gathered_filters: list[dict]) -> dict[str, tuple[str, set[str]]]:
```
allenrobel: `_build_gathered_query_plan` / `_query_interfaces_with_lucene` / `_MAX_EXPRESSIONS_PER_SWITCH` here are near-byte-identical to the versions this PR adds to `NDBaseInterfaceOrchestrator` — the only functional difference is loopback setting `endpoint_params.config_only = False`. Two independently maintained copies of the pagination/fan-out logic will drift (any future fix to one is easily missed in the other). Suggest the same consolidation PR #390 did for `_switches_to_query`: a small hook on the base class for the `config_only` divergence, and delete loopback's local copies in favor of the inherited implementation.

plugins/module_utils/orchestrators/ethernet_trunk_host_interface.py
```diff
@@ -114,7 +126,9 @@ def query_all(self, model_instance: NDBaseModel | None = None, **kwargs) -> Resp
 
         - Propagated from `EthernetBaseOrchestrator.query_all` on query failure.
         """
-        result = super().query_all(model_instance=model_instance, **kwargs)
+        result = super().query_all(model_instance=model_instance, gathered_filters=gathered_filters, **kwargs)
         if not isinstance(result, list):
             return result
+        if self.rest_send.params.get("state") == "gathered":
+            return result
         return [iface for iface in result if not self._is_unconfigured_default(iface)]
```
allenrobel: This early return skips `_is_unconfigured_default`, so an unscoped `state: gathered` run returns every factory-default trunkHost interface on the queried switches (ND echoes `policyType: trunkHost` for ports the user never configured). That contradicts the module DOCUMENTATION's promise that gathered reads "user-managed trunkHost interfaces", and diverges from loopback, where the managed-only filter applies to both the management and gathered paths. If returning raw defaults for gathered is intentional, the docs and this method's docstring should say so and a unit test should pin the behavior; otherwise the filter should apply on the gathered path too.

allenrobel (commented):
### Code review

The gathered-state extension looks solid overall — I traced the Lucene value escaping, per-switch fan-out, pagination, legacy-vs-lucene mutual exclusion, and identifier dedup and found no logic bugs. The inline comments below are consolidation/convention items plus one behavioral question on trunk-host gathered scope. One item couldn't be anchored inline: `query_all`'s docstring in `ethernet_base.py` (https://github.com/CiscoDevNet/ansible-nd/blob/cc4367da5ef77d493281aefc9b4d1d96340ec78f/plugins/module_utils/orchestrators/ethernet_base.py#L675-L677) shares the stale "fabric-wide for `state: overridden`" claim raised on `_switches_to_query`.

plugins/module_utils/gathered_filter.py
deekpand-cisco: I have added # TODO(4.2.1) interface-lucene-or-silently-empty right before the loop in [build_lucene_expressions](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html). This way it's easy to grep for when ND fixes the OR behavior.

plugins/module_utils/orchestrators/base_interface.py
deekpand-cisco: I have added the same marker as the first line inside [_build_gathered_query_plan](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) body. Thank you!!

plugins/module_utils/orchestrators/loopback_interface.py
deekpand-cisco: This one function is removed now — loopback's [_build_gathered_query_plan](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) was deleted as part of the consolidation. The marker lives in the single base-class copy.

plugins/module_utils/orchestrators/loopback_interface.py
deekpand-cisco: I do agree with this, thank you for pointing it out. 
I have deleted all three duplicated members from [LoopbackInterfaceOrchestrator](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) — it now inherits them from [NDBaseInterfaceOrchestrator](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html). The only real difference was loopback needing [endpoint_params.config_only = False](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), so I added a [_configure_lucene_endpoint](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) hook on the base class (no-op by default) that loopback overrides with a one-liner. Also cleaned up the unused [build_lucene_expressions](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) import.

plugins/module_utils/orchestrators/ethernet_base.py
deekpand-cisco: Thanks Allen , this makes sense. I have dropped the override entirely and updated the [query_all](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) docstring to mention the gathered path.

plugins/module_utils/orchestrators/ethernet_trunk_host_interface.py
deekpand-cisco: Hey Allen, Thank you for pointing this out. It was miss from my part. I believe it would be better to return only managed trunk hosts, and not all factory default interfaces. I have removed the early-return so [_is_unconfigured_default](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) now runs on all paths including gathered.

plugins/module_utils/models/base.py
deekpand-cisco: Changed the declaration as per suggested.

plugins/module_utils/nd_output.py
deekpand-cisco: Chnages with [self._gathered_spec](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) as [dict[str, Any]](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) now, and [gathered_spec](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) param in [assign()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) uses [dict[str, Any] | None](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html).

tests/unit/module_utils/test_utils.py
deekpand-cisco: Dropped these imports.

tests/unit/module_utils/models/test_base_secrets.py
deekpand-cisco: Cleaned up — dropped [List](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)/[Optional](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) imports, using [ClassVar[list[str] | None]](http://vscodecontentref/38) and [str | None](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) throughout.

nikhilsrikrishna:
`state: gathered` currently appears to run the same fabric prerequisite validation as mutation states.

The gathered path initializes the state machine and calls the interface orchestrator's `query_all()`. Both the loopback and shared Ethernet query paths call `validate_prerequisites()`, which delegates to `FabricContext.validate_for_mutation()`. That validation rejects a fabric when deployment freeze is enabled.

As a result, read-only gathering fails on a deployment-frozen fabric, even though no configuration is created, updated, deleted, saved, or deployed. This currently affects loopback, Ethernet access, and Ethernet trunk-host gathered operations, and the same behavior would be inherited by future interface modules using this base path.

Could we separate read prerequisites from mutation prerequisites? For example:

- Add `FabricContext.validate_for_read()` to validate that the fabric exists and is accessible through the targeted controller.
- Have `validate_for_mutation()` call `validate_for_read()` and then additionally enforce the deployment-freeze restriction.
- Use `validate_for_read()` when `state: gathered`.
- Continue using `validate_for_mutation()` for `merged`, `replaced`, `overridden`, and `deleted`.

The state machine could pass the operation intent explicitly to the orchestrator, or the base interface orchestrator could select the appropriate validation based on the current state.

This would preserve the existing mutation protection while allowing gathered state to inspect configuration during deployment freeze, which seems consistent with its read-only contract.

deekpand-cisco:
> 

Hey Nikhil, Thank you for the feedback.
I have worked on this. 

FabricContext now has validate_for_read() (fabric exists + belongs to this controller), and validate_for_mutation() calls it then adds the freeze check. Same behaviour as before for mutations, just built from two pieces.

Went with your second option for the dispatch — read_only_states on NDBaseOrchestrator plus a branch in NDBaseInterfaceOrchestrator.validate_prerequisites. 
I tried moving validation into preflight(), which looks cleaner. It doesn't work — manage_state only calls preflight() for merged/replaced/overridden, so deletes on a frozen fabric would go straight through. query_all is the only entry point every state hits, so it has to stay there. Added test_loopback_interface_00820 to pin it.

The state check is an allowlist, so only gathered takes the relaxed path — an unknown or missing state falls through to the mutation checks. merged, replaced, overridden and deleted are untouched.

Issue #466, opened 2026-08-03
[Staged configuration audit] VRF module


sivakasi-cisco:
## Parent Scope

Sub-issue of #453: staged configuration addition and removal audit.

This issue must evaluate **both staged additions and staged removals** for `replaced`, `overridden`, and `deleted` wherever those states are supported. An unsupported state must be recorded as `Not Applicable` with a technical reason; it must not be silently skipped.

## Applicability Assessment - Required First

> **Before implementation, first evaluate whether staged configuration addition and removal are technically applicable to this module and its Nexus Dashboard endpoints.**

- [ ] Identify whether the operation produces deployable switch intent, controller-only inventory or topology state, or an automatically deployed action.
- [ ] Identify whether ND provides a separate save or recalculation boundary and a separate deploy boundary.
- [ ] Record the applicability result as `Applicable`, `Partially Applicable`, or `Not Applicable`, with technical evidence.
- [ ] If staging is not applicable, document the endpoint limitation and obtain agreement in parent issue #453; do not force unsupported behavior into the module.

## Module Scope

- `nd_manage_vrfs`

## Module-Specific Problem Statement

VRF removal can require Network or L3Out dependency handling before deletion. Determine whether VRF additions, replacement resets, overridden omissions, and deletion can be staged without deploying dependent switch configuration.

The audit must establish the current behavior for additions, updates, removals, detachments, and deletions that this module supports. Pass or fail must be based on observed controller state, switch synchronization state, and API evidence rather than only the Ansible input flags.

## Required State Validation

| State | Required validation for this module |
|---|---|
| `replaced` | Validate staged additions and updates, plus any setting removal or reset caused by replacement. |
| `overridden` | Validate staged additions and removal or detachment of existing resources omitted from the desired configuration. |
| `deleted` | Validate whether the module deletes or detaches the selected resource, creates pending intent, performs dependency cleanup, or deploys automatically. |

For a state the module does not support, record `Not Applicable`, the technical reason, and any endpoint or architecture limitation.

## Save and Deploy Validation

- [ ] If `save` is supported, test `save=false` and `deploy=false`: no save or deploy action should run, and associated switches should remain `inSync`. Record whether controller data still changes.
- [ ] If `save` is supported, test `save=true` and `deploy=false`: intent should be saved without deployment, the first expected switch state is `pending`, and no deploy endpoint should run.
- [ ] Preserve and validate the existing `save=true` and `deploy=true` behavior.
[...]

Issue #465, opened 2026-08-03
[Staged configuration audit] Network module


sivakasi-cisco:
## Parent Scope

Sub-issue of #453: staged configuration addition and removal audit.

This issue must evaluate **both staged additions and staged removals** for `replaced`, `overridden`, and `deleted` wherever those states are supported. An unsupported state must be recorded as `Not Applicable` with a technical reason; it must not be silently skipped.

## Applicability Assessment - Required First

> **Before implementation, first evaluate whether staged configuration addition and removal are technically applicable to this module and its Nexus Dashboard endpoints.**

- [ ] Identify whether the operation produces deployable switch intent, controller-only inventory or topology state, or an automatically deployed action.
- [ ] Identify whether ND provides a separate save or recalculation boundary and a separate deploy boundary.
- [ ] Record the applicability result as `Applicable`, `Partially Applicable`, or `Not Applicable`, with technical evidence.
- [ ] If staging is not applicable, document the endpoint limitation and obtain agreement in parent issue #453; do not force unsupported behavior into the module.

## Module Scope

- `nd_manage_networks`

## Module-Specific Problem Statement

Network overridden and deleted workflows can detach switch attachments and may perform undeploy operations before removing the Network. Determine whether Network additions, attachment removals, and deletion can be staged in ND without automatically deploying or undeploying switch configuration.

The audit must establish the current behavior for additions, updates, removals, detachments, and deletions that this module supports. Pass or fail must be based on observed controller state, switch synchronization state, and API evidence rather than only the Ansible input flags.

## Required State Validation

| State | Required validation for this module |
|---|---|
| `replaced` | Validate staged additions and updates, plus any setting removal or reset caused by replacement. |
| `overridden` | Validate staged additions and removal or detachment of existing resources omitted from the desired configuration. |
| `deleted` | Validate whether the module deletes or detaches the selected resource, creates pending intent, performs dependency cleanup, or deploys automatically. |

For a state the module does not support, record `Not Applicable`, the technical reason, and any endpoint or architecture limitation.

## Save and Deploy Validation

- [ ] If `save` is supported, test `save=false` and `deploy=false`: no save or deploy action should run, and associated switches should remain `inSync`. Record whether controller data still changes.
- [ ] If `save` is supported, test `save=true` and `deploy=false`: intent should be saved without deployment, the first expected switch state is `pending`, and no deploy endpoint should run.
[...]

Issue #453, opened 2026-07-28
Support staged removal for all modules 


sivakasi-cisco:
## Summary

Audit and support staged configuration removal for mutating `ansible-nd` modules, including resources removed or detached through `overridden` and `deleted` workflows.

The immediate requirement is to determine how every in-scope module behaves today when configuration is added or removed without deployment. The audit results will then be used to identify rework, prioritize implementation, and assess the impact on the initial release.

This applies to modules with and without `config_actions`. The audit is separate from standardizing the collection-wide `config_actions` interface in #368.

## Priority and Delivery Approach

This is a P0/P1 discovery item for the initial release.

The work has two phases:

1. **Current-behavior audit:** Test and document how each mutating module handles staged additions and removals.
2. **Implementation:** Prioritize and implement required changes after the audit identifies the gaps and rework level.

Completing the audit does not mean every gap must be fixed immediately. Findings will determine implementation priority, effort, dependencies, and release impact.

**Audit target: 7 August 2026.**

## Definition of Staged Configuration

A configuration addition or removal is considered staged when:

- The intended change is accepted by Nexus Dashboard.
- The change is not deployed to the switches.
- For switch-associated configuration, the first expected synchronization state is `pending`.
- A later compliance or configuration-difference evaluation may show `outOfSync`.
- A later explicit deployment can apply the staged change.
- No deploy endpoint is called while deployment is disabled.

An `outOfSync` state after a later compliance check does not mean that the configuration was deployed.

## Use Cases

Examples include:

- Detaching a VRF.
- Detaching a Network.
- Detaching Network port attachments.
- Removing a line from an ACL or routing policy.
- Deleting a policy.
- Removing resources omitted from an `overridden` configuration.
- Deleting or tearing down a vPC pair.
- Changing fabric settings without immediately deploying them.

## States to Validate

| State | Behavior to validate |
|---|---|
| `replaced` | Validate staged additions, updates, and any setting removal or reset caused by replacement. |
| `overridden` | Validate staged additions and removal of resources omitted from the desired configuration. |
| `deleted` | Document whether the module deletes or detaches the resource, creates pending configuration, or deploys automatically. |

The `deleted` state requires particular attention because module families may perform dependency cleanup, detach, save, and deployment operations differently.

### Modules that support `save`

| Save | Deploy | Expected validation |
|---|---|---|
[...]

PR #509, opened 2026-08-13
ND Manage VRFs & Networks: Add Internal Staged Workflow

AKDRG:
## Related Issue(s)

Related to:

- https://github.com/CiscoDevNet/ansible-nd/issues/465
- https://github.com/CiscoDevNet/ansible-nd/issues/466

## Proposed Changes

Add a private/internal `_staged` workflow state for `nd_manage_vrfs` and `nd_manage_networks`.

The new workflow supports staged configuration removal where omitted resources should have attachments detached, but the omitted VRFs or Networks themselves should not be deleted.

Key behavior:

- `_staged` follows `overridden` attachment scope.
- Current VRF/Network attachments are queried across the full current resource set.
- Attachments omitted from desired config are detached.
- Generic CRUD execution is mapped to `replaced`, preventing overridden-style delete requests.
- Deployment is always suppressed for `_staged`.
- Delete deploy payloads, pending deploy fallback, and delete-readiness wait flows are not invoked.
- The implementation stays localized to the VRF/Network state machines and attachment managers.

## Test Notes

Commands run:

PYTHONPATH=/Users/achengam/Documents/Ansible_Dev/NDNetwork_empty_interfaces \
/Users/achengam/.pyenv/versions/ndfclab/bin/python \
-m pytest tests/unit/module_utils/orchestrators/test_networks.py -k staged -q

Result:

2 passed, 80 deselected

PYTHONPATH=/Users/achengam/Documents/Ansible_Dev/NDNetwork_empty_interfaces \
/Users/achengam/.pyenv/versions/ndfclab/bin/python \
-m pytest tests/unit/module_utils/orchestrators/test_vrf_workflow_coordinator.py -k staged -q

Result:

2 passed, 54 deselected

black -l 159 --check \
  plugins/module_utils/orchestrators/network_attachment_manager.py \
  plugins/module_utils/orchestrators/network_state_machine.py \
  plugins/module_utils/orchestrators/vrf_attachment_manager.py \
  plugins/module_utils/orchestrators/vrf_state_machine.py \
  plugins/modules/nd_manage_networks.py \
  plugins/modules/nd_manage_vrfs.py \
  tests/unit/module_utils/orchestrators/test_networks.py \
  tests/unit/module_utils/orchestrators/test_vrf_workflow_coordinator.py

Result:

8 files would be left unchanged.

Workflow-layer change. No new API endpoint schema dependency introduced.

- [ ] analyze
- [ ] infra
- [x] manage
- [x] onemanage
- [ ] other

## Checklist

### DIFF
--- a/plugins/module_utils/orchestrators/network_state_machine.py
+++ b/plugins/module_utils/orchestrators/network_state_machine.py
@@ -176,10 +176,19 @@ class NetworkStateMachine:
         desired_attachments = self.coordinator._desired_attachment_map(module_args, strategy)
         desired_network_names = self.coordinator._configured_network_names(config)
 
-        sm, original_config, original_state = self.coordinator._new_state_machine(query_module_args(module_args), strategy)
+        current_network_names = []
+        if state == "staged":
+            query_sm, query_original_config, query_original_state = self.coordinator._new_state_machine(query_module_args(module_args), strategy)
+            try:
+                current_network_names = self._network_names_from_models(query_sm.existing)
+            finally:
+                self.coordinator._restore_state_machine_params(query_original_config, query_original_state)
+
+        sm, original_config, original_state = self.coordinator._new_state_machine(crud_module_args(module_args), strategy)
         try:
             prepare_crud_state(sm, state)
-            current_network_names = self._network_names_from_models(sm.existing)
+            if state != "staged":
+                current_network_names = self._network_names_from_models(sm.existing)
             current_network_name_set = set(current_network_names)
             desired_network_name_set = set(desired_network_names)
             current_desired_network_names = [network_name for network_name in desired_network_names if network_name in current_network_name_set]
--- a/plugins/module_utils/orchestrators/network_state_machine.py
+++ b/plugins/module_utils/orchestrators/network_state_machine.py
@@ -12,9 +12,9 @@ This wrapper owns the Network-specific workflow around that CRUD phase:
   - post-attach for merged/replaced/overridden
   - deploy of pending attachment changes
 
-The private ``staged`` state follows overridden attachment handling but maps
-the CRUD phase to replaced and suppresses deploy so omitted Networks are
-detached without deploy or delete calls.
+The ``staged`` state follows overridden attachment handling but maps the CRUD
+phase to replaced and suppresses deploy so omitted Networks are detached
+without deploy or delete calls.
 
 It deliberately composes the generic NDStateMachine through the workflow
 coordinator helpers instead of changing the shared state-machine contract.
@@ -161,9 +161,9 @@ class NetworkStateMachine:
 
         The pre-detach phase must only query attachments for Networks that
         already exist. First-create ``state=replaced`` tasks can then create
-        the Network before the post-attach phase runs. The private ``staged``
-        state uses overridden detach scope but runs CRUD as replaced to
-        suppress deletes.
+        the Network before the post-attach phase runs. The ``staged`` state
+        uses overridden detach scope but runs CRUD as replaced to suppress
+        deletes.
         """
         state = module_args.get("state", "merged")
         config = [dict(network) for network in module_args.get("config") or []]
@@ -310,7 +310,7 @@ class NetworkStateMachine:
 
     @staticmethod
     def _prepare_crud_state(sm: Any, requested_state: str) -> None:
-        """Switch private staged workflows to replacement CRUD after query."""
+        """Switch staged workflows to replacement CRUD after query."""
         if requested_state != "staged":
             return
         sm.state = "replaced"
--- a/plugins/module_utils/orchestrators/network_state_machine.py
+++ b/plugins/module_utils/orchestrators/network_state_machine.py
@@ -13,7 +13,8 @@ This wrapper owns the Network-specific workflow around that CRUD phase:
   - deploy of pending attachment changes
 
 The private ``staged`` state follows overridden attachment handling but maps
-the CRUD phase to replaced so omitted Networks are detached without delete calls.
+the CRUD phase to replaced and suppresses deploy so omitted Networks are
+detached without deploy or delete calls.
 
 It deliberately composes the generic NDStateMachine through the workflow
 coordinator helpers instead of changing the shared state-machine contract.
@@ -259,6 +260,9 @@ class NetworkStateMachine:
             self._trace("attachment_phase_post_end", changed=post_attach.get("changed"), payload_count=len(post_attach.get("payloads", [])))
             self.coordinator._merge_api_trace(result, post_attach)
 
+            if state == "staged":
+                return result
+
             return self._deploy_after_attachment_changes(
                 result,
                 config,
@@ -403,6 +407,8 @@ class NetworkStateMachine:
             omitted_network_names,
             attachment_details=omitted_attachment_details,
         )
+        if module_args.get("state") == "staged":
+            return [detach_trace] if detach_trace else []
         return self._deploy_detach_traces(
             api_args=module_args,
             wait_args=module_args,
--- a/plugins/module_utils/orchestrators/network_state_machine.py
+++ b/plugins/module_utils/orchestrators/network_state_machine.py
@@ -12,7 +12,7 @@ This wrapper owns the Network-specific workflow around that CRUD phase:
   - post-attach for merged/replaced/overridden
   - deploy of pending attachment changes
 
-The private ``_staged`` state follows overridden attachment handling but maps
+The private ``staged`` state follows overridden attachment handling but maps
 the CRUD phase to replaced and suppresses deploy so omitted Networks are
 detached without deploy or delete calls.
 
@@ -88,12 +88,12 @@ class NetworkStateMachine:
         if state == "deleted":
             return self.run_deleted(module_args, active_strategy)
 
-        if state in ("replaced", "overridden", "_staged"):
+        if state in ("replaced", "overridden", "staged"):
             return self._run_state_machine_aware_with_attachments(module_args, active_strategy, defer_deploy)
 
         desired_attachments = None
         desired_network_names = None
-        if state in ("merged", "replaced", "overridden", "_staged"):
+        if state in ("merged", "replaced", "overridden", "staged"):
             desired_attachments = self.coordinator._desired_attachment_map(
                 module_args,
                 active_strategy,
@@ -157,11 +157,11 @@ class NetworkStateMachine:
         defer_deploy: bool = False,
     ) -> dict[str, Any]:
         """
-        Run replaced/overridden/_staged after deriving current Network names from the state machine.
+        Run replaced/overridden/staged after deriving current Network names from the state machine.
 
         The pre-detach phase must only query attachments for Networks that
         already exist. First-create ``state=replaced`` tasks can then create
-        the Network before the post-attach phase runs. The private ``_staged``
+        the Network before the post-attach phase runs. The private ``staged``
         state uses overridden detach scope but runs CRUD as replaced to
         suppress deletes.
         """
@@ -177,7 +177,7 @@ class NetworkStateMachine:
             current_network_name_set = set(current_network_names)
             desired_network_name_set = set(desired_network_names)
             current_desired_network_names = [network_name for network_name in desired_network_names if network_name in current_network_name_set]
-            attachment_query_network_names = current_network_names if state in ("overridden", "_staged") else current_desired_network_names
+            attachment_query_network_names = current_network_names if state in ("overridden", "staged") else current_desired_network_names
             current_attachment_details = (
                 self.coordinator._current_attachment_details_ignore_missing(
                     module_args,
@@ -193,7 +193,7 @@ class NetworkStateMachine:
             )
 
             pre_delete_traces: list[dict[str, Any]] = []
-            if state in ("overridden", "_staged"):
+            if state in ("overridden", "staged"):
                 omitted_network_names = [network_name for network_name in current_network_names if network_name not in desired_network_name_set]
                 pre_delete_traces = self._prepare_overridden_deletions(
                     module_args,
@@ -232,7 +232,7 @@ class NetworkStateMachine:
             post_attachment_details = list(current_attachment_details or [])
             post_current_attachments = dict(current_attachments or {})
             new_desired_network_names = [network_name for network_name in desired_network_names if network_name not in current_network_name_set]
-            if state in ("replaced", "overridden", "_staged") and desired_attachments and new_desired_network_names and not self._check_mode():
+            if state in ("replaced", "overridden", "staged") and desired_attachments and new_desired_network_names and not self._check_mode():
                 self._trace("attachment_phase_post_new_current_query_start", network_names=new_desired_network_names)
                 new_attachment_details = self.coordinator._current_attachment_details(
                     module_args,
@@ -261,7 +261,7 @@ class NetworkStateMachine:
             self._trace("attachment_phase_post_end", changed=post_attach.get("changed"), payload_count=len(post_attach.get("payloads", [])))
             self.coordinator._merge_api_trace(result, post_attach)
 
-            if state == "_staged":
+            if state == "staged":
                 return result
 
             return self._deploy_after_attachment_changes(
@@ -293,7 +293,7 @@ class NetworkStateMachine:
     @staticmethod
     def _crud_module_args(module_args: dict) -> dict:
         """Return module args for the generic CRUD state machine."""
-        if module_args.get("state") != "_staged":
+        if module_args.get("state") != "staged":
             return module_args
         crud_args = dict(module_args)
         crud_args["state"] = "replaced"
@@ -302,7 +302,7 @@ class NetworkStateMachine:
     @staticmethod
     def _query_module_args(module_args: dict) -> dict:
         """Return module args for the current-state query phase."""
-        if module_args.get("state") != "_staged":
+        if module_args.get("state") != "staged":
             return module_args
         query_args = dict(module_args)
         query_args["state"] = "overridden"
@@ -311,7 +311,7 @@ class NetworkStateMachine:
     @staticmethod
     def _prepare_crud_state(sm: Any, requested_state: str) -> None:
         """Switch private staged workflows to replacement CRUD after query."""
-        if requested_state != "_staged":
+        if requested_state != "staged":
             return
         sm.state = "replaced"
 
@@ -424,7 +424,7 @@ class NetworkStateMachine:
             omitted_network_names,
             attachment_details=omitted_attachment_details,
         )
-        if module_args.get("state") == "_staged":
+        if module_args.get("state") == "staged":
             return [detach_trace] if detach_trace else []
         return self._deploy_detach_traces(
             api_args=module_args,
--- a/plugins/module_utils/orchestrators/network_state_machine.py
+++ b/plugins/module_utils/orchestrators/network_state_machine.py
@@ -170,8 +170,9 @@ class NetworkStateMachine:
         desired_attachments = self.coordinator._desired_attachment_map(module_args, strategy)
         desired_network_names = self.coordinator._configured_network_names(config)
 
-        sm, original_config, original_state = self.coordinator._new_state_machine(self._crud_module_args(module_args), strategy)
+        sm, original_config, original_state = self.coordinator._new_state_machine(self._query_module_args(module_args), strategy)
         try:
+            self._prepare_crud_state(sm, state)
             current_network_names = self._network_names_from_models(sm.existing)
             current_network_name_set = set(current_network_names)
             desired_network_name_set = set(desired_network_names)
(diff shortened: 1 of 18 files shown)


plugins/modules/nd_manage_vrfs.py
```diff
@@ -41,8 +41,11 @@
       - V(deleted) removes specified VRFs (or all if config is empty).
       - V(gathered) returns current VRF state (the only state allowed on child
         fabrics when targeted directly).
+      - V(staged) is an internal/private workflow state. It follows
```
akinross: Why is staged called “internal/private” while also being included in the module’s public choices?

akinross: Should we add example for this state with description what it is actually doing?

akinross: I am a bit confused on what staged is supposed to be doing, could you clarify my overall understanding which might be wrong.

Should `state: staged` use full overridden attachment semantics while running definition CRUD as replaced and suppressing deployment, so that:

- Removing an attachment from a VRF that remains in config submits an undeployed `attach:false`; the current code appears to skip this attachment removal.
- Omitting an existing VRF entirely submits undeployed detach requests for its current attachments while retaining the VRF definition; the current code appears to handle this case as expected.
- Adding an attachment to an existing VRF submits an undeployed `attach:true`; the current code appears to skip this attachment addition.
- Creating a new VRF with attachments creates the VRF definition and submits undeployed `attach:true` requests; the current code appears to create the definition but skip its attachments.

It appears to me we that we do not test all of these scenarios but could have missed this.

akinross (changes requested):

plugins/modules/nd_manage_vrfs.py
AKDRG: Since "state" comes under the argspec, the choices for the state needs documentation else get's flagged by validate-module sanity.

AKDRG: Since it's an private state and not intended for customer use cases, I guess we need not document examples.

akinross: did you consider exposing this "feature" through an undocumented environment variable that can be specified in at task level? this way the documentation would remain clean

mtarking: @AKDRG just remove the usage of internal/private as last conversation on this was to fully document it as a state.

mtarking: @AKDRG customers can use it since we'll support it as a state. Original intention was undocumented state, but it's documented and in arg spec now.

AKDRG: My bad, was thinking we needed to document it as an internal state. Amended the wordings now.

AKDRG: Got it, fixed!

AKDRG: Hey Akini,
Thanks a lot for your feedback and testing. You had identified a critical bug here. I have fixed the gaps that were left out due to my commit miss and we are fully covered now. 

1. Removing an attachment from a VRF that remains in config submits an undeployed attach:false; the current code appears to skip this attachment removal. 
> Not skipped, attachment is removed with attach:false.  Not un-deployed -> Staged.

2. Adding an attachment to an existing VRF submits an undeployed attach:true; the current code appears to skip this attachment addition.
> Not skipped, attachment is posted with attach:true. Not deployed -> Staged.

3. Creating a new VRF with attachments creates the VRF definition and submits undeployed attach:true requests; the current code appears to create the definition but skip its attachments.
> Not skipped, attachment is posted with attach:true. Not deployed -> Staged.

Added the integration tests too!

Thanks,
Akshay

plugins/module_utils/orchestrators/vrf_state_machine.py
```diff
@@ -356,6 +390,8 @@ def _prepare_overridden_deletions(
             omitted_vrf_names,
             attachment_details=omitted_attachment_details,
         )
+        if module_args.get("state") == "staged":
```
allenrobel, line 384: This staged short-circuit comes after the unconditional `self.coordinator._ensure_vrfs_have_no_networks(...)` call at the top of `_prepare_overridden_deletions` (line 378). That guard exists to "Fail before VRF deletion when networks still reference the VRFs" (`vrf_dependency_checker.py::ensure_no_networks`) and calls `fail_json("Cannot delete VRF(s) because network(s) still reference them ...")`.

Since `staged` never deletes ("omitted VRFs are detached but not removed"), a staged task that omits a VRF still carrying networks hard-fails with a deletion error instead of performing the legitimate detach-only reconciliation. Suggest skipping the dependency check when `module_args.get("state") == "staged"`, i.e. hoisting this guard above the `_ensure_vrfs_have_no_networks` call.

Note the new unit test (`test_vrf_staged_detaches_omitted_vrfs_without_running_overridden_crud_delete`) stubs `_ensure_vrfs_have_no_networks` as a call logger, so this path is not exercised by the added tests. The network twin (`_ensure_networks_have_no_networks`) is currently a no-op placeholder, so only the VRF side fails today — but the same reordering there would keep the two symmetric if that checker is ever implemented.

plugins/module_utils/orchestrators/vrf_state_machine.py
```diff
@@ -276,6 +285,31 @@ def _current_after_pre_detach(
...
+        query_args = dict(module_args)
+        query_args["state"] = "overridden"
+        return query_args
+
+    @staticmethod
+    def _prepare_crud_state(sm: Any, requested_state: str) -> None:
+        """Switch staged workflows to replacement CRUD after query."""
+        if requested_state != "staged":
+            return
+        sm.state = "replaced"
```
allenrobel: These three helpers (`_crud_module_args`, `_query_module_args`, `_prepare_crud_state`) are byte-for-byte identical to the copies added in `network_state_machine.py` (lines 293-316), and the two classes share no base. This is the same two-independent-copies-may-drift situation raised on PR #505, which was resolved there by extracting the shared `attachment_vpc_peer_expander.py`. Consider the same treatment here — a small shared helper module or mixin — so a future change to the staged state mapping lands in one place instead of two.

plugins/module_utils/orchestrators/vrf_state_machine.py
```diff
@@ -149,24 +153,26 @@ def _run_state_machine_aware_with_attachments(
...
+        before the post-attach phase runs. The ``staged`` state uses overridden
+        detach scope but runs CRUD as replaced to suppress deletes.
         """
         state = module_args.get("state", "merged")
         config = module_args.get("config") or []
         desired_attachments = self.coordinator._desired_attachment_map(module_args, strategy)
         desired_vrf_names = self.coordinator._configured_vrf_names(config)
 
-        sm, original_config, original_state = self.coordinator._new_state_machine(module_args, strategy)
+        sm, original_config, original_state = self.coordinator._new_state_machine(self._query_module_args(module_args), strategy)
```
allenrobel: Because `_query_module_args` remaps `state` to `"overridden"` before `_new_state_machine` constructs the `NDStateMachine`, `results.state` (documented in `rest/results.py` as "The Ansible state for the current task") is captured as `"overridden"` for the whole staged run, and every API call's `metadata["state"]` in the `-vvv` `api_metadata` output reports `overridden` rather than `staged`. Metadata-only (requests, diff, and `changed` are unaffected), but it will mislead anyone debugging a staged task from verbose output. Worth restoring the user-requested state on `sm.results` after construction (e.g. in `_prepare_crud_state`). Same applies to the network twin.

plugins/module_utils/orchestrators/vrf_state_machine.py
```diff
@@ -149,24 +153,26 @@ def _run_state_machine_aware_with_attachments(
...
         try:
+            self._prepare_crud_state(sm, state)
```
allenrobel: Minor/latent: `NDStateMachine.__init__` threads `context={"state": self.state}` into pydantic validation "so models can apply state-aware validation", so `sm.proposed` here was validated under `"overridden"`; `_prepare_crud_state` then flips `sm.state` to `"replaced"` without re-validating. No current VRF/Network model reads the validation context, so this has no effect today — flagging it so a future state-aware validator doesn't silently validate staged runs under the wrong state. A comment noting the assumption would be enough.

allenrobel (commented):
### Code review

Four findings, all on the staged workflow plumbing in `vrf_state_machine.py` (the first is a functional bug; the network-side twin of each is noted inline where it applies).

akinross (approved):

plugins/module_utils/orchestrators/vrf_state_machine.py
AKDRG: The staged path now skips the VRF network-dependency deletion guard before processing omitted VRFs. Since `staged` does not delete omitted VRF definitions, it should not fail with a delete-safety error when those VRFs still have dependent networks.

The flow now performs only the staged reconciliation behavior:
- detach omitted VRF attachments with `deploy: false`
- retain the omitted VRF definition
- skip definition deletion

I made the same ordering adjustment on the Network state machine side for symmetry, even though the Network dependency checker is currently a no-op placeholder.

plugins/module_utils/orchestrators/vrf_state_machine.py
AKDRG: The duplicated staged helper logic was extracted into shared staged-state helpers and both VRF and Network state machines now use the same implementation.

The shared helper owns:
- mapping staged CRUD execution to replaced semantics
- mapping staged query discovery to overridden semantics
- preserving the user-requested staged state in result metadata

This avoids the two state machines drifting if staged mapping behavior changes later.

plugins/module_utils/orchestrators/vrf_state_machine.py
AKDRG: The staged state machine now restores the public result state to `staged` after constructing the CRUD state machine. The implementation can still use replacement semantics internally, but verbose/debug metadata now reflects the state the user actually ran.

That keeps `results.state` and API metadata from reporting `overridden` for a staged task.

plugins/module_utils/orchestrators/vrf_state_machine.py
AKDRG: The staged query and CRUD flows are now split more cleanly. Query discovery uses overridden-style current-state collection with empty config, while CRUD execution is constructed through the staged CRUD argument path instead of relying on a broad post-construction state mutation pattern.

This avoids validating desired staged config under the wrong state context and then changing the operational state afterward. No current VRF/Network model depends on state-aware validation here, but this makes the flow safer for future validators.

allenrobel (commented):
### Code review

Re-review of head 56ce4773 after the 2026-09-08 fixes. All four earlier findings are addressed (dependency guard skipped for staged with a real regression test, shared `staged_state_helpers.py`, result state restored, CRUD machine constructed under `replaced`). One new item inline, non-blocking. One body-level note:

1. The PR body links #465 / #466 as "Related to", so `closingIssuesReferences` is empty. Team convention is a closing keyword in every PR body (CLAUDE.md PR template: "Link issues with `Fixes #xxx`, `Closes #xxx`, or `Resolves #xxx`"). If this PR only partially covers them, `Closes` on the one it does close plus "Related to" for the other is fine.

plugins/module_utils/orchestrators/vrf_state_machine.py
```diff
@@ -149,24 +159,35 @@ def _run_state_machine_aware_with_attachments(
...
-        sm, original_config, original_state = self.coordinator._new_state_machine(module_args, strategy)
+        current_vrf_names = []
+        if state == "staged":
+            query_sm, query_original_config, query_original_state = self.coordinator._new_state_machine(query_module_args(module_args), strategy)
+            try:
+                current_vrf_names = self._vrf_names_from_models(query_sm.existing)
+            finally:
+                self.coordinator._restore_state_machine_params(query_original_config, query_original_state)
+
+        sm, original_config, original_state = self.coordinator._new_state_machine(crud_module_args(module_args), strategy)
```
allenrobel, line 182: For `staged` this now builds two `NDStateMachine`s back to back, and each constructor runs `query_all()` (`nd_state_machine.py:70`). With two or more VRFs in the task, `NDVrfOrchestrator.query_all` falls through to `_query_all_unfiltered()` for the `replaced`-scoped machine too (`vrfs.py:517-520`), so a staged run issues two full-inventory GETs; with one VRF it is one full plus one filtered GET. CLAUDE.md Performance: "Fetch each controller resource at most once per module run." `_new_state_machine`'s own docstring says the same ("should use this helper rather than issuing a second query").

This split came out of my earlier note on the validation context, so to be clear about what I meant there: `sm.existing` / `sm.before` are built by `from_api_response` with no state context, and no VRF or Network model reads the context today, so a single construction is safe. Either of these keeps it to one GET:

- (a) the shape before "Split staged query and CRUD state handling": one `_new_state_machine(...)` built with the unscoped state, `prepare_crud_state` flipping to `replaced`, plus a one-line comment stating the context assumption; or
- (b) keep the CRUD machine as is and use the coordinator's existing `_query_current_vrfs()` (a plain `query_all`, already used by `run_deleted` below) for discovery instead of a second state machine.

Same on the network side. The new tests assert both `new_state_machine` calls, so they would move with whichever you pick. Non-blocking.

plugins/module_utils/orchestrators/network_state_machine.py
```diff
@@ -153,24 +163,36 @@ def _run_state_machine_aware_with_attachments(
...
-        sm, original_config, original_state = self.coordinator._new_state_machine(module_args, strategy)
+        current_network_names = []
+        if state == "staged":
+            query_sm, query_original_config, query_original_state = self.coordinator._new_state_machine(query_module_args(module_args), strategy)
+            try:
+                current_network_names = self._network_names_from_models(query_sm.existing)
+            finally:
+                self.coordinator._restore_state_machine_params(query_original_config, query_original_state)
+
+        sm, original_config, original_state = self.coordinator._new_state_machine(crud_module_args(module_args), strategy)
```
allenrobel, line 187: Network twin of the note on `vrf_state_machine.py`: two `NDStateMachine` constructions, two `query_all()` GETs per staged run (`networks.py` scoping falls back to the unfiltered fetch for two or more networks as well). Same two options apply here.

plugins/module_utils/orchestrators/vrf_state_machine.py
allenrobel: Confirmed on 56ce4773 that the CRUD machine is built under `replaced` before validation, so this is addressed. The split does add a second full-inventory GET; I left a separate note on the new block with what I would suggest instead. Resolving this one.

plugins/module_utils/orchestrators/vrf_state_machine.py
AKDRG: Addressed. The fix removes the extra staged-only NDStateMachine construction for both VRFs and Networks. staged now builds one state machine using the original staged state, so the initial current-state query remains unscoped and discovers omitted resources in a single GET. After that discovery, the workflow flips only the internal CRUD execution state to replaced, preserving staged behavior without issuing a second inventory query. I also removed the unused query-state helper and updated the tests to assert there is only one state-machine construction, with no hidden overridden or replaced machine.

plugins/module_utils/orchestrators/network_state_machine.py
AKDRG: Fixed with the above

allenrobel (approved):
Confirmed on "Avoid duplicate staged state queries": the staged path now builds a single NDStateMachine under the user-requested state, the VRF and network query scoping both fall through to one unfiltered GET for `staged`, and `prepare_crud_state` flips only the internal CRUD state to replaced while keeping `results.state` as staged. The two workflow tests assert exactly one construction with no hidden overridden/replaced machine. That closes out my remaining threads, which I have resolved. Approving.

plugins/modules/nd_manage_networks.py
```diff
@@ -24,8 +24,10 @@
     description:
       - Desired state of Network resources.
       - V(query) is accepted as a compatibility alias for V(gathered).
+      - V(staged) creates or updates Networks and reconciles attachments without
```
mikewiebe: ### Medium: Document the full hybrid semantics of the public `staged` state

**Issue**

The new public documentation says that `staged` creates or updates definitions, reconciles attachments without deployment, and retains omitted definitions. It does not explain the state’s complete contract:

- listed definitions use `replaced` CRUD semantics, so unspecified definition properties can be reset;
- attachment reconciliation uses `overridden` scope across the full current VRF or Network set;
- item-level `deploy: true` is ignored;
- omitted definitions are retained after their attachments are detached; and
- an empty desired list stages detachment of every current attachment while retaining all definitions.

This behavior is intentional and internally consistent, but it is broader than a reader can safely infer from “creates or updates” and “reconciles attachments.” Existing PR discussion explicitly established `staged` as customer-supported, so the exact scope belongs in the public contract.

**Evidence**

- [`nd_manage_networks.py` lines 27-28](https://github.com/CiscoDevNet/ansible-nd/pull/509/files#diff-28c2ca22beff4f6a57eb86c269dffc3e68f3a06ef845597d91e11b9e8427f89eR27-R28) gives only the abbreviated create/update, attachment, no-deploy, and omitted-definition description. The same module defaults `config` to an empty list at lines 32-37.
- [`nd_manage_vrfs.py` lines 44-45](https://github.com/CiscoDevNet/ansible-nd/pull/509/files#diff-6d79f58f19fba55394110e94932a02118394c3ff0113c32209bd7960f793a957R44-R45) uses the same abbreviated wording.
- [`staged_state_helpers.py` lines 20-37](https://github.com/CiscoDevNet/ansible-nd/pull/509/files#diff-6422622b1ac27ce07dee0bd061e79103c2bfec26d99bead4de9c727dff0b3e26R20-R37) maps staged definition CRUD to `replaced`.
- [`vrf_state_machine.py` lines 161-180](https://github.com/CiscoDevNet/ansible-nd/pull/509/files#diff-877c7cec30271df39312f708718668ef3dcf85c29adf94856ef5c887a01943c1R161-R180) documents and implements overridden attachment scope for `staged`; the Network state machine mirrors it.
- [`network_state_machine.py` lines 269-270](https://github.com/CiscoDevNet/ansible-nd/pull/509/files#diff-60d7f5607b3c1ec58a2d086fccafee72fa4f6991b665f7ef324a20a63a1ad5f4R269-R270) returns before every deploy path. The VRF path does the same at lines 264-265.

**Practical example**

Suppose ND currently contains:

| Network | Definition | Attachments |
| --- | --- | --- |
| `Network_BLUE` | `mtu: 9000`, `arp_suppression: true` | `leaf-1`, `leaf-2` |
| `Network_GREEN` | Existing configuration | `leaf-3` |

The user wants only to stage removal of `Network_BLUE` from `leaf-2`:

```yaml
- name: Stage one attachment removal
  cisco.nd.nd_manage_networks:
    fabric_name: fab1
    state: staged
    config:
      - network_name: Network_BLUE
        network_id: 50010
        vlan_id: 2001
        is_l2only: true
        deploy: true
        attach:
          - ip_address: 192.0.2.11  # leaf-1
            interfaces:
              - interface_range: Ethernet1/1
                mode: trunk
```

They might expect the existing `Network_BLUE` definition to remain unchanged, only the `leaf-2` detachment to be staged, `Network_GREEN` to remain untouched, and `deploy: true` to deploy the change.

What actually happens:

1. `Network_BLUE` receives `replaced` definition semantics. Because `mtu` and `arp_suppression` were omitted, their proposed values become the model defaults, `9216` and `false`, instead of preserving `9000` and `true`.
2. `Network_BLUE` is staged for detachment from `leaf-2`.
3. Because attachment scope is fabric-wide and `Network_GREEN` is absent from `config`, its `leaf-3` attachment is also staged for detachment.
4. The `Network_GREEN` definition itself is retained.
5. `deploy: true` is ignored. These changes remain pending until a later central deployment.

Thus `staged` expects a complete desired set, not merely the Network being changed. To preserve the original intent, the task must include all desired Networks, all definition properties that must remain set, and every attachment that must remain attached.

**Existing PR overlap**

Related to [the original contract and missing-scenario review thread](https://github.com/CiscoDevNet/ansible-nd/pull/509#discussion_r3852209802), which described the hybrid semantics accurately. The author fixed the runtime gaps and added a short description and examples, but the current public docs still omit replacement resets, empty-list scope, and deploy precedence.

Related to [#453](https://github.com/CiscoDevNet/ansible-nd/issues/453), [#465](https://github.com/CiscoDevNet/ansible-nd/issues/465), and [#466](https://github.com/CiscoDevNet/ansible-nd/issues/466). Those issues define the intended behavior and call for staged-workflow documentation, but they do not make the abbreviated module documentation safe on their own.

**Impact**

A user may reasonably read `staged` as “merge without deploy” and unintentionally reset omitted definition fields or stage fabric-wide attachment removals. The highest-risk case is `config: []`, which stages detachment of every current attachment while retaining the definitions.

**Suggested fix**

Expand both state descriptions and examples to document the replacement semantics, fabric-wide attachment scope, ignored deploy flags, retained definitions, empty-configuration behavior, and supported follow-up deployment workflow.

AKDRG: Updated the public staged docs for both modules.

Added the missing contract details:
- listed definitions use replaced semantics
- attachments use fabric-wide overridden scope
- deploy: true is ignored
- omitted definitions are retained after detach
- config: [] detaches all current attachments while retaining definitions

Also expanded the staged examples to show a complete desired attachment set.

mikewiebe (approved):
LGTM

PR #513, opened 2026-08-17
nd_manage_vpc_pair: Add resource deploy scope and make config_actions.deploy opt-in

### LINKED ISSUE #428: vpc_pair Module Support for config_actions + resource level deploy

sivakasi-cisco:
## Summary

Adds a new `resource` deploy scope to `nd_manage_vpc_pair`'s `config_actions`,
and makes deploy opt-in (`deploy: false` by default) so the module never pushes
configuration to the fabric unless explicitly requested.

### New `config_actions.type: resource` scope
- Adds `resource` as a third choice for `config_actions.type` (alongside
  `switch` and `global`).
- `resource` deploys only the peer switches that make up the managed vPC pair(s)
  and are left out-of-sync — not every out-of-sync switch in the fabric
  (`switch`), nor the entire fabric (`global`).
- Nexus Dashboard has no vPC-pair-level deploy primitive, so a vPC pair resource
  is mapped to its two peer switches (`switch_id` / `peer_switch_id`), which are
  already resolved to serial numbers upstream.
- Peers not found in fabric inventory (mistyped serial, wrong fabric) raise a
  `module.warn` instead of being silently skipped, so a resource deploy never
  becomes a silent no-op. An already in-sync pair is a clean no-op.

### `deploy` now defaults to `false` (opt-in)
- `config_actions.deploy` defaults to `false` everywhere: the pydantic
  `ConfigActionsModel`, the Ansible `argument_spec`, `get_config_actions()`
  (explicit read + fallback), `custom_vpc_deploy()`, and module DOCUMENTATION.
- `save` still defaults to `true`. Set `deploy: true` explicitly to push
  configuration to the fabric.
- Prevents unintended fabric deploys from a bare `config_actions:` block.

## Files
- `plugins/module_utils/manage_vpc_pair/common.py` — `resource` added to
  choices; `deploy` fallback → `false`.
- `plugins/module_utils/manage_vpc_pair/deploy.py` — resource-scoped deploy
  (`_pair_identifiers`, `_get_managed_pair_switches_needing_deploy`);
  `switch`/`resource` share the per-switch deploy path.
- `plugins/module_utils/models/manage_vpc_pair/vpc_pair_model.py` — model +
  argument_spec: `deploy` default `false`, `type` Literal gains `resource`.
- `plugins/modules/nd_manage_vpc_pair.py` — DOCUMENTATION: `resource`
  choice/description, `deploy` default `false`.
- `tests/unit/module_utils/test_manage_vpc_pair_deploy.py`,
  `test_manage_vpc_pair_model.py` — unit coverage for resource scope + defaults.

## Behavior / compatibility
Changes the deploy default from true to false and adds the resource scope in `config_actions` so defaulting `deploy` to `false`
sets the initial contract rather than changing released behavior. Playbooks must
set `config_actions.deploy: true` to trigger a deploy.

## Testing
- Unit: `test_manage_vpc_pair_deploy.py` / `test_manage_vpc_pair_model.py`
  (resource-scope deploy, in-sync no-op, unresolved-peer warning, model &
  argument_spec defaults).
- Live integration: `nd_vpc_pair` target (merge/replace/override/delete/gather)
  against a lab fabric.

Closes #428


plugins/module_utils/manage_vpc_pair/deploy.py
```diff
@@ -288,17 +347,22 @@ def custom_vpc_deploy(nrm: Any, fabric_name: str, result: dict[str, Any]) -> dic
...
     #   - "global": deploy the entire fabric via .../actions/deploy.
-    #   - "switch": deploy only the switches left out-of-sync by the vPC pair
-    #     changes above, via .../switchActions/deploy.
+    #   - "switch": deploy every switch left out-of-sync by the vPC pair changes
+    #     above, via .../switchActions/deploy.
+    #   - "resource": deploy only the managed pair's out-of-sync peer switches,
+    #     via .../switchActions/deploy.
     if deploy_enabled:
-        if action_type == "switch":
+        if action_type in ("switch", "resource"):
```
akinross: why not define a constant for this similar as you did for DEFAULT_CONFIG_ACTION_TYPE and CONFIG_ACTION_TYPE_CHOICES? something like SWITCH_DEPLOY_PATH_ACTION

plugins/module_utils/manage_vpc_pair/deploy.py
```diff
@@ -136,6 +136,63 @@ def _get_switches_needing_deploy(nd_v2: Any, fabric_name: str) -> list[str]:
...
+    Deploy runs after custom_vpc_query_all -> normalize_vpc_playbook_switch_identifiers
+    has rewritten module.params["config"], so entries carry runtime keys
+    (switch_id/peer_switch_id) with any management IPs already resolved to serials.
+    """
+    peer1 = config_entry.get("switch_id")
+    peer2 = config_entry.get("peer_switch_id")
+    return [identifier for identifier in (peer1, peer2) if identifier]
+
+
+def _get_managed_pair_switches_needing_deploy(nd_v2: Any, fabric_name: str, config_entries: Any) -> list[str]:
```
akinross: Why is config_entries of type any? Is the expected contract not known by the time this function runs because the config has been validated/normalized?

plugins/module_utils/manage_vpc_pair/deploy.py
```diff
@@ -136,6 +136,63 @@ def _get_switches_needing_deploy(nd_v2: Any, fabric_name: str) -> list[str]:
...
+    """
+    switches = _validate_fabric_switches(nd_v2, fabric_name)
+    if not switches:
+        return []
+    managed_serials: set[str] = set()
+    for config_entry in config_entries or []:
+        if not isinstance(config_entry, dict):
+            continue
+        identifiers = _pair_identifiers(config_entry)
+        unresolved = [identifier for identifier in identifiers if identifier not in switches]
```
akinross: why not do a single loop to identify the unresolved and managed_serials? also i think if you use managed_serials we should also uses unresolved_serials to indicate these 2 are related.

plugins/module_utils/manage_vpc_pair/deploy.py
```diff
@@ -136,6 +136,63 @@ def _get_switches_needing_deploy(nd_v2: Any, fabric_name: str) -> list[str]:
...
+        if unresolved:
+            nd_v2.module.warn(
+                f"config_actions.type=resource: vPC pair peer(s) {unresolved} not found in fabric "
+                f"'{fabric_name}' inventory; skipping switch-scoped deploy for those switches."
+            )
+        managed_serials.update(identifier for identifier in identifiers if identifier in switches)
+    needing = [serial_number for serial_number in managed_serials if _is_switch_config_in_sync(switches.get(serial_number)) is not True]
```
akinross: why assign to variable needing when it can be sorted returned directly? if readability would `switches_needing_deploy` more clear?

every entry in managed_serials has already been confirmed present in switches, so why use the get(serial_number) over indexing? would you not want a KeyError if an invariant is broken in some unlikely case?

plugins/module_utils/manage_vpc_pair/deploy.py
```diff
@@ -136,6 +136,63 @@ def _get_switches_needing_deploy(nd_v2: Any, fabric_name: str) -> list[str]:
     return sorted(set(switch_ids))
 
 
+def _pair_identifiers(config_entry: dict[str, Any]) -> list[str]:
```
akinross: I am very confused by this function, because you appear to be silently filtering out None and empty values, is this intended? Does a VPC not always need to have exactly these 2 IDs and should the module not error when it cannot identify these?

if we are not unit testing this function what is the purpose of this function when it is only called a single time?

akinross (changes requested):

plugins/module_utils/manage_vpc_pair/deploy.py
```diff
@@ -288,17 +347,22 @@ def custom_vpc_deploy(nrm: Any, fabric_name: str, result: dict[str, Any]) -> dic
...
             deploy_path = fabric_utils.switch_deploy_path
         else:
             deploy_path = fabric_utils.config_deploy_path(force_show_run=True)
 
         try:
-            if action_type == "switch":
-                switch_ids = _get_switches_needing_deploy(nd_v2, fabric_name)
+            if action_type in ("switch", "resource"):
+                if action_type == "resource":
+                    switch_ids = _get_managed_pair_switches_needing_deploy(nd_v2, fabric_name, nrm.module.params.get("config"))
```
mikewiebe: ### High: Resource deploy ignores vPC pairs removed by overridden

**Issue**

Resource-action eligibility considers the complete state-machine result, but the deployment target is built only from the desired `module.params["config"]`. A vPC pair deleted because it was omitted from an overridden declaration is absent from that desired list, so the module saves the deletion and then deploys the wrong pair or no pair at all.

**Evidence**

- [`deploy.py` line 363](https://github.com/CiscoDevNet/ansible-nd/pull/513/files#diff-19a2454e1cceba649584ec33548e67bdb054942db409b29af53ecf59a72ff8beR363) passes only `module.params.get("config")` to the resource target resolver.
- [`resources.py` lines 442-472](https://github.com/CiscoDevNet/ansible-nd/blob/cbc3fa738a85f57c369502146cb0b250961c9bbf/plugins/module_utils/manage_vpc_pair/resources.py#L442-L472) deletes pairs omitted from the desired overridden collection, while [`_build_class_diff()`](https://github.com/CiscoDevNet/ansible-nd/blob/cbc3fa738a85f57c369502146cb0b250961c9bbf/plugins/module_utils/manage_vpc_pair/resources.py#L226-L270) retains their identifiers in `class_diff.deleted`.
- Practical example: assume the fabric has deployed pairs `KEEP1/KEEP2` and `DROP1/DROP2`, and the user authoritatively keeps only the first pair:

  ```yaml
  - name: Keep one vPC pair and remove every other pair
    cisco.nd.nd_manage_vpc_pair:
      fabric_name: fabric1
      state: overridden
      config_actions:
        save: true
        deploy: true
        type: resource
      config:
        - peer1_switch_id: KEEP1
          peer2_switch_id: KEEP2
  ```

  `overridden` correctly records:

  ```text
  class_diff.deleted=[("DROP1", "DROP2")]
  ```

  Expected deployment:

  ```json
  {"switchIds": ["DROP1", "DROP2"]}
  ```

  Actual deployment:

  ```json
  {"switchIds": ["KEEP1", "KEEP2"]}
  ```

  If the retained pair is already in sync, no deployment request is sent. The removal can remain pending while `DROP1/DROP2` continue running the old vPC configuration.

- With `state: overridden` and `config: []`, the deletion is recorded but no `switchActions/deploy` request occurs.
- An unrelated out-of-sync pair can also trigger a fabric-level config-save even when the managed pair is unchanged and in sync.

**Existing PR overlap**

No matching existing PR comment found. The five active review threads do not identify the missing overridden-deletion targets.

Related open issues are [#428](https://github.com/CiscoDevNet/ansible-nd/issues/428) and [#467](https://github.com/CiscoDevNet/ansible-nd/issues/467). Neither records this specific target-selection defect or its override-all failure mode.

**Impact**

The module can report success while a removed vPC pair remains pending. This breaks resource-scoped authoritative reconciliation, including purge-all workflows.

**Suggested fix**

Pass the run's `class_diff` into `_get_managed_pair_switches_needing_deploy()` and build one affected-pair set before checking switch sync state:

```python
affected_pairs = (
    pairs_from(module.params.get("config"))
    | pairs_from(class_diff.get("created", []))
    | pairs_from_updates(class_diff.get("updated", []))
    | pairs_from(class_diff.get("deleted", []))
)
```

For the example above, this produces both `KEEP1/KEEP2` and `DROP1/DROP2`. Resolve the switch IDs for only those affected pairs, then filter them by their current sync state. If `KEEP1/KEEP2` is in sync and `DROP1/DROP2` is out of sync after config-save, the deploy request must be `{"switchIds": ["DROP1", "DROP2"]}`.

Use this same run-scoped set for resource-action eligibility, runtime targeting, and check-mode preview. That both preserves delete-only targets and prevents an unrelated fabric-wide out-of-sync pair from triggering resource actions. Add regressions for overridden keep-one/delete-one, overridden `config: []`, explicit deletion, and an unchanged in-sync managed pair beside an unrelated out-of-sync pair.

tests/integration/targets/nd_vpc_pair/tasks/nd_vpc_pair_merge.yaml
```diff
@@ -1589,6 +1589,48 @@
   tags: merge
 
 # TC18 - config_actions save-only path (save=true, deploy=false)
+# Settle the controller after TC17's deploy so the save-only configSave is not dropped by a busy controller (RemoteDisconnected).
```
mikewiebe, line 1594: ### Medium: Resource scope lacks integration test regression coverage

**Issue**

This is a test-coverage finding, not an independent runtime defect. The PR adds an API-dependent deploy scope but does not call `type: resource` anywhere in the integration target tests. Its integration-file change only stabilizes an existing save-only test, so CI does not exercise the new feature's controller request sequence, target selection, deletion flows, or idempotency.

**Evidence**

- [`nd_vpc_pair_merge.yaml` lines 1591-1632](https://github.com/CiscoDevNet/ansible-nd/pull/513/files#diff-a2041261b6916a3ca38168d34a2a2a23c787c3693a9a18c40c06f84d512c5b5bR1592) adds a delay and convergence wait before existing TC18; it does not exercise the new action type.
- Repository search found no `type: resource` invocation under `tests/integration/targets/nd_vpc_pair` and no resource-scope task in the module `EXAMPLES` block.
- The added unit tests cover desired-pair filtering, in-sync no-op, both peers absent, and model defaults, but not overridden omissions, override-all, partial inventory, the exact config-save body, or unrelated out-of-sync pairs.
- The two exact-head probes above demonstrate that the omitted state/contract cases contain defects despite all 80 focused unit tests passing.

**Existing PR overlap**

Related to `@akinross`'s active [`_pair_identifiers()` thread](https://github.com/CiscoDevNet/ansible-nd/pull/513#discussion_r3851467857), which asks why that helper is not unit-tested. This finding is broader: it concerns the complete controller workflow and state matrix.

Related open issues are [#428](https://github.com/CiscoDevNet/ansible-nd/issues/428) for the resource feature and [#467](https://github.com/CiscoDevNet/ansible-nd/issues/467) for staged vPC additions/removals. Neither is a substitute for PR-level regression coverage.

**Impact**

The risk is that CI cannot detect the concrete target-selection, config-save, and incomplete-inventory problems described in the preceding findings. Those defects can remain or later regress while the focused unit suite stays green. Users also lack a copyable example showing the intended opt-in resource deployment contract.

**Suggested fix**

Treat this as test acceptance criteria for the substantive fixes above rather than as a separate product bug. Add a resource-scope example and an integration lifecycle that proves merged, replaced, explicit deleted, overridden omission, overridden empty, check mode, and second-run idempotency. Assert both peer serials and no unrelated switch in the deploy request/result, and cover controller error handling for unresolved peers.

plugins/module_utils/manage_vpc_pair/deploy.py
sivakasi-cisco: Added SWITCH_DEPLOY_ACTION_TYPES and imported it

plugins/module_utils/manage_vpc_pair/deploy.py
sivakasi-cisco: This is modified now - Post normalisation:  config_entries: list[dict[str, Any]]

plugins/module_utils/manage_vpc_pair/deploy.py
sivakasi-cisco: Agreed. Collapsed it into a single pass. Now, each of the peer goes into either managed_serials or unresolved_serials in the same loop. Also renamed 'unresolved' to 'unresolved_serials'

plugins/module_utils/manage_vpc_pair/deploy.py
sivakasi-cisco: Both fixed. Dropped the interim variable and returning the sorted generator directly.

plugins/module_utils/manage_vpc_pair/deploy.py
sivakasi-cisco: The silent filtering of None or empty valiues are wrong. 

Removed the "_pair_identifiers" function fully and inlined direct indexing "config_entry".

A vPC pair always has exactly these two IDs — the model guarantees them as distinct, non-empty serials so a missing key now raises instead of being quietly dropped. UT is now not required.

allenrobel (commented):
### Code review

Four inline comments (one behavioral edge case in the new resource-scope helper, three DOCUMENTATION follow-ups), plus one note on the PR description:

1. **PR body — "`config_actions` is introduced in this PR"** is not accurate: `config_actions` (with `deploy` defaulting to `true`) has been on `develop` since #197 and is documented in the module today. The compatibility conclusion still holds because it has not shipped in a published release, but the framing should be "changes the `deploy` default from `true` to `false` and adds the `resource` scope", not "introduces `config_actions`" — that is what a changelog reader needs to know.

plugins/module_utils/manage_vpc_pair/deploy.py
```diff
@@ -136,6 +137,73 @@ def _get_switches_needing_deploy(nd_v2: Any, fabric_name: str) -> list[str]:
...
+        config_entries: Managed vPC pair config entries (module.params["config"])
+        class_diff: Run diff with created/updated/deleted identifiers; deleted
+            pairs recover peers removed this run that no longer appear in config
+
+    Returns:
+        Sorted list of unique switch serial numbers needing deployment
+    """
+    switches = _validate_fabric_switches(nd_v2, fabric_name)
+    if not switches:
+        return []
```
allenrobel: This early return bypasses the per-peer warning loop below, so the one case the docstring promises never happens ("a resource deploy never becomes a silent no-op") can still happen: `_validate_fabric_switches` returns `{}` without warning on an empty/falsy response (`validation.py` `if not switches_response: return {}`), this returns `[]`, and `custom_vpc_deploy` records `success=True, changed=False, "No switches required resource-scoped deployment"` with no warning at all.

Dropping the `if not switches: return []` guard fixes it — with an empty `switches` dict the loop warns for every configured peer, which is the documented contract. Worth a unit test with an empty inventory fixture alongside 00120.

plugins/modules/nd_manage_vpc_pair.py
```diff
@@ -51,15 +51,17 @@
...
                 - Deploy configuration after save.
                 type: bool
-                default: true
+                default: false
             type:
                 description:
                 - Deploy scope for configuration actions.
-                - C(switch) deploys only the switches left out-of-sync by the vPC pair changes using the per-switch deploy action.
+                - C(resource) deploys only the switches that make up the managed vPC pair(s) and are left out-of-sync, using the per-switch deploy action.
+                - Nexus Dashboard has no vPC-pair-level deploy primitive, so a vPC pair resource maps to its two peer switches.
```
allenrobel: This sentence describes an ND API limitation rather than module behavior, and the team direction (2026-07-27) is to keep ND API quirk/limitation statements out of module DOCUMENTATION — they belong in the PR body (where this already is) and code comments. Line 58 already tells the user everything they need (`resource` deploys the managed pair's out-of-sync peer switches), so this line can simply be removed.

plugins/modules/nd_manage_vpc_pair.py
```diff
@@ -51,15 +51,17 @@
                 description:
                 - Deploy configuration after save.
                 type: bool
-                default: true
+                default: false
             type:
                 description:
                 - Deploy scope for configuration actions.
-                - C(switch) deploys only the switches left out-of-sync by the vPC pair changes using the per-switch deploy action.
+                - C(resource) deploys only the switches that make up the managed vPC pair(s) and are left out-of-sync, using the per-switch deploy action.
```
allenrobel: Follow-up to the overridden-deletion fix in "Fix resource-scope deploy skipping removed vPC pair peers": after that change `resource` also deploys the peers of pairs *removed* by this run (`state: overridden` omissions, `config: []`, `state: deleted`) via `class_diff["deleted"]`, but the option text still reads as if only the pairs listed in `config` are deployed. Suggest adding a sentence such as "Peers of vPC pairs removed by this task are also deployed so the removal takes effect." so users are not surprised by a deploy on switches they omitted.

plugins/modules/nd_manage_vpc_pair.py
```diff
@@ -51,15 +51,17 @@
                 description:
                 - Deploy configuration after save.
                 type: bool
-                default: true
+                default: false
```
allenrobel, line 55: When #520 flipped this same default for the `nd_interface_*` family it added the line `- Deployment is opt-in. Set O(config_actions.deploy=true) explicitly to push changes to switches.` to each module's `deploy` description. Suggest doing the same here so the description, not just the `default:` field, tells users deploy is now off unless requested.

plugins/module_utils/manage_vpc_pair/deploy.py
sivakasi-cisco: Right. Fixed it in _get_managed_pair_switches_needing_deploy. This now takes the run's class_diff and the affected set is the union of `module.params["config"]` (create/update/unchanged) and `class_diff["deleted"]` (pairs removed this run — overridden omissions, `config: []`, explicit delete). 

Peers are resolved for that set and then filtered by current sync state, so for the KEEP/DROP example it deploys `{"switchIds": ["DROP1", "DROP2"]}` and the retained in-sync pair is a no-op instead of being the deploy target. The same run-scoped set feeds runtime targeting and the check-mode preview.

Regressions: unit 00160/00170/00180 (overridden keep-one/delete-one, `config: []`, explicit delete) plus integration TC5 (explicit deleted) and TC6 (overridden empty purge), which assert `switchIds` is exactly the removed pair's peers and never an unrelated switch.

tests/integration/targets/nd_vpc_pair/tasks/nd_vpc_pair_merge.yaml
sivakasi-cisco: Added a dedicated - nd_vpc_pair_resource_deploy.yaml target (auto-discovered by main.yaml) that exercises `type: resource` end to end: check-mode preview, merged create+deploy, second-run idempotency, replaced, explicit deleted, and overridden `config: []` purge. 

Every deploy assertion checks the `/switchActions/deploy` request's `DATA.switchIds` equals exactly the managed pair's two peers and contains no unrelated switch. Also added the `type: resource` example block to the module.

Overridden keep-one/delete-one needs a second pair that our shared lab (2 switches = 1 pair) can't provide, and the unresolved-peer warning isn't reliably reproducible against a live fabric, so those two stay as unit tests (00170, and 00120/00190). I noted that in the file header.

plugins/module_utils/manage_vpc_pair/deploy.py
sivakasi-cisco: Done. Dropped the - "if not switches: return []" guard so that an empty inventory falls through and the loop warns for every configured peer. There will be no more silent no-op. 

Added unit test 00190 with an empty-inventory fixture alongside 00120; it asserts both peers are named in the warning and that no switchActions/deploy is posted.

plugins/modules/nd_manage_vpc_pair.py
sivakasi-cisco: Removed. Line 58 already covers the behaviour - "resource deploys the managed pair's out-of-sync peers", so I dropped the ND-API-limitation sentence from the documentatiuon

plugins/modules/nd_manage_vpc_pair.py
sivakasi-cisco: Added the deploy suboption. It now includes "Peers of vPC pairs removed by this task (via O(state=overridden) omissions, an empty O(config), or O(state=deleted)) are also deployed so the removal takes effect"

plugins/modules/nd_manage_vpc_pair.py
sivakasi-cisco: Added the same line  "-Deployment is opt-in. Set O(config_actions.deploy=true) explicitly to push changes to switches." to the deploy description.

allenrobel (approved):

mikewiebe:
### Design: Collapse `resource` deployment into `switch`

**Issue**

For this module, `type: resource` does not invoke a resource-level Nexus Dashboard API. It converts each managed vPC pair into its peer switch IDs and calls the same `/switchActions/deploy` endpoint used by `type: switch`.

Maintaining both values creates an unnecessary distinction:

- `resource` deploys the switches associated with the vPC pairs managed by this task.
- `switch` deploys every out-of-sync switch in the fabric.
- Both ultimately use switch-level deployment.

The current `switch` behavior also has a broader blast radius than users are likely to expect from a vPC-pair module. `global` already exists when the caller intentionally wants to deploy all pending fabric configuration.

**Practical example**

Given two vPC pairs:

```text
PAIR-A: LEAF1/LEAF2 — managed by this task
PAIR-B: LEAF3/LEAF4 — unrelated but out of sync
```

This input:

```yaml
config_actions:
  save: true
  deploy: true
  type: switch
config:
  - peer1_switch_id: LEAF1
    peer2_switch_id: LEAF2
```

should deploy only `LEAF1/LEAF2`. It should not deploy `LEAF3/LEAF4` merely because they are also out of sync.

For `state: overridden` or `state: deleted`, the affected switches must also include peers recovered from `class_diff.deleted`, so removals are deployed.

**Suggested change**

Remove `resource` as a supported `config_actions.type` and redefine the remaining scopes as:

| Type | Behavior |
| --- | --- |
| `switch` | Deploy only the affected peer switches for vPC pairs created, updated, or removed by this task. |
| `global` | Deploy all pending configuration for the fabric. |

This should include:

- Removing `resource` from the model, argspec, documentation, examples, constants, changelog, and tests.
- Moving the current resource-targeting logic into the `switch` path.
- Using the same run-scoped affected-pair set for action eligibility, check-mode preview, and runtime targeting.
- Retaining `class_diff.deleted` handling for explicit and overridden removals.
- Updating tests to verify that unrelated out-of-sync switches are excluded from `type: switch`.

Because this interface has not yet shipped in a collection release, this is the best time to simplify the contract without requiring a later compatibility or deprecation path.

**Existing overlap**

Related to issues #368 and #428, but this proposed two-scope contract is not currently captured by an existing PR comment.

mikewiebe:
### Medium: Deploy-only actions report `changed: false`

**Issue**

When an unchanged vPC pair has previously staged configuration, a later deployment can successfully save and deploy both switches while the module reports `changed: false`.

This issue remains even if `resource` is removed and its behavior becomes `type: switch`.

**Practical example**

First, stage the pair without deployment:

```yaml
- name: Stage the vPC pair
  cisco.nd.nd_manage_vpc_pair:
    fabric_name: fabric1
    state: merged
    config_actions:
      save: true
      deploy: false
    config:
      - peer1_switch_id: LEAF1
        peer2_switch_id: LEAF2
```

Later, deploy the unchanged pending configuration:

```yaml
- name: Deploy the staged vPC pair
  cisco.nd.nd_manage_vpc_pair:
    fabric_name: fabric1
    state: merged
    config_actions:
      save: true
      deploy: true
      type: resource
    config:
      - peer1_switch_id: LEAF1
        peer2_switch_id: LEAF2
```

The second task calls config-save and deploys `LEAF1/LEAF2`, but returns:

```yaml
changed: false
deployment:
  changed: false
```

After the proposed scope simplification, the same problem would occur with `type: switch`.

**Evidence**

- `register_action_api_call()` supplies `changed=True` but leaves the Results operation type at its default `QUERY`.
- `Results._determine_if_changed()` treats `QUERY` as read-only before considering the explicit changed flag.
- The action records also incorrectly retain the default `GET`, empty path, and null payload metadata.
- `VpcPairResourceService.execute()` does not merge `deployment.changed` into the top-level result.
- An exact-head regression reproduced the successful deployment followed by `deployment.changed == false`.

**Existing PR overlap**

No matching PR comment found.

Related to the later-explicit-deployment requirement in issue #467, but that issue does not track the incorrect changed result.

**Impact**

Automation cannot reliably determine whether staged configuration was deployed. Handlers, reporting, and subsequent tasks may incorrectly treat a real switch mutation as an idempotent no-op.

**Suggested fix**

Register action calls with their actual operation type, path, HTTP verb, and payload before calling `register_api_call()`. Then propagate `deployment.changed` into the module’s top-level `changed` value.

Add a regression that:

1. Stages a vPC pair with deployment disabled.
2. Reruns the unchanged pair with deployment enabled.
3. Confirms the switch deployment request occurred.
4. Requires both `deployment.changed` and top-level `changed` to be `true`.

mikewiebe:
### Medium: Config-save receives an undocumented deploy-scope body

**Issue**

The module sends its deploy-scope selection to the fabric config-save endpoint:

```json
{"type": "resource"}
```

Nexus Dashboard’s config-save operation is bodyless. Deployment scope should only control which deployment path the module selects after saving.

If the proposed design change removes `resource`, the problem still exists because the same code would send:

```json
{"type": "switch"}
```

That body is also undocumented.

**Practical example**

With the current PR:

```yaml
- name: Save and deploy the affected vPC pair
  cisco.nd.nd_manage_vpc_pair:
    fabric_name: fabric1
    state: merged
    config_actions:
      save: true
      deploy: true
      type: resource
    config:
      - peer1_switch_id: LEAF1
        peer2_switch_id: LEAF2
```

The module currently sends:

```http
POST /api/v1/manage/fabrics/fabric1/actions/configSave
Content-Type: application/json

{"type": "resource"}
```

After the proposed scope simplification, it would send:

```http
POST /api/v1/manage/fabrics/fabric1/actions/configSave
Content-Type: application/json

{"type": "switch"}
```

The documented request should be bodyless in either case:

```http
POST /api/v1/manage/fabrics/fabric1/actions/configSave
```

The scope is then applied to the separate deployment request:

```http
POST /api/v1/manage/fabrics/fabric1/switchActions/deploy
Content-Type: application/json

{"switchIds": ["LEAF1", "LEAF2"]}
```

**Evidence**

- `deploy.py` creates `action_payload = {"type": action_type}`.
- It passes that payload to `FabricUtils.save_config()`.
- An exact-head probe captured `{"type":"resource"}` on the config-save request.
- OpenAPI operation `executeConfigSave` defines no request body.
- The shared `ConfigActionsMixin.config_save()` implementation deliberately sends a bodyless request.

**Existing PR overlap**

No matching PR comment found.

Issues #428 and #467 are related to resource deployment and staged configuration, but neither defines or approves a config-save request body.

**Regression status**

The custom vPC action code already sends `switch` and `global` bodies. PR #513 newly adds the unsupported `resource` value, but the appropriate fix should make config-save bodyless for every deploy type.

**Impact**

A strict controller can reject config-save before deployment begins. A permissive controller may ignore the payload, but success then depends on undocumented behavior.

**Suggested fix**

Make `FabricUtils.save_config()` accept no payload—or an optional payload defaulting to `None`—and invoke config-save without a request body. Keep `type` only as local dispatch information for selecting switch-scoped or global deployment.

Add unit coverage asserting:

```python
assert config_save_request.payload is None
```

for both `type: switch` and `type: global`.

mikewiebe:
### Medium: Incomplete inventory can cause an asymmetric vPC deployment

**Issue**

The deployment target resolver evaluates each vPC peer independently. If Nexus Dashboard returns only one member of a configured pair, the module warns about the missing peer but still deploys the member it found.

For pairs removed through `deleted` or `overridden`, missing peers are silently discarded without even generating the warning.

This remains applicable if `resource` is removed and its behavior moves to `type: switch`.

**Practical example**

Assume the task manages this pair:

```yaml
config_actions:
  save: true
  deploy: true
  type: switch
config:
  - peer1_switch_id: LEAF1
    peer2_switch_id: LEAF2
```

A transient, malformed, or incomplete inventory response contains only:

```json
{
  "switches": [
    {
      "serialNumber": "LEAF1",
      "configSyncStatus": "Out-of-Sync"
    }
  ]
}
```

The module currently:

1. Adds `LEAF1` to the deployment targets.
2. Warns that `LEAF2` was not found.
3. Sends:

```json
{
  "switchIds": ["LEAF1"]
}
```

4. Reports the deployment as successful.

For a removed pair recovered from:

```python
class_diff["deleted"] = [("LEAF1", "LEAF2")]
```

an empty inventory response produces no warning and no deployment. A partial response again deploys only the peer it contains.

**Evidence**

- `_get_managed_pair_switches_needing_deploy()` adds each configured peer independently.
- The warning does not prevent deployment of the resolved peer.
- Deleted peers are silently intersected with the returned inventory.
- The fabric-switch inventory endpoint supports pagination, but this path makes one request and does not verify `meta.counts.remaining`.
- Current unit tests cover both configured peers being absent, but not exactly one configured peer being absent or deleted-pair inventory being empty or partial.
- Exact-head probes reproduced both one-peer deployment and the silent deleted-pair no-op.

**Existing PR overlap**

Related to the resolved `_pair_identifiers()` thread from `@akinross` and the resolved empty-inventory thread from `@allenrobel`.

Those changes fixed required-key access and warnings when both configured peers are absent, but they did not enforce atomic pair resolution.

Related to #428 and #467, but neither tracks incomplete inventory or asymmetric deployment.

**Regression status**

The inventory helper is pre-existing, but using an unverified inventory snapshot to construct resource-scoped deployment targets is introduced by this PR.

**Impact**

The module can report success after deploying only one vPC member or after silently failing to deploy a pair removal. That can leave the two peers in inconsistent states.

**Suggested fix**

Treat configured vPC peers as an atomic unit:

- Load a complete, paginated inventory snapshot.
- Require both configured peers to resolve before adding either one to the deployment payload.
- Fail with a clear error when only one peer resolves.
- For removed pairs, distinguish a switch confirmed to have been removed from an incomplete inventory response.
- Do not silently interpret an absent inventory entry as proof that deployment is unnecessary.

Add tests for:

- Both configured peers present.
- Both configured peers absent.
- Exactly one configured peer absent.
- Empty inventory for `class_diff.deleted`.
- Exactly one deleted peer present.
- Paginated inventory with a peer on a later page.

sivakasi-cisco (commented):
@akinross the changes requested are addressed.

sivakasi-cisco:
> ### Design: Collapse `resource` deployment into `switch`
> **Issue**
> 
> For this module, `type: resource` does not invoke a resource-level Nexus Dashboard API. It converts each managed vPC pair into its peer switch IDs and calls the same `/switchActions/deploy` endpoint used by `type: switch`.
> 
> Maintaining both values creates an unnecessary distinction:
> 
> * `resource` deploys the switches associated with the vPC pairs managed by this task.
> * `switch` deploys every out-of-sync switch in the fabric.
> * Both ultimately use switch-level deployment.
> 
> The current `switch` behavior also has a broader blast radius than users are likely to expect from a vPC-pair module. `global` already exists when the caller intentionally wants to deploy all pending fabric configuration.
> 
> **Practical example**
> 
> Given two vPC pairs:
> 
> ```
> PAIR-A: LEAF1/LEAF2 — managed by this task
> PAIR-B: LEAF3/LEAF4 — unrelated but out of sync
> ```
> 
> This input:
> 
> ```yaml
> config_actions:
>   save: true
>   deploy: true
>   type: switch
> config:
>   - peer1_switch_id: LEAF1
>     peer2_switch_id: LEAF2
> ```
> 
> should deploy only `LEAF1/LEAF2`. It should not deploy `LEAF3/LEAF4` merely because they are also out of sync.
> 
> For `state: overridden` or `state: deleted`, the affected switches must also include peers recovered from `class_diff.deleted`, so removals are deployed.
> 
> **Suggested change**
> 
> Remove `resource` as a supported `config_actions.type` and redefine the remaining scopes as:
> 
> Type	Behavior
> `switch`	Deploy only the affected peer switches for vPC pairs created, updated, or removed by this task.
> `global`	Deploy all pending configuration for the fabric.
> This should include:
> 
> * Removing `resource` from the model, argspec, documentation, examples, constants, changelog, and tests.
> * Moving the current resource-targeting logic into the `switch` path.
> * Using the same run-scoped affected-pair set for action eligibility, check-mode preview, and runtime targeting.
> * Retaining `class_diff.deleted` handling for explicit and overridden removals.
> * Updating tests to verify that unrelated out-of-sync switches are excluded from `type: switch`.
> 
> Because this interface has not yet shipped in a collection release, this is the best time to simplify the contract without requiring a later compatibility or deprecation path.
> 
> **Existing overlap**
> 
> Related to issues #368 and #428, but this proposed two-scope contract is not currently captured by an existing PR comment.

Agreed and done in this PR.

Removed resource as a config_actions.type completely — model, argspec (config_actions_spec + the vpc_pair model), documenatation, examples and tests now expose only switch and global.

The resource-targeting logic moved into the switch path: _get_managed_pair_switches_needing_deploy() scopes switchActions/deploy to only the peers of the pairs this run created/updated/removed — the union of module.params["config"] and class_diff["deleted"] — so an unrelated out-of-sync pair (LEAF3/LEAF4 in your example) is never pulled in. The same run-scoped set drives action eligibility, the check-mode preview and runtime targeting, and class_diff.deleted is retained for overridden/deleted removals.

Unit coverage added

sivakasi-cisco:
> ### Medium: Deploy-only actions report `changed: false`
> **Issue**
> 
> When an unchanged vPC pair has previously staged configuration, a later deployment can successfully save and deploy both switches while the module reports `changed: false`.
> 
> This issue remains even if `resource` is removed and its behavior becomes `type: switch`.
> 
> **Practical example**
> 
> First, stage the pair without deployment:
> 
> ```yaml
> - name: Stage the vPC pair
>   cisco.nd.nd_manage_vpc_pair:
>     fabric_name: fabric1
>     state: merged
>     config_actions:
>       save: true
>       deploy: false
>     config:
>       - peer1_switch_id: LEAF1
>         peer2_switch_id: LEAF2
> ```
> 
> Later, deploy the unchanged pending configuration:
> 
> ```yaml
> - name: Deploy the staged vPC pair
>   cisco.nd.nd_manage_vpc_pair:
>     fabric_name: fabric1
>     state: merged
>     config_actions:
>       save: true
>       deploy: true
>       type: resource
>     config:
>       - peer1_switch_id: LEAF1
>         peer2_switch_id: LEAF2
> ```
> 
> The second task calls config-save and deploys `LEAF1/LEAF2`, but returns:
> 
> ```yaml
> changed: false
> deployment:
>   changed: false
> ```
> 
> After the proposed scope simplification, the same problem would occur with `type: switch`.
> 
> **Evidence**
> 
> * `register_action_api_call()` supplies `changed=True` but leaves the Results operation type at its default `QUERY`.
> * `Results._determine_if_changed()` treats `QUERY` as read-only before considering the explicit changed flag.
> * The action records also incorrectly retain the default `GET`, empty path, and null payload metadata.
> * `VpcPairResourceService.execute()` does not merge `deployment.changed` into the top-level result.
> * An exact-head regression reproduced the successful deployment followed by `deployment.changed == false`.
> 
> **Existing PR overlap**
> 
> No matching PR comment found.
> 
> **Existing open issue overlap**
> 
> Related to the later-explicit-deployment requirement in issue #467, but that issue does not track the incorrect changed result.
> 
> **Impact**
> 
> Automation cannot reliably determine whether staged configuration was deployed. Handlers, reporting, and subsequent tasks may incorrectly treat a real switch mutation as an idempotent no-op.
> 
> **Suggested fix**
> 
> Register action calls with their actual operation type, path, HTTP verb, and payload before calling `register_api_call()`. Then propagate `deployment.changed` into the module’s top-level `changed` value.
> 
> Add a regression that:
> 
> 1. Stages a vPC pair with deployment disabled.
> 2. Reruns the unchanged pair with deployment enabled.
> 3. Confirms the switch deployment request occurred.
> 4. Requires both `deployment.changed` and top-level `changed` to be `true`.

Fixed in 2 parts.

1) The register_action_api_call() now sets results.operation_type = OperationType.UPDATE before register_api_call(), so save/deploy are no longer treated as read-only QUERY and the explicit changed=True is honoured (the real path/verb/payload are registered too, not the GET/empty defaults).
2) The VpcPairResourceService.execute() merges the deploy result back up: if deploy_result.get("changed"): result["changed"] = True.
So the stage-then-deploy sequence now reports changed: true at both the deployment and top level. The declarative state with no diff with changed=False and a real switch deploy of a staged pair asserts top-level changed is True (and that a repeated in-sync run stays False).

sivakasi-cisco:
> ### Medium: Config-save receives an undocumented deploy-scope body
> **Issue**
> 
> The module sends its deploy-scope selection to the fabric config-save endpoint:
> 
> ```json
> {"type": "resource"}
> ```
> 
> Nexus Dashboard’s config-save operation is bodyless. Deployment scope should only control which deployment path the module selects after saving.
> 
> If the proposed design change removes `resource`, the problem still exists because the same code would send:
> 
> ```json
> {"type": "switch"}
> ```
> 
> That body is also undocumented.
> 
> **Practical example**
> 
> With the current PR:
> 
> ```yaml
> - name: Save and deploy the affected vPC pair
>   cisco.nd.nd_manage_vpc_pair:
>     fabric_name: fabric1
>     state: merged
>     config_actions:
>       save: true
>       deploy: true
>       type: resource
>     config:
>       - peer1_switch_id: LEAF1
>         peer2_switch_id: LEAF2
> ```
> 
> The module currently sends:
> 
> ```httpspec
> POST /api/v1/manage/fabrics/fabric1/actions/configSave
> Content-Type: application/json
> 
> {"type": "resource"}
> ```
> 
> After the proposed scope simplification, it would send:
> 
> ```httpspec
> POST /api/v1/manage/fabrics/fabric1/actions/configSave
> Content-Type: application/json
> 
> {"type": "switch"}
> ```
> 
> The documented request should be bodyless in either case:
> 
> ```httpspec
> POST /api/v1/manage/fabrics/fabric1/actions/configSave
> ```
> 
> The scope is then applied to the separate deployment request:
> 
> ```httpspec
> POST /api/v1/manage/fabrics/fabric1/switchActions/deploy
> Content-Type: application/json
> 
> {"switchIds": ["LEAF1", "LEAF2"]}
> ```
> 
> **Evidence**
> 
> * `deploy.py` creates `action_payload = {"type": action_type}`.
> * It passes that payload to `FabricUtils.save_config()`.
> * An exact-head probe captured `{"type":"resource"}` on the config-save request.
> * OpenAPI operation `executeConfigSave` defines no request body.
> * The shared `ConfigActionsMixin.config_save()` implementation deliberately sends a bodyless request.
> 
> **Existing PR overlap**
> 
> No matching PR comment found.
> 
> **Existing open issue overlap**
> 
> Issues #428 and #467 are related to resource deployment and staged configuration, but neither defines or approves a config-save request body.
> 
> **Regression status**
> 
> The custom vPC action code already sends `switch` and `global` bodies. PR #513 newly adds the unsupported `resource` value, but the appropriate fix should make config-save bodyless for every deploy type.
> 
> **Impact**
> 
> A strict controller can reject config-save before deployment begins. A permissive controller may ignore the payload, but success then depends on undocumented behavior.
> 
> **Suggested fix**
> 
> Make `FabricUtils.save_config()` accept no payload—or an optional payload defaulting to `None`—and invoke config-save without a request body. Keep `type` only as local dispatch information for selecting switch-scoped or global deployment.
> 
> Add unit coverage asserting:
> 
> ```python
> assert config_save_request.payload is None
> ```
> 
> for both `type: switch` and `type: global`.

Done. FabricUtils.save_config() no longer accepts or sends a payload — it POSTs to .../actions/configSave bodyless (self.nd.request(path, HttpVerbEnum.POST)), and the save action is registered with payload=None. The type is now only local dispatch info for choosing the switch-scoped vs global deploy request; it never reaches config-save, so neither {"type":"resource"} (now gone) nor {"type":"switch"} is sent to configSave.

Asserted in unit 00010 (type: global) and 00110 (type: switch): payloads_for(SAVE_PATH) == [None]

sivakasi-cisco:
> ### Medium: Incomplete inventory can cause an asymmetric vPC deployment
> **Issue**
> 
> The deployment target resolver evaluates each vPC peer independently. If Nexus Dashboard returns only one member of a configured pair, the module warns about the missing peer but still deploys the member it found.
> 
> For pairs removed through `deleted` or `overridden`, missing peers are silently discarded without even generating the warning.
> 
> This remains applicable if `resource` is removed and its behavior moves to `type: switch`.
> 
> **Practical example**
> 
> Assume the task manages this pair:
> 
> ```yaml
> config_actions:
>   save: true
>   deploy: true
>   type: switch
> config:
>   - peer1_switch_id: LEAF1
>     peer2_switch_id: LEAF2
> ```
> 
> A transient, malformed, or incomplete inventory response contains only:
> 
> ```json
> {
>   "switches": [
>     {
>       "serialNumber": "LEAF1",
>       "configSyncStatus": "Out-of-Sync"
>     }
>   ]
> }
> ```
> 
> The module currently:
> 
> 1. Adds `LEAF1` to the deployment targets.
> 2. Warns that `LEAF2` was not found.
> 3. Sends:
> 
> ```json
> {
>   "switchIds": ["LEAF1"]
> }
> ```
> 
> 4. Reports the deployment as successful.
> 
> For a removed pair recovered from:
> 
> ```python
> class_diff["deleted"] = [("LEAF1", "LEAF2")]
> ```
> 
> an empty inventory response produces no warning and no deployment. A partial response again deploys only the peer it contains.
> 
> **Evidence**
> 
> * `_get_managed_pair_switches_needing_deploy()` adds each configured peer independently.
> * The warning does not prevent deployment of the resolved peer.
> * Deleted peers are silently intersected with the returned inventory.
> * The fabric-switch inventory endpoint supports pagination, but this path makes one request and does not verify `meta.counts.remaining`.
> * Current unit tests cover both configured peers being absent, but not exactly one configured peer being absent or deleted-pair inventory being empty or partial.
> * Exact-head probes reproduced both one-peer deployment and the silent deleted-pair no-op.
> 
> **Existing PR overlap**
> 
> Related to the resolved `_pair_identifiers()` thread from `@akinross` and the resolved empty-inventory thread from `@allenrobel`.
> 
> Those changes fixed required-key access and warnings when both configured peers are absent, but they did not enforce atomic pair resolution.
> 
> **Existing open issue overlap**
> 
> Related to #428 and #467, but neither tracks incomplete inventory or asymmetric deployment.
> 
> **Regression status**
> 
> The inventory helper is pre-existing, but using an unverified inventory snapshot to construct resource-scoped deployment targets is introduced by this PR.
> 
> **Impact**
> 
> The module can report success after deploying only one vPC member or after silently failing to deploy a pair removal. That can leave the two peers in inconsistent states.
> 
> **Suggested fix**
> 
> Treat configured vPC peers as an atomic unit:
> 
> * Load a complete, paginated inventory snapshot.
> * Require both configured peers to resolve before adding either one to the deployment payload.
> * Fail with a clear error when only one peer resolves.
> * For removed pairs, distinguish a switch confirmed to have been removed from an incomplete inventory response.
> * Do not silently interpret an absent inventory entry as proof that deployment is unnecessary.
> 
> Add tests for:
> 
> * Both configured peers present.
> * Both configured peers absent.
> * Exactly one configured peer absent.
> * Empty inventory for `class_diff.deleted`.
> * Exactly one deleted peer present.
> * Paginated inventory with a peer on a later page.

Addressed for the config-pair path. _get_managed_pair_switches_needing_deploy() now resolves each vPC pair atomically — if either peer is missing from the inventory snapshot it warns and skips the whole pair instead of deploying the one it found, so a partial/incomplete inventory can no longer produce an asymmetric half-pair deploy or a silent no-op. Deleted-pair peers still present are deployed; ones already absent are treated as removed.

It is kept as warn-and-skip rather than a hard fail so a transient inventory so that it  doesn't fail an otherwise valid run. Pagination of the inventory taken as a follow-up.

Tests: 00120 (peer not in inventory warns), 00200 (skips the pair when exactly one peer is absent), 00205 (_get_managed_pair_switches drops the asymmetric pair), 00190 (empty inventory warns per configured pair).

tests/integration/targets/nd_vpc_pair/tasks/nd_vpc_pair_merge.yaml
sivakasi-cisco: Removed them as we removed resource type.

akinross (approved):

sivakasi-cisco (commented):
Changes made to address the review comments.

Also made it to go hand-in-hand with [PR503. ](https://github.com/CiscoDevNet/ansible-nd/pull/503). Integ test passed successfully with combined changes

Issue #516, opened 2026-08-19
Bump Ansible Galaxy Version to 2.0.0


mtarking:
Bump Ansible Galaxy Version to 2.0.0

PR #518, opened 2026-08-20
Update release/changelog generation workflow after main/develop split (DCNE-890)

samiib:
## Proposed Changes

- Made the changelog comparison ref configurable via a new `compare_ref` workflow_dispatch input (defaults to `develop`), instead of hard-coding `develop` in both the collection checkout step and the release script's compare URL. This allows generating a changelog against any branch, tag, or commit SHA.
- Pinned the `release_script` checkout to a specific commit (`c3405128f9ff87cf5231cf2aac8e03424921b1a6`) rather than a mutable ref, so the workflow always runs a known-good version of the automation script.
- Added an optional `target_version` workflow_dispatch input, allowing the release version to be manually overridden (e.g. `2.0.0`) instead of relying solely on the automatically calculated version derived from bug/minor/major commit types.
- Fixed the `python-version` input being unintentionally parsed as `3.1` instead of `3.10` due to YAML interpreting the unquoted value as a float; it is now quoted as a string.

## Test Notes

- Test run of the changelog generation -> https://github.com/samiib/ansible-nd/actions/runs/32930754394/job/98062388169
- Example release PR result -> https://github.com/samiib/ansible-nd/pull/4/changes
Note: The release PR will still need manual adjustment after generation.

* [ ] analyze
* [ ] infa
* [ ] manage
* [ ] onemanage
* [x] other

## Checklist


.github/workflows/sync_main.yml
```diff
@@ -0,0 +1,23 @@
...
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Fast-forward main to develop
+        run: git push origin refs/remotes/origin/develop:refs/heads/main
```
mikewiebe: ### High: The new sync job cannot update `main` under the active ruleset

**Issue**

The workflow directly pushes `origin/develop` to `main`, but the live repository rules do not permit the workflow's `GITHUB_TOKEN` to update `main`. The release flow therefore fails at its final synchronization step.

**Evidence**

- [`sync_main.yml` line 23](https://github.com/CiscoDevNet/ansible-nd/pull/518/files#diff-3ac73592f8a6f745257685d4ba6004e629c63a0f2f7ac5419605e7bedcf88587R23) runs:

  ```shell
  git push origin refs/remotes/origin/develop:refs/heads/main
  ```

- Active repository ruleset [14381430](https://github.com/CiscoDevNet/ansible-nd/rules/14381430) targets `refs/heads/main` and enforces restricted updates plus a pull-request workflow.
- Its only bypass actor is repository role `admin`; there is no GitHub App or Actions integration bypass.
- `actions/checkout` configures the push to use the job's `GITHUB_TOKEN`, which is a GitHub App installation token. `contents: write` grants write permission but does not make that token an admin-role bypass actor.
- `main` is already an ancestor of `develop`, so this is an authorization failure rather than a non-fast-forward problem.

**Practical example**

1. A release PR is merged into `develop` by a repository administrator.
2. The `sync_main` workflow starts and runs as `github-actions[bot]`, not as the administrator who merged the PR.
3. The workflow attempts to fast-forward `main` with `git push origin refs/remotes/origin/develop:refs/heads/main`.
4. GitHub evaluates the push against the `main` ruleset using the workflow's `GITHUB_TOKEN` identity.
5. The token has `contents: write`, so it may attempt a repository write, but it is not the ruleset's `admin` bypass actor.
6. GitHub rejects the protected-branch update, the workflow fails, and `main` remains on the previous release.

In short: the Git history permits the fast-forward, but the identity performing it is not authorized by the ruleset.

**Existing PR overlap**

No matching open issue found. Issue #516 is adjacent release work but does not track branch authorization or `main` synchronization.

**Impact**

Every qualifying merged release PR reaches the direct push and is rejected, leaving `main` on the previous release snapshot.

**Suggested fix**

Use a purpose-built, narrowly scoped GitHub App whose installation is explicitly listed as a bypass actor for this ruleset, ideally behind a protected release environment. Alternatively, update `main` through an authorized PR flow that satisfies the repository's rules. Do not assume `contents: write` bypasses repository policy.

.github/workflows/changelog-generation.yml
```diff
@@ -17,6 +19,7 @@ jobs:
         uses: actions/checkout@v6
         with:
           path: ./collection
+          ref: develop
```
mikewiebe: ### High: Changelog generation and publication do not use one immutable release SHA

**Issue**

The workflow generates a release from one observation of the moving `develop` branch, then later publishes whichever commit `origin/develop` points to when the sync job runs. There is no immutable SHA binding the changelog, release PR, and `main` update together.

**Evidence**

- [`changelog-generation.yml` line 22](https://github.com/CiscoDevNet/ansible-nd/pull/518/files#diff-c2c5aafeb335d5ced1b63d08c95dc945905789e705c4f8ca1efb5b8e23309023R22) checks out the moving `develop` branch.
- The external release script independently compares the latest tag with the server-side moving `develop` ref.
- [`sync_main.yml` line 23](https://github.com/CiscoDevNet/ansible-nd/pull/518/files#diff-3ac73592f8a6f745257685d4ba6004e629c63a0f2f7ac5419605e7bedcf88587R23) later pushes the runner's current `origin/develop`, not the release PR head or another recorded release SHA.
- At review time, `main` exactly matches release tag `v1.5.0`, while `develop` is 54 commits ahead, confirming that `main` is being used as a release snapshot rather than as a continuously synchronized branch.

A practical race is:

1. The changelog job checks out `develop` at commit A.
2. Another PR merges as commit B while the external script performs its separate comparison, or after the changelog is generated.
3. The release PR contains artifacts based on A, but the later sync job pushes B because B is now `origin/develop`.

The published `main` snapshot can therefore contain B without B appearing in that release's generated changelog. A merge between checkout and comparison can produce the inverse mismatch: the changelog mentions B while the release branch was created from A.

**Existing PR overlap**

No matching open issue found.

**Impact**

The code, changelog, and release branch can describe different snapshots. Queue delays or retries make this possible even without two jobs running simultaneously.

**Suggested fix**

Select and record one immutable release SHA when the release starts, then use it throughout the release pipeline. This requires coordinated changes in three places:

1. In `.github/workflows/changelog-generation.yml`, resolve the checked-out `develop` commit once with `git rev-parse HEAD`, expose it as a step output such as `release_sha`, and pass it to the release script as `RELEASE_SHA`.
2. In `ciscoecosystem/release_script`, replace the hard-coded comparison against `develop` with a comparison against `RELEASE_SHA`, and create the `release_pr` branch from that exact commit. For example:

   ```python
   release_sha = os.environ["RELEASE_SHA"]
   compare_url = f"{repo_url}/compare/{latest_version}...{release_sha}"
   ```

   The branch helper should likewise use the recorded commit, for example `git checkout -B release_pr "$RELEASE_SHA"`.
3. Add a required release-PR check that verifies the recorded source SHA still matches the intended `develop` release head. If `develop` advances before approval, fail the check and require deliberate changelog regeneration rather than silently adding the new commits.
4. In `.github/workflows/sync_main.yml`, fetch and promote the exact merged release commit from `github.event.pull_request.merge_commit_sha` instead of the moving `origin/develop` ref. For example:

   ```yaml
   - name: Promote exact release commit
     env:
       RELEASE_SHA: ${{ github.event.pull_request.merge_commit_sha }}
     run: |
       git fetch origin "$RELEASE_SHA"
       git push origin "$RELEASE_SHA:refs/heads/main"
   ```

The separate `main` ruleset finding must also be resolved: this exact-SHA push still needs an authorized GitHub App bypass or an approved PR-based update to `main`.

.github/workflows/sync_main.yml
```diff
@@ -0,0 +1,23 @@
...
+on:
+  pull_request:
+    types:
+      - closed
+    branches:
+      - develop
+
+jobs:
+  sync-main:
+    if: github.event.pull_request.merged == true && github.event.pull_request.head.ref == 'release_pr'
```
mikewiebe: ### Medium: A fork branch named `release_pr` passes the release authorization gate

**Issue**

The post-merge job identifies a trusted release solely by the source branch name. It does not verify that the branch belongs to this repository.

**Evidence**

- [`sync_main.yml` line 12](https://github.com/CiscoDevNet/ansible-nd/pull/518/files#diff-3ac73592f8a6f745257685d4ba6004e629c63a0f2f7ac5419605e7bedcf88587R12) checks only:

  ```yaml
  github.event.pull_request.merged == true &&
  github.event.pull_request.head.ref == 'release_pr'
  ```

- `head.ref` contains only the branch name. A merged pull request from a fork whose branch is also named `release_pr` satisfies this condition.

**Practical example**

1. A contributor creates a branch named `release_pr` in their fork.
2. They open an ordinary pull request from that branch into `develop`.
3. Maintainers review and merge it as a normal contribution.
4. The closed-PR event contains `merged=true` and `head.ref=release_pr`, so both current conditions pass.
5. The workflow incorrectly treats that contribution as the official release and starts the `main` synchronization job.

**Existing PR overlap**

No matching open issue found.

**Impact**

An ordinary merged contribution can be mistaken for the release PR and trigger publication logic. Existing review rules reduce malicious exploitation, but they do not make branch-name-only authorization correct. The direct-push ruleset problem currently prevents the final update, but this authorization flaw remains once that blocker is fixed.

**Suggested fix**

Require both the expected branch and repository:

```yaml
if: >-
  github.event.pull_request.merged == true &&
  github.event.pull_request.head.ref == 'release_pr' &&
  github.event.pull_request.head.repo.full_name == github.repository
```

.github/workflows/changelog-generation.yml
```diff
@@ -1,11 +1,13 @@
...
 jobs:
   generate-changelog:
     name: Run automation script
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
     steps:
       - name: Check out script code
         uses: actions/checkout@v6
```
mikewiebe, line 23: ### Medium: Release-critical executable behavior is not pinned in this repository

**Issue**

The workflow checks out and executes the default branch of `ciscoecosystem/release_script`. PR #518 depends on a recent out-of-band change in that repository to compare against `develop`, but this PR does not pin or review the exact executable revision it requires.

I understand that this unpinned checkout predates PR #518, but the new explicit permission block narrows the repository's broader default workflow token rather than expanding it. The current concern is reproducibility and the new release flow's dependence on an unversioned external contract.

**Evidence**

- [`changelog-generation.yml` lines 12–16](https://github.com/CiscoDevNet/ansible-nd/pull/518/files#diff-c2c5aafeb335d5ced1b63d08c95dc945905789e705c4f8ca1efb5b8e23309023R13) checks out `ciscoecosystem/release_script` without `ref`.
- The reviewed external HEAD was `894a44483a85778e7cc44218be10a13753154586`, whose change from `master` to `develop` landed shortly before this PR.
- The workflow installs that checkout's unpinned `requirements.txt` and executes its Python code with contents and pull-request write permissions.
- The external script force-pushes the fixed `release_pr` branch and closes/recreates its PR, so changes to that mutable dependency directly alter this repository's release behavior.

**Practical example**

1. PR #518 is reviewed while `release_script` points to commit `894a44483a85778e7cc44218be10a13753154586`.
2. Someone later changes the external repository's default branch; no corresponding change is made to `ansible-nd`.
3. A maintainer manually runs the changelog workflow again from the same `ansible-nd` commit.
4. The workflow downloads and executes the new external code automatically.
5. That code can change files, force-push `release_pr`, or close and recreate its pull request using the repository token.

The same `ansible-nd` commit therefore has different release behavior depending on when it is run.

**Existing PR overlap**

No matching open issue found.

**Impact**

The same ansible-nd commit can behave differently on later manual dispatches, and future external changes can modify branches or pull requests without an ansible-nd review.

**Suggested fix**

Pin the external checkout to the audited full commit SHA:

```yaml
- name: Check out script code
  uses: actions/checkout@v6
  with:
    repository: ciscoecosystem/release_script
    ref: 894a44483a85778e7cc44218be10a13753154586
    path: ./release_script
```

Update that SHA only through reviewed pull requests. Pin the external Python dependencies as well, preferably with hashes, so dependency resolution is reproducible.

.github/workflows/sync_main.yml
samiib: Yes, this was a known restriction. This workflow had been used in [terraform-provider-aci](https://github.com/CiscoDevNet/terraform-provider-aci/blob/develop/.github/workflows/sync_main.yml#L20) but uses our specific automation gen user and separate rulesets. With the changes to the CiscoDevNet org this will need to be re-evaluated or done manually as it no longer works in that repo either.

For now I will remove the develop->main auto sync step and keep this PR scoped to changelog generation.

.github/workflows/sync_main.yml
samiib: This is a good suggestion if we re-add the main sync job later. It is now gone from this PR scope.

.github/workflows/changelog-generation.yml
samiib: Done, pinned to https://github.com/ciscoecosystem/release_script/commit/edd5e186aff0e854dbe128209624c11e5db15bb2

.github/workflows/changelog-generation.yml
samiib: Main sync is gone from this PR scope but this was definitely a possibility. We can still make the changelog generation comparison ref configurable and default to develop instead.

samiib: Its now possible to specify a comparison SHA/branch/tag when running the workflow. I also added an option to specify the exact target version of the bump.

changelogs/fragments/relocate_nd_argument_spec.yml
```diff
@@ -1,7 +0,0 @@
----
```
allenrobel, line 1: This fragment's content was never released, so deleting it here permanently loses both changelog entries (the `nd_argument_spec()` relocation and the `config_actions_spec()` addition from #387). `changelog.yaml`/`CHANGELOG.rst`/`galaxy.yml` all top out at 1.5.0 (2026-03-12), which predates the fragment (added 2026-07-19) — and it's the only pending fragment. With `keep_fragments: false` in `changelogs/config.yaml`, `antsibull-changelog release` deletes fragments only *after* folding them into the changelog at release time; deleting it by hand beforehand means the entries never appear in any future release notes. Was this an accidental side effect of a local test run of the release script? If so, please restore the fragment and drop this hunk from the PR.

allenrobel (commented):
### Code review

The workflow changes look good — the `compare_ref`/`target_version` inputs are wired exactly to what the pinned `release_script` commit (`edd5e186`) reads, and the added `permissions` block matches what the script needs. One inline finding below on the deleted changelog fragment, plus one note on the PR body:

1. The PR body's "Related Issue(s)" section is still the template placeholder — per the team convention, every PR should link an issue with `Closes #N` (DCNE-890 in the title doesn't auto-link/close anything on GitHub). Also, none of the "Related ND API Resource Category" checkboxes is checked — `other` would apply for a CI-only change.

changelogs/fragments/relocate_nd_argument_spec.yml
samiib: The general rule for our changelogs is only to add user facing changes. This makes it easier for users to understand changes that have impact on their operations. These changes only appear to be code refactoring that didn't change any behaviour for the users.

mikewiebe (approved):
LGTM

Issue #379, opened 2026-07-02
nd_interface_svi, nd_interface_subinterface_*: missing broad exception handler in main() — unhandled exceptions traceback instead of fail_json


allenrobel:
## Summary

Raised by Copilot review on PR #268 (https://github.com/CiscoDevNet/ansible-nd/pull/268#discussion_r3500646390): the `main()` docstring states `None (catches all exceptions and calls module.fail_json)`, but the `try` block only handles `NDStateMachineError`. Any other exception — including the explicit `AssertionError` raised by the orchestrator-type narrowing check in `nd_interface_svi.main()`, or a `ValueError`/`pydantic.ValidationError` escaping model validation — bubbles up as a raw traceback instead of a clean `fail_json` with the module's structured output.

The claim is verified, and it is not limited to `nd_interface_svi`. A sweep of all `NDStateMachine`-based modules shows three are missing the broad handler while every sibling module has it:

## Affected modules (docstring claims "catches all exceptions", only `NDStateMachineError` is caught)

- `plugins/modules/nd_interface_svi.py` (PR #268)
- `plugins/modules/nd_interface_subinterface_managed.py` (PR #296)
- `plugins/modules/nd_interface_subinterface_unmanaged.py` (PR #297)

## Compliant modules (for reference — all carry the two-handler pattern)

`nd_interface_loopback`, `nd_interface_ethernet_access`, `nd_interface_ethernet_trunk_host`, `nd_interface_port_channel_access`, `nd_interface_port_channel_trunk_host`, `nd_interface_vpc_access`, `nd_interface_vpc_trunk_host`, `nd_local_user`, `nd_maintenance_mode`, `nd_manage_fabric_{ebgp_vxlan,ibgp_vxlan,external}`, `nd_fabric_update_group`.

## Proposed change

Add the standard broad fallback handler used by the compliant modules (e.g. `nd_interface_loopback.main()`), after the existing `NDStateMachineError` handler:

```python
    except Exception as e:  # pylint: disable=broad-except
        module_log.exception("Unhandled exception during module execution")
        output = nd_state_machine.output.format() if nd_state_machine else {}
        error_msg = f"Module failed: {str(e)}"
        if module.params.get("output_level") == "debug":
            error_msg += f"\nTraceback:\n{traceback.format_exc()}"
        module.fail_json(msg=error_msg, **output)
```

[...]

PR #547, opened 2026-09-01
nd_interface_*: finalize controller-accepted mutations on the failure path

### LINKED ISSUE #546: nd_interface_*: failure path strands controller-accepted mutations (adopt the #403 finalizer beyond nd_interface_loopback)
### Description

PR #403 (review thread https://github.com/CiscoDevNet/ansible-nd/pull/403#discussion_r3905703008) found and fixed the following in `nd_interface_loopback` only:

When a task runs with `config_actions.deploy: true` and the module fails after the controller has already accepted some of the requested mutations (for example a later grouped bulk create is rejected), the accepted subset is left **staged but undeployed**. A retry then classifies those interfaces as unchanged (`no_diff`), never re-queues their deploy, and finishes successfully while controller intent and switch running state remain divergent.

The #403 fix added `NDBaseInterfaceOrchestrator.deploy_accepted_mutations()` plus a module-side failure-path helper that the loopback module's `except` handlers call. The nine sibling interface modules share the exact same code path and are still affected:

- `nd_interface_ethernet_trunk_host`
- `nd_interface_port_channel_access`
- `nd_interface_port_channel_trunk_host`
- `nd_interface_subinterface_managed`
- `nd_interface_subinterface_unmanaged`
- `nd_interface_svi`
- `nd_interface_vpc_access`
- `nd_interface_vpc_trunk_host`

Each `main()` calls `remove_pending()` / `deploy_pending()` only on the success path; their `except` handlers report the error and exit, so any deploy queued by an already-accepted mutation is dropped.

### Proposed change

- Lift the loopback-private helper into `plugins/module_utils/orchestrators/base_interface.py` as `finalize_accepted_intent(orchestrator, check_mode, module_log)` so all interface modules share one implementation (orchestrator-shaped signature, so the orchestrator layer does not import the state machine).
- Call it from every `except` handler in all ten `nd_interface_*` modules and append its NOTE sentence to the failure message. It honors `config_actions.deploy: false` (staged intent stays the contract), does nothing in check mode, and folds a failure-path deploy error into the message so it cannot mask the original error.
- Carry the loopback `config_actions.deploy` documentation sentence about the accepted subset over to the nine siblings.
- Unit tests for the shared helper, plus a parametrized module-level test that drives each module's `main()` with `AnsibleModule` / `NDStateMachine` stand-ins to prove the handlers call it (both handlers, `deploy: false`, check mode).

### Notes

- `nd_interface_svi` and the two `nd_interface_subinterface_*` modules carry only the `NDStateMachineError` handler; the broad `Exception` fallback is tracked by #379 and should call the finalizer too when it lands.
- The routed-ethernet work for #447 (branch stacked on #403) needs the same one-line adoption after it rebases.
- Depends on #403 (introduces `deploy_accepted_mutations`); the fix PR is stacked on the #403 branch.

### Which resource area of Cisco Nexus Dashboard does this belong to?

manage

### Related

- #403 (origin of the finalizer)
- #379 (missing broad handler in svi / subinterface modules)
- #523 (interface `config_actions` deploy-bridge centralization, same module family)

allenrobel:
## Related Issue(s)

Closes #546

## Proposed Changes

- Lift the `nd_interface_loopback`-private failure-path helper into `plugins/module_utils/orchestrators/base_interface.py` as `finalize_accepted_intent(orchestrator, check_mode, module_log)`. The signature is orchestrator-shaped so the orchestrator layer keeps no dependency on `NDStateMachine`.
- Call it from every `except` handler in all ten `nd_interface_*` modules and append its NOTE sentence to the failure message. Behavior is unchanged for loopback; the nine siblings gain the fix: when a task with `config_actions.deploy: true` fails after the controller has already accepted some mutations, the accepted subset is deployed and named instead of being left staged-but-undeployed (where a retry would classify it as unchanged and never deploy it).
- `nd_interface_svi` and the two `nd_interface_subinterface_*` modules carry only the `NDStateMachineError` handler, so they get the call there; their missing broad handler is #379 and should call the finalizer when it lands.
- Docs: carry the loopback `config_actions.deploy` sentence about the accepted subset over to the nine siblings.

Review round (mikewiebe, 2026-09-09), both fixed orchestrator-side in `base_interface.py` so the module wrappers are unchanged:

- **Mixed 207 on `interfaceActions/remove`** ("Reconcile mixed-207 removes and skip the finalizer after a failed normal deploy"): the remove queue was cleared only on whole-request success, so on a 207 with one `success` and one `Failed` item the accepted removal stayed queued, was classified as unsent by `_unsent_delete_pairs`, and was excluded from the failure-path deploy. `remove_pending` now snapshots the response count before the request; on failure it reads the response only if a new one was recorded (a sender exception leaves the previous response in place, #554), dequeues the pairs reported as an exact `success` matched on `(interfaceName, switchId)` (new `_accepted_multistatus_pairs`, the pair-keyed counterpart of `_accepted_multistatus_names`), leaves rejected / status-less / unknown pairs queued, and names the accepted pairs in the raised message.
- **Failed normal deploy resubmitted by the finalizer** (same commit): `deploy_pending` retains its queue on failure and runs inside the same `try` as the broad handler's finalizer, which then sent the identical deploy again. `deploy_pending` now sets a `_deploy_attempted` stage flag before sending and `deploy_accepted_mutations` returns early once the normal deploy has been attempted. This was already reachable on `develop` through loopback and routed; the fix covers all eleven modules.
- Dropped the changelog fragment: this repo's changelog is generated from squash-commit subjects by the release script (`changelog-generation.yml`), which never reads `changelogs/fragments/`.

This PR builds on the `NDBaseInterfaceOrchestrator.deploy_accepted_mutations` finalizer that landed with #403 (merged to `develop` 2026-09-02). It was originally stacked on that branch; it is now based directly on `develop`.

## Test Notes

- New `tests/unit/module_utils/orchestrators/test_base_interface.py` 00680–00685: `finalize_accepted_intent` with `None` orchestrator, check mode, non-interface orchestrator, `deploy: false`, success (sorted NOTE, queue drained), and failure-path deploy error (folded into the message, logged, queue retained).
- New `tests/unit/modules/test_nd_interface_finalize_accepted_intent.py`: parametrized over all eleven `nd_interface_*` modules (routed included, since it already carries the shared helper from #550), drives each `main()` with `AnsibleModule` / `NDStateMachine` stand-ins and asserts the `NDStateMachineError` handler (all eleven) and the broad handler (the eight that have one) deploy the accepted pair and name it, and that `deploy: false` and check mode issue no deploy.
- Review-round tests (`test_base_interface.py`): 00671 `deploy_pending` fails with 500, then `deploy_accepted_mutations` makes no request and retains the queue; 00730 mixed 207 with `loopback10` on two switches (only the `success` pair dequeued, both rejected pairs retained in order, message names both sets); 00740 untrusted item statuses (` Success ` accepted case/whitespace-tolerantly; `error`, missing `status`, a non-dict item, and an unsubmitted pair ignored); 00750 stale-response guard (an all-success 207 from a prior remove, then a sender exception on the next remove dequeues nothing). Removing the freshness guard makes 00750 fail, so it is load-bearing.
- Review-round module-boundary test 00040 across the eight two-handler modules: `manage_state` succeeds, the normal deploy raises, `fail_json` carries only the `Bulk deploy failed` message, and exactly one deploy request was made.
- Full unit suite: 4657 passed in the nd-dev machine.
- `ndtest --test validate-modules`, `ndtest --test pylint`, and `ndtest --test pep8` pass for all touched modules and the test files (`ndtest --test pylint` re-run on the three files touched by the review round).
- black / isort / pylint / mypy clean on the changed files (mypy's three pre-existing `ModelType` attribute errors in `base_interface.py` are unchanged).
- CI: the first `develop`-based run (after #403 merged) passed the full unit-test matrix, Black, lint, and build/import. The two Sanity jobs flagged one pep8 E501 (a 172-character string literal in the new test file) that the stacked configuration had hidden, since CI does not run on non-`develop`-based PRs. Fixed in the "Wrap ACCEPTED_NOTE literal to satisfy CI pep8 E501" commit; no behavior change. The run on the current head ("Drop the changelog fragment; this repo's changelog is generated from commit subjects") is fully green: all 56 checks pass, Integration Tests skipped as usual.
- Not lab-run: the loopback two-run convergence probe in the (now merged) #403 thread is the live evidence for the shared code path; the sibling adoption is a one-line call into the same base-class method. The mixed-207 remove reconciliation is unit-tested against the spec's `multiDeleteResponse` shape and the vault's observed `Failed` status; it has not been reproduced on the lab.

4.2.1

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

* [x] Latest commit is rebased from develop with merge conflicts resolved (branch is directly on `develop` at #548's squash commit `67197ec9`)

https://claude.ai/code/session_01H5ogXuT7RjKHTF8Ty5RXpL
https://claude.ai/code/session_01GtjGrZGWEtY2dfDPuVVfkb

### DIFF
--- a/plugins/modules/nd_interface_ethernet_access.py
+++ b/plugins/modules/nd_interface_ethernet_access.py
@@ -268,6 +268,9 @@ options:
         - When V(true), all queued interface changes are deployed in a single bulk API call at the end of module
           execution via the C(interfaceActions/deploy) API. Only the interfaces modified by this task are deployed.
         - When V(false), changes are staged but not deployed. Use a separate deploy module or task to deploy later.
+        - When V(true) and the module fails after the controller has already accepted a subset of the requested changes, that
+          accepted subset is still deployed and is named in the failure message, so a failed task does not leave accepted
+          changes staged but undeployed.
         - Setting O(config_actions.deploy=false) is useful when batching changes across multiple interface tasks before a single deploy.
         - Deployment is opt-in. Set O(config_actions.deploy=true) explicitly to push changes to switches.
         type: bool
@@ -422,7 +425,7 @@ from ansible_collections.cisco.nd.plugins.module_utils.common.pydantic_compat im
 from ansible_collections.cisco.nd.plugins.module_utils.models.interfaces.ethernet_access_interface import EthernetAccessInterfaceModel
 from ansible_collections.cisco.nd.plugins.module_utils.nd_argument_specs import config_actions_spec, nd_argument_spec
 from ansible_collections.cisco.nd.plugins.module_utils.nd_state_machine import NDStateMachine
-from ansible_collections.cisco.nd.plugins.module_utils.orchestrators.base_interface import NDBaseInterfaceOrchestrator
+from ansible_collections.cisco.nd.plugins.module_utils.orchestrators.base_interface import NDBaseInterfaceOrchestrator, finalize_accepted_intent
 from ansible_collections.cisco.nd.plugins.module_utils.orchestrators.ethernet_access_interface import EthernetAccessInterfaceOrchestrator
 
 
@@ -622,6 +625,7 @@ def main() -> None:
         module_log.exception("NDStateMachineError during module execution")
         output = nd_state_machine.output.format() if nd_state_machine else {}
         error_msg = f"Module execution failed: {str(e)}"
+        error_msg += finalize_accepted_intent(nd_state_machine.model_orchestrator if nd_state_machine else None, module.check_mode, module_log)
         if module.params.get("output_level") == "debug":
             error_msg += f"\nTraceback:\n{traceback.format_exc()}"
         module.fail_json(msg=error_msg, **output)
@@ -630,6 +634,7 @@ def main() -> None:
         module_log.exception("Unhandled exception during module execution")
         output = nd_state_machine.output.format() if nd_state_machine else {}
         error_msg = f"Module failed: {str(e)}"
+        error_msg += finalize_accepted_intent(nd_state_machine.model_orchestrator if nd_state_machine else None, module.check_mode, module_log)
         if module.params.get("output_level") == "debug":
             error_msg += f"\nTraceback:\n{traceback.format_exc()}"
         module.fail_json(msg=error_msg, **output)
--- a/changelogs/fragments/interface_finalize_accepted_intent.yml
+++ b/changelogs/fragments/interface_finalize_accepted_intent.yml
@@ -1,0 +1,9 @@
+---
+bugfixes:
+  - >-
+    nd_interface_* - when a task with ``config_actions.deploy`` set to ``true`` fails after the controller has already
+    accepted some of the requested interface changes, deploy that accepted subset and name it in the failure message instead
+    of leaving it staged but undeployed. A retry previously classified those interfaces as unchanged and never deployed them,
+    so controller intent and switch running state stayed divergent even after a successful retry. The failure-path finalizer
+    introduced for ``nd_interface_loopback`` is now shared (``finalize_accepted_intent`` in ``base_interface.py``) and
+    adopted by all ten ``nd_interface_*`` modules.
--- a/changelogs/fragments/interface_finalize_accepted_intent.yml
+++ b/changelogs/fragments/interface_finalize_accepted_intent.yml
@@ -1,9 +1,0 @@
----
-bugfixes:
-  - >-
-    nd_interface_* - when a task with ``config_actions.deploy`` set to ``true`` fails after the controller has already
-    accepted some of the requested interface changes, deploy that accepted subset and name it in the failure message instead
-    of leaving it staged but undeployed. A retry previously classified those interfaces as unchanged and never deployed them,
-    so controller intent and switch running state stayed divergent even after a successful retry. The failure-path finalizer
-    introduced for ``nd_interface_loopback`` is now shared (``finalize_accepted_intent`` in ``base_interface.py``) and
-    adopted by all ten ``nd_interface_*`` modules.
--- a/plugins/module_utils/orchestrators/base_interface.py
+++ b/plugins/module_utils/orchestrators/base_interface.py
@@ -71,7 +71,8 @@ class NDBaseInterfaceOrchestrator(NDBaseOrchestrator[ModelType]):
 
         Initialize mutable private state after Pydantic model construction. Pydantic disallows `Field()` on
         underscore-prefixed names, so these are set here to ensure each instance gets its own container: the
-        deploy/remove queues and the per-switch interface cache read by `_switch_interfaces`.
+        deploy/remove queues, the `_deploy_attempted` stage flag read by `deploy_accepted_mutations`, and the per-switch interface
+        cache read by `_switch_interfaces`.
 
         ## Raises
 
@@ -79,6 +80,7 @@ class NDBaseInterfaceOrchestrator(NDBaseOrchestrator[ModelType]):
         """
         self._pending_deploys: list[tuple[str, str]] = []
         self._pending_removes: list[tuple[str, str]] = []
+        self._deploy_attempted: bool = False
         self._switch_interfaces_cache: dict[str, dict[str, dict]] = {}
 
     def apply_config_actions(self, params: Mapping[str, Any]) -> bool:
@@ -409,14 +411,18 @@ class NDBaseInterfaceOrchestrator(NDBaseOrchestrator[ModelType]):
 
         When `deploy` is `False`, returns `None` without making any API call.
 
+        Sets `_deploy_attempted` before sending so that, if this request fails, the failure-path finalizer
+        (`deploy_accepted_mutations`) does not resubmit the identical deployment (PR #547 review).
+
         ## Raises
 
         ### RuntimeError
 
-        - If the deploy API request fails.
+        - If the deploy API request fails. The queue is retained.
         """
         if not self.deploy or not self._pending_deploys:
             return None
+        self._deploy_attempted = True
         try:
             result = self._deploy_interfaces(self._pending_deploys)
             self._pending_deploys = []
@@ -442,8 +448,9 @@ class NDBaseInterfaceOrchestrator(NDBaseOrchestrator[ModelType]):
         exactly one that was not accepted.
 
         Returns the deployed `(interface_name, switch_id)` pairs so the caller can name them in the failure report. Returns an
-        empty list without any API call when `deploy` is `False` (staged intent is the documented contract in that case) or when
-        no accepted-mutation pairs are queued.
+        empty list without any API call when `deploy` is `False` (staged intent is the documented contract in that case), when
+        the normal `deploy_pending` request was already attempted (a failed normal deploy is reported by its own error and must not
+        be resubmitted — PR #547 review), or when no accepted-mutation pairs are queued.
 
         ## Raises
 
@@ -451,7 +458,7 @@ class NDBaseInterfaceOrchestrator(NDBaseOrchestrator[ModelType]):
 
         - If the failure-path deploy API request fails. The accepted pairs remain queued in that case.
         """
-        if not self.deploy:
+        if not self.deploy or self._deploy_attempted:
             return []
         unsent = self._unsent_delete_pairs()
         accepted = [pair for pair in self._pending_deploys if pair not in unsent]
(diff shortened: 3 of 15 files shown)


allenrobel:
⚠️ Merge-ordering note: #551 (Closes #379) adds the broad ``except Exception`` handler to ``nd_interface_svi`` / ``nd_interface_subinterface_managed`` / ``nd_interface_subinterface_unmanaged``. Whichever of #551 / this PR lands second must add the ``finalize_accepted_intent(...)`` line to that new broad handler in all three modules during rebase — git will not merge it in automatically since the block is new.

akinross (approved):

plugins/modules/nd_interface_ethernet_access.py
```diff
@@ -630,6 +636,7 @@ def main() -> None:
         module_log.exception("Unhandled exception during module execution")
         output = nd_state_machine.output.format() if nd_state_machine else {}
         error_msg = f"Module failed: {str(e)}"
+        error_msg += finalize_accepted_intent(nd_state_machine.model_orchestrator if nd_state_machine else None, module.check_mode, module_log)
```
mikewiebe, line 637: ### High: Reconcile exact-success removals from a mixed 207 before finalizing

**Issue**

The failure finalizer assumes every pair left in `_pending_removes` was not accepted. The bulk-remove endpoint can return HTTP 207 with independent per-interface results, so one removal can succeed while another fails. `remove_pending()` leaves the whole queue intact on that exception; the successful deletion is therefore classified as unsent and excluded from deployment.

**Evidence**

- [The new broad exception path invokes the failure finalizer after `remove_pending()` raises](https://github.com/CiscoDevNet/ansible-nd/pull/547/files#diff-d9dedc3c5c695765362c861c47123a3a9b3f0354811540374561ab575ab91513R639).
- [`deploy_accepted_mutations()` excludes every pair reported by `_unsent_delete_pairs()`](https://github.com/CiscoDevNet/ansible-nd/blob/a2b52be383024964d3d8c97b2109244df48fa8be/plugins/module_utils/orchestrators/base_interface.py#L411-L449), and the base implementation defines every queued remove as unsent.
- [`remove_pending()` clears the queue only when the entire request succeeds](https://github.com/CiscoDevNet/ansible-nd/blob/a2b52be383024964d3d8c97b2109244df48fa8be/plugins/module_utils/orchestrators/base_interface.py#L515-L536).
- The bundled Manage OpenAPI contract defines `POST /fabrics/{fabricName}/interfaceActions/remove` as HTTP 207 `multiDeleteResponse`; each result carries `interfaceName`, `switchId`, `status`, and `message`.
- An exact-head probe returned `loopback10/SW1: success` and `loopback20/SW1: failed`. Both pairs remained in `_pending_removes`, the failure finalizer returned an empty list, and no deploy request was made.

**Existing PR overlap**

No matching existing PR comment found. The author's only PR #547 conversation comment concerns PR #551 merge ordering, not mixed per-item removal results or queue reconciliation.

No matching open issue found. [Issue #546](https://github.com/CiscoDevNet/ansible-nd/issues/546) is the umbrella behavior this PR claims to complete, but it does not identify mixed-remove result reconciliation. [Issue #397](https://github.com/CiscoDevNet/ansible-nd/issues/397) classifies a mixed 207 as request failure without reconciling accepted item identities into orchestrator queues. [Issue #554](https://github.com/CiscoDevNet/ansible-nd/issues/554) tracks stale-response risk in the separate normalize path. None tracks this exact remove-specific bookkeeping defect, although #554's response-freshness requirement must shape the fix.

**Impact**

For example, an `overridden` task removes two omitted loopbacks with `config_actions.deploy: true`. ND removes `loopback10` from controller intent but rejects `loopback20`. The task fails and deploys neither. On retry, `loopback10` is already absent from controller inventory, so it is not queued again. The retry can succeed while the switch still has `loopback10` and the controller does not.

**Suggested fix**

Before `_remove_interfaces()`, snapshot the response/result history. On failure, inspect only a newly appended HTTP 207 response belonging to that request. Extract exact-`success` `(interfaceName, switchId)` pairs, intersect them with the submitted queue, dequeue only those accepted pairs, and retain failed/missing/unknown pairs. Re-raise the original failure so the existing finalizer deploys the accepted subset. Add coverage for duplicate interface names on two switches, missing/unknown statuses, and an exception before a fresh response is appended.

---

plugins/modules/nd_interface_ethernet_access.py
```diff
@@ -630,6 +634,7 @@ def main() -> None:
...
```
mikewiebe, line 637: ### Medium: Do not resubmit the normal deployment after it has already failed

**Issue**

The wrapper runs `deploy_pending()` inside the same `try` covered by the new catch-all finalizer. If normal deployment fails, `deploy_pending()` retains its queue and raises. The exception handler immediately calls `deploy_accepted_mutations()`, which submits the identical deployment a second time.

**Evidence**

- [The new broad exception path invokes the failure finalizer after normal `deploy_pending()` raises](https://github.com/CiscoDevNet/ansible-nd/pull/547/files#diff-d9dedc3c5c695765362c861c47123a3a9b3f0354811540374561ab575ab91513R637), while normal execution calls `remove_pending()` and then `deploy_pending()` immediately beforehand.
- [`deploy_pending()` clears `_pending_deploys` only after success](https://github.com/CiscoDevNet/ansible-nd/blob/8e36356bce621d3842ac148ab7a945d35a6d9eb8/plugins/module_utils/orchestrators/base_interface.py#L403-L425).
- [`deploy_accepted_mutations()` then submits the retained accepted pairs](https://github.com/CiscoDevNet/ansible-nd/blob/8e36356bce621d3842ac148ab7a945d35a6d9eb8/plugins/module_utils/orchestrators/base_interface.py#L427-L465).
- An exact-head probe made two identical deploy submissions after the first returned a terminal failure. The new module-boundary tests raise only from `manage_state()` and therefore do not exercise this stage.

**Existing PR overlap**

No matching existing PR comment found. The same behavior was documented locally during the PR #403 re-review but was never posted as a PR comment.

No matching open issue found. [Issue #457](https://github.com/CiscoDevNet/ansible-nd/issues/457) discusses transport-layer mutation retries, while [issue #546](https://github.com/CiscoDevNet/ansible-nd/issues/546) owns accepted-intent finalization. Neither tracks this stage-specific second deployment initiated by the module exception handler.

**Impact**

For example, an operator configures the same server-facing port on two leaf switches and requests immediate deployment:

```yaml
- name: Configure and deploy server ports
    fabric_name: site1
    config:
      - switch_ip: 192.168.1.11
        interface_names: [Ethernet1/20]
        config_data:
          network_os:
            policy:
              access_vlan: 100
      - switch_ip: 192.168.1.12
        interface_names: [Ethernet1/20]
        config_data:
          network_os:
            policy:
              access_vlan: 100
    config_actions:
      deploy: true
    state: merged
```

ND accepts and stages both interface changes. The normal deploy request then reports `LEAF1/Ethernet1/20: success` and `LEAF2/Ethernet1/20: failed` because the second switch rejects the configuration. `deploy_pending()` raises while retaining both pairs. The broad exception handler calls `finalize_accepted_intent()`, which submits both pairs again: LEAF1 is needlessly redeployed, and LEAF2 repeats the same deterministic failure.

For a retryable 500 response, the first deployment can instead consume its full retry budget before failing, after which the failure finalizer starts a second deployment attempt and potentially another complete retry window.

**Suggested fix**

Run accepted-intent finalization only for failures that occur before normal deployment begins. Track the execution stage, or separate `manage_state()`, `remove_pending()`, and `deploy_pending()` exception handling. Add a module-boundary test that makes normal `deploy_pending()` fail and asserts exactly one deploy request.

plugins/modules/nd_interface_ethernet_access.py
allenrobel: Confirmed against the branch and the 4.2.1 spec: the remove 207 keys each result by `interfaceName` + `switchId`, and the queue was only cleared on whole-request success, so an accepted removal was classified as unsent and skipped by the finalizer.

Fixed in "Reconcile mixed-207 removes and skip the finalizer after a failed normal deploy": `remove_pending` now snapshots the response count before the request; on failure it reads the response only if a new one was recorded, dequeues the pairs reported as an exact `success` (matched on the `(interfaceName, switchId)` pair, so a duplicate name on two switches is told apart), leaves rejected/status-less/unknown pairs queued, and names the accepted pairs in the raised message. The finalizer then deploys the accepted subset as before.

Tests cover the duplicate-name-on-two-switches case, untrusted statuses (`Failed`, `error`, missing key, non-dict item, unsubmitted pair), and a sender exception with no new response, which per #554 leaves the previous 207 in place and must dequeue nothing.

plugins/modules/nd_interface_ethernet_access.py
allenrobel: Confirmed: `deploy_pending` retained its queue on failure and the broad handler's finalizer resubmitted the same deploy. Note this was already reachable on develop via the loopback and routed modules; this PR extended it to the rest.

Fixed in the same commit on the orchestrator side rather than in the module wrappers: `deploy_pending` sets a `_deploy_attempted` stage flag before sending, and `deploy_accepted_mutations` returns early once the normal deploy has been attempted, so the failure is reported by its own error only. Keeping this in `base_interface` leaves the eleven wrappers unchanged.

Added a unit test (`deploy_pending` 500, then the finalizer makes no request and the queue is retained) and a module-boundary test across the two-handler modules that makes the normal deploy fail and asserts exactly one deploy request.

mikewiebe (approved):
LGTM pending CI fixes

Issue #407, opened 2026-07-17
Add reusable, safe pagination for ND collection queries in the Orchestrators framework


gmicol:
## Summary

Add a reusable, opt-in offset-pagination component to the Orchestrators framework for Nexus Dashboard collection endpoints that expose `max` and `offset` query parameters.

This enhancement will be implemented only after all currently developed Routing Policies modules have been merged into `develop`.

Until then, Route Maps, Community Lists, and Extended Community Lists will use the existing ACL pagination approach within their respective feature pull requests. No shared pagination abstraction or broader framework refactor should be introduced into those pull requests.

## Dependencies and implementation gate

The framework enhancement must not begin until the current Routing Policies module work is merged into `develop`.

Related pull requests:

- [x] [#266 — ACL module](https://github.com/CiscoDevNet/ansible-nd/pull/266)
- [ ] [#286 — Prefix Lists](https://github.com/CiscoDevNet/ansible-nd/pull/286)
- [ ] [#287 — Community Lists and Extended Community Lists](https://github.com/CiscoDevNet/ansible-nd/pull/287)
- [x] [#288 — Route Maps](https://github.com/CiscoDevNet/ansible-nd/pull/288)

The intended sequence is:

1. Complete the individual Routing Policies modules.
2. Use the ACL pagination method temporarily for Route Maps, Community Lists, and Extended Community Lists.
3. Merge all Routing Policies pull requests into `develop`.
4. Re-evaluate the pagination implementations from the common `develop` baseline.
5. Implement the shared pagination enhancement in a dedicated framework pull request.
6. Migrate the merged Routing Policies modules to the shared implementation.
7. Evaluate other existing collection modules for follow-up migration.

This sequencing avoids rebasing multiple feature branches around a new framework abstraction and keeps pagination refactoring separate from module delivery.

## Background

`NDStateMachine` initializes its current-state collection using `model_orchestrator.query_all()`. Collection reconciliation is therefore correct only when `query_all()` retrieves the complete resource collection.

Several ND Manage APIs use the same pagination contract:

- `max` and `offset` query parameters.
- A resource-specific collection wrapper such as:
  - `accessControlLists`
  - `ipv4PrefixLists`
  - `ipv6PrefixLists`
  - `communityLists`
  - `extendedCommunityLists`
  - `routeMaps`
- Optional metadata containing:
  - `counts.total`
  - `counts.remaining`
  - `links.next`
  - `links.previous`

[...]

PR #548, opened 2026-09-02
nd_interface_port_channel_*: member-already-in-use preflight

### LINKED ISSUE #369: nd_interface_port_channel_*: add member-already-in-use preflight (convert ND's opaque 500 to fail-fast validation)
## Summary

Add a client-side **member-already-in-use preflight** to the port-channel interface orchestrators (`nd_interface_port_channel_access`, `nd_interface_port_channel_trunk_host`) so that configuring a port-channel whose member ethernet is already owned by another port-channel fails fast with a clear validation error, instead of surfacing ND's opaque mid-run HTTP 500.

**Status (rebaselined 2026-09-01 against `develop`):** no longer blocked. Both port-channel modules and the shared interface base are merged to `develop`, and PR #390 hoisted the config-scoped `_switches_to_query()` plus a per-switch interface cache (`_switch_interfaces()`) into `NDBaseInterfaceOrchestrator`. Tagged `2.0.0` as a user quality-of-life fix to land before the release. All file/line anchors below are from `develop` as of the rebaseline date. Start from a fresh branch off `develop`.

## Background — what ND does today (lab-verified, ND 4.2.1)

Verified by direct REST against the live controller (SITE1, `192.168.7.7`, switch `S1_LE1`/`93ZOLKD04T7`) on 2026-06-30. Full write-up in the bug-tracker vault: **`port-channel-member-conflict-returns-500`**.

1. Create `port-channel701` with a free member `Ethernet1/35` → **HTTP 207** `{"results":[{"status":"success",...}]}`. The member's intent `policyType` immediately flips `trunkHost → accessPoMember`.
2. Create `port-channel702` reusing the **same** member `Ethernet1/35` →

   ```
   HTTP 500
   {"code":500,"description":"",
    "message":"Switch [S1_LE1/93ZOLKD04T7]: Member port Ethernet1/35 is a member of port-channel701, remove Ethernet1/35 from port-channel701 first. "}
   ```
3. `GET .../interfaces/port-channel702` → **404**: the conflicting port-channel was never created (atomic; no silent re-home).

Two properties make this worth a client-side preflight:

- **The error bypasses `results[]`.** Success is a 207 with a `results[]` array; this conflict is a **flat 500 `{code,message}`** with no `results[]`. It also masquerades as a transient server error when it is really a permanent 409-class conflict (retry logic keyed on 5xx would loop).
- **The conflict is checked against ND *intent*, not deployed state** (the owning PO in the repro was never deployed). So a preflight that reads the existing interface intent reliably predicts the conflict **before** any write — no deploy required.

Today the port-channel orchestrators do **no** membership preflight; `PortChannelBaseOrchestrator.create` wraps `_request` (which raises on non-success) into `RuntimeError("Create failed for <po>: ...")`, so the user sees the 500 surfaced mid-run rather than as a fail-fast validation error.

## Proposed design (agreed)

Three axes; the chosen bundle is **1a + 2a + 3a**, hard-failing in check mode.

### 1. Where — override `preflight()` in `PortChannelBaseOrchestrator`

The hook and its single call site already exist: `NDStateMachine.manage_state` calls `model_orchestrator.preflight(proposed_items)` (`nd_state_machine.py:118`) for **merged/replaced/overridden** only, before any mutation, and it runs in check mode. Deletes correctly skip it (`nd_state_machine.py:130`). Since PR #362 the state machine wraps both preflight hooks and normalizes a bare `RuntimeError` into `NDStateMachineError` (`nd_state_machine.py:105-118`), so the new check should raise `RuntimeError` like the existing capability preflight does.

```python
# orchestrators/port_channel_base.py
def preflight(self, model_instances):
    super().preflight(model_instances)          # existing capability preflight (currently inert, see aside)
    self._validate_members_available(model_instances)
```

The membership check **cannot** live in the Pydantic model — `NDBaseModel`/validators are pure and have no `rest_send`/API reach (`models/base.py`). It must be in the orchestrator.

### 2. Data source — the shared per-switch interface cache; **read the unfiltered list**

The per-switch interface GET already runs once at state-machine init (`nd_state_machine.py:71` → `query_all()`), scoped to config switches for merged/replaced and fabric-wide for overridden. Its response **already contains the ethernet members** — but `query_all()` filters to managed port-channel types before returning and throws the rest away.

- **Rebaseline note:** `NDBaseInterfaceOrchestrator._switch_interfaces(switch_id)` (`base_interface.py:124-148`) already fetches the per-switch `interfaceList` at most once per run and caches the **unfiltered** result keyed by lower-cased interface name in `_switch_interfaces_cache`. Every other interface orchestrator's `query_all` goes through it. `PortChannelBaseOrchestrator.query_all` (`port_channel_base.py:246-286`) is the **one remaining orchestrator that still issues its own `_request` per switch** and never populates the cache.
- Step one is therefore to switch `PortChannelBaseOrchestrator.query_all` to `list(self._switch_interfaces(switch_id).values())` (mirroring `ethernet_base.py:652` / `loopback_interface.py:257`). That is a no-behavior-change refactor that warms the cache for the config switches during the state machine's initial `query_all`.
- The preflight then reads membership from `self._switch_interfaces(switch_id)` for each proposed switch → **0 extra requests** for merged/replaced/overridden, because those switches were already fetched by `query_all`. (Do **not** add a per-switch or per-member GET in preflight, and do not introduce a second private inventory attribute — the base cache is the single source of truth.)
- **Critical:** read membership from the **unfiltered** records. A member can be owned by a *different* policy type than the orchestrator manages. Real example: `Ethernet1/2` is owned by `port-channel500` (`vpcPeerlinkPo`); the access orchestrator's filtered `self.before` would show it as free and **miss the conflict**. Use every interface whose `policyType` ends in `Member` (or every `portChannel` record's `ports`, any type).

### 3. Strictness — reject cross-PO conflicts; allow same-PO (idempotent)

A proposed member is rejected when its current `policyType` ends in `*Member` **and** its owning port-channel ≠ the PO being configured in this task. A member already owned by the PO under management is allowed (preserves idempotent re-apply of `merged`). The error names the member and its current owner, mirroring ND's message but fail-fast.

- Intra-task re-home (member moving PO-A→PO-B, both in the same config) is **out of scope for v1** — ND rejects it at create without mutation ordering we don't have yet; treat as a conflict with a clear message.
- **Check mode:** hard-fail even under `--check`. The existing capability preflight downgrades to a warning in check mode (`base_interface.py:295-300`) only because its `capableSwitches` endpoint is unpublished/unreliable. Membership comes from the standard, reliable interfaces GET, so a hard fail gives a truer dry-run.

## Implementation sketch

- `orchestrators/port_channel_base.py` — (1) `query_all()` reads from `self._switch_interfaces(switch_id)` instead of its own `_request`; (2) new `preflight()` override calling `super().preflight(model_instances)` then `_validate_members_available(model_instances)`, which walks `self._switch_interfaces(self._resolve_switch_id(item.switch_ip))` for each proposed item.
- No change needed at the call site (`nd_state_machine.py:118`), to the base `preflight()` contract (`base_interface.py:194`), or to `_switch_interfaces` itself.
- Existing unit tests: `tests/unit/module_utils/orchestrators/test_port_channel_{access,trunk_host}_interface.py` with fixtures under `tests/unit/module_utils/fixtures/fixture_data/`. The `query_all` refactor must keep those green before the preflight is added.

## Test plan (TDD)

- Fixture: per-switch interfaces GET where `Ethernet1/35` is a `trunkHost` (free) and `Ethernet1/2` is `vpcPeerLinkMember` of `port-channel500` (the cross-type trap).
- Cases: (a) free member → passes; (b) member owned by a different PO of the **same** type → fails with member+owner named; (c) member owned by a different PO of a **different** type (`vpcPeerlinkPo`) → fails (guards against reading the filtered view); (d) idempotent re-apply where the member is already owned by the PO under management → passes; (e) `--check` mode → still hard-fails.

## Scope / blocking

- ~~**Blocked until** `nd_interface_port_channel_access` / `nd_interface_port_channel_trunk_host` (and the shared interface-orchestrator base) are merged into `develop`.~~ **Unblocked** — all interface stacks merged to `develop` (see status note at top). Two orchestrators inherit `PortChannelBaseOrchestrator` (`port_channel_access_interface.py`, `port_channel_trunk_host_interface.py`); both pick up the preflight automatically.
- Conceptually extends to vPC host port-channels (`accessVpcHost`/`trunkVpcHost`); the vPC variant's exact 500 shape is **not yet lab-confirmed** (vault documents the same `Member port ... is a member of VPC_PAIR` family). Worth a follow-up once vPC interface modules land.

## Aside (separate, optional)

The PC orchestrators don't set the `interface_type` / `interface_mode` ClassVars, so the inherited **capability** preflight (`super().preflight()`) is currently inert for port-channels. Independent of this member check, but worth deciding whether to wire it up in the same pass.

## Related

- bug-tracker vault note: `port-channel-member-conflict-returns-500`
- Tracking index: #357
- Same message family on a different endpoint: vault `vpc-physical-peer-link-configsave-fails`

allenrobel:
## Related Issue(s)

Closes #369

## Proposed Changes

Add a client-side member-already-in-use preflight to the port-channel orchestrators (`nd_interface_port_channel_access`, `nd_interface_port_channel_trunk_host`), per the design agreed in #369.

ND 4.2.1 rejects a port-channel create whose `ports` member already belongs to another port-channel with a flat HTTP 500 `{code, message}` envelope (no `results[]`) -- a permanent 409-class conflict dressed as a server error (vault: `port-channel-member-conflict-returns-500`). Today that surfaces mid-run as `Create failed for <po>: ...`, after any earlier interfaces in the same POST batch were already accepted. The module now fails fast with a validation error naming each offending member and its current owner, before any write, in check mode too.

- **`PortChannelBaseOrchestrator.query_all`** now reads each switch's inventory through the shared `NDBaseInterfaceOrchestrator._switch_interfaces` cache (mirroring the ethernet/loopback/SVI orchestrators). No behavior change; it was the last orchestrator still issuing its own per-switch `_request`. This warms the cache with the **unfiltered** records during the state machine's initial `query_all`.
- **`PortChannelBaseOrchestrator.preflight`** override: runs the inherited capability preflight, then `_validate_members_available`. Because it reads the same cache, it issues **zero additional requests** for merged/replaced/overridden.
  - Owners are derived from every `portChannel` record's `ports` list regardless of policy type, so a member of an unmanaged flavor (e.g. `Ethernet1/2` owned by `port-channel500`/`vpcPeerlinkPo`) is still a conflict -- the filtered `before` view would miss it.
  - Membership is read from **intent**, not `operData.portChannelId` (which stays `-1` until the owner is deployed; ND still rejects the conflict).
  - A member whose own `policyType` ends in `Member` but that no port-channel record lists is treated as owned (owner reported as unknown) rather than free.
  - A member already owned by the port-channel under management passes (idempotent re-apply). Two proposed port-channels on one switch claiming the same member also fail (ND would accept the first and 500 on the second).
  - Hard-fails in `--check` mode: unlike the capability preflight (unpublished endpoint), this data comes from the standard interfaces GET.
  - Intra-task re-home (remove from A, add to B in one config) stays out of scope per the issue and is reported as a conflict.
- Workaround site carries `TODO(4.2.1) port-channel-member-conflict-returns-500`.
- Module `description` gets one line stating the fail-fast behavior; changelog fragment added.
- Not done: the optional "aside" in #369 about wiring the capability `interface_type`/`interface_mode` ClassVars for port-channels.

## Test Notes

- New unit tests in `tests/unit/module_utils/orchestrators/test_port_channel_access_interface.py` (00900-00970): free member passes; cross-PO conflict aggregates every offender naming member + owner; cross-type (`vpcPeerlinkPo`) owner conflicts; same-PO re-apply passes (case-insensitive); check mode still fails; intra-task duplicate claim fails; `query_all` then `preflight` consumes no extra response (generator exhaustion would raise) and the cache retains the unfiltered inventory; orphan `*Member` policy type fails. `test_port_channel_trunk_host_interface.py` 00900-00910 confirm inheritance.
- Full unit suite: 4205 passed (`ndpytest tests/unit/`).
- `ndtest --test pylint` / `--test validate-modules` on the changed plugin files pass; `ndblack`, `ndisort` clean; `ndmypy` error count on `port_channel_base.py` unchanged from develop (pre-existing `NDBaseModel` attr-defined pattern noted at the top of the file).
- Not lab-run: the conflict shape itself was lab-verified on 2026-06-30 (issue body / vault note); the preflight is exercised against that recorded shape in unit tests.

4.2.1

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

https://claude.ai/code/session_012LpZPyFp9vCBb1mH97CXj4

### DIFF
--- a/changelogs/fragments/port_channel_member_preflight.yml
+++ b/changelogs/fragments/port_channel_member_preflight.yml
@@ -1,0 +1,5 @@
+---
+bugfixes:
+  - nd_interface_port_channel_access, nd_interface_port_channel_trunk_host - fail fast with a validation error naming the
+    member and its current owner when a port-channel lists a member ethernet already belonging to a different port-channel,
+    instead of surfacing the controller's opaque mid-run HTTP 500 (https://github.com/CiscoDevNet/ansible-nd/issues/369).
--- a/changelogs/fragments/port_channel_member_preflight.yml
+++ b/changelogs/fragments/port_channel_member_preflight.yml
@@ -1,5 +1,0 @@
----
-bugfixes:
-  - nd_interface_port_channel_access, nd_interface_port_channel_trunk_host - fail fast with a validation error naming the
-    member and its current owner when a port-channel lists a member ethernet already belonging to a different port-channel,
-    instead of surfacing the controller's opaque mid-run HTTP 500 (https://github.com/CiscoDevNet/ansible-nd/issues/369).
--- a/plugins/module_utils/orchestrators/port_channel_base.py
+++ b/plugins/module_utils/orchestrators/port_channel_base.py
@@ -19,6 +19,8 @@ Inherits shared interface lifecycle operations (deploy queuing, fabric validatio
 resolution) from `NDBaseInterfaceOrchestrator` and adds port-channel-specific functionality:
 - Standard remove-based deletion (port-channels are virtual interfaces and are deletable)
 - Fabric-wide `query_all()` filtered by `interfaceType: "portChannel"` and per-type policy filtering
+- A member-already-in-use `preflight()` that rejects, before any write, a port-channel whose member ethernet is
+  already owned by a different port-channel (issue #369)
 
 Member ethernet interfaces are not managed by this orchestrator — the port-channel policy is the
 single source of truth for member configuration via the `ports` list. Member field restrictions
@@ -28,6 +30,7 @@ on standalone ethernet modules are enforced separately by the ethernet orchestra
 from __future__ import annotations
 
 from collections import defaultdict
+from collections.abc import Sequence
 from typing import ClassVar
 
 from ansible_collections.cisco.nd.plugins.module_utils.endpoints.base import NDEndpointBaseModel
@@ -67,6 +70,7 @@ class PortChannelBaseOrchestrator(NDBaseInterfaceOrchestrator[ModelType]):
 
     - Via `validate_prerequisites` if the fabric does not exist or is in deployment-freeze mode.
     - Via `_resolve_switch_id` if no switch matches the given IP in the fabric.
+    - Via `preflight` if a proposed member ethernet is already owned by a different port-channel.
     - Via `create` if the create API request fails.
     - Via `update` if the update API request fails.
     - Via `remove_pending` if the bulk remove API request fails.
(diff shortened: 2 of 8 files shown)


mikewiebe (approved):
Re-reviewed at 2777face. The port-channel member-ownership preflight is sound for the complete inventory it receives, fails before writes, and behaves correctly in check mode. Focused tests and CI are green. The inherited inventory-pagination limitation remains explicitly tracked by #407 and is not a blocker for this PR.

changelogs/fragments/port_channel_member_preflight.yml
```diff
@@ -0,0 +1,5 @@
+---
+bugfixes:
```
akinross: remove the changelog

plugins/module_utils/orchestrators/port_channel_base.py
```diff
@@ -223,6 +227,144 @@ def delete_bulk(self, model_instances: list[ModelType], **kwargs) -> None:
...
+        for iface in inventory.values():
+            if iface.get("interfaceType") != "portChannel":
+                continue
+            for member in self._policy_of(iface).get("ports") or []:
+                if isinstance(member, str) and member:
+                    owners[member.lower()] = iface.get("interfaceName", "")
+        for name, iface in inventory.items():
+            policy_type = self._policy_of(iface).get("policyType")
+            if name not in owners and isinstance(policy_type, str) and policy_type.endswith("Member"):
+                owners[name] = f"unknown port-channel (member policyType {policy_type})"
```
akinross: This is not something you have to change because the logic here is correct but I personally found the 2 loop structure without comments very hard to follow. I think we do not need 2 loops here and could simplify with something like example below.

```
owners: dict[str, str] = {}

for name, iface in inventory.items():
    policy = self._policy_of(iface)
    policy_type = policy.get("policyType")

    # Default unset member interfaces to an unknown owner.
    if isinstance(policy_type, str) and policy_type.endswith("Member"):
        owners.setdefault(
            name,
            f"unknown port-channel (member policyType {policy_type})",
        )

    # Replace the default with the explicit owner when available.
    if iface.get("interfaceType") == "portChannel":
        for member in policy.get("ports") or []:
            if isinstance(member, str) and member:
                owners[member.lower()] = iface.get("interfaceName", "")

akinross (changes requested):

changelogs/fragments/port_channel_member_preflight.yml
allenrobel: Removed in "Collapse member-ownership map to a single inventory pass; drop changelog fragment".

plugins/module_utils/orchestrators/port_channel_base.py
allenrobel: Agreed, the single pass reads much better. Adopted your shape in "Collapse member-ownership map to a single inventory pass; drop changelog fragment": setdefault for the unknown-owner default and a plain assignment for the explicit owner, so the result is the same whichever order the inventory lists the member and its port-channel (the inventory is keyed by lowercased interfaceName, so the two lookups agree). Kept a short comment on each branch. Focused tests (42) and the full unit suite (4598) pass.

akinross (approved):

PR #551, opened 2026-09-03
nd_interface_svi, nd_interface_subinterface_*: add broad exception handler in main()

### LINKED ISSUE #379: nd_interface_svi, nd_interface_subinterface_*: missing broad exception handler in main() — unhandled exceptions traceback instead of fail_json

allenrobel:
## Related Issue(s)

Closes #379

## Proposed Changes

- Add the standard broad ``except Exception`` fallback handler after the existing ``NDStateMachineError`` handler in
  ``nd_interface_svi``, ``nd_interface_subinterface_managed``, and ``nd_interface_subinterface_unmanaged``.
  The handler body is identical to the one every sibling ``NDStateMachine``-based module already carries
  (e.g. ``nd_interface_loopback.main()``), so the existing docstring contract
  (``None (catches all exceptions and calls module.fail_json)``) now holds.
- Add a ``bugfixes`` changelog fragment.

⚠️ **Merge-ordering note (PR #547):** #547 adds a ``finalize_accepted_intent(...)`` line to *both* handlers in the
compliant interface modules. The new broad handlers introduced here do not have that line because it does not exist
on develop yet. Whichever of #547 / this PR lands second must add that one line to the new broad handler in each of the
three modules during rebase — git will not do it automatically.

## Test Notes

- ``black``, ``isort``, and ``ansible-test sanity --test pylint`` pass on the three modules; ``yamllint`` sanity passes on the fragment.
- ``pylint`` (wrapper) output is byte-identical before/after the change (all findings pre-existing).
- ``mypy`` reports only the pre-existing unresolvable ``ansible.module_utils.basic`` import present on develop.
- Unit tests referencing these modules (``test_nd_interface_deploy_default.py``, ``test_nd_state_machine.py``) pass (20 tests).

4.2.1

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

### DIFF
--- a/changelogs/fragments/379_broad_exception_handler.yml
+++ b/changelogs/fragments/379_broad_exception_handler.yml
@@ -1,0 +1,5 @@
+---
+bugfixes:
+  - nd_interface_svi, nd_interface_subinterface_managed, nd_interface_subinterface_unmanaged - add the broad fallback exception
+    handler in ``main()`` so unhandled exceptions (e.g. model validation errors) are reported via ``fail_json`` with the module's
+    structured output instead of a raw traceback, matching the sibling interface modules (https://github.com/CiscoDevNet/ansible-nd/issues/379).
--- a/changelogs/fragments/379_broad_exception_handler.yml
+++ b/changelogs/fragments/379_broad_exception_handler.yml
@@ -1,5 +1,0 @@
----
-bugfixes:
-  - nd_interface_svi, nd_interface_subinterface_managed, nd_interface_subinterface_unmanaged - add the broad fallback exception
-    handler in ``main()`` so unhandled exceptions (e.g. model validation errors) are reported via ``fail_json`` with the module's
-    structured output instead of a raw traceback, matching the sibling interface modules (https://github.com/CiscoDevNet/ansible-nd/issues/379).
--- a/plugins/modules/nd_interface_subinterface_managed.py
+++ b/plugins/modules/nd_interface_subinterface_managed.py
@@ -468,6 +468,14 @@ def main():
             error_msg += f"\nTraceback:\n{traceback.format_exc()}"
         module.fail_json(msg=error_msg, **output)
 
+    except Exception as e:  # pylint: disable=broad-except
+        module_log.exception("Unhandled exception during module execution")
+        output = nd_state_machine.output.format() if nd_state_machine else {}
+        error_msg = f"Module failed: {str(e)}"
+        if module.params.get("output_level") == "debug":
+            error_msg += f"\nTraceback:\n{traceback.format_exc()}"
+        module.fail_json(msg=error_msg, **output)
+
 
 if __name__ == "__main__":
     main()
--- a/plugins/modules/nd_interface_subinterface_unmanaged.py
+++ b/plugins/modules/nd_interface_subinterface_unmanaged.py
@@ -299,6 +299,14 @@ def main():
             error_msg += f"\nTraceback:\n{traceback.format_exc()}"
         module.fail_json(msg=error_msg, **output)
 
+    except Exception as e:  # pylint: disable=broad-except
+        module_log.exception("Unhandled exception during module execution")
+        output = nd_state_machine.output.format() if nd_state_machine else {}
+        error_msg = f"Module failed: {str(e)}"
+        if module.params.get("output_level") == "debug":
+            error_msg += f"\nTraceback:\n{traceback.format_exc()}"
+        module.fail_json(msg=error_msg, **output)
+
 
 if __name__ == "__main__":
     main()
--- a/plugins/modules/nd_interface_svi.py
+++ b/plugins/modules/nd_interface_svi.py
@@ -606,6 +606,14 @@ def main():
             error_msg += f"\nTraceback:\n{traceback.format_exc()}"
         module.fail_json(msg=error_msg, **output)
 
+    except Exception as e:  # pylint: disable=broad-except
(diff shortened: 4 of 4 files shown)


plugins/modules/nd_interface_subinterface_managed.py
```diff
@@ -477,6 +477,14 @@ def main():
             error_msg += f"\nTraceback:\n{traceback.format_exc()}"
         module.fail_json(msg=error_msg, **output)
 
+    except Exception as e:  # pylint: disable=broad-except
```
akinross, line 471: I was cross-checking modules and notice that exception handling is done differently across modules, is this something that should be discussed and standardised?

Since most of the code is identical in the two exceptions ( just minor log and message difference from which I am not sure it is very meaningful in the logs ) should we considering a generic module error function that is called on the broad exception that takes in a few arguments, then logs and fails with the desired standard?

akinross (changes requested):

plugins/modules/nd_interface_subinterface_managed.py
allenrobel: Agreed the handlers are worth standardizing. A sweep of the `NDStateMachine` modules on develop shows five distinct shapes today (this two-handler block in 14 modules; `format_with_verbosity` with no logging in `nd_manage_fabric_*`; a `traceback=` kwarg in `nd_manage_policy_group`; a guarded `output.format()` in `nd_manage_l3out`; a bare one-liner in `nd_local_user` / `nd_fabric_update_group`), and a shared `fail_json` helper is the right fix.

Lets do this as its own PR rather than fold it into this one:

1. **Scope**: #379 is that these three modules diverge from the 14 siblings that already carry this exact block. This PR makes them match. A helper only these three call would leave them diverging in a new way until the sweep lands.
2. **Sequencing**: #547 adds a `finalize_accepted_intent(...)` line to both handlers in all eleven `nd_interface_*` modules. A helper introduced before #547 merges conflicts with every one of those hunks; one introduced after can absorb the finalizer so interface modules get a single call.

On the log/message wording: "Unhandled exception" / "Module failed" is the maintainer's grep for a bug in our code vs a controller-reported failure, so I'd keep that distinction as a helper parameter.

Filed #556 for the helper + sweep (with the proposed signature and the five-shape inventory), sequenced after #547. Happy to do it here instead if you'd prefer.

akinross: No separate PR is easier I think, so fine with this or if you want to address these exceptions already in the new form is also ok.

changelogs/fragments/379_broad_exception_handler.yml
```diff
@@ -0,0 +1,5 @@
+---
+bugfixes:
```
akinross: remove this changelog

akinross (changes requested):

changelogs/fragments/379_broad_exception_handler.yml
allenrobel: Removed in "Drop the changelog fragment; this repo's changelog is generated from commit subjects".

akinross (approved):

mikewiebe (approved):
LGTM

PR #557, opened 2026-09-10
Standardize main() exception handling behind fail_from_exception

### LINKED ISSUE #556: Standardize main() exception handling behind a shared fail-from-exception helper
## Summary

Raised by @akinross on PR #551 (https://github.com/CiscoDevNet/ansible-nd/pull/551#discussion_r3955749473): the `except NDStateMachineError` / `except Exception` blocks in `main()` are near-identical copy-paste across modules, and exception handling is not uniform across the collection. Replace the copied blocks with one shared helper that logs and calls `module.fail_json` in a single standard shape.

## Current state (develop, 2026-09-09)

A sweep of the `NDStateMachine`-based modules shows five distinct handler shapes:

1. **Two-handler block** (`module_log.exception(...)`, `nd_state_machine.output.format()`, `Module execution failed:` / `Module failed:` prefix, traceback appended when `output_level: debug`) — all `nd_interface_*` modules, `nd_manage_acl`, `nd_manage_route_map`, `nd_manage_prefix_list`, `nd_manage_community_list`, `nd_manage_extended_community_list`, `nd_maintenance_mode` (14 modules; #551 brings `nd_interface_svi` and the two `nd_interface_subinterface_*` modules onto this shape).
2. **`format_with_verbosity`, no logging** — `nd_manage_fabric_{ebgp,ibgp,external,ai_ebgp,ai_ibgp}_vxlan`.
3. **`traceback=` kwarg + per-line `log.error`** — `nd_manage_policy_group`.
4. **Guarded `output.format()`** (`try/except` around the format call) — `nd_manage_l3out`.
5. **Bare one-liner** (`fail_json(msg=..., **output.format())`, no logging, no debug traceback) — `nd_local_user`, `nd_fabric_update_group`.

The interface modules additionally append `finalize_accepted_intent(...)` to both handlers (PR #403 for loopback, PR #550 for ethernet routed, PR #547 for the rest).

## Proposed change

Add a helper to `plugins/module_utils/` (exact home TBD — `nd_output.py` or a small new `module_failure.py`), roughly:

```python
def fail_from_exception(
    module: AnsibleModule,
    module_log: logging.Logger,
    nd_state_machine: NDStateMachine | None,
    error: BaseException,
    *,
    log_message: str,
    prefix: str,
) -> NoReturn:
```

- Logs via `module_log.exception(log_message)`.
- Builds `output` from `nd_state_machine.output.format()` when the state machine exists (guarded like `nd_manage_l3out`, so a formatting failure cannot mask the original error).
- Appends `finalize_accepted_intent(...)` when the state machine's orchestrator is an `NDBaseInterfaceOrchestrator` (no-op otherwise), so the interface modules collapse to a single call per handler.
- Appends the traceback when `output_level == "debug"`.
- Keeps the two-tier wording as parameters: `"Unhandled exception during module execution"` / `Module failed:` is the maintainer's grep for a bug in our code, versus `"NDStateMachineError during module execution"` / `Module execution failed:` for a controller- or validation-reported failure.

Then sweep every `NDStateMachine`-based module onto it. Modules owned by other contributors (`nd_manage_policy_group`, `nd_manage_l3out`, `nd_manage_fabric_*`) should be coordinated with their owners rather than changed unilaterally.

## Sequencing

**After PR #547 merges.** #547 edits both handlers in all eleven `nd_interface_*` modules; a helper landed before it conflicts with every one of those hunks, and a helper landed after it can absorb the finalizer.

## Related

- PR #551 review thread (origin): https://github.com/CiscoDevNet/ansible-nd/pull/551#discussion_r3955749473
- Issue #379 (the three modules missing the broad handler; fixed by #551)
- Issue #546 / PR #547 (`finalize_accepted_intent` adoption across `nd_interface_*`)

allenrobel:
## Related Issue(s)

Closes #556

Origin: @akinross's review thread on #551 (https://github.com/CiscoDevNet/ansible-nd/pull/551#discussion_r3955749473).

## Proposed Changes

- Add `plugins/module_utils/module_failure.py` with a single `fail_from_exception(module, module_log, nd_state_machine, error)` helper. It logs via `module_log.exception`, merges the state machine's formatted output into `fail_json` (guarded like `nd_manage_l3out`, so a formatting failure cannot mask the original error), appends the `finalize_accepted_intent` note when the orchestrator is an `NDBaseInterfaceOrchestrator` (no-op otherwise), and appends the traceback when `output_level` is `debug`.
- The tier is inferred from the exception type, so the two wordings live in one place: `NDStateMachineError` → `Module execution failed: ...` (controller- or validation-reported), anything else → `Module failed: ...` (unhandled, the maintainer's grep for a bug in our code). Each module's two copied handlers collapse to one `except Exception` clause that delegates to the helper. Failure messages and log lines are unchanged for every swept module.
- Swept: the 11 `nd_interface_*` modules, `nd_maintenance_mode`, `nd_manage_acl`, `nd_manage_route_map`, `nd_manage_prefix_list`, `nd_manage_community_list`, `nd_manage_extended_community_list`, and the state-machine handler in `nd_fabric_update_group` (its separate auto-assign handler wraps a different flow with no state machine and is left as is). In `nd_fabric_update_group` the `NDStateMachine` construction now also sits inside the guarded `try` (review finding, commit "nd_fabric_update_group: construct NDStateMachine inside the fail_from_exception guard"): the constructor runs the initial `query_all`, and on develop a failure there escaped `main()` as a raw exception because the module was the only one built before its `try`.
- Not swept: modules owned by other contributors (`nd_manage_fabric_*`, `nd_manage_l3out`, `nd_manage_policy_group`, `nd_manage_links`, `nd_local_user`). Converting them changes their behavior (adds logging, the debug traceback, and the standard prefixes), so that is left for a follow-up coordinated with their owners.

## Release Classification

**Refactor, targeted at 2.0.1** (label changed from 2.0.0 by @allenrobel). It qualifies as neither a bug fix nor missing functionality under the 2.0.0 triage rule:

- For 17 of the 18 swept modules the change is behavior-preserving: same log lines, same failure messages, same output keys, same finalizer behavior. The unit tests that drive each interface module's `main()` pin those messages and pass unchanged.
- `nd_fabric_update_group` gains the standard tiering: an unhandled exception now reads `Module failed:` instead of `Module execution failed:`, and the `output_level: debug` traceback is new there. Cosmetic alignment, not a reported defect. Its `NDStateMachine` construction failures (initial query, orchestrator construction, proposed-config validation) now also reach `fail_json` with the standard message instead of escaping as a raw exception; pre-existing on develop, unreported, and covered by a new module-boundary test.
- The guarded `output.format()` call is a latent robustness improvement borrowed from `nd_manage_l3out`; it only matters if formatting the failure result itself raises, which no issue or lab run has shown.

Deferring this PR holds nothing else back: #547 is the bug fix in this stack and stands alone, and #551 (issue #379) can land in 2.0.0 on its own; this PR supersedes its handler hunks later without needing to land first.

## Sequencing

Was stacked on #547 (`nd_interface_finalize_accepted_intent`), which adds the finalizer calls this helper absorbs; #547 and #551 have both merged and this branch is rebased onto develop, so it stands alone.

## Test Notes

- New `tests/unit/module_utils/test_module_failure.py` (8 tests): tier wording and log line per exception type, state-machine output merged into `fail_json`, `None` state machine, guarded `output.format()` failure keeps the original message, debug traceback only at `output_level: debug`, finalizer note for an interface orchestrator, and check mode skipping the finalizer.
- `tests/unit/modules/test_nd_interface_finalize_accepted_intent.py` now drives the unhandled-exception tier for all 11 interface modules (the one-handler / two-handler split disappears).
- `tests/unit/modules/test_nd_fabric_update_group.py::test_nd_fabric_update_group_00300` drives the real `main()` with an `NDStateMachine` stand-in whose constructor raises and asserts exactly one `fail_json` call carrying `Module execution failed: ...`; it failed on the previous head (the exception escaped `main()`) and passes with the fix.
- Full unit suite: 4672 passed (`ndpytest tests/unit/`).
- `ansible-test sanity --test pylint` on all changed files: pass. `ansible-lint plugins/`: pass. black, isort, pylint, mypy on the helper: clean (mypy's only remaining error is the environment's unresolvable `ansible.module_utils.basic` import).

4.2.1

* [ ] analyze
* [ ] infra
* [ ] manage
* [ ] onemanage
* [x] other

## Checklist

https://claude.ai/code/session_01DV9sCsmsDCPXUrPu2GSxz5


gmicol (approved):
LGTM

plugins/modules/nd_fabric_update_group.py
```diff
@@ -376,9 +377,8 @@ def main():
         nd_state_machine.manage_state()
         module_log.debug("manage_state end")
         module.exit_json(**nd_state_machine.output.format())
-    except Exception as e:
-        module_log.exception("Unhandled exception during module execution")
-        module.fail_json(msg=f"Module execution failed: {str(e)}", **nd_state_machine.output.format())
+    except Exception as e:  # pylint: disable=broad-except
```
mikewiebe, line 380: ### Medium: Fabric update-group initialization bypasses the shared failure handler

**Issue**

`nd_fabric_update_group` constructs `NDStateMachine` before entering the `try` guarded by the new helper. Any current-state query, orchestrator construction, or proposed-config validation failure raised during initialization therefore escapes `fail_from_exception()` entirely. This behavior already exists on `develop`, so it is not a regression, but the PR's conversion of this module does not actually give its initialization path the standardized failure behavior that the new helper and its `nd_state_machine is None` branch are designed to provide.

**Evidence**

- [`plugins/modules/nd_fabric_update_group.py` line 380](https://github.com/CiscoDevNet/ansible-nd/pull/557/files#diff-420d145705ca6ec1d2908c1629d85cdd9b670d2f3d71ed56b5695f6398dc7d0cR380) adds the shared handler, but the guarded `try` begins at line 375, after `NDStateMachine(...)` is constructed on lines 370-373.
- [`tests/unit/module_utils/test_module_failure.py` lines 246-263](https://github.com/CiscoDevNet/ansible-nd/pull/557/files#diff-f837b4ff63fd4067346d63c75a58b6ec517f22598474153ecad78f1fce819d00R250) explicitly tests the helper with `nd_state_machine=None`, but no module-boundary test exercises an update-group constructor failure.
- [`NDStateMachine.__init__` lines 93-132](https://github.com/CiscoDevNet/ansible-nd/blob/0e3156977d14bec2ef594dd1c2fc19aceca61e67/plugins/module_utils/nd_state_machine.py#L93-L132) performs the initial controller query and model construction, then wraps failures as `NDStateMachineError`.
- A focused wrapper probe replaced `NDStateMachine` with one that raises `NDStateMachineError("Initialization failed: controller unavailable")`; `target.main()` propagated that exception and recorded zero `fail_json` calls (`PYENV_VERSION=ansible_2_19 python -m pytest /private/tmp/pr557_init_probe.py -q`: **1 passed** while asserting the escape).
- A 72-case base/head parity probe (18 wrappers × two exception types × normal/debug output) found semantic parity for all 17 two-handler conversions. The only semantic differences were the four expected update-group cases, which reinforces that this wrapper is the sole outlier in the converted set.

**Existing PR overlap**

Related open issue [#556](https://github.com/CiscoDevNet/ansible-nd/issues/556) owns the broader shared-handler rollout, but it does not identify this `NDStateMachine` construction-before-`try` gap. The gap is specific to this PR's update-group conversion.

**Impact**

For example, an `nd_fabric_update_group` task can fail while the constructor performs its initial `query_all()` because ND returns HTTP 500. Instead of returning the standard `Module execution failed: Initialization failed: ...` result through `fail_json`, with traceback exposure controlled by `output_level`, the exception escapes the module entry point. Users get an inconsistent/raw module failure and none of the standardized handler behavior this PR promises.

**Suggested fix**

This issue should be fixed in PR #557. Although the constructor placement is pre-existing, #557 explicitly converts `nd_fabric_update_group` to the shared failure handler. Leaving `NDStateMachine(...)` outside the `try` means that conversion is incomplete and contradicts the helper's existing `nd_state_machine=None` support.

The fix is small and directly within this PR's scope:

```python
nd_state_machine = None

try:
    nd_state_machine = NDStateMachine(...)
    ...
except Exception as error:
    fail_from_exception(..., nd_state_machine, error, ...)
```

Add a module-boundary regression test that makes the constructor raise and asserts one `fail_json` call with the standard message; this will exercise the helper's already-tested `None` path through the real wrapper. The alternative would be to remove `nd_fabric_update_group` from this PR's claimed conversion scope and explicitly track it as remaining #556 work, but fixing it here is clearer.

allenrobel: Confirmed and fixed in commit "nd_fabric_update_group: construct NDStateMachine inside the fail_from_exception guard".

I checked the other 11 modules in the sweep: every one initializes `nd_state_machine = None` and constructs inside the `try`; `nd_fabric_update_group` was the only outlier, and it was that way on develop as you noted. The module now follows the same shape, so an `NDStateMachineError` from the constructor's initial `query_all` (or orchestrator construction / proposed-config validation) reaches `fail_json` through the helper's `None` path with `Module execution failed: ...`.

Regression test: `test_nd_fabric_update_group_00300` drives the real `main()` with a fake `AnsibleModule` and an `NDStateMachine` stand-in whose constructor raises `NDStateMachineError("Initialization failed: controller unavailable")`, and asserts exactly one `fail_json` call with `{"msg": "Module execution failed: Initialization failed: controller unavailable"}` (no output keys, matching the helper's `None` branch). It failed on the previous head with the exception escaping `main()` at the constructor line, and passes now. Full unit suite 4672 passed; sanity pylint clean on both files. PR body updated (Proposed Changes, Release Classification, Test Notes).

mikewiebe (approved):
Reviewed exact head 71c5084c. The constructor-boundary issue is fixed, the 137-test targeted suite passes, and the status-rollup CI is green apart from the intentionally skipped integration job. Approving. Nonblocking follow-up: keep the nine deferred #556 modules explicitly tracked (change Closes #556 to Refs #556 or create a replacement issue) and carry the empty RETURN documentation debt.

PR #550, opened 2026-09-02
nd_interface_ethernet_routed: L3 routed ethernet interfaces (routedHost / iosXeRoutedHost)

allenrobel:
## Related Issue(s)

Closes #447

## Proposed Changes

Second module at the "(interfaceType × mode)" boundary from the interface-granularity design, following the discriminated-union pattern that landed for loopback in #403. It answers the recurring "where is L3 Routed for VXLAN?" question: the GUI create wizard lists 0 capable switches, but the API accepts `routedHost` on a VXLAN leaf (lab-verified in #447).

- **New module `nd_interface_ethernet_routed`** with a two-level union: outer `network_os_type` (`nx-os` | `ios-xe`), inner `policy_type` per OS branch.
  - NX-OS: `routedHost` (`intRoutedHostTemplate`) - ip/prefix, vrf, routing_tag, PIM, netflow, QoS, FEC, speed, mtu, admin_state, extra_config.
  - IOS-XE: `iosXeRoutedHost` (`iosXeIntRoutedHostTemplate`).
  - States: merged, replaced, overridden, deleted. EXAMPLES cover every state.
- **`EthernetRoutedInterfaceOrchestrator`** on `EthernetBaseOrchestrator`:
  - `query_all` filters to the managed set (`routedHost`, `iosXeRoutedHost`) only, so `overridden` can never touch fabric underlay links, vPC keepalives, multisite members, or other system-provisioned routed types.
  - `deleted` normalizes the interface back to the fabric default (physical ethernet cannot be deleted), with a C8000V-safe reset path for IOS-XE.
  - Mode flips (trunk -> routed) ride the existing create path: a trunk-intent interface is not in `before[]`, so the state machine classifies it as a create and the POST rewrites the intent.
  - Capability-preflight opt-out for the VXLAN "0 capable switches" wizard filter.
  - Unconfigured-default signature filter so fabric-default interfaces are not reported as managed. The signature is derived from each policy model's `reverse_diff_defaults` table plus an explicit overlay for ND-injected read keys (`ptp`), so the query-scope filter and the replaced/overridden reverse pass share one source of truth.
- **Models**: `policy_base.py` (`InterfacePolicyStrictBase`, promoted from `LoopbackPolicyStrictBase`; carries the shared `adminState` default and honours the framework's `mode=response` read context), `ethernet_routed_interface.py` with per-policy `reverse_diff_defaults` tables sourced from the ND 4.2.1 OpenAPI templates, and new enums.
[...]

Issue #530, opened 2026-08-27
NDStateMachine modules never populate diff or logs in module output despite documenting them


allenrobel:
## Summary

Every `NDStateMachine`-backed (Gen-3) module returns `diff: []` and — at `output_level: debug` — `logs: []`, unconditionally. Neither key is ever populated, yet 24 modules document `diff` as `returned: always` with a populated sample, and 19 document `logs`. Surfaced by review of #403 (thread https://github.com/CiscoDevNet/ansible-nd/pull/403#discussion_r3861995959), but it is a framework gap, not a module one.

## Root cause

- `NDStateMachine` only ever calls `self.output.assign(after=..., before=..., proposed=...)` (`nd_state_machine.py:85`) and later `assign(after=...)`. It never passes `diff` or `logs`.
- `NDOutput.format()` computes `changed` by comparing `before` and `after` (`get_diff_collection`) but does not derive a diff from the same comparison; it emits `self._diff`, which stays at its `[]` default.
- Nothing in `module_utils` collects log records into a list: `common/log.py` only permits file handlers, so `NDOutput._logs` has no producer.
- `diff=` has never been assigned since the initial ND 4.2 state-machine commit (`git log -S"diff="` on `nd_state_machine.py`).

The only diff that exists today is API-level: `Results` stores a per-call `diff` that `format_with_verbosity()` surfaces as `api_diff` at `-vvv`. That is the payload sent, not a before/after config diff.

## Scope

25 modules construct `NDStateMachine` (the `nd_interface_*` family, `nd_manage_fabric_*`, `nd_local_user`, `nd_maintenance_mode`, `nd_manage_vpc_pair`, `nd_manage_policy_group`, ...). Five of them do not document `diff` at all; the rest promise it.

## Related

- #304 — output-consistency discussion. The resource-manager module populates `diff` in the cisco.dcnm shape (`{merged, deleted, gathered, debugs}`), so the collection currently has two conventions: dcnm-style vs. empty. **The shape decision belongs to #304**; this issue is about the state-machine layer never producing one at all.
- #389 / PR #525 — the changed/after-on-failure rewrite touches `nd_output.py` and already gates `diff` emission on "final state provable". A fix here should land on top of (or inside) that work rather than collide with it.
- #358 — RETURN documentation audit; whichever shape is chosen, the per-module samples need updating.

## Suggested direction

[...]

Issue #409, opened 2026-07-20
nd_interface_* bulk create: group payloads by (switch, policyType) — ND rejects mixed-policyType interfaces arrays


allenrobel:
## Summary

ND 4.2.1 rejects a bulk interface create (`POST .../switches/{switchId}/interfaces`) whose `interfaces` array mixes `policyType` values. The response is HTTP 207 with a single `results[]` item, `status: "failed"`:

```text
Mixed policy types [iosXeLoopback, csrLoopback] are not allowed in bulk interface creation for interface loopback206[loopback]
```

Nothing in the request is created — not even the interfaces of the "first" policy type. The OpenAPI create schema places no same-type constraint on the `interfaces` array (its items are the full discriminated union), so this is an undocumented runtime constraint. Lab-verified 2026-07-18 on fabric `ISN`, IOS-XE switch WAN1, via a `cisco.nd` integration run. Bug-tracker vault note: `bulk-interface-create-rejects-mixed-policy-types`.

Every Gen-3 interface orchestrator's `create_bulk` batches **per switch only** (`loopback_interface.py` and `ethernet_base.py` both build `groups: dict[switch_id, ...]` and send one POST per switch), so any task that mixes policy types on one switch sends exactly this rejected shape.

## Current behavior vs. after #398

- **Today (develop):** the 207 rejection is silently classified as success — the module reports `changed: true` while creating nothing, and the failure masquerades as an idempotency bug on the next run (#295).
- **After #398 merges:** `NdV1Strategy` detects the failed per-item status and the module fails with ND's error. Correct failure surfacing — but the module still *fails* a task that is perfectly legitimate at the module interface. Users should be able to mix policy types across items in one task.

## Proposed fix

1. In `create_bulk`, group by **(switch_id, policyType)** and send one POST per group. This belongs in the shared `base_interface.py` write path that #348 proposes lifting out of the per-feature orchestrators — do it there rather than patching each copy, since the bug is identical in every module sharing the pattern (loopback, ethernet, and any future adopter of the bulk-create endpoint).
2. Keep the centralized per-item 207 validation from #398 as the backstop for other per-item failures within a now-homogeneous batch.

Request-count impact: POST count only increases when a task actually mixes policy types on one switch (which today produces zero successful requests anyway); homogeneous tasks are unchanged, so this stays within the CLAUDE.md performance rules.

## Lab confirmation — done (2026-07-22)

[...]

PR #558, opened 2026-09-11
nd_interface_ethernet_{access,trunk_host}: add IOS-XE policy types (iosXeAccess, iosXeTrunkHost)

### LINKED ISSUE #534: [ND 4.x] Extend nd_interface_ethernet_access for NX-OS and IOS-XE policy types
## Summary

Extend `nd_interface_ethernet_access` for Nexus Dashboard 4.3 Campus VXLAN EVPN with platform-aware support for NX-OS and IOS-XE interface policy types where ND 4.3 exposes them.

The module must not assume that a policy type supported by one operating system, switch role, model, or topology is valid for another. Applicability must be established from the ND 4.3 contract before implementation.

## Applicability Assessment - Required First

- [ ] Inventory the ND 4.3 policy templates and policy types applicable to this module.
- [ ] Record applicability by fabric type, `network_os_type`, switch role, model/release prerequisite, and topology.
- [ ] Classify each requested combination as `Supported`, `Partially Supported`, or `Not Applicable` with API/template evidence.
- [ ] Reject unsupported combinations before any controller read-modify-write or deployment request.
- [ ] Do not invent an IOS-XE equivalent when ND 4.3 does not expose one.

## Module-Specific Scope

Validate access-interface templates, Catalyst StackWise/StackWise Virtual interface identity, Nexus Ethernet identity, access VLAN behavior, and every platform-specific field exposed by ND 4.3.

## Required Implementation

- Expose and validate `policy_type` and `network_os_type` using platform-aware discriminated models where applicable.
- Keep controller-response parsing tolerant of read-only/defaulted fields while keeping write validation strict.
- Normalize platform-specific identifiers and controller defaults so gathered output is replayable and repeat runs are idempotent.
- Group bulk create/update operations by switch and policy type where the ND API requires homogeneous payloads.
- Preserve predictable `merged`, `replaced`, `overridden`, `deleted`, and `gathered` behavior where those states are supported.
- Preserve check mode, sanitized before/after/diff output, per-item HTTP 207 failure handling, and staged/deploy behavior.
- Update module documentation, examples, return documentation, and the supported policy-type matrix.
- Add focused unit tests and Campus integration tests for Catalyst/IOS-XE, Nexus/NX-OS, mixed valid inputs, and invalid cross-platform inputs.

## Acceptance Criteria

- [ ] Every supported ND 4.3 policy type for this module is documented and validated.
- [ ] Cross-OS, unsupported-role, unsupported-model, and unsupported-topology combinations fail before controller mutation.
- [ ] Supported lifecycle states reconcile correctly and remain idempotent after ND injects defaults or reorders data.
- [ ] Gathered output uses the module argspec shape and is suitable for replay where gathered is supported.
- [ ] Mixed policy types are separated into valid API payloads rather than sent in one incompatible batch.
- [ ] Check mode performs no mutation and predicts the same normalized result as apply mode.
- [ ] Deployment-disabled operations do not call deploy endpoints; a later explicit deployment is validated where staging applies.
- [ ] Unit, sanity, documentation, and Catalyst/Nexus Campus integration validation pass with deterministic cleanup.

## References

- Loopback policy-type baseline: #452
- Shared mixed-policy bulk-operation behavior: #409

### LINKED ISSUE #535: [ND 4.x] Extend nd_interface_ethernet_trunk_host for NX-OS and IOS-XE policy types
## Summary

Extend `nd_interface_ethernet_trunk_host` for Nexus Dashboard 4.3 Campus VXLAN EVPN with platform-aware support for NX-OS and IOS-XE interface policy types where ND 4.3 exposes them.

The module must not assume that a policy type supported by one operating system, switch role, model, or topology is valid for another. Applicability must be established from the ND 4.3 contract before implementation.

## Applicability Assessment - Required First

- [ ] Inventory the ND 4.3 policy templates and policy types applicable to this module.
- [ ] Record applicability by fabric type, `network_os_type`, switch role, model/release prerequisite, and topology.
- [ ] Classify each requested combination as `Supported`, `Partially Supported`, or `Not Applicable` with API/template evidence.
- [ ] Reject unsupported combinations before any controller read-modify-write or deployment request.
- [ ] Do not invent an IOS-XE equivalent when ND 4.3 does not expose one.

## Module-Specific Scope

Validate trunk-host templates, allowed-VLAN normalization, Catalyst StackWise/StackWise Virtual interface identity, Nexus Ethernet identity, and platform-specific trunk policy fields.

## Required Implementation

- Expose and validate `policy_type` and `network_os_type` using platform-aware discriminated models where applicable.
- Keep controller-response parsing tolerant of read-only/defaulted fields while keeping write validation strict.
- Normalize platform-specific identifiers and controller defaults so gathered output is replayable and repeat runs are idempotent.
- Group bulk create/update operations by switch and policy type where the ND API requires homogeneous payloads.
- Preserve predictable `merged`, `replaced`, `overridden`, `deleted`, and `gathered` behavior where those states are supported.
- Preserve check mode, sanitized before/after/diff output, per-item HTTP 207 failure handling, and staged/deploy behavior.
- Update module documentation, examples, return documentation, and the supported policy-type matrix.
- Add focused unit tests and Campus integration tests for Catalyst/IOS-XE, Nexus/NX-OS, mixed valid inputs, and invalid cross-platform inputs.

## Acceptance Criteria

- [ ] Every supported ND 4.3 policy type for this module is documented and validated.
- [ ] Cross-OS, unsupported-role, unsupported-model, and unsupported-topology combinations fail before controller mutation.
- [ ] Supported lifecycle states reconcile correctly and remain idempotent after ND injects defaults or reorders data.
- [ ] Gathered output uses the module argspec shape and is suitable for replay where gathered is supported.
- [ ] Mixed policy types are separated into valid API payloads rather than sent in one incompatible batch.
- [ ] Check mode performs no mutation and predicts the same normalized result as apply mode.
- [ ] Deployment-disabled operations do not call deploy endpoints; a later explicit deployment is validated where staging applies.
- [ ] Unit, sanity, documentation, and Catalyst/Nexus Campus integration validation pass with deterministic cleanup.

## References

- Loopback policy-type baseline: #452
- Shared mixed-policy bulk-operation behavior: #409

allenrobel:
## Related Issue(s)

Closes #534
Closes #535
Refs #409 (ethernet half: `create_bulk` now groups by `(switch_id, policy_type)`)

## Proposed Changes

Extends `nd_interface_ethernet_access` and `nd_interface_ethernet_trunk_host` with an IOS-XE branch on the pattern `nd_interface_ethernet_routed` shipped in #550: a discriminated union on `network_os_type` with per-OS policy models, `policy_type` derived from `network_os_type` when omitted, and write-strict / read-tolerant XE branch models.

**Applicability (ND 4.2.1 + 4.3.1 OpenAPI, verified 2026-09-10):**

| Module | NX-OS policy | IOS-XE policy → template | Fields XE / NX |
|---|---|---|---|
| ethernet_access | `accessHost` | `iosXeAccess` → `iosXeIntAccessHostTemplate` | 9 / 36 |
| ethernet_trunk_host | `trunkHost` | `iosXeTrunkHost` → `iosXeIntTrunkHostTemplate` | 9 / 39 |

The create-side `policyType` enums are identical on 4.2.1 and 4.3.1, so the discriminators need no version branch. The XE models carry the seven fields both releases agree on (`access_vlan` / `allowed_vlans`, `admin_state`, `bpdu_guard`, `description` max 200, `extra_config`, `mtu` int 1500-9216, `speed` XE enum). The 4.3.1-only `deviceTrackingPolicy` / `flowMonitors` are deliberately **not** modeled while both releases are supported (no Gen-3 version detection exists yet); follow-up for the 4.3.1 compatibility work.

**Backward compatibility:** `network_os_type` defaults to `nx-os` in the argspec and is injected on input when absent, so existing playbooks select the NX-OS branch unchanged. Two visible changes for NX-OS users: `policy_type` is now exposed in the argspec and echoed in `before` / `after` output (routed parity; additive), and `mtu` is a plain `str` option with no argspec `choices` because the two branches take different forms (the branch models still enforce `default` / `jumbo` vs 1500-9216).

**Shared infrastructure (`EthernetBaseOrchestrator`):**

- `create_bulk` groups by `(switch_id, policy_type)`, never by switch alone (ND rejects a mixed-policyType `interfaces[]` array; #409). Routed inherits this.
- The IOS-XE delete-side contract is hoisted from the routed orchestrator so it is not copy-pasted a third time: merge-only under `state: overridden` (unnamed XE interfaces dropped at **query** scope), per-interface reset PUT under `state: deleted` tracked in `_pending_xe_resets` and covered by `_unsent_delete_pairs`. The reset target is a class-level `(XE_RESET_MODE, XE_RESET_POLICY_TYPE)`: the host-facing modules reset to a defaults-only `iosXeTrunkHost` in trunk mode (the XE mirror of the NX-OS normalize-to-`trunkHost`); routed keeps its lab-verified `iosXeRoutedHost`.
- The IOS-XE fabric-link ownership guard from #550 (`_fabric_link_endpoints` index fetched once per run + `_check_xe_fabric_link`) is hoisted from the routed orchestrator too (review finding, mikewiebe): a Catalyst fabric-link endpoint reads as a plain `iosXeRoutedHost`, which the policy-type guard alone treats as convertible, so without it an access/trunk task naming an uplink could replace and deploy the link's intent. `_check_fabric_ownership` runs it for IOS-XE targets on create / update / `create_bulk` / `preflight`, and `preflight_delete` / `delete` / `delete_bulk` call it directly (the `_check_xe_delete_guard` hook is gone). The check is a no-op for non-IOS-XE models, so NX-OS-only runs still never fetch the links.
- `default_policy_type` / `normalize_ethernet_interface_name` (+ new `default_network_os_type`) move to `models/interfaces/ethernet_common.py`, shared by access, trunk-host, and routed. Unrecognized interface-name prefixes now pass through verbatim in access/trunk (previously Title-cased), so correctly-typed XE families like `TenGigabitEthernet1/1/1` are never corrupted.
- The trunk-host unconfigured-default filter learns the `iosXeTrunkHost` default signature, so a reset Catalyst port leaves managed scope like an NX-OS one. Interfaces named in the task are exempt from that filter outside `state: deleted` (review finding, mikewiebe, 2026-09-17): a defaults-only desired config of either OS family reads back as existing and converges with `changed: false` instead of being classified as new and re-posted on every run. The NX-OS filter had that defect since #265.
- `NDBaseInterfaceOrchestrator.preflight` refuses a `network_os_type` that disagrees with the `platformType` the switch reports (review finding, mikewiebe, 2026-09-17), in check and normal mode alike, before any interfaceList, links or mutation request. It compares each unique `(switch_ip, network_os_type)` pair once and aggregates every mismatch into one error; `FabricContext` now builds an IP-keyed platform index from the switch-inventory GET it already issues, so no request is added and the lookup is O(1). A switch with no recognizable `platformType` is skipped. Every interface orchestrator on the shared preflight gets the check, including the stacked XE branches.
- `nd_interface_ethernet_access` docs completed (review advisories, 2026-09-17): RETURN block mirroring trunk-host, `replaced` / `overridden` examples, and a note on both modules naming `deviceTrackingPolicy` / `flowMonitors` as intentionally unexposed.

**Queue invariant answers (per CLAUDE.md, for the new `_pending_xe_resets` stage):** a mixed 207 cannot occur (per-interface PUT, one item per request); the finalizer never resubmits (pairs are dequeued only after the PUT is accepted, so anything still queued is unsent); the except-path reader is not used (no response parsing on failure); items are keyed by `(interface_name, switch_id)`.

## Test Notes

- `tests/unit/` full suite: **4816 passed** (`ndpytest tests/unit/`).
- **Review round of 2026-09-17, lab-verified on both controllers the same day** (commits "refuse a network_os_type that disagrees with the switch's platformType before any write" and "keep a named default-signature interface visible so a defaults-only config is idempotent"): `nd_interface_ethernet_trunk_host` full target on ND 4.2.1.10 (SITE1 `S1_BG1`, CAMPUS1 `C1_LE1`) and on ND 4.3.1.175 (this branch + #566 + #563 merged in a throwaway worktree) — **ok=187 changed=47 failed=0 skipped=7 ignored=3 on both**, up from ok=162 with the new `defaults_only.yaml` (NX-OS, after `deleted.yaml`'s cleanup) and the `XE DEFAULTS-ONLY` block at the end of `xe.yaml`: a desired policy made only of template defaults under merged, replaced and overridden, check and normal mode, with the interface asserted present in `before[]` on every run and `changed: false` for merged, overridden and the second replaced pass. Finding from the first attempt: on a port that carried real config before it was normalized, `interfaceActions/normalize` leaves `nativeVlan: 1` and the six `stormControl*Level(Pps)` keys at 0 on the record (both builds; a never-configured port or one last written by the module does not carry them). The default-signature filter tolerates `nativeVlan` 1 by design, so the port is read as existing, but the replaced path's removal check clears those leftovers with one full-payload PUT (`changed: true`, empty `diff`, the #530 reporting gap); from the next run on it is idempotent, which is why replaced is asserted on its second pass. The platform check ran on every task of both runs without adding a request; its refusal paths are unit-tested (access 00900-00940, trunk-host 00900-00920).
- New unit coverage: XE branch model tests (injection, wrong-branch field rejection, ranges, enums, backward-compat injection of `nx-os`, reverse-diff defaults, argspec contract, XE name normalization) for both modules; orchestrator tests for `(switch, policy_type)` grouping, XE reset queue + payload, `remove_pending` flush, overridden query scope, the partial-failure finalizer path, and the trunk XE default-signature truth table. Fabric-link guard on the access/trunk IOS-XE paths (five tests per module): a plain `iosXeRoutedHost` record the links API marks as a fabric-link endpoint is refused before any write in `create_bulk`, `preflight` (check mode), `preflight_delete` (check mode), `delete_bulk`, and a mixed bulk batch (nothing posted, links fetched once). Routed suites unchanged and green after the hoists.
- **Fabric-link guard lab-verified 2026-09-14 on both controllers** (probe playbook naming the ISN core router's `GigabitEthernet3`, the `ebgpVrfLite` VRF-Lite endpoint, in `nd_interface_ethernet_access` and `nd_interface_ethernet_trunk_host` with `network_os_type: ios-xe`): `merged` check mode, `merged` normal mode with deploy on, `replaced`, for both modules — all five refused in preflight with nothing written (interface and link records byte-identical before/after via GET). On 4.2.1.10 WAN1 Gi3 currently carries NO policy on its interface record, so the policy-type guard passes and the links check is the only thing that refuses it (message names `LINK-UUID-122220 (ebgpVrfLite: WAN1 GigabitEthernet3 -> S2_BG1 Ethernet1/3)`); on 4.3.1.175 WAN2 Gi3 carries `userDefined`, so the earlier policy-type guard refuses first. `state: deleted` in check mode is a harmless `changed: false` on both (a routed-typed endpoint is outside the access/trunk managed sets). Negative control: the full access and trunk-host targets re-run on CAMPUS1 (Cat9kv, zero fabric links, the new links GET returns an empty list) — 4.2.1 access ok=116/changed=37, trunk ok=144/changed=46; 4.3.1 (this branch + #566 + #563 merged in a throwaway worktree) access ok=116/changed=37, trunk ok=144/changed=46; failed=0 everywhere, all 26 XE tasks executed per target.
- **Fabric-link guard integration scenario, lab-verified 2026-09-15 on both controllers** (commit "nd_interface_ethernet_*: add the IOS-XE fabric-link guard integration scenario"). `CAMPUS1` now has a Cat9kv spine on each controller (`C1_SP1` / `C3_SP1`) and one `iosXeNumbered` leaf-spine link whose leaf endpoint is `C1_LE1` / `C3_LE1` `GigabitEthernet1/0/8` (`LINK-UUID-226450` on 4.2.1, `LINK-UUID-47050` on 4.3.1). `xe_fabric_link_guard.yaml` in the access, trunk-host and routed targets, gated on `nd_test_xe_fabric_link_switch_ip` / `nd_test_xe_fabric_link_interface_name`, snapshots the link record and the endpoint's interface record, names the endpoint in `merged` (check and normal mode) and `replaced`, asserts each run is refused in preflight with `changed: false`, and asserts both records are byte-identical afterwards. Result on 4.2.1.10 and 4.3.1.175: access, trunk-host and routed each ok=18 changed=0 failed=0 (the three module calls are the expected `ignored=3`). On this link ND stamps the endpoint's interface record with `iosXeNumbered`, so the policy-type check refuses first on both builds and the links check is the second layer; the scenario accepts either refusal and requires the message to name the port and switch. **Delete/reset side cannot be lab-exercised:** an intent-only `PUT` of `iosXeAccess` on the link endpoint (deploy never called) gets HTTP 400 `Policy type change from iosXeNumbered to iosXeAccess is not allowed` on both builds, and 4.3.1 also refuses an unchanged `PUT` of the record as read-only, so the managed-set-plus-link-endpoint state the delete-side guard protects against cannot be manufactured through the interface API; that side stays covered by the unit tests above and the controller's own refusal is the first barrier. Same commit: the routed target's default XE interface moves from `GigabitEthernet3` (the C8000V name) to `GigabitEthernet1/0/1`; the old default had written `iosXeRoutedHost` intent for a port the Catalyst does not have, which broke the `CAMPUS1` deploy until it was deleted.
- `black`, `isort`, `pylint` (only the repo-wide tolerated `C0413` / `W0201` / `C0302` classes remain), `mypy` (no new errors; flagged lines are pre-existing `NDBaseModel` attribute accesses), `ansible-lint`, and `ansible-test sanity` `validate-modules` / `pylint` / `yamllint` on the changed files: clean.
- Integration: `xe.yaml` added for both targets, gated on `nd_test_xe_switch_ip` exactly like routed. **Catalyst 9000v run done 2026-09-11 on ND 4.2.1.10:** a Cat9kv (CML 17.15.03 UADP, `C9KV-UADP-8P`) leaf in a new `vxlanCampus` fabric `CAMPUS1`, test port `GigabitEthernet1/0/1` (fabric default `iosXeTrunkHost`). With the XE scenarios enabled: `nd_interface_ethernet_access` ok=116 changed=37 failed=0; `nd_interface_ethernet_trunk_host` ok=144 changed=46 failed=0; `nd_interface_ethernet_routed` (already on develop) ok=93 failed=0. Every XE step passed on the real switch: baseline converge, merged idempotency, merged update with `mtu`, replaced back to base, overridden merge-only, empty override leaving the XE port alone, and the `deleted` reset to `iosXeTrunkHost` plus its idempotent re-reset. One test-data change fell out of the lab run (commit "lower the XE integration-test MTU to 8000"): the Cat9kv caps port and system MTU at 8978, so the merged-update step now uses 8000 instead of 9000. **ND 4.3.1.175 (C3_LE1 in `CAMPUS1`), this branch with #563 cherry-picked:** the NX-OS portion of `nd_interface_ethernet_access` passes end to end (ok=90), then both `xe.yaml` scenarios stop at `XE BASELINE: Converge` because ND 4.3.1 rejects an `iosXeAccess` / `iosXeTrunkHost` create without `mtu` (`Validation failed for following fields: [mtu]`; 4.2.1 defaults it to 1500). That is the 4.3.1 create-side regression tracked in #564, which now lists the XE types; the fix (always emit the template default) belongs there, not here. #563 is a one-hunk overlap with this branch in `ethernet_base.py` `model_post_init` (keep both lines).
- NX-OS integration targets are unchanged in intent (only the `policy_type` echo and the name-normalizer pass-through are visible NX-side changes). **SITE1 regression run done 2026-09-11 on ND 4.2.1.10 (N9K-C9300v):** `ansible-test network-integration` for both targets at this head (c13cca88) — `nd_interface_ethernet_access` ok=89 changed=30 failed=0 skipped=2; `nd_interface_ethernet_trunk_host` ok=117 changed=39 failed=0 skipped=8. Skips are the gated IOS-XE block and the `supports_vlan_mapping` block (9000v rejects dot1q-tunnel mapping). Every NX-OS before/after item echoes `policy_type` (`accessHost` / `trunkHost`).

4.2.1 and 4.3.1 (OpenAPI only for the IOS-XE branch)

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

https://claude.ai/code/session_01DV9sCsmsDCPXUrPu2GSxz5


plugins/module_utils/orchestrators/ethernet_access_interface.py
```diff
@@ -44,10 +49,11 @@ def _managed_policy_types(self) -> set[str]:
...
-        Return the set of API-side policy type values managed by this orchestrator.
+        Return the set of API-side policy type values managed by this orchestrator: the union of the NX-OS (`accessHost`)
+        and IOS-XE (`iosXeAccess`) access policy types.
 
         ## Raises
 
         None
         """
-        return {e.value for e in AccessHostPolicyTypeEnum}
+        return {e.value for e in AccessHostPolicyTypeEnum} | {e.value for e in XeAccessHostPolicyTypeEnum}
```
mikewiebe, line 59: ### High: Protect IOS-XE fabric-link endpoints in the new access and trunk paths

**Issue**

The access and trunk-host orchestrators now manage IOS-XE policies, but they inherit only the shared policy-type allowlist. That allowlist deliberately treats `iosXeRoutedHost` as user-convertible. On IOS-XE, however, a fabric-link endpoint can also appear in the interface inventory as plain `iosXeRoutedHost`; only the routed orchestrator performs the additional links-API ownership check. The new access and trunk paths therefore allow a named fabric link to reach POST/PUT and deployment.

**Evidence**

- [The access orchestrator adds `iosXeAccess` to its managed policy types without adding the routed orchestrator's IOS-XE fabric-link guard](https://github.com/CiscoDevNet/ansible-nd/pull/558/files#diff-c2cf12bf526cee93d3e38aff74080267af200ddc47df7af413d5394ac2335e69R59).
- [`EthernetBaseOrchestrator.CONVERTIBLE_POLICY_TYPES` includes `iosXeRoutedHost`, so policy type alone permits the conversion](https://github.com/CiscoDevNet/ansible-nd/blob/ebb113cca9a736dcaccb2e51a13eccc89eb35e6f/plugins/module_utils/orchestrators/ethernet_base.py#L116-L140).
- [Only `EthernetRoutedInterfaceOrchestrator` extends `_check_fabric_ownership()` with `_check_xe_fabric_link()` and a paginated fabric-links index](https://github.com/CiscoDevNet/ansible-nd/blob/ebb113cca9a736dcaccb2e51a13eccc89eb35e6f/plugins/module_utils/orchestrators/ethernet_routed_interface.py#L209-L255).
- A focused request-boundary probe supplied an existing IOS-XE interface record with `policyType: iosXeRoutedHost` that was also a fabric-link endpoint. The access orchestrator accepted it, grouped it as `iosXeAccess`, issued the POST, and queued deployment; no links query or ownership rejection occurred.

**Existing PR overlap**

No matching open issue found. [Issue #533](https://github.com/CiscoDevNet/ansible-nd/issues/533) provides related interface-ownership background, but it does not track extending the IOS-XE links-API guard to the access/trunk paths added by this PR.

**Impact**

For example, suppose `GigabitEthernet1/0/48` is an IOS-XE underlay or inter-fabric link. ND's interface record can identify it only as `iosXeRoutedHost`, while the separate links record carries the fabric ownership. If an operator mistypes that port in `nd_interface_ethernet_access` with `network_os_type: ios-xe`, `access_vlan: 100`, and deployment enabled, this branch can replace and deploy the endpoint as `iosXeAccess`, disrupting the fabric link.

**Suggested fix**

Move the routed orchestrator's cached, paginated fabric-link endpoint index and IOS-XE ownership check into `EthernetBaseOrchestrator` (or a shared mixin), and apply it before every IOS-XE access/trunk/routed create, update, and delete/reset. Add access and trunk tests showing that a plain `iosXeRoutedHost` record is rejected before any write when the links API identifies it as a fabric-link endpoint, including check mode and bulk input.

---

allenrobel: Confirmed and fixed in commit "nd_interface_ethernet_*: apply the IOS-XE fabric-link ownership guard to the access and trunk-host paths".

The paginated fabric-link endpoint index (`_fabric_link_endpoints`, fetched once per run) and the IOS-XE ownership check (`_check_xe_fabric_link`) now live in `EthernetBaseOrchestrator`. `_check_fabric_ownership` runs the link check once the policy-type check passes, so it covers `create`, `update`, `create_bulk`, and `preflight` (check mode) for every host-facing module. `preflight_delete`, `delete`, and `delete_bulk` call it directly on the XE path, and the `_check_xe_delete_guard` hook is gone. The routed overrides collapse to nothing and are deleted.

The check is a no-op for non-IOS-XE models, so NX-OS-only runs still never fetch the links; an IOS-XE run adds one links GET per module run, the same as routed already did.

Tests: five new orchestrator tests per module (`test_ethernet_access_orchestrator_00800`-`00840`, `test_ethernet_trunk_host_orchestrator_00800`-`00840`) show a plain `iosXeRoutedHost` record that the links API marks as a fabric-link endpoint is refused before any write in `create_bulk`, `preflight` (check mode), `preflight_delete` (check mode), and `delete_bulk`, plus a mixed bulk batch where nothing is posted and the links are fetched once. Full `tests/unit/` suite: 4682 passed.

plugins/module_utils/models/interfaces/ethernet_access_interface.py
```diff
@@ -180,77 +205,152 @@ class EthernetAccessPolicyModel(StormControlMutexMixin):
...
+        None
+        """
+        return default_policy_type(data, XeAccessHostPolicyTypeEnum.IOS_XE_ACCESS.value)
 
-        ## Raises
+    access_vlan: int | None = Field(default=None, alias="accessVlan", ge=1, le=4094, description="VLAN for this access port")
+    bpdu_guard: BpduGuardEnum | None = Field(default=None, alias="bpduGuard", description="Enable spanning-tree BPDU guard")
+    description: AsciiDescription = Field(default=None, alias="description", min_length=1, max_length=200, description="Interface description")
+    extra_config: str | None = Field(default=None, alias="extraConfig", description="Additional CLI for the interface")
+    mtu: int | None = Field(default=None, alias="mtu", ge=1500, le=9216, description="Interface MTU (1500-9216)")
```
mikewiebe, line 255: ### Advisory: Track the ND 4.3.1 IOS-XE `mtu` requirement in #564

**Issue**

The new IOS-XE policy models leave `mtu` unset by default, so baseline create payloads omit it. The PR's own live results show that ND 4.3.1 rejects both `iosXeAccess` and `iosXeTrunkHost` creates without `mtu`. This is a known cross-version default-emission gap tracked outside this PR, not a request to fix it here.

**Evidence**

- [The new IOS-XE access model declares `mtu` with `default=None`](https://github.com/CiscoDevNet/ansible-nd/pull/558/files#diff-bf25adb755fd81952898d30904f931925c91c029f3c4b2993fee5a33799a5602R255), causing ordinary payload serialization to omit it.
- [The new IOS-XE trunk-host model has the same `default=None` contract](https://github.com/CiscoDevNet/ansible-nd/pull/558/files#diff-aec2c04ae468fb9c9f05bee0765635325601113fce4945ad5c37b32fd8076340R441).
- The PR body reports both ND 4.3.1 integration scenarios failing at `XE BASELINE: Converge` with `Validation failed for following fields: [mtu]`; the work is recorded in issue #564.

**Existing PR overlap**

No matching existing PR comment found. The author already discloses the failure in the PR description; this advisory records the external tracking and temporary user impact.

Tracked by open issue [#564](https://github.com/CiscoDevNet/ansible-nd/issues/564). Issues [#534](https://github.com/CiscoDevNet/ansible-nd/issues/534) and [#535](https://github.com/CiscoDevNet/ansible-nd/issues/535) remain open and contain the broader IOS-XE lifecycle acceptance work.

**Impact**

Until #564 lands, the simplest IOS-XE access or trunk-host configuration succeeds on ND 4.2.1 but requires the user to set `mtu: 1500` explicitly on ND 4.3.1. Otherwise ND rejects the create before applying the policy.

**Suggested fix**

No change is requested in this PR for this advisory. Keep #564 as the source of truth for emitting the version-compatible default, and validate both IOS-XE lifecycle targets on ND 4.3.1 when that fix lands. Keep #534/#535 open until their acceptance criteria are satisfied.

plugins/module_utils/orchestrators/ethernet_access_interface.py
allenrobel: Lab-verified 2026-09-14 on both controllers, details in the PR body's Test Notes. Summary: naming the ISN core router's `GigabitEthernet3` (the `ebgpVrfLite` VRF-Lite endpoint) in `nd_interface_ethernet_access` and `nd_interface_ethernet_trunk_host` with `network_os_type: ios-xe` is refused in preflight for `merged` (check and normal mode, deploy on) and `replaced`, with the interface and link records unchanged. On 4.2.1.10 that port currently carries no policy at all on its interface record, so the links check is the only guard that stops it; on 4.3.1.175 it carries `userDefined` and the policy-type guard refuses first. The full access and trunk-host targets re-run clean on the Cat9kv in CAMPUS1 on both releases (zero fabric links there, so the new links GET is an empty list on the happy path).

allenrobel: Lab result for the second half of your ask (delete/reset side), plus an integration scenario for the create side, in commit "nd_interface_ethernet_*: add the IOS-XE fabric-link guard integration scenario".

**Create side, now a repeatable integration scenario.** The lab gained a Cat9kv spine in `CAMPUS1` on both controllers, so there is an `iosXeNumbered` leaf-spine link whose leaf endpoint (`GigabitEthernet1/0/8`) nothing else depends on. `xe_fabric_link_guard.yaml` in the access, trunk-host and routed targets (gated on `nd_test_xe_fabric_link_switch_ip` / `nd_test_xe_fabric_link_interface_name`) snapshots the link record and the endpoint's interface record, names the endpoint in `merged` (check and normal mode) and `replaced`, asserts every run is refused in preflight with `changed: false`, and asserts both records are byte-identical afterwards. Passed on 4.2.1.10 and 4.3.1.175 for all three modules on 2026-09-15.

One detail worth knowing: on this link ND stamps the endpoint's interface record with `iosXeNumbered`, so the policy-type check refuses first and the links check is the second layer. On the ISN VRF-Lite leg the record carried no policy on 4.2.1 (links check is the only barrier) and `userDefined` on 4.3.1. The scenario therefore accepts either refusal and requires the message to name the port and switch.

**Delete/reset side cannot be lab-exercised, and the reason is ND's.** The reset path only runs for a port already in the module's managed set (record reads `iosXeAccess`, or `iosXeTrunkHost` with a non-default field) that is simultaneously a link endpoint. To manufacture that state we tried an intent-only `PUT` of `iosXeAccess` on the link endpoint, deploy never called:

- 4.2.1 and 4.3.1 both answer HTTP 400: `Policy type change from iosXeNumbered to iosXeAccess is not allowed for interface GigabitEthernet1/0/8.`
- 4.3.1 additionally refuses an unchanged `PUT` of the same record: `Interface [GigabitEthernet1/0/8] is not eligible for edit as it has read only parameters`.
- Interface and link records unchanged, zero pending config on both fabrics after Recalculate.

So the controller itself blocks the policy-type change on a link endpoint, which makes the delete-side check in `preflight_delete` / `delete_bulk` defense in depth rather than the only barrier. That side stays covered by the unit tests (`test_ethernet_{access,trunk_host}_orchestrator_00820` / `00830`: refused before any write in check mode and in `delete_bulk`, links fetched once).

Also in the same commit: the routed target's default XE interface moves from `GigabitEthernet3` (the C8000V name) to `GigabitEthernet1/0/1`. The old default had written `iosXeRoutedHost` intent for a port the Catalyst does not have, which broke the fabric deploy until it was deleted.

plugins/module_utils/models/interfaces/ethernet_access_interface.py
allenrobel: Agreed, and no change here. #564 stays the tracker; the fix is PR #566 ("nd_interface_{ethernet_*,vpc_trunk_host}: always emit the template defaults ND 4.3.1 requires (allowedVlans, mtu)"), stacked on this branch and labeled Ready for Review / 2.0.0. It always emits the template default on the wire when the user set nothing, and since the defaults are identical on both releases there is no version branch.

Both IOS-XE lifecycle targets are already validated on ND 4.3.1 with that fix applied. From #566's Test Notes (verification merge of #566 + #563, 2026-09-14, ND 4.2.1.10 and 4.3.1.175, `CAMPUS1` Cat9kv on both): `nd_interface_ethernet_access` ok=116 failed=0, `nd_interface_ethernet_trunk_host` ok=144 failed=0, `nd_interface_ethernet_routed` ok=93 failed=0, task counts identical on both releases, every `xe.yaml` step included. Before #566 both Catalyst scenarios on 4.3.1 stopped at `XE BASELINE: Converge` with `Validation failed for following fields: [mtu]`, which is the failure this PR's description discloses.

#534 and #535 stay open until their acceptance criteria are met; this PR closes neither.

allenrobel:
Lab observation from the port-channel IOS-XE work (PR #570), for the ethernet side rather than this PR's scope: on both ND 4.2.1.10 and 4.3.1.175 a Catalyst member of a port-channel reads `policyType: iosXeAccessPoMember` / `iosXeTrunkPoMember` with the parent only in `configData.networkOS.policy.portChannelId` (`"Port-channel101"`), while `operData.portChannelId` stays `-1`. `EthernetBaseOrchestrator._existing_port_channel_id` reads the operData field, so `_check_port_channel_restrictions` and the delete-side member guard never fire for XE members. It also lags the removal deploy by about a minute on 4.2.1 (the delete preflight refused a just-detached member once). Probably belongs with the membership work in #561; noting it here so it is not lost.

plugins/modules/nd_interface_ethernet_access.py
```diff
@@ -59,203 +60,266 @@
             - Network OS specific configuration.
             type: dict
             suboptions:
+              network_os_type:
```
mikewiebe, line 63: ### Medium: Check mode accepts a platform mismatch that normal execution rejects

**Issue**

The discriminated models validate `network_os_type` against `policy_type`, but runtime preflight never compares the chosen OS with the target switch's actual `platformType`. Normal execution sends the incompatible request and ND rejects it before persisting intent; check mode skips that mutation request, never sees the controller rejection, and can report the same operation as viable.

**Evidence**

- [The public contract makes `network_os_type` the branch selector](https://github.com/CiscoDevNet/ansible-nd/pull/558/files#diff-d9dedc3c5c695765362c861c47123a3a9b3f0354811540374561ab575ab91513R63).
- [Ethernet preflight resolves the switch and checks ownership/membership but never calls `FabricContext.get_platform_type()`](https://github.com/CiscoDevNet/ansible-nd/pull/558/files#diff-923f5fa5e7f2b2e100e7dc90ee24f3bc89d080b057dcc7a052c35498a462ae06R579-R603).
- The shared capability check is keyed only by interface type/mode and these Ethernet orchestrators deliberately opt out; it is not an OS-binding check.
- A focused exact-head probe confirmed that an IOS-XE desired model aimed at an NX-OS inventory passes this preflight without a platform lookup.
- [Shared interface preflight already resolves every `switch_ip` through `FabricContext.get_switch_id()`](https://github.com/CiscoDevNet/ansible-nd/blob/09dd496b9d37f0a4464db7b73dc5e33b67635630/plugins/module_utils/orchestrators/base_interface.py#L214-L263). That resolution triggers the existing switch-inventory GET; this finding does not require another request.
- [`FabricContext._load_switch_maps()` retains the complete response from that same GET in `_switches` while building the existing ID maps](https://github.com/CiscoDevNet/ansible-nd/blob/09dd496b9d37f0a4464db7b73dc5e33b67635630/plugins/module_utils/fabric_context.py#L240-L272). Each cached record already contains `additionalData.platformType`.
- [`get_platform_type()` currently reuses that cached response but scans the list`](https://github.com/CiscoDevNet/ansible-nd/blob/09dd496b9d37f0a4464db7b73dc5e33b67635630/plugins/module_utils/fabric_context.py#L375-L404), making a per-interface implementation unnecessarily expensive at scale unless the existing load pass also builds a platform index.

**Existing PR overlap**

No matching existing PR comment found. The PR's “cross-platform rejection” tests cover wrong-branch fields, not desired OS versus actual switch platform.

Directly covered by open issues [#534](https://github.com/CiscoDevNet/ansible-nd/issues/534) and [#535](https://github.com/CiscoDevNet/ansible-nd/issues/535), whose acceptance criteria require cross-OS combinations to fail before mutation.

**Impact**

The controller guard keeps normal execution state-safe: the incompatible policy is neither staged nor deployed. The defect is behavioral parity. A CI dry run can approve a playbook that normal execution will reject, and the normal run incurs an avoidable controller round trip before returning a less targeted API error.

For example, assume ND reports switch `10.10.10.21` as `platformType: nx-os`, but the playbook accidentally selects IOS-XE:

```yaml
- name: Stage an access-interface policy
    fabric_name: DC_FABRIC
    config:
      - switch_ip: 10.10.10.21  # Nexus switch running NX-OS
        interface_names:
          - Ethernet1/10
        config_data:
          network_os:
            network_os_type: ios-xe  # Incorrect for this switch
            policy:
              policy_type: iosXeAccess
              admin_state: true
              access_vlan: 100
              mtu: 1500
    config_actions:
      deploy: false
    state: merged
```

Today, `ios-xe` and `iosXeAccess` agree with each other, so model validation passes. With `--check`, the module can report the proposed change as valid because it never issues the request that ND would reject. Without `--check`, ND rejects the POST/PUT before persistence; no incompatible intent is staged or deployed, regardless of `deploy: false`. Both paths should instead return the same early, module-level failure, such as:

```text
Switch 10.10.10.21 reports platformType 'nx-os', but the requested
network_os_type is 'ios-xe'. No changes were made.
```

The validation itself requires no new controller traffic because switch resolution has already loaded the needed `platformType` values. It should only add in-memory indexing and comparison. Calling the current linear-scan `get_platform_type()` for every interface would still cost `O(interfaces × switches)` comparisons—up to roughly 10 million comparisons for 10,000 interfaces in a 1,000-switch fabric—even though the inventory was already fetched.

**Suggested fix**

Reuse—not repeat—the switch-inventory GET already performed for `switch_ip` resolution. While `_load_switch_maps()` processes that existing response, build a cached `{switch_ip: platform_type}` map from each record's `additionalData.platformType`; make `get_platform_type()` read that map in `O(1)` time; then compare each unique `(switch_ip, desired_os)` pair during preflight and aggregate mismatches before any write in normal or check mode.

Do not add a second `EpManageSwitchesListGet`, a new platform endpoint call, or a per-interface GET. Add tests covering both mismatch directions, omission defaulting to NX-OS on a Catalyst, and an interface with no current policy. The scale test should submit multiple interfaces across multiple switches and assert that the entire run still makes exactly the same single switch-inventory GET used for resolution, followed only by in-memory platform lookups—one validation per unique switch/OS pair.

---

plugins/module_utils/orchestrators/ethernet_trunk_host_interface.py
```diff
@@ -100,6 +115,28 @@ def _is_unconfigured_default(iface: dict) -> bool:
             return False
         return True
 
+    @staticmethod
+    def _is_unconfigured_xe_default(policy: dict) -> bool:
```
mikewiebe, line 120: ### Medium: A named defaults-only IOS-XE trunk is recreated on every run

**Issue**

`query_all()` unconditionally removes every `iosXeTrunkHost` record matching the default signature, even when the task explicitly names that interface as desired. The state machine consequently classifies the already-converged interface as new on every run.

**Evidence**

- [The exact default signature treats all default-valued IOS-XE fields as unconfigured](https://github.com/CiscoDevNet/ansible-nd/pull/558/files#diff-24f4c1d42187e77c5d7a104077d7e00fd050ec638abf4a743c9d50b83e27facfR119-R138).
- [The query filter drops every matching record without consulting desired names or state](https://github.com/CiscoDevNet/ansible-nd/pull/558/files#diff-24f4c1d42187e77c5d7a104077d7e00fd050ec638abf4a743c9d50b83e27facfR140-R157).
- A focused exact-head probe built the normal IOS-XE defaults (`adminState=true`, `allowedVlans=none`, `bpduGuard=default`, `mtu=1500`, `speed=auto`) and confirmed the filter classifies it as unconfigured.
- Integration fixtures always add a non-default value, so they do not exercise a second defaults-only run.

**Existing PR overlap**

Tracked by [#535](https://github.com/CiscoDevNet/ansible-nd/issues/535), which requires lifecycle idempotency after ND injects template defaults.

**Impact**

A deliberately defaults-only trunk continually reports changed, reposts intent, and may redeploy despite already matching.

For example, assume `GigabitEthernet1/0/10` on a Catalyst switch should intentionally use only the standard IOS-XE trunk defaults:

```yaml
- name: Stage a default IOS-XE trunk-host policy
  cisco.nd.nd_interface_ethernet_trunk_host:
    fabric_name: CAMPUS_FABRIC
    config:
      - switch_ip: 10.10.20.21
        interface_names:
          - GigabitEthernet1/0/10
        config_data:
          network_os:
            network_os_type: ios-xe
            policy:
              policy_type: iosXeTrunkHost
              admin_state: true
              allowed_vlans: none
              bpdu_guard: default
              mtu: 1500
              speed: auto
    config_actions:
      deploy: false
    state: merged
```

Run the playbook twice:

```bash
ansible-playbook trunk.yml
ansible-playbook trunk.yml
```

After the first run, ND returns approximately this policy:

```yaml
policyType: iosXeTrunkHost
adminState: true
allowedVlans: none
bpduGuard: default
mtu: 1500
speed: auto
description: ""
```

Those values exactly match the module's unconfigured-default signature. On the second run, `query_all()` filters out the existing record, the state machine sees no current interface, and the module posts the same intent again:

```text
ND contains GigabitEthernet1/0/10
        ↓
Policy matches all defaults
        ↓
query_all() filters it out
        ↓
State machine sees no existing interface
        ↓
Module creates/posts it again
```

The expected second-run result is `changed: false`; the current result is `changed: true`. With `deploy: false`, the module repeats the staging request. With `deploy: true`, it can also queue another unnecessary deployment.

The filter must distinguish a default interface omitted from the playbook, which should remain hidden, from a default interface explicitly named in the desired configuration, which must remain visible for idempotency.

**Suggested fix**

Retain a default-valued record when it is explicitly named in an active desired state, while continuing to hide omitted default interfaces and to treat an already-default interface as satisfied for deletion. Add two-run `merged`, `replaced`, and named-`overridden` tests.

---

plugins/modules/nd_interface_ethernet_access.py
```diff
@@ -281,12 +345,15 @@
     - Use O(state=merged) to create new resources and update existing ones as defined in your configuration.
       Resources on ND that are not specified in the configuration will be left unchanged.
     - Use O(state=replaced) to replace the resources specified in the configuration.
-    - Use O(state=overridden) to enforce the configuration as the single source of truth.
-      The resources on ND will be modified to exactly match the configuration.
-      Any resource existing on ND but not present in the configuration will be deleted. Use with extra caution.
-    - Use O(state=deleted) to reset the specified interfaces to their fabric default configuration via the
-      C(interfaceActions/normalize) API. Physical ethernet interfaces cannot be truly deleted from a switch;
-      this operation is the API equivalent of the NX-OS C(default interface) CLI command.
+    - Use O(state=overridden) to enforce the configuration as the single source of truth. Named interfaces are
```
mikewiebe, line 348: ### Advisory: Complete the access module's public documentation

**Issue**

The access module exposes two OS branches and four lifecycle states but has no `replaced`/`overridden` examples and its `RETURN` block is empty.

**Evidence**

- [The access module documents all four states](https://github.com/CiscoDevNet/ansible-nd/pull/558/files#diff-d9dedc3c5c695765362c861c47123a3a9b3f0354811540374561ab575ab91513R348-R359), but [its examples stop without demonstrating `replaced` or `overridden`](https://github.com/CiscoDevNet/ansible-nd/pull/558/files#diff-d9dedc3c5c695765362c861c47123a3a9b3f0354811540374561ab575ab91513R374-R505).
- [Its `RETURN` block is empty](https://github.com/CiscoDevNet/ansible-nd/pull/558/files#diff-d9dedc3c5c695765362c861c47123a3a9b3f0354811540374561ab575ab91513R507-R508), unlike the trunk-host sibling.

**Existing PR overlap**

Tracked by [#534](https://github.com/CiscoDevNet/ansible-nd/issues/534). Issue [#530](https://github.com/CiscoDevNet/ansible-nd/issues/530) separately tracks whether the shared `diff` and `logs` values are populated at runtime.

**Impact**

Generated Ansible documentation does not describe the returned IOS-XE discriminators or demonstrate the two destructive convergence states.

Users therefore cannot determine from the module documentation:

- What `replaced` preserves or removes.
- The scope of `overridden`.
- Which result fields and IOS-XE discriminator values are returned.

**Suggested fix**

Mirror the trunk-host RETURN schema and add concise access-specific examples for `replaced` and `overridden`.

---

plugins/module_utils/models/interfaces/ethernet_access_interface.py
```diff
@@ -180,77 +205,152 @@ class EthernetAccessPolicyModel(StormControlMutexMixin):
         description="Unicast storm control level in packets per second",
     )
 
-    @model_serializer(mode="wrap")
-    def _strip_policy_type_in_config(self, handler, info: SerializationInfo):
+
+class XeEthernetAccessPolicyModel(InterfacePolicyStrictBase):
```
mikewiebe, line 209: ### Advisory: Disclose the ND 4.3.1 fields intentionally unavailable in compatibility mode

**Issue**

The IOS-XE access and trunk-host models intentionally omit the ND 4.3.1-only `deviceTrackingPolicy` and `flowMonitors` properties to retain ND 4.2.1 compatibility.

However, the public module documentation does not tell users that these controller capabilities cannot be configured through the modules.

**Evidence**

- [The access model records the deliberate omission only in source](https://github.com/CiscoDevNet/ansible-nd/pull/558/files#diff-bf25adb755fd81952898d30904f931925c91c029f3c4b2993fee5a33799a5602R209-R216).
- [The trunk-host model records the same compatibility choice](https://github.com/CiscoDevNet/ansible-nd/pull/558/files#diff-aec2c04ae468fb9c9f05bee0765635325601113fce4945ad5c37b32fd8076340R387-R394).

**Existing PR overlap**

No matching open issue found.

**Impact**

Users on ND 4.3.1 can see these properties in the controller but cannot express them through the modules' strict models. Generated Ansible documentation provides no explanation that the fields were intentionally excluded for cross-version compatibility.

**Suggested fix**

Add a module note naming both omitted properties and explaining that the shared ND 4.2.1/4.3.1 write contract intentionally does not expose them.

plugins/modules/nd_interface_ethernet_access.py
allenrobel: Confirmed and fixed in commit "nd_interface_*: refuse a network_os_type that disagrees with the switch's platformType before any write".

`NDBaseInterfaceOrchestrator.preflight` now runs `_check_platform_match` right after switch resolution, so it covers every interface orchestrator that inherits the shared preflight (both ethernet host modules here, and the stacked port-channel, SVI and subinterface XE branches for free). Each unique `(switch_ip, network_os_type)` pair is compared once against `FabricContext.get_platform_type`; every mismatch is aggregated into one `RuntimeError` in both check and normal mode, before any interfaceList, links or mutation request. The message is the one you proposed: `Switch 10.10.10.21 reports platformType 'nx-os', but the requested network_os_type is 'ios-xe' (Ethernet1/10). No changes were made.` A switch whose record carries no recognizable `platformType` is skipped, since there is no evidence of a mismatch to act on.

No request is added. `FabricContext._load_switch_maps` now also builds an IP-keyed platform index from the same switch-inventory GET it already issues for `switch_ip` resolution, and `get_platform_type` reads that map in O(1) instead of scanning the switch list, so the check is in-memory only regardless of fabric size.

Tests (access 00900-00940, trunk-host 00900-00920): both mismatch directions, an item that omits `network_os_type` and so defaults to `nx-os` aimed at a Catalyst, a matching platform on an interface with no current policy, and four IOS-XE targets across two switches (one of them without a `platformType`) asserting the whole run issues exactly one switch-inventory GET followed only by the per-switch interfaceList and the single links GET. Full `tests/unit/` suite: 4816 passed.

plugins/module_utils/orchestrators/ethernet_trunk_host_interface.py
allenrobel: Confirmed and fixed in commit "nd_interface_ethernet_trunk_host: keep a named default-signature interface visible so a defaults-only config is idempotent".

One thing worth stating plainly: the NX-OS `trunkHost` filter has had exactly this behavior since PR #265 (a named defaults-only `trunkHost` was also re-posted every run), and the IOS-XE filter this branch added inherited it. Both families are fixed by the same change, since they share `query_all`.

Interfaces named in the task config are now exempt from the default-signature filter for every state except `deleted`. Under `deleted` a named default interface stays hidden: it is already at the reset target, so the deletion is satisfied and nothing is queued for normalize or the XE reset PUT. Unnamed default interfaces stay hidden as before, so `state: overridden` still produces no delete intent for them and IOS-XE stays merge-only.

Unit tests (trunk-host 00450-00480): merged retains the named NX-OS default (named as `eth1/2`, exercising name normalization) and hides the unnamed one; replaced retains the named IOS-XE default echo; overridden retains the named default of both families while hiding the unnamed NX-OS default (no delete intent) and the unnamed IOS-XE one; deleted still hides a named default of either family.

Integration, both OS families: `defaults_only.yaml` (runs after `deleted.yaml`'s cleanup leaves Ethernet1/41 at the template default) and an `XE DEFAULTS-ONLY` block at the end of `xe.yaml` (after the XE reset) apply a desired policy made only of template defaults under merged, replaced and overridden, each in check and normal mode, and assert the interface is present in `before[]` on every run, `changed: false` for merged and overridden, and `changed: false` on the second replaced pass.

Why replaced is asserted on its second pass, found while running this: on a port that carried configuration before it was normalized, `interfaceActions/normalize` leaves `nativeVlan: 1` and the six `stormControl*Level` / `*LevelPps` keys at 0 on the record (both ND 4.2.1.10 and 4.3.1.175; a port that was never configured, or was last written by the module, does not carry them). The default-signature filter tolerates `nativeVlan` 1 by design, so the interface is correctly read as existing, but the replaced path's removal check sees those keys as device-side values absent from the desired config and issues one full-payload PUT that clears them. That is a real one-time convergence of normalize's leftovers, not a re-create, and it is idempotent from the next run on. merged never sees it (it compares only desired keys) and overridden runs after the PUT.

Lab results: `nd_interface_ethernet_trunk_host` full target on ND 4.2.1.10 and on ND 4.3.1.175 (this branch + #566 + #563 in a throwaway worktree), 2026-09-17: ok=187 changed=47 failed=0 skipped=7 ignored=3 on both (the three ignored are the fabric-link guard's expected refusals), up from ok=162 with the 24 new defaults-only tasks, every one executed. Details in the PR body's Test Notes.

plugins/modules/nd_interface_ethernet_access.py
allenrobel: Done in commit "nd_interface_ethernet_{access,trunk_host}: document RETURN, replaced/overridden examples, and the unexposed ND 4.3.1 fields". The access module's RETURN block now mirrors the trunk-host schema with access samples (`changed`, `output_level`, `before`, `after`, `diff`, `proposed`, `logs`, `msg`), and EXAMPLES gained a `replaced` example and an `overridden` example whose comment states the NX-OS fabric-wide reset scope and the IOS-XE merge-only exception. Both the empty RETURN and the missing examples predate this branch (they are the same on develop), which is why they slipped past the diff. `validate-modules` and `yamllint` pass on both modules.

plugins/module_utils/models/interfaces/ethernet_access_interface.py
allenrobel: Done in the same docs commit. Both modules now carry a note naming `deviceTrackingPolicy` and `flowMonitors` as intentionally unexposed, with the reason: the module writes one policy shape that both ND 4.2.1 and ND 4.3.1 accept, so those properties cannot be configured through it. Phrased as a scope statement rather than an API-behavior note, per the team's rule on keeping ND quirks out of module DOCUMENTATION.

mikewiebe (approved):
Re-reviewed at c9d9263ff4a2fe7f77379ff18dec005b151c6a1c. The prior platform/check-mode, named-default idempotency, and documentation findings are resolved. ND 4.2.1/4.3.1 contract checks and focused verification pass.

mikewiebe (approved):
Re-reviewed current head 3bc748994ca9fdfde5c2ce903c638fa232e6da91. No blocking findings remain.

Issue #553, opened 2026-09-04
[Architecture] Add ND 4.3.1 bulk interface updates with ND 4.2.1 singular compatibility


mikewiebe:
## Summary

Implement a shared bulk interface-update architecture for Nexus Dashboard 4.3.1 while preserving the existing singular interface-update path and behavior for Nexus Dashboard 4.2.1.

This is the implementation architecture following the performance finding documented as IFACE-002 in #533. Related shared interface CRUD consolidation is tracked in #348.

## Version compatibility and normative scope

This architecture is version-gated and must preserve support for both Nexus Dashboard 4.2.1 and Nexus Dashboard 4.3.1.

- **ND 4.2.1:** The bulk interface modification endpoint is not available. All standalone interface modules and `cisco.nd.nd_interfaces_workflow` must continue using the existing singular update endpoint and current update execution logic:

  ```text
  PUT /api/v1/manage/fabrics/{fabricName}/switches/{switchId}/interfaces/{interfaceName}
  ```

- **ND 4.3.1:** Eligible updates should use the bulk interface modification endpoint:

  ```text
  POST /api/v1/manage/fabrics/{fabricName}/interfaceActions/modify
  ```

  Standalone modules should consolidate eligible updates supplied within one module invocation. The aggregator should consolidate compatible updates across interface families and switches within the same fabric and dependency wave.

Controller-version or capability selection must occur before any update request is sent. ND 4.2.1 must never be sent the bulk request. Unsupported, unrecognized, or not-yet-qualified controller versions must conservatively use the existing singular path unless explicitly added to the supported bulk-capability matrix.

The version decision must not require one probe per interface or otherwise negate the discovery and performance benefits of the aggregator. It should use controller version information already available to the module execution context, with robust handling of full build strings such as `4.2.1.x` and `4.3.1.x`.

This compatibility section is normative. It corrects the API-version statement currently recorded in [IFACE-002 of issue #533](https://github.com/CiscoDevNet/ansible-nd/issues/533), which describes the bulk endpoint as available in ND 4.2.1 and later. Issue #533 remains the originating performance finding; this issue defines the version-correct implementation architecture.

## Architecture analysis record — preserved verbatim

Yes. The right architecture is “prepare globally, send once.”

The aggregator should not call each family’s future `update_bulk()` method directly, because that would still produce one bulk request per family. Instead, standalone modules and the aggregator should share a lower-level bulk-update service.

```mermaid
flowchart TD
[...]

Issue #549, opened 2026-09-02
[ND 4.x] Add L3 routed port-channel interface support for NX-OS and IOS-XE policy types


sivakasi-cisco:
## Summary

Add declarative L3 routed port-channel interface support for both NX-OS and IOS-XE policy types, following the cross-platform discriminated-union approach used for routed Ethernet in #447 and loopback policy types in #452.

## Problem

The collection documents that managed and unmanaged subinterfaces require a routed physical or routed port-channel parent. Routed Ethernet is tracked by #447, but there is no equivalent issue covering the routed port-channel policy family.

Campus and mixed Catalyst/Nexus fabrics need a supported way to configure routed port-channel parents without relying on raw API calls or first converting an L2 access/trunk port-channel outside Ansible.

## Applicability Check

Evaluate the exact applicability first for each supported Nexus Dashboard 4.3 fabric type, operating system, switch role, model, release, and interface template. Unsupported combinations must fail before any controller mutation.

## Scope

- Validate the ND 4.3 schemas and API behavior for the NX-OS `l3PortChannel` policy and the corresponding IOS-XE routed port-channel policy type or types.
- Define one user-facing routed port-channel resource shape with policy-type and network-OS discriminators where appropriate.
- Add the required argument specification, models, orchestrator/module handling, documentation, examples, and return data.
- Support merged, replaced, overridden, deleted, and gathered workflows where applicable.
- Keep write validation strict while tolerating controller-added read-only fields.
- Reject cross-OS, unsupported-role, unsupported-model, and unsupported-template combinations before write calls.
- Group bulk operations safely by switch and policy type where the controller requires homogeneous payloads.
- Preserve the shared `config_actions` contract, including staged changes when save/deploy actions are disabled.
- Validate per-item HTTP 207 responses and report partial failures correctly.
- Cover routed port-channel parents used by managed and unmanaged subinterfaces.

## Acceptance Criteria

- [ ] Supported NX-OS and IOS-XE routed port-channel policy types are identified from ND 4.3 schema and live-controller evidence.
- [ ] Supported policy types reconcile correctly through the documented states.
- [ ] Cross-platform and unsupported-capability combinations fail before controller mutation.
- [ ] Repeated create/update/gather/delete workflows are idempotent.
- [ ] Check mode and sanitized before/after diff behave correctly.
- [ ] `config_actions` and staged-configuration behavior match the shared collection contract.
- [ ] HTTP 207 partial failures are surfaced per item.
[...]

PR #561, opened 2026-09-11
nd_interface_ethernet_*: support safe updates to port-channel and vPC members

### LINKED ISSUE #533: [Tracking] Standalone interface-module defects found during nd_interfaces_workflow testing
## Purpose

This issue tracks problems found in individual `cisco.nd` interface modules while testing `cisco.nd.nd_interfaces_workflow`.

The aggregator is the discovery context only. Each finding documents behavior attributable to a standalone interface module or its shared implementation.

### Finding conventions

- Every finding receives a permanent `IFACE-###` identifier.
- Every finding is added to the index.
- Every finding uses visible headings and HTML start/end delimiters.
- Finding identifiers will not be renumbered.
- If a finding is moved into a separate GitHub issue, that issue will be linked from the index while the summary remains here.

## Initial test environment

- Ansible Core: `2.16.3`
- Python: `3.12.13`
- `cisco.nd`: `1.5.0` development checkout
- Tested branch: `interface_aggregator`
- Tested revision: `cdcb8de1488e30a635a93c1cfd9444fed9b14da0`
- Nexus Dashboard: `4.2.1` (`build_version: 4.2.1.10`)
- NDFC: `12.5.0.475`
- NX-OS version: Not relevant to IFACE-001 because the failure occurs during module planning before any controller mutation or deployment request.

Future findings should document any differing environment details inside their respective finding block.

## Finding index

| ID | Module | Summary | Status |
| --- | --- | --- | --- |
| [IFACE-001](#iface-001) | `cisco.nd.nd_interface_ethernet_access` | `state: deleted` silently ignores a named physical Ethernet interface whose current policy is `dot1qTunnelHost` | Open |
| [IFACE-002](#iface-002) | Shared interface update execution | ND 4.3.1 bulk-update opportunity; ND 4.2.1 retains singular PUTs ([architecture #553](https://github.com/CiscoDevNet/ansible-nd/issues/553)) | Open |

---

<a id="iface-001"></a>

#### Module

`cisco.nd.nd_interface_ethernet_access`

#### Summary

When `cisco.nd.nd_interface_ethernet_access` is called with `state: deleted` for an explicitly named physical Ethernet interface whose current `policyType` is `dot1qTunnelHost`, the interface is silently excluded from deletion planning.

The task succeeds with:

```yaml
changed: false
```

The requested interface is absent from both `before` and `after`, and no interface normalization is planned.

This behavior was discovered through interface-aggregator testing and then reproduced directly with `cisco.nd.nd_interface_ethernet_access` in check mode. The problem originates in the standalone module's shared Ethernet orchestration path.

#### Preconditions

- The target is a physical Ethernet interface.
- The target is explicitly identified by `switch_ip` and `interface_names`.
- The target is not a port-channel member.
- The interface is not already at fabric-default configuration.
- Its current controller state includes:

```yaml
interfaceType: ethernet
configData:
  mode: dot1qTunnel
  networkOS:
    policy:
      policyType: dot1qTunnelHost
```

#### Ansible playbook

The addresses and fabric name below are sanitized.

```yaml
---
- name: Reproduce cross-policy Ethernet deletion
  hosts: nd_controller
  gather_facts: false

  tasks:
    - name: Reset an explicitly named physical Ethernet interface
        fabric_name: test_fabric
        config:
          - switch_ip: 192.0.2.10
            interface_names:
              - Ethernet1/40
        config_actions:
          deploy: false
        state: deleted
      check_mode: true
      register: delete_result

    - name: Display the result
      ansible.builtin.debug:
        var: delete_result
```

#### Steps to reproduce

1. Configure a physical Ethernet interface with `policyType: dot1qTunnelHost` through Nexus Dashboard.
2. Confirm that the interface is not a port-channel member.
3. Run the playbook above.
4. Inspect `changed`, `before`, and `after`.
5. Observe that the explicitly requested interface is absent from both state collections and no normalization is planned.

#### Expected behavior

Because the user explicitly named an eligible physical Ethernet interface with `state: deleted`, the module should select it regardless of its current Ethernet policy family.

In check mode, the module should report a prospective change:

```yaml
changed: true
```

During normal execution, it should use the existing normalize/reset implementation to return the interface to the applicable fabric-default physical-interface configuration, currently represented by the `trunkHost` normalization template.

Existing safeguards must remain enforced, including the protection against normalizing port-channel members. A subsequent identical deletion against an interface already at fabric default must be idempotent.

If cross-policy normalization is intentionally outside the module contract, the module should instead return a clear policy-mismatch error. A successful silent no-op should not occur.

#### Actual behavior

The direct standalone check-mode invocation completes successfully with:

```text
changed: false

before:
  # Contains unrelated accessHost interfaces only.
  # The explicitly requested Ethernet1/40 is absent.

after:
  # Identical to before.
  # Ethernet1/40 remains absent.
```

The requested interface is treated as though it does not exist even though it is present in the controller's physical Ethernet inventory.

#### Suspected cause

`EthernetAccessInterfaceOrchestrator._managed_policy_types()` returns only:

```text
accessHost
```

`EthernetBaseOrchestrator.query_all()` then filters the physical Ethernet inventory using that managed-policy set before deletion planning.

Relevant paths:

```text
plugins/module_utils/orchestrators/ethernet_access_interface.py
    EthernetAccessInterfaceOrchestrator._managed_policy_types()

plugins/module_utils/orchestrators/ethernet_base.py
    EthernetBaseOrchestrator.query_all()

plugins/module_utils/nd_state_machine.py
    NDStateMachine._manage_delete_state()
```

The standalone delete logic only selects proposed identifiers found in the filtered current-state collection. Because `dot1qTunnelHost` was removed by `query_all()`, the explicitly requested interface never reaches `delete_bulk()` or `remove_pending()`.

The existing delete execution path already supports normalization once an interface reaches it:

```text
plugins/module_utils/orchestrators/ethernet_base.py
    EthernetBaseOrchestrator.delete_bulk()
    EthernetBaseOrchestrator.remove_pending()
```

#### Suggested implementation scope

Deletion discovery should retain explicitly named, eligible, non-default physical Ethernet interfaces independently of their current policy family.

The existing policy-family filtering should remain unchanged for:

- `state: merged`
- `state: replaced`
- `state: overridden`

A foreign-policy response cannot simply be deserialized unchanged as `EthernetAccessInterfaceModel`, because that model fixes `mode` to `access` and `policyType` to `accessHost`. Identity-only deletion candidates can be used while retaining access to the cached raw interface record for:

- Port-channel membership protection
- Detection of fields requiring the PUT reset path
- Normalization payload construction

#### Acceptance criteria

- An explicitly named, eligible physical Ethernet interface using `dot1qTunnelHost` is not silently filtered from `state: deleted`.
- Check mode reports `changed: true` when normalization is required.
- Normal execution resets the interface to fabric-default configuration.
- Existing port-channel-member protection remains enforced.
- A subsequent identical deletion is idempotent.
- Policy-family filtering remains unchanged for create and update states.
- Standalone-module regression coverage is added.
- Aggregate-workflow regression coverage verifies the inherited behavior.

Recommended regression cases:

1. A named `dot1qTunnelHost` interface becomes a deletion candidate.
2. The candidate produces `changed: true` in check mode.
3. Normal execution resets the interface.
4. A second deletion is idempotent.
5. An interface already at fabric default remains unchanged.
6. A port-channel member remains protected.
7. `merged`, `replaced`, and `overridden` remain scoped to `accessHost`.

## ===== END FINDING IFACE-001 =====

---

---

<a id="iface-002"></a>

#### Module

Shared implementation used by:

- `cisco.nd.nd_interfaces_workflow`
- Standalone `cisco.nd.nd_interface_*` modules that update existing interfaces

#### Classification

Performance enhancement for ND 4.3.1. This is not a defect on ND 4.2.1, where the bulk endpoint is unavailable and the existing singular PUT path is required. This is not currently known to cause incorrect interface configuration.

Implementation architecture: [Issue #553 — Add ND 4.3.1 bulk interface updates with ND 4.2.1 singular compatibility](https://github.com/CiscoDevNet/ansible-nd/issues/553).

#### Summary

Existing-interface updates are currently sent to Nexus Dashboard one interface at a time through the singular update endpoint:

```text
PUT /api/v1/manage/fabrics/{fabricName}/switches/{switchId}/interfaces/{interfaceName}
```

For the releases currently in scope, the bulk interface-modification capability is available only in Nexus Dashboard 4.3.1:

```text
POST /api/v1/manage/fabrics/{fabricName}/interfaceActions/modify
```

It is not available in Nexus Dashboard 4.2.1. ND 4.2.1 must continue using the existing singular PUT endpoint and current update execution logic. Eligible ND 4.3.1 updates can use the bulk endpoint, which accepts between 1 and 10,000 interfaces and can span multiple interface types and switches within the same fabric. It returns HTTP `207 Multi-Status` with a result for each attempted update.

On ND 4.3.1, the current implementation therefore leaves substantial update throughput available, particularly for large `state: merged` and `state: replaced` operations and aggregate tasks containing multiple interface families. On ND 4.2.1, individual PUTs remain the required controller path.

API references:

- [Update Interfaces](https://developer.cisco.com/docs/nexus-dashboard/latest/update-interfaces/)
- [Update Interface](https://developer.cisco.com/docs/nexus-dashboard/latest/update-interface/)

#### Preconditions

- The target controller is Nexus Dashboard 4.3.1 and supports the bulk interface-update API. ND 4.2.1 is outside this bulk-path precondition and remains on the existing singular PUT path.
- A task plans updates to two or more existing interfaces in the same fabric.
- Existing model validation, capability checks, membership protection, structural checks, and policy-transition safety checks have succeeded.
- Interface deployment is independent of this finding. The performance limitation exists with both `deploy: false` and `deploy: true`.

#### Performance evidence

A live calibration on Nexus Dashboard `4.2.1` used the same 50 interface objects on one leaf switch with:

```yaml
state: replaced
config_actions:
  deploy: false
```

Because `interfaceActions/modify` is unavailable on ND 4.2.1, these measurements characterize the required singular path and the aggregator's discovery/shared-context savings. They are not evidence that the ND 4.2.1 implementation failed to use an available bulk endpoint.

The workload covered seven interface families:

- Ethernet access
- Ethernet trunk host
- Port-channel access
- Port-channel trunk host
- Managed subinterface
- Loopback
- SVI

Measured A-to-B task durations were:

| Implementation | Duration |
| --- | ---: |
| `cisco.nd.nd_interfaces_workflow` | 68.190945 seconds |
| Seven standalone ND module tasks | 100.111529 seconds |
| Bulk-capable `cisco.dcnm.dcnm_interface` path | 49.555543 seconds |

The aggregator completed exactly 50 successful mutations with zero deployments. It was 31.885% faster than the seven standalone ND task processes because it shared discovery and orchestration context, but it remained 27.328% slower than the bulk-capable DCNM path.

The DCNM implementation used its separate legacy NDFC bulk-modification endpoint. Its result illustrates the potential performance benefit of request consolidation for the ND 4.3.1 path, but it neither demonstrates bulk-endpoint availability on ND 4.2.1 nor directly measures the ND 4.3.1 `interfaceActions/modify` endpoint.

##### Latest scale calibration after `verify.enabled` — 2026-09-04

A follow-up single measured sample used the same ND 4.2.1 controller, fabric, leaf switch, 50 interface objects, seven interface families, `state: replaced`, and `deploy: false`. The aggregator's new `verify.enabled` option defaults to `false`; it was explicitly set to `false` for the measured aggregator task. Guarded setup, replay, restoration, and independent inventory qualification remained verified.

| Implementation | Measured A-to-B task | Whole playbook |
| --- | ---: | ---: |
| `cisco.nd.nd_interfaces_workflow` (`verify.enabled: false`) | 59.179432 seconds | 190.244188 seconds |
| Seven standalone ND module tasks | 98.303064 seconds | 215.873817 seconds |
| `cisco.dcnm.dcnm_interface` | 44.591481 seconds | 130.469970 seconds |

| Comparison | Seconds saved | Duration reduction | Speedup |
| --- | ---: | ---: | ---: |
| Aggregator versus standalone ND | 39.123632 | 39.799% | 1.661x |
| DCNM versus aggregator | 14.587951 | 24.650% | 1.327x |
| DCNM versus standalone ND | 53.711583 | 54.639% | 2.205x |

Compared with the earlier verified aggregator sample of 68.190945 seconds, the `verify.enabled: false` sample completed in 59.179432 seconds: an observed reduction of 9.011513 seconds (13.215%, 1.152x). Successful-run interface inventory traffic fell from four GETs/pages to two, a 50% reduction.

| Aggregator metric | Earlier verified sample | `verify.enabled: false` sample |
| --- | ---: | ---: |
| Interface inventory GETs/pages | 4 / 4 | 2 / 2 |
| Interface inventory refreshes | 1 | 0 |
| Interface inventory dirty refetches | 1 | 0 |
| Interface inventory cache hits | 29 | 22 |

The latest aggregator result reported `mutation_count: 50`, `mutations_sent: 50`, `deployments_sent: 0`, `interface_summary_gets: 0`, and `vpc_pair_gets: 0`. All seven resource results correctly reported `after_verified: false`. The standalone and DCNM rows do not consume the aggregator's `verify.enabled` option; they are the comparison samples from the same follow-up series.

This remains ND 4.2.1 singular-path evidence. It demonstrates the aggregator's shared-discovery benefit and the additional savings from making successful post-write verification optional. It does not exercise or measure the ND 4.3.1 `interfaceActions/modify` bulk endpoint.

Both datasets are single calibration samples rather than warmup-plus-three-run statistical medians. HTTP capture was disabled for timing integrity, so exact standalone and DCNM wire-request totals were not independently measured. The DCNM comparison retained its known 15-replacement replay drift. Source inspection independently confirms that the current ND update paths iterate once per interface.

#### Expected behavior

On ND 4.3.1, compatible interface updates within the same fabric should be consolidated into the minimum safe number of bulk requests. On ND 4.2.1, the existing singular PUT path and behavior must remain unchanged.

For an eligible group of no more than 10,000 independent updates, the normal target should be one bulk `POST` instead of one singular `PUT` per interface. Requests should be separated only where required by controller limits, dependency ordering, response correlation, or interface safety rules.

Check mode, idempotency, transition reporting, deployment queuing, and per-interface error visibility must remain unchanged.

#### Actual behavior

Standalone interface modules execute updates individually:

```text
plugins/module_utils/nd_state_machine.py
    NDStateMachine._manage_create_update_state()
```

That method loops over `items_to_update` and invokes `model_orchestrator.update()` separately for each interface.

The aggregate workflow also executes ordinary updates and policy transitions individually:

```text
plugins/module_utils/interface_workflow_executor.py
    InterfaceWorkflowExecutor._execute_updates()
    InterfaceWorkflowExecutor._execute_transitions()
```

The family orchestrators then send each payload through the singular interface update endpoint.

Bulk create, delete/reset, normalization, and deployment paths already exist in portions of the interface implementation, but there is no corresponding shared bulk-update path for existing-interface changes. That is the required behavior on ND 4.2.1; the missing optimization applies to eligible ND 4.3.1 updates.

#### Suggested implementation scope

- Implement the version-gated shared bulk-update architecture tracked in [issue #553](https://github.com/CiscoDevNet/ansible-nd/issues/553), using the following endpoint only on ND 4.3.1:

  ```text
  POST /api/v1/manage/fabrics/{fabricName}/interfaceActions/modify
  ```

- Select the controller update strategy once per execution context before any update request is sent; do not probe once per interface.
- Preserve the existing singular-update implementation as the required ND 4.2.1 path and as the conservative fallback for unknown or unqualified controller versions.

- Allow the aggregate workflow to combine compatible updates across interface families and switches within the same fabric.
- Allow standalone modules to bulk-update multiple interfaces supplied in one task.
- Preserve dependency boundaries for port-channel members, vPC peers, parent/child interfaces, and policy transitions.
- Retain all existing validation and safety checks before constructing a bulk request.
- Enforce the documented 10,000-interface request limit and split larger plans deterministically.
- Parse HTTP `207` responses and require an explicit successful outcome for every requested interface.
- Report partial failures with enough switch and interface identity to identify every failed item. HTTP `207` alone must not be treated as overall success.
- Live-qualify response correlation when identical interface names occur on different switches.
- On ND 4.3.1, preserve a singular-update fallback for interface combinations that have not been qualified for bulk modification.
- Extend request statistics so regression tests can verify request consolidation directly rather than relying only on elapsed time.

#### Acceptance criteria

- ND 4.2.1 never calls `interfaceActions/modify` and retains the existing singular PUT payloads, execution behavior, state semantics, idempotency, safety checks, deployment behavior, and optional verification behavior.
- Eligible ND 4.3.1 updates use the version-gated bulk architecture tracked in [issue #553](https://github.com/CiscoDevNet/ansible-nd/issues/553).
- Controller version/capability selection occurs once per execution context and before the first update write; unknown or unqualified versions select the singular path.
- Two or more eligible ND 4.3.1 updates in one task use the bulk update endpoint.
- An eligible ND 4.3.1 group of up to 10,000 independent updates is not sent as one singular PUT per interface.
- Requests larger than 10,000 interfaces are split into valid, deterministic chunks.
- Every interface family supported by the standalone modules and aggregate workflow is regression-tested.
- Supported cross-policy transitions are live-qualified through the bulk path.
- Existing port-channel, vPC, parent/child, structural-collision, and capability protections remain enforced.
- HTTP `207` success and mixed success/failure responses are handled and reported per interface.
- Duplicate interface names on different switches are correlated without ambiguity.
- Check mode performs no mutations.
- `deploy: false` performs no deployment.
- `deploy: true` preserves the existing targeted deployment behavior.
- A subsequent identical task is idempotent.
- Standalone-module and aggregate-workflow integration coverage verifies the same resulting controller intent as the singular PUT path.
- Unit and integration coverage explicitly exercises ND 4.2.1 singular routing and ND 4.3.1 bulk routing.
- A 50-interface ND 4.3.1 request-count regression test demonstrates a material reduction from approximately 50 update requests to the minimum safe number of bulk requests.

## ===== END FINDING IFACE-002 =====

---

mikewiebe:
## Related Issue(s)

Implements [IFACE-004 in #533](https://github.com/CiscoDevNet/ansible-nd/issues/533#iface-004).

This PR addresses the standalone-module implementation tracked by IFACE-004. It does not close #533 because that issue tracks multiple independent interface findings.

## Proposed Changes

Enable the standalone Ethernet access, trunk-host, and routed modules to recognize and safely update authentic port-channel and vPC member policies.

### Supported member policies

| Module | Supported existing member policies |
| --- | --- |
| `cisco.nd.nd_interface_ethernet_access` | `accessPoMember`, `accessVpcPoMember` |
| `cisco.nd.nd_interface_ethernet_trunk_host` | `poMember`, `vpcMember` |
| `cisco.nd.nd_interface_ethernet_routed` | `l3PoMember`, `iosXeL3PoMember` |

Member interfaces have the following deliberately restricted contract:

- Only explicitly targeted members using `state: merged` are eligible.
- Only `admin_state`, `description`, and `extra_config` can be changed.
- The update starts from the authentic controller member record and preserves its exact policy type, parent identity, port-channel or vPC membership, and all other modeled configuration.
- Existing members use the singular per-interface `PUT`; they can never fall back to a host-policy `POST`.
- `state: replaced` and `state: deleted` reject explicit members. `state: overridden` leaves omitted members outside the host-policy scope and rejects explicitly named members.
- Check mode reports the planned safe-field update without issuing a write.

### Ownership and safety validation

- Build a shared membership index from cached switch inventories and require one compatible parent for ordinary port-channel members.
- Validate configured and operational port-channel identifiers, parent policy, mode, network OS, parent member lists, duplicate claims, orphaned members, and wrong-family or fabric-owned policies.
- For vPC members, require reciprocal peer identity, consistent literal parent records on both peers, exact member ownership, and an unambiguous mapping of peer member slots to switches.
- Reject `extra_config` commands, including common abbreviations, that could change channel-group membership, change interface context, default an interface, or escape configuration context.
- Permit and discard only qualified response-only `ptp` and access-member `portMode` echoes. Reject any other unrecognized nested member configuration before constructing the full `PUT` payload.
- Queue deployment only after a confirmed successful write, while retaining previously accepted deployment intent if a later operation fails.

### Scale behavior

- Reuse one in-memory membership index for the normal state-machine batch.
- Reuse cached interface inventory and fetch each unique vPC peer inventory at most once.
- Add no discovery `GET` per member.
- Issue zero `PUT` requests for idempotent members.
- Retain one singular `PUT` per changed member on ND 4.2.1 and ND 4.3.1. Version-gated bulk interface updates remain separate work in #553.

## Test Notes

Local validation was completed after rebasing on `develop` commit `c6369920`:

- `black --check --line-length 159` passed for all 17 changed Python files.
- `git diff --check` passed.
- Ethernet member model unit suite: **86 passed**.
- Focused IFACE-004 unit suite: **406 passed**.
- Complete unit suite: **4,893 passed**.
- Targeted Pylint sanity for `ethernet_member_interface.py` passed with the actual `ansible-core` stable-2.18 (`2.18.19`) and stable-2.19 (`2.19.13`) test environments using Docker target Python 3.13.
- Complete `ansible-test sanity --docker --python 3.13` passed with both `ansible-core` stable-2.18 (`2.18.19`) and stable-2.19 (`2.19.13`).
- The CI compatibility correction replaces the PEP 604 runtime model union with `typing.Union`, preserving all five supported member-policy model variants while avoiding the Pylint `unsupported-binary-operation` failure.
- GitHub Actions completed on head `24cb0db1`: **56 checks passed**, one intentionally disabled Integration Tests job was skipped, and there were zero failures or cancellations.
- Added guarded integration coverage for standalone access and trunk members, pair-aware access and trunk vPC members, and NX-OS and IOS XE routed members. The scenarios verify exact baselines, preserve non-safe fields and membership, exercise idempotence, and restore state after attempted mutation.

Live Nexus Dashboard qualification of this implementation has not yet been run. The guarded live integration scenarios remain opt-in.

IFACE-004 was identified while testing Nexus Dashboard 4.2.1 (`build_version: 4.2.1.10`). This implementation preserves the existing singular interface-update path for ND 4.2.1 and ND 4.3.1. Live qualification of the new member-update behavior remains pending on both versions.

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

### DIFF
--- a/changelogs/fragments/iface-004-port-channel-member-updates.yml
+++ b/changelogs/fragments/iface-004-port-channel-member-updates.yml
@@ -1,0 +1,18 @@
+---
+# Copyright: (c) 2026, Mike Wiebe (@mikewiebe) mwiebe@cisco.com
+
+bugfixes:
+  - >-
+    nd_interface_ethernet_access, nd_interface_ethernet_trunk_host, and
+    nd_interface_ethernet_routed - recognize authentic accessPoMember,
+    accessVpcPoMember, poMember, vpcMember, l3PoMember, and iosXeL3PoMember
+    controller policies and safely update admin_state, description, and
+    extra_config with state=merged while preserving port-channel or vPC
+    membership. Member validation rejects abbreviated membership or
+    interface-context commands and requires a compatible parent policy, mode,
+    and network OS. vPC validation compares the literal shared peer
+    configuration returned in both parent echoes, validates both switches from
+    cached inventories, and uses one reusable pair proof without a GET per
+    member. Unrecognized nested member configuration fails closed instead of
+    being silently omitted from the full PUT
+    (https://github.com/CiscoDevNet/ansible-nd/issues/533#iface-004).
--- a/plugins/module_utils/interface_membership.py
+++ b/plugins/module_utils/interface_membership.py
@@ -1,0 +1,1074 @@
+# Copyright: (c) 2026, Mike Wiebe (@mikewiebe) mwiebe@cisco.com
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+"""Controller-independent ethernet membership ownership validation.
+
+``EthernetMembershipIndex`` consumes interface inventories that have already been
+fetched and cached by an orchestrator.  It never sends a REST request.  This keeps
+membership validation reusable by both the standalone ethernet orchestrators and
+the interface aggregator without adding a per-interface discovery cost.
+
+The index deliberately fails closed.  A member update is safe only when its exact
+member-policy descriptor agrees with one unambiguous parent, the configured and
+operational port-channel identities do not conflict, and every required parent
+membership reference is internally consistent.  vPC members additionally require
+reciprocal evidence from both peers' cached inventories.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Mapping, Sequence
+from dataclasses import dataclass
+from types import MappingProxyType
+from typing import Any, TypeAlias
+
+from ansible_collections.cisco.nd.plugins.module_utils.models.interfaces.ethernet_member_interface import (
+    MemberPolicyDescriptor,
+    MemberPolicyDisposition,
+    classify_member_policy,
+    get_member_policy_descriptor,
+    normalize_port_channel_id,
+    parse_member_interface_response,
+    policy_type_from_interface_record,
+)
+
+MemberKey: TypeAlias = tuple[str, str]
+InterfaceRecord: TypeAlias = Mapping[str, Any]
+SwitchInventory: TypeAlias = Mapping[str, InterfaceRecord]
+InterfaceInventories: TypeAlias = Mapping[str, SwitchInventory]
+
+VpcParentSignature: TypeAlias = tuple[
+    str,
+    str,
+    str,
+    tuple[Any, ...],
+]
+_LOCAL_PARENT_MEMBER_FIELDS = (
+    "ports",
+    "peer1MemberPorts",
+    "peer2MemberPorts",
+)
+
+
+class MembershipValidationError(ValueError):
+    """Raised when cached inventory cannot prove one safe member owner."""
+
+
+class MissingPeerInventoryError(MembershipValidationError):
+    """Raised when pair-aware validation needs a peer inventory not yet cached."""
+
+    def __init__(self, peer_switch_id: str) -> None:
+        self.peer_switch_id = peer_switch_id
+        super().__init__(f"Pair-aware membership validation requires cached interface inventory " f"for peer switch {peer_switch_id!r}")
+
+
+class MissingPeerIdentityError(MembershipValidationError):
+    """Raised when a vPC parent omits its peer and cached evidence is insufficient."""
+
+    def __init__(self, switch_id: str, parent_name: str) -> None:
+        self.switch_id = switch_id
+        self.parent_name = parent_name
+        super().__init__(
+            f"Pair-aware membership validation cannot resolve the peer for vPC "
+            f"parent {parent_name!r} on switch {switch_id!r} from cached inventory "
+            "or pair evidence"
+        )
+
+
+@dataclass(frozen=True)
+class IndexedEthernetMember:
+    """One member-policy interface found in the supplied inventories."""
+
+    switch_id: str
+    interface_name: str
+    policy_type: str
+    disposition: MemberPolicyDisposition
+    descriptor: MemberPolicyDescriptor | None
+    record: InterfaceRecord
+
+    @property
+    def key(self) -> MemberKey:
+        """Return the canonical index key."""
+
+        return self.switch_id, self.interface_name.lower()
+
+
+@dataclass(frozen=True)
+class MemberOwner:
+    """Validated parent metadata for one member on one switch."""
+
+    switch_id: str
+    interface_name: str
+    interface_type: str
+    policy_type: str
+    port_channel_id: int
+    claim_fields: tuple[str, ...]
+    record: InterfaceRecord
+
+
+@dataclass(frozen=True)
+class ValidatedMemberOwnership:
+    """Proof that a member's cached ownership is safe for an in-place update."""
+
+    member: IndexedEthernetMember
+    owner: MemberOwner
+    configured_port_channel_id: int
+    operational_port_channel_id: int | None
+    peer_owner: MemberOwner | None = None
+    peer_members: tuple[IndexedEthernetMember, ...] = ()
+    pair_validated: bool = False
+
+
+@dataclass(frozen=True)
+class _ParentClaim:
+    """Internal representation of a parent record that names a member."""
+
+    switch_id: str
+    interface_name: str
+    interface_type: str
+    policy_type: str | None
+    claim_fields: tuple[str, ...]
+    record: InterfaceRecord
+
+
+@dataclass(frozen=True)
+class _VpcParentCopy:
+    """One valid, literal vPC parent signature used for O(1) peer inference."""
+
+    switch_id: str
+    signature: VpcParentSignature
+    record: InterfaceRecord
+
+
+@dataclass(frozen=True)
+class _VpcParentSlot:
+    """One literal peer slot from a vPC parent's pair-wide configuration."""
+
+    number: int
+    port_channel_id: int
+    member_names: frozenset[str]
+    fingerprint: tuple[Any, ...]
+
+    @property
+    def member_field(self) -> str:
+        """Return the ND policy key that names this slot's members."""
+
+        return f"peer{self.number}MemberPorts"
+
+
+@dataclass(frozen=True)
+class _ValidatedVpcSlotMember:
+    """One vPC member whose configured, operational, and parent evidence agrees."""
+
+    member: IndexedEthernetMember
+    configured_port_channel_id: int
+    operational_port_channel_id: int | None
+
+
+@dataclass(frozen=True)
+class _ValidatedVpcParentPair:
+    """Cached pair-wide ownership proof for every member of one vPC parent."""
+
+    results: Mapping[MemberKey, ValidatedMemberOwnership]
+
+
+class EthernetMembershipIndex:
+    """Index and validate ethernet members from already-cached inventories.
+
+    ``inventories`` has the same shape as the interface orchestrator cache:
+    ``{switch_id: {lower_interface_name: raw_interface_record}}``.  The mapping
+    keys are treated as lookup hints; canonical keys come from each record's
+    ``interfaceName`` and the enclosing switch serial.
+    """
+
+    def __init__(
+        self,
+        inventories: InterfaceInventories,
+        *,
+        peer_switch_ids: Mapping[str, str] | None = None,
+    ) -> None:
+        self._inventories: dict[str, dict[str, InterfaceRecord]] = {}
+        self._members: dict[MemberKey, IndexedEthernetMember] = {}
+        self._vpc_members_by_parent: dict[MemberKey, tuple[IndexedEthernetMember, ...]] = {}
+        self._claims: dict[MemberKey, tuple[_ParentClaim, ...]] = {}
+        self._peer_switch_ids = self._normalize_peer_switch_ids(peer_switch_ids)
+        self._vpc_parents_by_signature: dict[VpcParentSignature, tuple[_VpcParentCopy, ...]] = {}
+        self._vpc_peer_cache: dict[tuple[str, str], str] = {}
+        self._vpc_pair_validation_cache: dict[tuple[tuple[str, str], str, str], _ValidatedVpcParentPair] = {}
+        self._validation_cache: dict[MemberKey, ValidatedMemberOwnership] = {}
+        self._build(inventories)
+
+    @property
+    def members(self) -> Mapping[MemberKey, IndexedEthernetMember]:
+        """Return the indexed member records keyed by serial and lower-case name."""
+
+        return MappingProxyType(self._members)
+
+    @property
+    def member_keys(self) -> tuple[MemberKey, ...]:
+        """Return canonical member keys in deterministic order."""
+
+        return tuple(sorted(self._members))
+
+    def get_member(self, switch_id: str, interface_name: str) -> IndexedEthernetMember | None:
+        """Return one indexed member, or ``None`` for a non-member/missing record."""
+
+        return self._members.get(self._key(switch_id, interface_name))
+
+    def require_member(self, switch_id: str, interface_name: str) -> IndexedEthernetMember:
+        """Return one indexed member or raise a closed ownership error."""
+
+        member = self.get_member(switch_id, interface_name)
+        if member is None:
+            raise MembershipValidationError(f"Interface {interface_name!r} on switch {switch_id!r} is not an " "indexed ethernet member")
+        return member
+
+    def validate(self, switch_id: str, interface_name: str) -> ValidatedMemberOwnership:
+        """Validate and return ownership proof for one indexed member.
+
+        Standalone port-channel membership requires only the target switch's
+        inventory.  A vPC member requires both peer inventories and raises
+        ``MissingPeerInventoryError`` with the exact missing serial when the
+        orchestrator must populate another cached inventory before retrying.
+        """
+
+        key = self._key(switch_id, interface_name)
+        cached = self._validation_cache.get(key)
+        if cached is not None:
+            return cached
+
+        member = self.require_member(switch_id, interface_name)
+        descriptor = member.descriptor
+        if descriptor is None:
+            raise MembershipValidationError(
+                f"Interface {interface_name!r} on switch {switch_id!r} uses " f"protected or unsupported member policy {member.policy_type!r}"
+            )
+
+        if descriptor.pair_aware:
+            claim = self._require_vpc_parent_claim(member)
+            result = self._validate_vpc_pair(member=member, claim=claim)
+        else:
+            configured_id, operational_id = self._member_port_channel_ids(member)
+            claim = self._require_single_claim(member)
+            owner = self._validate_parent_claim(
+                member,
+                claim,
+                configured_id,
+                local_member_field=descriptor.parent_member_fields[0],
+            )
+            result = ValidatedMemberOwnership(
+                member=member,
+                owner=owner,
+                configured_port_channel_id=configured_id,
+                operational_port_channel_id=operational_id,
+            )
+        self._validation_cache[key] = result
+        return result
+
+    def _build(self, inventories: InterfaceInventories) -> None:
+        if not isinstance(inventories, Mapping):
+            raise TypeError("inventories must be a mapping keyed by switch serial")
+
+        for switch_id, inventory in inventories.items():
+            if not isinstance(switch_id, str) or not switch_id:
+                raise MembershipValidationError(f"Inventory switch identifier must be a non-empty string; got {switch_id!r}")
+            if not isinstance(inventory, Mapping):
+                raise TypeError(f"Inventory for switch {switch_id!r} must be a mapping keyed by interface name")
+
+            normalized_inventory: dict[str, InterfaceRecord] = {}
+            for record in inventory.values():
+                if not isinstance(record, Mapping):
+                    raise TypeError(f"Interface inventory entry for switch {switch_id!r} must be a mapping")
+                interface_name = record.get("interfaceName")
+                if not isinstance(interface_name, str) or not interface_name:
+                    raise MembershipValidationError(f"Interface inventory entry for switch {switch_id!r} lacks a valid interfaceName")
+                record_switch_id = record.get("switchId")
+                if record_switch_id is not None and record_switch_id != switch_id:
+                    raise MembershipValidationError(
+                        f"Interface {interface_name!r} is stored under switch {switch_id!r} " f"but its record declares switchId {record_switch_id!r}"
+                    )
+
+                normalized_name = interface_name.lower()
+                if normalized_name in normalized_inventory:
+                    raise MembershipValidationError(
+                        f"Inventory for switch {switch_id!r} contains duplicate interface " f"name {interface_name!r} after case normalization"
+                    )
+                normalized_inventory[normalized_name] = record
+
+                policy_type = policy_type_from_interface_record(record)
+                disposition = classify_member_policy(policy_type)
+                if disposition == MemberPolicyDisposition.NOT_MEMBER:
+                    continue
+                if not isinstance(policy_type, str) or not policy_type:
+                    raise MembershipValidationError(
+                        f"Interface {interface_name!r} on switch {switch_id!r} " "was classified as an ethernet member without a valid policyType"
+                    )
+                member = IndexedEthernetMember(
+                    switch_id=switch_id,
+                    interface_name=interface_name,
+                    policy_type=policy_type,
+                    disposition=disposition,
+                    descriptor=get_member_policy_descriptor(policy_type),
+                    record=record,
+                )
+                if member.key in self._members:
+                    raise MembershipValidationError(f"Duplicate member index key {member.key!r}")
+                self._members[member.key] = member
+            self._inventories[switch_id] = normalized_inventory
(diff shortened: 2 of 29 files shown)


allenrobel:
Two heads-ups for when this leaves draft, neither is a review finding:

**Conflict with #558 (`feat_534_535_ethernet_iosxe`, Ready for Review).** A `git merge-tree --write-tree` of this branch against #558 (head c13cca88, also on develop c6369920) reports content conflicts in 8 files:

- `plugins/module_utils/orchestrators/ethernet_base.py`
- `plugins/module_utils/orchestrators/ethernet_access_interface.py`
- `plugins/module_utils/orchestrators/ethernet_trunk_host_interface.py`
- `plugins/module_utils/orchestrators/ethernet_routed_interface.py`
- `plugins/modules/nd_interface_ethernet_access.py`
- `plugins/modules/nd_interface_ethernet_trunk_host.py`
- `tests/integration/targets/nd_interface_ethernet_access/tasks/main.yaml`
- `tests/integration/targets/nd_interface_ethernet_trunk_host/tasks/main.yaml`

#558 moved the IOS-XE delete/reset contract and `(switch_id, policy_type)` create grouping into `EthernetBaseOrchestrator`, which is the same region this PR extends. Since #558 is smaller and already in review, my suggestion is to land it first and rebase this one on top; happy to help with that rebase if useful.

**`iosXeL3PoMember` overlap.** The IOS-XE L3 port-channel parent (`iosXeL3PortChannel`) is scoped under #549, and the port-channel access/trunk-host parents under #536/#537. The member descriptors this PR introduces for `iosXeL3PoMember` (and the NX-OS `poMember`/`accessPoMember`/`l3PoMember` family) are the natural shared source of truth for those parent modules' member handling, so I would rather build on `interface_membership.py` than grow a second registry. Flagging so we keep one descriptor table.

allenrobel:
Lab evidence for this PR from the #549 work (PR #577), plus a refresh of my earlier heads-up. None of it is a review finding.

**A real trigger for the inconsistent-evidence guard.** While cleaning up after an IOS-XE routed port-channel removal on ND 4.3.1.175, `nd_interface_ethernet_trunk_host` (`state: replaced`) failed on the released members with a raw validation error instead of a member message:

```text
Preflight failed: 1 validation error for XeEthernetTrunkHostPolicyModel
policyType
  Input should be 'iosXeTrunkHost' [type=literal_error, input_value='iosXeRoutedHost', input_type=str]
  ...
  File "plugins/module_utils/orchestrators/ethernet_base.py", in _check_port_channel_restrictions
    existing_model = type(policy).from_response(existing_policy)
```

Cause: the removal had been deployed (`deploymentHistory` shows `no channel-group 110 mode active` and `no interface Port-channel110`, all `success`, and the Catalyst confirmed it over SSH) and the members' intent already read `iosXeRoutedHost`, but `operData.portChannelId` was still positive. `_existing_port_channel_id` therefore reported membership, and the existing policy was parsed with the proposed policy's class. The same line is on develop, where `policy_type` is a single-member enum, so it should fail the same way there (read from the code, not reproduced on develop).

This PR already fixes it: the `NOT_MEMBER` branch you added ahead of that parse ("has operational port-channel membership N, but its configured policy is 'X'. Refusing a host-policy write because membership evidence is inconsistent") is exactly this case. So I am not patching it separately; it lands with this PR.

One suggestion for that message: in this case nothing needs reconciling, the evidence is inconsistent only because `operData` lags the removal deploy. It is intermittent: on 4.3.1.175 it outlasted 8 retries at 15 s on one run, and on the next run an explicit poll found the members unbundled immediately; on 4.2.1.10 the member cleanup was never refused in five runs. A hint that the condition can be transient after a port-channel removal ("retry once ND reports the interface as unbundled") would save the next person the investigation. Vault note: `xe-port-channel-removal-inventory-lag`.

**The premise holds on real data.** Every deployed member in the lab carries a member policy type, never a host type: 9 members per controller on both 4.2.1.10 and 4.3.1.175 (`vpcPeerLinkMember`, `vpcUplinkMember`, `uplinkPoMember`, `l3PoMember`), each with `policy.portChannelId` set and a matching positive `operData.portChannelId`. The existing orchestrator fixtures give members `accessHost` with a positive `operData.portChannelId`, a shape I could not find on either controller. On IOS-XE, joined members read `iosXeAccessPoMember` / `iosXeTrunkPoMember` (2026-09-15) and `iosXeL3PoMember` (2026-09-18), all with `policy.portChannelId: Port-channel<N>`; for the access and trunk members `operData.portChannelId` stayed `-1` while joined (observed 2026-09-15, not re-checked for the routed type), so on IOS-XE the configured id is the only reliable one.

**Descriptor table.** Following up on my 2026-09-11 note about keeping one registry: #577 adds a single `iosXeL3PortChannel` row to the existing `PortChannelBaseOrchestrator.XE_MEMBER_HOST_POLICY` from #570 (required host type `iosXeRoutedHost`, joined member type `iosXeL3PoMember`) rather than anything new. That table can fold into `interface_membership.py` when this lands.

**Conflict refresh (2026-09-18).** `git merge-tree` of this branch now reports conflicts against develop itself in the three module files (`nd_interface_ethernet_{access,routed,trunk_host}.py`, most likely from the `fail_from_exception` sweep in #557), and against #558 in 10 files (the 8 listed before plus `nd_interface_ethernet_routed.py` and the routed target's `tasks/main.yaml`). The offer to help with the rebase once #558 lands still stands.

allenrobel:
More lab evidence on the `operData.portChannelId` lag from my 2026-09-18 comment, this time with a customer angle. Still not a review finding; per our agreement I'll review after the sync.

**What happened (ND 4.3.1.175, CAMPUS1 / C3_LE1, 2026-09-21).** The port-channel access XE scenario removed `port-channel102` under `state: overridden`. `deploymentHistory` shows both commands pushed and `success` at 19:24:08Z:

```text
interface GigabitEthernet1/0/4 ; no channel-group 102 mode active ; ...
no interface Port-channel102
```

The member's intent already read `iosXeAccess`. ND's inventory nevertheless kept reporting `operData.portChannelId: 102` on `GigabitEthernet1/0/4`, and a policy-less `Port-channel102` record (empty per-interface `diff`, so nothing on the switch), for **about 7 minutes**. The scenario's member reset (`nd_interface_ethernet_access`, `state: deleted`) was refused through all 8 retries at 15 s:

```text
Interface GigabitEthernet1/0/4 is a member of port-channel 102. Refusing to normalize a port-channel member
(this would strip its channel-group membership). Remove the interface from the port-channel first, then re-run the delete.
```

On 2026-09-18 the same lag outlasted 2 minutes once and was absent on the next run; on 4.2.1.10 it has stayed under a minute. So the window is controller-load dependent and much longer than I reported before.

**Why I think it matters beyond our tests.** "Remove a port-channel, then repurpose its member ports" in one playbook is an ordinary workflow, and for the length of that window the ethernet modules refuse it with a message that is wrong on both counts: the interface is not a member, and there is nothing left to remove it from. It fails safe, but the only remedy is to retry blind for minutes.

**How this branch handles it today (read at `24cb0db1`).** The write path is covered: `classify_member_policy` puts a host policy with a positive operational ID in the `NOT_MEMBER` / inconsistent-evidence branch, which is the case above. The delete path is not: `_check_port_channel_delete_restriction` classifies by policy type first, but when that says `NOT_MEMBER` it still falls through to `_existing_port_channel_id` and refuses on the operational ID alone, with the old message. That is exactly the refusal quoted above, so it would survive this PR.

**Suggestion.** Let intent decide membership on the delete path too, the way `EthernetMembershipIndex` and the port-channel orchestrator's `_member_owners` already do (ND itself validates a conflicting create against intent, not `operData`):

- A member policy type, or a port-channel record whose `ports` lists the interface: a member. Refuse, as now.
- A host policy type that no port-channel claims, with a positive `operData.portChannelId`: stale discovery data after a removal deploy. I'd let the reset proceed. The risk the guard exists for is stripping a live `channel-group`, and in this state ND's own intent has no bundle to strip; the normalize body only restates what intent already says.

If you would rather keep refusing on inconsistent evidence, then the same transient-condition hint I suggested for the write path belongs here as well ("ND may still report the old bundle for several minutes after a port-channel removal; retry once `operData.portChannelId` reads -1"), so the message at least tells the user it will clear by itself.

The opposite gap is still there for IOS-XE, for what it is worth: a joined Catalyst member reads `operData.portChannelId: -1` (the parent is only in `policy.portChannelId`), so any check that keys on the operational ID alone never fires for it. Intent-first membership closes both.

Vault: `xe-port-channel-removal-inventory-lag` (today's 7-minute row is in). None of this blocks #570; its scenarios now retry the member reset for up to ten minutes under a `TODO(4.3.1)` marker.

Issue #554, opened 2026-09-09
A stale HTTP 207 response can mark an unsent group as accepted.


akinross:
### Ansible Version

```shell
n/a
```

### Ansible Collection Versions

```shell
n/a
```

### Ansible Collection Module

```shell

```

### Cisco Nexus Dashboard Version

```shell
n/a
```

### Cisco NX-OS Version

```shell

```

### Cisco ACI Version

```shell

```

### Expected Behavior

```shell
If a normalize request fails before receiving a response, every interface in that request should remain in _pending_normalizes. The failure-path finalizer must not deploy those interfaces because their normalization was never accepted.
```

### Actual Behavior

```shell
_dequeue_accepted_normalizes() reads rest_send.response_current. If the current request raises inside sender.commit() before response_current is updated, it still contains the previous request’s HTTP 207 response.
When both groups contain the same interface name on different switches, the stale success result matches the unsent group by name. That group is removed from _pending_normalizes and can subsequently be deployed even though its normalize request was never sent successfully.
```

### Ansible Playbook

```yaml
n/a
```

### Steps to Reproduce

```shell
1. Queue the same interface name on two switches:
[
    ("Ethernet1/31", "FDO11111AAA"),
    ("Ethernet1/31", "FDO22222BBB"),
]
2. Return HTTP 207 with Ethernet1/31: success for the first switch’s normalize request.
3. For the second request, make sender.commit() raise before assigning a new response.
4. Call _normalize_interfaces().
5. Observe that _dequeue_accepted_normalizes() reads the first request’s stale 207 response.
6. The second switch’s pair is incorrectly removed from _pending_normalizes.
7. deploy_accepted_mutations() may then deploy both switches instead of only the first.
The existing HTTP-500 test does not reproduce this because the 500 response replaces response_current; the second request specifically needs to fail before response assignment.
```

### Relevant Debug Output

```shell

```

Issue #349, opened 2026-06-25
Interface query_all fans out one GET per switch; no fabric-wide interface-list endpoint


allenrobel:
## Summary

Deferred finding from the per-policy interface stacks. The interface orchestrators implement `query_all` as a
loop that issues one `interfaceList` GET per switch, because ND 4.2.1 exposes no fabric-wide interface-list
endpoint (only `GET /api/v1/manage/fabrics/{fabric}/switches/{switchId}/interfaces`). On large fabrics this makes
the request count proportional to the number of switches queried. Performance / scalability, not a correctness bug.

## Current state

`query_all` in `plugins/module_utils/orchestrators/{port_channel_base,ethernet_base,svi_interface,subinterface_*}`
iterates a switch map and calls `_request` (or the cached `_switch_interfaces`) once per switch.

PR #269 reduced the blast radius for `nd_interface_port_channel_access` by scoping the loop to the switches named
in `config` for all non-`overridden` states (`_switches_to_query()`), mirroring `ethernet_base`. That keeps the
request count proportional to **config size** rather than fabric size for `merged` / `replaced` / `deleted`. But
`state: overridden` is fabric-wide by definition and still issues one GET per switch, and the same per-switch
fan-out exists in the other interface modules.

## Why it matters

For a `state: overridden` run (or any module that has not yet adopted `_switches_to_query`), a single-interface
change still walks every switch in the fabric — slow runs and controller-side HTTP timeouts on large fabrics, the
exact symptom the CLAUDE.md performance rule ("no N+1 requests") targets.

## Affected PRs

- #268 `nd_interface_svi`
- #296 `nd_interface_subinterface_managed`
- #297 `nd_interface_subinterface_unmanaged`
- #269 `nd_interface_port_channel_access` (config-scoping landed; overridden still fabric-wide)
- #270 `nd_interface_port_channel_trunk_host` (config-scoping inherited; overridden still fabric-wide)
- #280 `nd_interface_vpc_trunk_host` (per-switch fan-out inherited from `VpcInterfaceBaseOrchestrator`; config-scoping for non-overridden, fabric-wide only for overridden — no fabric-wide vPC-interface endpoint exists)

## Proposed approach

1. Short term: ensure every interface orchestrator adopts the `_switches_to_query()` config-scoping pattern so
   non-`overridden` states never fan out beyond the named switches, and that each switch's interface list is
   fetched at most once per run (the `_switch_interfaces` cache in `ethernet_base`).
[...]

PR #563, opened 2026-09-11
nd_interface_ethernet_*: omit description from the normalize body (ND 4.3.1 rejects "")

### LINKED ISSUE #562: nd_interface_ethernet_*: state deleted fails on ND 4.3.1 (normalize rejects description: "", HTTP 400 minimum string length is 1)
## Summary

Every ethernet interface module's `state: deleted` (`nd_interface_ethernet_access`, `nd_interface_ethernet_trunk_host`, `nd_interface_ethernet_routed`, and the IOS-XE branches that share `EthernetBaseOrchestrator`) fails on **ND 4.3.1** at `remove_pending()`:

```text
Bulk normalize failed for ['Ethernet1/48']: Request failed (400): Bad Request - ND Error 400: Request validation failed. None of these interfaces were reset.
```

ND's error detail:

```json
{"errors":[{"code":0,"description":"Error at /configData/networkOS/policy/description: minimum string length is 1","message":"Schema validation failed"}]}
```

## Cause

`InterfaceDefaultConfig.to_normalize_payload()` serializes the full `int_trunk_host` template, including `description: ""`. The spec types the ethernet policy `description` as `interfaceDescription` (`minLength: 1`) on **both** 4.2.1 (manage 1.1.411) and 4.3.1 (manage 1.2.224). ND 4.2.1.10 never enforced it; ND 4.3.1.175 does. So the empty string was a spec violation that 4.2.1 tolerated.

## Lab evidence (2026-09-11)

Same full-template normalize body for `Ethernet1/48`, fabric `SITE1`:

| Controller | `description: ""` | `description` omitted |
|---|---|---|
| ND 4.3.1.175 (S3_BG1) | HTTP 400 as above | HTTP 207 `Interface normalized successfully` |
| ND 4.2.1.10 (S1_BG1) | HTTP 207 success | HTTP 207 success |

A GET after the omitted-key normalize on 4.3.1 shows `description` absent and trunkHost defaults in place, so omitting the key still resets it. Only `description` is affected; the other empty-string template fields have no `minLength`.

Integration runs against 4.3.1 on `develop` (c6369920): `nd_interface_ethernet_access`, `nd_interface_ethernet_trunk_host`, and `nd_interface_ethernet_routed` all fail at their first cleanup task.

## Fix

Default `InterfaceDefaultPolicyModel.description` to `None` so `to_payload` (`exclude_none`) drops the key. This is spec conformance, not an ND workaround, so no `TODO(X.Y.Z)` marker. Vault note: `empty-interface-description-accepted`.

4.2.1.10 (tolerates) and 4.3.1.175 (enforces)

https://claude.ai/code/session_01LiYaRfNfQ6GR9a73NPHF8S

allenrobel:
## Related Issue(s)

Closes #562

## Proposed Changes

- **Why:** every ethernet `state: deleted` (access, trunk host, routed, and the IOS-XE branches sharing `EthernetBaseOrchestrator`) failed on ND 4.3.1 at `remove_pending()`: `interfaceActions/normalize` returned HTTP 400 `Error at /configData/networkOS/policy/description: minimum string length is 1`. The `int_trunk_host` template body carries `description: ""`. The spec types that field as `interfaceDescription` (`minLength: 1`) on both 4.2.1 (manage 1.1.411) and 4.3.1 (manage 1.2.224); 4.2.1.10 never enforced it, 4.3.1.175 does.
- **Why not just omit the key:** the releases also disagree on what normalize does with an omitted field. 4.3.1 resets it; 4.2.1 leaves it untouched, so the omit-only first version of this PR broke 4.2.1 `overridden` idempotency (descriptions survived the first pass). No single body works on both, and the collection has no ND version detection.
- **Change:** `EthernetBaseOrchestrator._post_normalize` sends the template body first (byte-identical to what 4.2.1 always received) and, on that exact 400 (`_rejected_empty_description`, read only when the response count grew, per issue #554), resends the same group once without `description` and keeps omitting it for the rest of the run (`_normalize_omits_description`). One extra request per module run on 4.3.1, none on 4.2.1. `InterfaceDefaultConfig.to_normalize_payload` gains `omit_description`.
- **Queue invariants (the four questions):** a 400 rejects the whole request, so nothing was accepted by the first attempt and the group stays queued until the resend's 207 is processed by the existing bookkeeping; the finalizer cannot resubmit anything because the normalize queue is unchanged on failure; the except-path read is gated on `rest_send.response_count` growing; result items are keyed by `name` as before. Any other failure, and a failure of the resend, propagates through the existing all-or-nothing message.
- Not a `TODO(X.Y.Z)` workaround: the fallback is the dual-release contract until 4.2.1 support ends (ND 4.5.3), when the empty-string first attempt can go. Vault note `empty-interface-description-accepted` (status workaround) records both behaviors.
- **Freshness token (review finding, commit "RestSend: add response_count and use it for the normalize and remove freshness snapshots"):** the snapshot was first written as `len(rest_send.responses)`, but that property deep-copies the whole response history on every read, and with a repeated interface name across switches (one normalize group per switch, one inventory response per switch already in the history) the copying grew quadratically with the switch count. `RestSend` gains a `response_count` property (the list is never cleared after `__init__`, so a count is a sound token); both normalize reads and the two reads on the #547 remove path in `base_interface.py` use it. Same primitive the issue #554 fix needs.
- Tests: new `tests/unit/module_utils/models/test_interface_default_config.py`; routed orchestrator tests 00990 (resend on the 4.3.1 rejection: three responses consumed, last payload has no `description`, queues dequeued, deploys kept) and 01000 (no resend on any other 400); scale regressions 01010 (Ethernet1/1 on 50 switches, 50 groups, `RestSend.responses` patched to raise if read, every pair dequeued, count grew by one per group) and 01020 (same setup, the 4th group's sender raises without a response: no resend, fail-fast message, count unchanged for that group, still no copy); `test_rest_send_response_count`.

## Test Notes

- Raw REST, full-template normalize body, fabric `SITE1` (2026-09-11): ND 4.3.1.175 returns 400 with `description: ""` and 207 without it, and the GET afterwards shows the description cleared. ND 4.2.1.10 returns 207 for both, but only the `""` body clears the description; with the key omitted the old description survives (Ethernet1/47, Ethernet1/49 on S1_BG1).
- `ndpytest tests/unit/`: **4671 passed**. `black`, `isort`, `pylint` (only the repo-tolerated `C0413` / `W0201` classes remain), `mypy` (no new errors; the 37 flagged lines in `ethernet_base.py` are pre-existing `NDBaseModel` attribute accesses): clean.
- Integration, this branch (`ansible-test network-integration`), all three ethernet targets:
  - **ND 4.2.1.10 (S1_BG1):** access ok=89 failed=0, trunk_host ok=117 failed=0, routed ok=93 failed=0 (ignored=1 is the expected unknown-switch negative test). Identical counts to the develop baselines run earlier the same day.
  - **ND 4.3.1.175 (S3_BG1):** access ok=89 failed=0 (end to end; the module log shows 14 normalize POSTs, 7 rejected with the description error and each immediately resent successfully). trunk_host ok=113 through every merged/replaced/overridden/deleted normalize step, then fails at `DELETED CLASS C: Seed Ethernet1/48` on the separate 4.3.1 create regression tracked in #564 (`trunkHost` create without `allowedVlans`). routed passes its pre-test cleanup normalize and fails at its first `MERGED CREATE` on the same #564 regression (`routedHost` create without `mtu`). Both are outside this PR.
- Lab hygiene: the 4.3.1 runs leave S3_BG1 test interfaces in the state #564 describes; the next run's pre-test cleanup resets them.

4.2.1.10 and 4.3.1.175

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

https://claude.ai/code/session_01LiYaRfNfQ6GR9a73NPHF8S


allenrobel:
Converted to draft. The 4.2.1 integration run of `nd_interface_ethernet_trunk_host` on this branch fails at `OVERRIDDEN IDEMPOTENT: Verify no change on second run`: the interfaces the first overridden pass normalized still carry their descriptions on the second pass. That points at a semantic difference between the releases: ND 4.3.1 resets a field the normalize body omits, ND 4.2.1 leaves it untouched. Verifying with raw REST now; the fix has to be version-agnostic for both, so the current "omit the key" shape is not sufficient on its own.

plugins/module_utils/orchestrators/ethernet_base.py
```diff
@@ -584,6 +587,65 @@ def _normalize_interfaces(self) -> list[ResponseType]:
...
+        attempt.
+
+        ## Raises
+
+        ### Exception
+
+        - If the normalize request fails for any reason other than the ND 4.3.1 empty-description rejection, or if the resend fails.
+        """
+        payload = InterfaceDefaultConfig.to_normalize_payload(group, omit_description=self._normalize_omits_description)
+        response_count = len(self.rest_send.responses)
```
mikewiebe, line 611: ### Medium: Do not deep-copy the complete response history for every normalize group

**Issue**

The new response-freshness snapshot evaluates `len(self.rest_send.responses)` before every normalize request. `RestSend.responses` returns a deep copy of the complete accumulated controller-response history, even though this caller needs only an integer sequence/count.

When any interface name repeats across switches—for example, `Ethernet1/1` is being reset on every leaf—`_normalize_groups()` changes the entire queue to one group per switch. With `S` switches, the new code therefore deep-copies the roughly `S` per-switch interface-query responses `S` times, adding quadratic local CPU and allocation work before/during the unavoidable API calls.

**Evidence**

- [The changed line snapshots freshness through `len(self.rest_send.responses)` for every normalize group](https://github.com/CiscoDevNet/ansible-nd/pull/563/files#diff-923f5fa5e7f2b2e100e7dc90ee24f3bc89d080b057dcc7a052c35498a462ae06R611).
- [`RestSend.responses` constructs `copy.deepcopy(self._response)` on every read](https://github.com/CiscoDevNet/ansible-nd/blob/c45e165a74e5c46f83e390875991e21ac687f7d1/plugins/module_utils/rest/rest_send.py#L584-L593).
- [`_normalize_groups()` uses a single request only when every queued interface name is globally unique; one duplicate changes the operation to per-switch groups](https://github.com/CiscoDevNet/ansible-nd/blob/c45e165a74e5c46f83e390875991e21ac687f7d1/plugins/module_utils/orchestrators/ethernet_base.py#L521-L541).
- A focused benchmark populated the actual `RestSend` history with one 64-interface query response per switch, then measured only the newly added count snapshots—no network, controller, JSON parsing, or normalize-request time. The local copies took 0.228s for 25 switches, 0.923s for 50, 3.782s for 100, and 15.959s for 200. The benchmark is preserved as `response_history_benchmark_pr563.py` beside this review.

**Existing PR overlap**

No matching existing PR comment found. The PR's only conversation comment concerns the superseded omit-only implementation's ND 4.2.1 idempotency failure, not the response-history copy cost.

No matching open issue found. [Issue #349](https://github.com/CiscoDevNet/ansible-nd/issues/349) tracks controller-side per-switch query fan-out, while [issue #554](https://github.com/CiscoDevNet/ansible-nd/issues/554) tracks stale response reuse after a sender exception. Neither tracks this PR's newly added repeated local deep copy, although the same cheap freshness primitive can help resolve #554.

**Impact**

A normal `deleted` or `overridden` task that resets the same numbered ports across a fabric—such as `Ethernet1/1` through `Ethernet1/64` on 200 leaf switches—takes about 16 seconds of additional local Python work in the isolated benchmark. That cost is additive to the existing controller requests and grows quadratically with switch count; real response bodies can be larger than the benchmark fixture.

**Suggested fix**

The code currently asks, “How many responses do we have?” by copying every stored response and then counting the copy:

```python
before = len(self.rest_send.responses)
```

But `responses` performs a deep copy. If 100 responses are stored, all 100 are copied just to obtain the number `100`. Repeating that for every switch causes the unnecessary quadratic slowdown.

The fix is to expose a cheap counter:

```python
@property
def response_count(self) -> int:
    return len(self._response)
```

Then each operation can determine whether it received a new response without copying anything:

```python
before = self.rest_send.response_count
self.rest_send.send(...)

if self.rest_send.response_count == before:
    # No new response was received; do not reuse an older response.
    return
```

A monotonically increasing `response_sequence` would be even safer if the stored response list can ever be cleared.

This solves two problems:

- It removes the repeated copying and associated slowdown.
- If a request fails before receiving a response, the counter does not advance, preventing an earlier `207` response from being mistaken for the current request's result.

The actual controller requests do not change; only the inexpensive bookkeeping used to identify fresh responses changes. Add a scale regression with a repeated interface name across many switches and assert that freshness checks never copy the response collection.

allenrobel: Confirmed and fixed in commit "RestSend: add response_count and use it for the normalize and remove freshness snapshots".

I reproduced the shape of your benchmark (30 KB 64-interface inventory response per switch, one full-history copy per switch) and got the same curve, slightly worse on my box: 0.29 s / 1.17 s / 4.78 s / 19.9 s at 25 / 50 / 100 / 200 switches. The premise holds end to end: `query_all` fans out one GET per switch, so the history already holds S inventory responses before the first normalize group, and a repeated name makes `_normalize_groups` send S groups.

Change: `RestSend.response_count` returns `len(self._response)` without copying. `_response` is assigned only in `__init__` and never cleared or replaced, so a plain count is a sound freshness token and I did not add a separate sequence. Both normalize reads (`_post_normalize` snapshot, `_rejected_empty_description` compare) use it, and I switched the two reads on the #547 remove path in `base_interface.py` as well so the collection has one freshness primitive; that is also the one the #554 fix needs.

Tests, per your ask: `test_ethernet_routed_orchestrator_01010` queues Ethernet1/1 on 50 switches with 50 inventory responses already in the history, patches `RestSend.responses` to raise if read, and asserts 50 groups sent, every pair dequeued, and `response_count` grew by exactly one per group. `01020` is the failure side: the 4th group's `_request` raises without recording a response (the #554 shape), and the test asserts no resend, the fail-fast message, the first three pairs dequeued, the rest still queued, and again no read of `responses`. Both failed on the previous head with `AssertionError: ... must not read the deep-copying RestSend.responses property`. `test_rest_send_response_count` pins the property itself. Full unit suite 4671 passed; sanity pylint clean on all five files. PR body updated.

mikewiebe (approved):
Reviewed exact head 8e5c3cf1. The prior response-history performance concern is fixed with the O(1) response_count path; 537 targeted tests pass, and the preserved probe reports 5 passed with one strict xfail for the separate #554 condition. Status-rollup CI is green apart from the intentionally skipped integration job. Approving; #554 remains separately tracked.

mikewiebe (approved):
Reviewed exact head 3ad0cbb007cc61f68201bbb4643082d09aa5d5ec. The rebase is patch-equivalent to the previously reviewed series, exact-head verification remains green, and the prior performance concern remains fixed.

mikewiebe (approved):
Re-reviewed at 484fad29578f09b3b31eec1b8ca0ed2cf1339f7a. No current-head correctness findings remain. The normalize compatibility fix and O(1) response freshness check pass focused verification for the ND 4.2.1/4.3.1 behavior split.

mikewiebe (approved):
Re-reviewed current head d6569daf5d62f16f11d6395ad4d346fc79e70fd7. No blocking findings remain.

PR #568, opened 2026-09-15
ci: optimize GitHub Actions to reduce runtime and Actions minutes

### LINKED ISSUE #567: Optimize GitHub Actions CI to reduce runtime and Actions minutes
## Summary

The `CI` workflow (`.github/workflows/ansible-test.yml`) currently runs a 12-cell matrix (4 ansible-core versions x 3 Python versions) across four chained jobs, producing roughly **49 jobs per PR** (integration is disabled via `if: false`). Much of this is redundant and consumes Actions minutes without adding test coverage.

## Current cost breakdown (per PR)

| Job | Cells | Matrix meaningful? |
|-----|-------|--------------------|
| `pep8-black` | 1 | n/a |
| `build` | 12 | No - collection tarball is identical for every cell |
| `ansible-galaxy-importer` | 12 | No - import check is version-independent |
| `ansible-test` (sanity) | 12 | Partly - only the ansible-core axis matters (`--docker` ignores host Python) |
| `units` | 12 | Yes - `--python` genuinely exercises each interpreter |

## Proposed optimizations

1. **De-matrix `build`** - build the collection tarball once and upload a single artifact (saves 11 jobs).
2. **De-matrix `ansible-galaxy-importer`** - run the import check once against the single artifact (saves 11 jobs).
3. **Collapse the sanity matrix to the ansible-core axis only** - `ansible-test sanity --docker` runs inside the default test container and ignores the runner host Python, so one cell per ansible-core version is sufficient (12 -> 4 cells).
4. **Keep the `units` full 12-cell matrix** - `--python` matters, so this coverage is retained.
5. **Add `concurrency` with `cancel-in-progress`** so superseded PR runs are cancelled (the expensive workflow lacks this today; `certification.yml` already has it).
6. **Add `paths-ignore`** so docs-only changes (`**.md`, `docs/**`, `changelogs/**`, `playbooks/**`) skip the full pipeline.
7. **Enable pip caching** on `actions/setup-python` steps to avoid re-downloading dependencies on every job.
8. **Align artifact action versions** - `build` currently uploads with `upload-artifact@v7` while consumers use `download-artifact@v8` (incompatible majors); standardize on v8.

### Net effect
Approximately **49 -> ~21 jobs per PR** (~55% fewer) with no loss of real test coverage.

## Follow-up (out of scope for the first PR)

- Fold the no-op `jira-issue-sync-pr-review.yml` bridge workflow into `jira-issue-sync.yml` via a direct `pull_request_review` trigger. Deferred because it changes `pull_request_target` trigger semantics and warrants its own focused security review.
- Review the overlap between `certification.yml` sanity and `ansible-test.yml` sanity, both of which run on every PR.

mtarking:
## Summary

Optimizes the `CI` workflow (`.github/workflows/ansible-test.yml`) to cut Actions minutes without reducing test coverage. Closes #567.

The workflow previously ran a 12-cell matrix (4 ansible-core x 3 Python) across four chained jobs, producing ~49 jobs per PR. Most of that was redundant.

## Changes

- **`build` — matrix reduced to the ansible-core axis (12 → 4 cells).** Still built on every supported ansible-core version so version-specific build / `galaxy.yml` breakage is caught, but the Python axis is dropped since the tarball does not vary by interpreter. Only the canonical cell (`stable-2.19`) uploads the shared `collection` artifact.
- **De-matrix `ansible-galaxy-importer` (12 → 1).** The import check is version-independent, and Galaxy / Automation Hub import with a single ansible-core version anyway. It now runs once against the shared artifact.
- **Collapse the sanity matrix to the ansible-core axis (12 → 4 cells).** See the coverage note below — this is not a coverage reduction.
- **Keep the full `units` matrix (12 cells) unchanged.** This is where the ansible-core × Python support-matrix coverage actually lives, and it is untouched.
- **Add `concurrency` with `cancel-in-progress`** so superseded PR runs are cancelled (`certification.yml` already had this; the expensive workflow did not).
- **Add `paths-ignore`** so docs-only changes (`**.md`, `docs/**`, `changelogs/**`, `playbooks/**`) skip the full pipeline.

Artifact action versions are left as they were (`upload-artifact@v7`, `download-artifact@v8`). These two actions are versioned independently — upload's latest is v7.0.1 and download's is v8.0.1 — so the existing pairing is correct and intentional.

## Coverage note: why the sanity reduction is not a loss

The sanity step does **not** pass `--python`:

```yaml
run: ansible-test sanity --docker -v --color --truncate 0 --coverage
```

With `--docker` and no `--python`, sanity runs inside the default test container and iterates its **own** set of target Python versions. It does not inherit the host interpreter installed by `actions/setup-python` — the host Python only runs the `ansible-test` CLI process. On ansible-core 2.19 the CLI advertises target versions `3.8`–`3.13`, a wider range than the `3.11`/`3.12`/`3.13` the matrix was supplying as host Python.

So `(2.19, 3.11)`, `(2.19, 3.12)` and `(2.19, 3.13)` were running identical sanity work — three jobs for one job's worth of coverage. Keeping one cell per ansible-core version preserves it.

This is corroborated by the repo's sanity ignore files, which are keyed per ansible-core version only (`ignore-2.16.txt` … `ignore-2.19.txt`) with no per-Python variants.

Contrast with `units`, which **does** pin the interpreter and therefore keeps its full 12-cell matrix:

```yaml
run: ansible-test units --docker -v --color --truncate 0 --python ${{ matrix.python_version }} --coverage
```

## Net effect

Approximately **49 → 22 jobs per PR** (~55% fewer), with the ansible-core × Python support-matrix coverage unchanged.

| Job | Before | After |
|-----|--------|-------|
| `pep8-black` | 1 | 1 |
| `build` | 12 | 4 |
| `ansible-galaxy-importer` | 12 | 1 |
| `ansible-test` (sanity) | 12 | 4 |
| `units` | 12 | 12 |

## Deferred (not in this PR)

- **pip caching** — evaluated but deferred: the test jobs install ansible-core from a URL and don't check out the repo, so `setup-python`'s `cache: pip` has no dependency file to key on. Doing it correctly needs a dedicated `actions/cache` step.
- **Fold `jira-issue-sync-pr-review.yml` into `jira-issue-sync.yml`** — changes `pull_request_target` trigger semantics and warrants its own focused security review.
- **`certification.yml` vs `ansible-test.yml` sanity overlap** — both run sanity on every PR; worth a separate decision.

## Testing

- Workflow YAML parses cleanly via `yaml.safe_load`.
- Verified job graph: `build` = 4 cells, `ansible-galaxy-importer` = 1, sanity = 4 cells, `units` = 12 cells, `concurrency` present; 22 active jobs per PR.
- All action references verified to exist: `checkout@v6`, `setup-python@v6` (v6.3.0), `upload-artifact@v7` (v7.0.1), `download-artifact@v8` (v8.0.1).
- `build` pins Python 3.12, the common controller version across ansible-core 2.16–2.19 (2.16/2.17 cap at 3.12; 2.18/2.19 start at 3.11).


.github/workflows/ansible-test.yml
```diff
@@ -32,86 +44,50 @@ jobs:
   build:
     name: Build Collection
     runs-on: ubuntu-latest
-    strategy: &ansible_strategy
```
akinross, line 35: Should we not test the build works for each stable version of ansible? I think difference between these versions could be missed, if I recall correctly there were issues with this in the past where a specific version was unable to build.

.github/workflows/ansible-test.yml
```diff
@@ -32,86 +44,50 @@ jobs:
...
+      - name: Set up Python
         uses: actions/setup-python@v6
         with:
-          python-version: ${{ matrix.python_version }}
+          python-version: "3.13"
 
-      - name: Install ansible-core (${{ matrix.ansible_version }})
-        run: pip install https://github.com/ansible/ansible/archive/${{ matrix.ansible_version }}.tar.gz --disable-pip-version-check
+      - name: Install ansible-core (stable-2.19)
+        run: pip install https://github.com/ansible/ansible/archive/stable-2.19.tar.gz --disable-pip-version-check
```
akinross, line 96: The ansible-galaxy-importer job now runs only under stable-2.19, but the importer's own README pins its supported range at "ansible-core up to 2.16" — meaning the ansible-doc / ansible-lint versions it invokes may not match what Galaxy's server-side importer actually uses at publish time. Could we pin this job to stable-2.16 (or also all 4 stable version) so the check mirrors the environment our tarball will hit when it's uploaded?

akinross (changes requested):

.github/workflows/ansible-test.yml
mtarking: Draft - still work-in-progress

.github/workflows/ansible-test.yml
mtarking: Draft - still work-in-progress

allenrobel:
Coverage follow-up (the `units` job, which is ~82% of the run's minutes) is tracked in #585 and implemented in #586, out of this PR's scope by design.

Issue #569, opened 2026-09-16
ND leaves the IOS-XE port-channel interface on the switch after a port-channel is removed (4.2.1 and 4.3.1)


allenrobel:
## Summary

On IOS-XE (Catalyst 9000) switches, removing a port-channel from Nexus Dashboard intent does not remove the `interface Port-channel<N>` object from the switch. This is ND behaviour on **both** 4.2.1.10 and 4.3.1.175, reproduced with raw REST on 2026-09-15 (CAMPUS1 `vxlanCampus` fabric, Catalyst 9000v leaves C1_LE1 / C3_LE1). NX-OS is unaffected: the same suite leaves no `port-channel50x` lines on S1_BG1.

Found while landing the IOS-XE policy types for `nd_interface_port_channel_access` / `nd_interface_port_channel_trunk_host` (#536 / #537). The modules' `state: deleted` / `state: overridden` therefore remove the ND record and detach the members, but leave a fully configured `Port-channel` interface (description, VLANs, bpduguard) in the Catalyst running config. The module DOCUMENTATION states this; this issue tracks the follow-up.

## What ND does

1. `POST .../switches/{sn}/interfaces` (`Port-channel111`, `iosXeTrunkPoHost`, member `GigabitEthernet1/0/5`) + `interfaceActions/deploy` → the switch has `interface Port-channel111` and the member carries `channel-group 111 mode active`.
2. `interfaceActions/remove` (what the modules use) — or `DELETE .../interfaces/port-channel111`, or `DELETE .../interfaces/Port-channel111` — + `interfaceActions/deploy` → the ND record is gone immediately, `GET .../switches/{sn}/diff` shows **no** pending deletion, and the interface object stays in the running config. With the bulk remove the members are detached (on 4.2.1 one member, Gi1/0/4, once stayed bundled and `operData.portChannelId` lagged by about a minute); with the per-interface DELETE on the live record the member keeps its `channel-group`.
3. Minutes later ND rediscovers the orphan as a policy-less record (`configData: {mode: unknown, networkOS: {networkOSType: ios-xe}}`, canonical name `Port-channel111`).
4. Only `DELETE .../interfaces/Port-channel111` on **that** rediscovered record (canonical capitalized name; the lowercase form silently drops the record) followed by `interfaceActions/deploy` `Port-channel111` queues `delete interface Port-channel111` in the diff and removes it from the switch.

[...]

PR #566, opened 2026-09-14
nd_interface_{ethernet_*,vpc_trunk_host}: always emit the template defaults ND 4.3.1 requires (allowedVlans, mtu)

allenrobel:
## Related Issue(s)

Closes #564

Stacked on #558 (`feat_534_535_ethernet_iosxe`): the two IOS-XE ethernet policy models only exist there. Rebase onto develop once #558 merges.

## Proposed Changes

ND 4.3.1 rejects interface bodies that omit template fields ND 4.2.1 silently defaulted (`Validation failed for following fields: [...]`). Neither spec marks them required, and the defaults are identical on both releases, so the fix is version-agnostic: always emit the template default on the wire when the user set nothing.

- `NDBaseModel.payload_defaults`: a per-model table of wire key to template default, injected by a wrap-mode serializer **only on payload-mode dumps** (`to_payload`). Config and diff dumps carry no payload context, so before/after output and merged/replaced/overridden classification are unchanged; the read side already normalizes the echoed default via `reverse_diff_defaults`. A model with its own wrap serializer calls the shared helper (`TrunkVpcHostPolicyModel` applies it before its per-peer fan-out, so both `peer1AllowedVlans` / `peer2AllowedVlans` carry `none` on POST and PUT).
- Tables, each with a `TODO(4.3.1)` marker pointing at the vault note:

  | policyType | key | default |
  |---|---|---|
  | `trunkHost` | `allowedVlans` | `none` |
  | `routedHost` | `mtu` | `9216` |
  | `iosXeAccess` | `mtu` | `1500` |
  | `iosXeTrunkHost` | `allowedVlans`, `mtu` | `none`, `1500` |
  | `iosXeRoutedHost` | `mtu` | `1500` |
  | `trunkVpcHost` | `allowedVlans` | `none` |

- The first 4.3.1 lab runs showed the same rejection on the delete-side reset PUTs, so those bodies are now sourced from the same tables: `InterfaceDefaultConfig.to_reset_payload` (NX-OS Class C reset) gains `allowedVlans: none`, and `EthernetBaseOrchestrator._xe_reset_payload` gains `XE_RESET_POLICY_DEFAULTS` (`iosXeTrunkHost`: `allowedVlans` + `mtu`; routed override: `mtu`). Values are the template defaults, so the reset lands on the same unconfigured-default signature on both releases.
[...]

Issue #564, opened 2026-09-11
nd_interface_{ethernet_trunk_host,ethernet_routed,ethernet_access (XE),vpc_trunk_host}: ND 4.3.1 rejects create/update bodies that omit fields 4.2.1 defaulted (allowedVlans, mtu)


allenrobel:
## Summary

First integration runs of the ethernet targets against **ND 4.3.1.175** (2026-09-11, fabric `SITE1`, S3_BG1) surface two create-side regressions relative to 4.2.1.10. Both are ND applying template-required fields that 4.2.1 silently defaulted; the modules only send what the user set.

| Module | Create body omits | ND 4.3.1 response (207 item) | 4.2.1 |
|---|---|---|---|
| `nd_interface_ethernet_trunk_host` | `allowedVlans` | `Policy [trunkHost] - Validation failed for following fields: [allowedVlans]` | accepted, defaults to `none` |
| `nd_interface_ethernet_routed` | `mtu` | `Policy [routedHost] - Validation failed for following fields: [mtu]` | accepted, defaults to `9216` |

Same family as the vault's `vpc-trunk-allowedvlans-required-431` (trunkVpcHost, found 2026-09-09). The ethernet flavors are new.

## Where the integration targets hit it

- `nd_interface_ethernet_trunk_host`: `DELETED CLASS C: Seed Ethernet1/48 with bandwidth=1500000` (task 114 of the target; everything before it, including every `deleted`/normalize step, passes on 4.3.1 with #563 applied).
- `nd_interface_ethernet_routed`: `MERGED CREATE: Configure routed Ethernet1/31 with full config` (the very first create; the pre-test cleanup normalize passes with #563).
- `nd_interface_ethernet_access` passes end to end on 4.3.1 with #563 (accessHost has no newly-required field).

## Proposed fix (version-agnostic, same approach as the vPC item)

Always emit the field with the template default when the user did not set it: `allowedVlans: "none"` for trunkHost, `mtu: 9216` for routedHost (verify the routedHost default against the 4.2.1 and 4.3.1 `intRoutedHost` templates before hardcoding). Both values are what 4.2.1 already stores, so before/after idempotency on 4.2.1 is unchanged, and no ND version detection is needed. Confirm on the lab that the read side already normalizes these via `reverse_diff_defaults` so the extra key does not create a spurious diff.

[...]

PR #570, opened 2026-09-16
nd_interface_port_channel_{access,trunk_host}: add IOS-XE policy types (iosXeAccessPoHost, iosXeTrunkPoHost)

### LINKED ISSUE #381: nd_interface_port_channel_access: expose 7 accessPoHost policy fields present in the schema and the trunk-host sibling
## Summary

In PR #269 review, @akinross noted the UI exposes several options (checked on an iBGP fabric leaf) that the module does not, and asked how we verify that modules expose all available options (https://github.com/CiscoDevNet/ansible-nd/pull/269#discussion_r3496560998).

Verified against the ND 4.2.1 OpenAPI spec: all seven fields exist in the `intPortChannelAccessHostTemplate` schema (`int_port_channel_access_host.template`, `policyType: accessPoHost`), and the trunk-host sibling model already implements all seven — the gap is only in the access model.

## Missing fields (schema alias — proposed module option)

| Schema field | Module option | Type | Schema default |
| --- | --- | --- | --- |
| `linkType` | `link_type` | enum `auto`/`pointToPoint`/`shared` | `auto` |
| `portTypeEdgeTrunk` | `port_type_edge_trunk` | bool | `true` |
| `negotiateAuto` | `negotiate_auto` | bool | `true` |
| `orphanPort` | `orphan_port` | bool | `false` |
| `bandwidth` | `bandwidth` | int 1-100000000 | — |
| `inheritBandwidth` | `inherit_bandwidth` | int 1-100000000 | — |
| `pfc` | `pfc` | bool | `false` |

## Proposed change

Add the seven fields to `plugins/module_utils/models/interfaces/port_channel_access_interface.py` (policy model + argument spec) and the module `DOCUMENTATION`, mirroring the existing implementations in `port_channel_trunk_host_interface.py` (which already carries all seven with correct aliases, ranges, and descriptions — copy those field definitions).

Field ranges/enums must be verified against the `accessPoHost` template schema (done above), not assumed from the sibling — interface policy schemas differ per policyType.

## Coverage verification (the meta-question)

@akinross also asked how we verify option coverage per module. Proposal: add a unit test per interface model that diffs the model's field aliases against the properties of the corresponding template schema (vendored as a test fixture from the OpenAPI spec), so a missing or misspelled field fails CI instead of surfacing in review. If we adopt this, it can be a follow-up issue — flagging it here so the discussion has a home.

## Related

- PR #269 review thread: https://github.com/CiscoDevNet/ansible-nd/pull/269#discussion_r3496560998

### LINKED ISSUE #536: [ND 4.x] Extend nd_interface_port_channel_access for NX-OS and IOS-XE policy types
## Summary

Extend `nd_interface_port_channel_access` for Nexus Dashboard 4.3 Campus VXLAN EVPN with platform-aware support for NX-OS and IOS-XE interface policy types where ND 4.3 exposes them.

The module must not assume that a policy type supported by one operating system, switch role, model, or topology is valid for another. Applicability must be established from the ND 4.3 contract before implementation.

## Applicability Assessment - Required First

- [ ] Inventory the ND 4.3 policy templates and policy types applicable to this module.
- [ ] Record applicability by fabric type, `network_os_type`, switch role, model/release prerequisite, and topology.
- [ ] Classify each requested combination as `Supported`, `Partially Supported`, or `Not Applicable` with API/template evidence.
- [ ] Reject unsupported combinations before any controller read-modify-write or deployment request.
- [ ] Do not invent an IOS-XE equivalent when ND 4.3 does not expose one.

## Module-Specific Scope

Validate access port-channel templates, member identity, aggregation constraints, Catalyst EtherChannel applicability, Nexus port-channel behavior, and platform-specific access policy fields.

## Required Implementation

- Expose and validate `policy_type` and `network_os_type` using platform-aware discriminated models where applicable.
- Keep controller-response parsing tolerant of read-only/defaulted fields while keeping write validation strict.
- Normalize platform-specific identifiers and controller defaults so gathered output is replayable and repeat runs are idempotent.
- Group bulk create/update operations by switch and policy type where the ND API requires homogeneous payloads.
- Preserve predictable `merged`, `replaced`, `overridden`, `deleted`, and `gathered` behavior where those states are supported.
- Preserve check mode, sanitized before/after/diff output, per-item HTTP 207 failure handling, and staged/deploy behavior.
- Update module documentation, examples, return documentation, and the supported policy-type matrix.
- Add focused unit tests and Campus integration tests for Catalyst/IOS-XE, Nexus/NX-OS, mixed valid inputs, and invalid cross-platform inputs.

## Acceptance Criteria

- [ ] Every supported ND 4.3 policy type for this module is documented and validated.
- [ ] Cross-OS, unsupported-role, unsupported-model, and unsupported-topology combinations fail before controller mutation.
- [ ] Supported lifecycle states reconcile correctly and remain idempotent after ND injects defaults or reorders data.
- [ ] Gathered output uses the module argspec shape and is suitable for replay where gathered is supported.
- [ ] Mixed policy types are separated into valid API payloads rather than sent in one incompatible batch.
- [ ] Check mode performs no mutation and predicts the same normalized result as apply mode.
- [ ] Deployment-disabled operations do not call deploy endpoints; a later explicit deployment is validated where staging applies.
- [ ] Unit, sanity, documentation, and Catalyst/Nexus Campus integration validation pass with deterministic cleanup.

## References

- Loopback policy-type baseline: #452
- Shared mixed-policy bulk-operation behavior: #409

### LINKED ISSUE #537: [ND 4.x] Extend nd_interface_port_channel_trunk_host for NX-OS and IOS-XE policy types
## Summary

Extend `nd_interface_port_channel_trunk_host` for Nexus Dashboard 4.3 Campus VXLAN EVPN with platform-aware support for NX-OS and IOS-XE interface policy types where ND 4.3 exposes them.

The module must not assume that a policy type supported by one operating system, switch role, model, or topology is valid for another. Applicability must be established from the ND 4.3 contract before implementation.

## Applicability Assessment - Required First

- [ ] Inventory the ND 4.3 policy templates and policy types applicable to this module.
- [ ] Record applicability by fabric type, `network_os_type`, switch role, model/release prerequisite, and topology.
- [ ] Classify each requested combination as `Supported`, `Partially Supported`, or `Not Applicable` with API/template evidence.
- [ ] Reject unsupported combinations before any controller read-modify-write or deployment request.
- [ ] Do not invent an IOS-XE equivalent when ND 4.3 does not expose one.

## Module-Specific Scope

Validate trunk-host port-channel templates, member identity, aggregation constraints, allowed-VLAN handling, and platform-specific trunk policy fields.

## Required Implementation

- Expose and validate `policy_type` and `network_os_type` using platform-aware discriminated models where applicable.
- Keep controller-response parsing tolerant of read-only/defaulted fields while keeping write validation strict.
- Normalize platform-specific identifiers and controller defaults so gathered output is replayable and repeat runs are idempotent.
- Group bulk create/update operations by switch and policy type where the ND API requires homogeneous payloads.
- Preserve predictable `merged`, `replaced`, `overridden`, `deleted`, and `gathered` behavior where those states are supported.
- Preserve check mode, sanitized before/after/diff output, per-item HTTP 207 failure handling, and staged/deploy behavior.
- Update module documentation, examples, return documentation, and the supported policy-type matrix.
- Add focused unit tests and Campus integration tests for Catalyst/IOS-XE, Nexus/NX-OS, mixed valid inputs, and invalid cross-platform inputs.

## Acceptance Criteria

- [ ] Every supported ND 4.3 policy type for this module is documented and validated.
- [ ] Cross-OS, unsupported-role, unsupported-model, and unsupported-topology combinations fail before controller mutation.
- [ ] Supported lifecycle states reconcile correctly and remain idempotent after ND injects defaults or reorders data.
- [ ] Gathered output uses the module argspec shape and is suitable for replay where gathered is supported.
- [ ] Mixed policy types are separated into valid API payloads rather than sent in one incompatible batch.
- [ ] Check mode performs no mutation and predicts the same normalized result as apply mode.
- [ ] Deployment-disabled operations do not call deploy endpoints; a later explicit deployment is validated where staging applies.
- [ ] Unit, sanity, documentation, and Catalyst/Nexus Campus integration validation pass with deterministic cleanup.

## References

- Loopback policy-type baseline: #452
- Shared mixed-policy bulk-operation behavior: #409

allenrobel:
## Related Issue(s)

Closes #536
Closes #537
Closes #381
Closes #569
Refs #409 (the ethernet copy of the grouping helper moves to the shared base once #558 lands; loopback and both port-channel orchestrators use it here)

## Proposed Changes

- `nd_interface_port_channel_access` / `nd_interface_port_channel_trunk_host`: `config_data.network_os` becomes a `network_os_type` discriminated union with new `iosXeAccessPoHost` / `iosXeTrunkPoHost` policy models (ten-field strict subsets of the NX-OS templates; the 4.3.1-only `deviceTrackingPolicy` / `flowMonitors` are excluded, as in #558). `policy_type` is optional (derived from `network_os_type`) and is now visible in before/after output for the NX-OS branch too.
- The IOS-XE create body carries the canonical `Port-channel<N>` name (ND rejects the lowercase spelling on create but echoes and accepts it on GET/PUT/remove); identifiers and output stay lowercase. Vault: `xe-port-channel-create-requires-canonical-name`.
- A preflight refuses an IOS-XE port-channel whose member's current policy mode does not match (ND returns a flat 500 for it on 4.2.1), pointing at the ethernet module that converts the member. Vault: `xe-port-channel-member-mode-mismatch`.
- `(switch, policyType)` bulk-create grouping hoisted into `NDBaseInterfaceOrchestrator.bulk_create_groups`; loopback and both port-channel orchestrators use it.
- The IOS-XE trunk model carries `payload_defaults = {"allowedVlans": "none"}` (ND 4.3.1 requires it; the #564 class), which is why this PR is stacked on #566.
- #381: the seven `accessPoHost` fields (`bandwidth`, `inherit_bandwidth`, `link_type`, `negotiate_auto`, `orphan_port`, `pfc`, `port_type_edge_trunk`) with schema-verified ranges.
- Member names normalize through the shared cross-OS helper (`gi1/0/2` -> `GigabitEthernet1/0/2`; unknown families verbatim).
- IOS-XE port-channels are deleted from the switch (#569): `PortChannelBaseOrchestrator._delete_side_name()` queues the switch-canonical `Port-channel<N>` for both the `interfaceActions/remove` and the deploy of an `ios-xe` model. ND keys the intent record by the lowercase name it echoes but the discovered switch object by the canonical one; a remove naming the lowercase record drops the intent and detaches the members only, while the canonical name flips the record to `userDefined` and queues `no interface Port-channel<N>` (the GUI delete does the same through the canonical per-interface DELETE). `TODO(4.2.1) xe-port-channel-remove-leaves-switch-interface`.
- Integration: an `xe.yaml` scenario per target (gated on `nd_test_xe_switch_ip`) covering the negative member-mode case, merged create/idempotency/update, replaced, overridden and deleted on a Catalyst 9000v; the trunk-host gate sits after the pre-existing `vlan_mapping.yaml` include (disjoint IDs and members). The member-reset tasks retry for up to two minutes because ND 4.2.1's `operData.portChannelId` lags the removal deploy, and the scenarios poll the switch interface list until ND has discovered the deployed port-channels before `overridden` / `deleted` (see the ND limitation below).

## Test Notes

- Unit: `ndpytest tests/unit/` — 4808 passed. black / isort / pylint / mypy clean on every touched file; `ansible-test sanity --test validate-modules` and `--test pylint` clean on both modules; yamllint clean.
- ND 4.2.1.10 — SITE1 NX-OS regression + CAMPUS1 Catalyst 9000v XE scenarios: access ok=95 changed=36 failed=0; trunk_host ok=96 changed=36 failed=0. The switch `deploymentHistory` shows `no interface Port-channel102` at the overridden step and `no interface Port-channel101` at the deleted step; the Catalyst running config holds no `Port-channel` or `channel-group` lines afterwards.
- ND 4.3.1.175 (with #558 + #566 on the tree) — same targets, identical counts: access ok=95 failed=0; trunk_host ok=96 failed=0; same clean running config.
- Lab findings behind the design (2026-09-15, both releases): lowercase create rejected; `interfaceActions/remove` accepted for XE port-channels; member mode enforced in both directions; 4.3.1 requires `allowedVlans` but not `mtu`.
- Lab finding that shaped the scenarios: the Catalyst 9000v rejects a per-port-channel `mtu` at deploy time ("Command mtu 8000 is invalid") even though ND accepts it in the template, so the XE scenarios never set `mtu` (the option stays available; the device decides).

## ND limitation that remains (#569, narrowed)

ND generates the switch-side deletion of an IOS-XE port-channel only once it has discovered the deployed interface (the record's `operData` leaves `unknown` 25-45 s after the create deploy, both releases). A remove issued inside that window drops the ND record and the deploy pushes nothing; ND rediscovers the interface as a policy-less record seconds later, and a second `state: deleted` run then removes it (the canonical name in the remove body is what makes that second run work). Real playbooks delete long after creating, so the modules do not wait; the XE scenarios, which create and delete in one run, poll for discovery instead. NX-OS is unaffected. Vault note `xe-port-channel-remove-leaves-switch-interface` carries the 4.2.1 and 4.3.1 REST tables and the GUI capture.

## Behaviour changes visible to existing NX-OS playbooks

- `policy_type` now appears in `before` / `after` / `gathered` output (it was stripped before; the ethernet modules already emit it).
- `mtu` and `port_channel_mode` are validated by the Pydantic branch model instead of Ansible `choices` (the option set is the union of both platforms), so an invalid value fails at model validation with the model's error text.
- `state: overridden` in a fabric that also holds IOS-XE port-channels now converges those too (they were invisible before).
- An unrecognized member-name prefix passes through verbatim instead of being Title-cased (`e1/1` / `eth1/1` still become `Ethernet1/1`; `gi1/0/2` becomes `GigabitEthernet1/0/2`).
- On the IOS-XE branch `description` has the template's `minLength 1`, so an empty string cannot clear it (same class as #562/#563).

## Follow-ups (not in this PR)

- Ethernet member guards never fire for Catalyst members (`operData.portChannelId` stays -1; the parent is only in `policy.portChannelId`) — belongs with #558 / #561.
- Pre-existing: on a mixed 207 inside one bulk-create group, `PortChannelBaseOrchestrator.create_bulk` (and loopback) never deploy-queue the accepted items; `EthernetBaseOrchestrator` already has `_post_bulk_group` / `_queue_accepted_bulk_items` for this. Adopting it for port-channels must key the XE items case-insensitively (the create 207 echoes `Port-channel101` against the `port-channel101` identifier).
- The `_canonical_xe_create_name` serializer is duplicated across the two port-channel models; hoist into `ethernet_common` when a third consumer (vPC / #378) appears.

## Sequencing

Stacked on #566 (itself on #558): no CI until those merge; the local verification above substitutes.

4.2.1.10 and 4.3.1.175

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Release Classification

2.0.0 (missing functionality: IOS-XE port-channel policy types)

## Checklist

* [x] Latest commit is rebased from develop with merge conflicts resolved (via the #558 -> #566 stack)

https://claude.ai/code/session_01UEmeduuKXCEBhnU8bsHc37


plugins/module_utils/orchestrators/port_channel_base.py
```diff
@@ -224,26 +239,57 @@ def delete_bulk(self, model_instances: list[ModelType], **kwargs) -> None:
...
+
+        None
+        """
+        # TODO(4.2.1) xe-port-channel-remove-leaves-switch-interface
+        # ND keys the IOS-XE intent record by the lowercase name it echoes but the discovered switch object by `Port-channel<N>`.
+        # A remove naming the lowercase record drops the intent and detaches the members only; the configured `interface Port-channel<N>`
+        # stays on the Catalyst and ND's diff never lists it. A remove naming the canonical spelling flips the record to `userDefined`
+        # and queues `no interface Port-channel<N>` for the next deploy (lab-verified 2026-09-16 on 4.2.1.10; the GUI delete does the
+        # same through the per-interface DELETE). NX-OS deletes correctly with either spelling. The deletion is generated only once ND has
+        # discovered the deployed interface (25-45 s after the create deploy), which is a separate ND-side race this rewrite cannot close.
```
mikewiebe, line 265: ### Medium: Fail IOS-XE port-channel removal until ND discovery is complete

**Issue**


## @allenrobel Would like to get your feedback on this.  What are your thoughts on this problem and how we should handle it?  In theory we could just leave this alone and require a user to use state gathered in a task to make sure the interface is in the correct state before a task calls the module to delete it.

The canonical-name workaround only generates switch-side deletion after ND has discovered the newly deployed port-channel, 25–45 seconds later. In the existing interface inventory, a discovered interface reports `operData.operationalStatus` as `up` or `down`; a pre-discovery interface reports `unknown` or may not expose the field.

Runtime nevertheless queues canonical removal without enforcing that prerequisite, while only the integration scenario polls externally.

The module should treat discovery as a fail-fast precondition, not wait for it. Polling, sleeping, or retrying inside the module would hide an unbounded controller transition inside a mutation and make its runtime policy implicit.

**Evidence**

- [`port_channel_base.py` lines 247–271](https://github.com/CiscoDevNet/ansible-nd/pull/570/files#diff-554d1f33a508b4a264f5d327da30ece3337d1b63b17c697d40f7471b41827e44R247) says the rewrite cannot close the discovery race but still queues canonical removal without a readiness check.
- [`xe.yaml` lines 197–214](https://github.com/CiscoDevNet/ansible-nd/pull/570/files#diff-1e4f1c8151064103ce5a7d1be041bd630a2e3209844d2325552fc68418db61a6R197) waits for discovery because an earlier removal drops intent and leaves the switch object; the trunk scenario has the same compensating wait.
- The PR body and [issue #569's lab results](https://github.com/CiscoDevNet/ansible-nd/issues/569#issuecomment-5702710722) confirm that a removal during this window pushes nothing and requires a later retry after rediscovery.

**Existing PR overlap**

No matching existing PR comment was found. The PR body discloses the narrowed ND limitation, but there are no issue comments, reviews, or inline threads covering this required fail-fast behavior.

Tracked by open issue [#569](https://github.com/CiscoDevNet/ansible-nd/issues/569). The PR says it closes #569, but the early-delete branch remains, so closure is premature.

Open enhancement [#340](https://github.com/CiscoDevNet/ansible-nd/issues/340) separately tracks future `state: gathered` support for orchestrator-backed modules.

**Impact**

A create/deploy followed promptly by cleanup can return an empty `after` while `interface Port-channel<N>` remains configured on the Catalyst. The integration suite passes only because it adds polling behavior users do not receive from the module.

Internally polling would also impose extra requests and an arbitrary timeout on every affected removal.

**Suggested fix**

Reuse the raw interface inventory already fetched by the module; do not add another GET.

Before any mutation for `state: deleted` or `state: overridden`, preflight every IOS-XE port-channel that would be removed and require its normalized `operData.operationalStatus` to be exactly `up` or `down`.

Treat a missing, `unknown`, or unrecognized value as not ready, and fail the entire operation before queuing any remove or deploy request. Apply the same validation in check mode. For example:

```text
Cannot remove IOS-XE port-channel Port-channel101 because Nexus Dashboard
has not finished discovering it (operationalStatus=unknown). Retry after
operationalStatus becomes up or down.
```

Do **not** make either interface module poll, sleep, or retry for discovery.

Once [#340](https://github.com/CiscoDevNet/ansible-nd/issues/340) adds `state: gathered`, users who want waiting behavior can implement it explicitly as a separate read-only `gathered` task with Ansible `until` and `retries`, followed by the removal task.

Add tests showing that `up` and `down` proceed, while missing, `unknown`, and unrecognized values fail without writes or extra GETs.

plugins/module_utils/orchestrators/port_channel_base.py
mikewiebe, line 1: ### Medium: Policy-less rediscovered port-channels are invisible to cleanup

**Issue**

The recovery described for the pre-discovery race depends on a second explicit `state: deleted` run finding ND's rediscovered, policy-less IOS-XE port-channel. The observed record has `configData.networkOS` but no `policy`; `query_all()` silently filters it out because it retains only known managed `policyType` values. If ND returns the equally valid explicit form `policy: null`, the same chained lookup raises `AttributeError` instead.

`NDStateMachine` builds its entire `existing` collection from `query_all()` and skips an explicit deletion when the requested identifier is absent from that collection. The second-run recovery therefore never reaches `_delete_side_name()` or queues the canonical removal.

**Evidence**

- [`port_channel_base.py` lines 551–564](https://github.com/CiscoDevNet/ansible-nd/pull/570/files#diff-554d1f33a508b4a264f5d327da30ece3337d1b63b17c697d40f7471b41827e44R551) filters exclusively on `configData.networkOS.policy.policyType` and calls `.get()` on an explicit null policy without an `or {}` guard.
- [`port_channel_base.py` lines 259–265](https://github.com/CiscoDevNet/ansible-nd/pull/570/files#diff-554d1f33a508b4a264f5d327da30ece3337d1b63b17c697d40f7471b41827e44R259) makes the canonical deletion depend on later policy-less rediscovery.
- [`nd_state_machine.py` lines 94–96](https://github.com/CiscoDevNet/ansible-nd/blob/f72bac2e33eb2efea027f6f9070cb941d8ff1554/plugins/module_utils/nd_state_machine.py#L94-L96) constructs the current-state snapshot only from `query_all()`, and [lines 305–314](https://github.com/CiscoDevNet/ansible-nd/blob/f72bac2e33eb2efea027f6f9070cb941d8ff1554/plugins/module_utils/nd_state_machine.py#L305-L314) skip a requested delete when that snapshot has no matching item.
- An exact-head probe using `policy: null` raises:

```text
AttributeError: 'NoneType' object has no attribute 'get'
```

**Existing PR overlap**

No matching existing PR comment was found. The PR's proposed second-run recovery conflicts with the current query filter.

This is in the scope of [issue #569](https://github.com/CiscoDevNet/ansible-nd/issues/569), but its latest update says the second `state: deleted` run removes the policy-less record. The current implementation prevents that claimed recovery, and PR #570 declares that it closes #569.

**Impact**

The known missing-policy form leaves the orphan invisible and undeletable through either module, even with an explicit `state: deleted` task. An explicit-null form is worse: one such record aborts every access or trunk-host port-channel task targeting that switch.

**Suggested fix**

Use null-safe policy traversal (`...get("policy") or {}`). For explicit `state: deleted`, retain a policy-less IOS-XE port-channel only when its switch and canonicalized interface name match an item the user explicitly requested, so `_delete_side_name()` can queue its canonical removal. Do not make policy-less records candidates for broad `state: overridden` deletion, because the module cannot prove ownership from their missing policy data.

PR #571's `_policy_type_of()` helper demonstrates the appropriate null-safe parsing pattern. Add missing-policy and explicit-null tests proving that explicit deletion can reach the remove/deploy path, while `overridden` leaves ambiguous policy-less records alone.

plugins/module_utils/orchestrators/port_channel_base.py
```diff
@@ -172,40 +190,37 @@ def delete(self, model_instance: ModelType, **kwargs) -> None:
...
                 # Guarded at runtime by @requires_bulk_support("supports_bulk_create")
-                api_endpoint = self._configure_endpoint(self.create_bulk_endpoint(), switch_sn=switch_id)  # pyright: ignore[reportOptionalCall]
-                request_body = {"interfaces": [payload for interface_name, payload in items]}
+                api_endpoint = self._configure_endpoint(self.create_bulk_endpoint(), switch_sn=group_key.switch_id)  # pyright: ignore[reportOptionalCall]
+                request_body = {"interfaces": [item.payload for item in items]}
                 result = self._request(path=api_endpoint.path, verb=api_endpoint.verb, data=request_body)
                 results.append(result)
-                for interface_name, payload in items:
-                    self._queue_deploy(interface_name, switch_id)
+                for item in items:
```
mikewiebe, line 222: ### Medium: A mixed HTTP 207 create strands successful siblings without deployment

**Issue**

Within one homogeneous `(switch, policyType)` create group, `_request()` raises when any item in an HTTP 207 response fails.

`create_bulk()` queues deployments only after `_request()` returns successfully. Therefore, items that ND accepted successfully in the same 207 response are never queued for deployment when one sibling fails.

**Evidence**

- [`port_channel_base.py` lines 197–226](https://github.com/CiscoDevNet/ansible-nd/pull/570/files#diff-554d1f33a508b4a264f5d327da30ece3337d1b63b17c697d40f7471b41827e44R197) queues deployments only after the potentially raising request.
- [`nd_interface_port_channel_access.py` lines 314–329](https://github.com/CiscoDevNet/ansible-nd/pull/570/files#diff-e005d60b69237b2686dd7ff0a46429a31e70c9fde798ae7bd213452bd90be5e4R314) promises that when `config_actions.deploy=true`, accepted items are still deployed if another item fails.
- Existing tests cover an entire group succeeding before a later group fails. They do not cover successful and failed items within the same HTTP 207 response.
- The inherited `_accepted_multistatus_names()` helper is not called by either port-channel orchestrator.

**Existing PR overlap**

No matching PR comment was found. The PR body identifies this behavior as a follow-up rather than resolving it.

Issues [#536](https://github.com/CiscoDevNet/ansible-nd/issues/536) and [#537](https://github.com/CiscoDevNet/ansible-nd/issues/537) require correct per-item HTTP 207 handling and staged/deploy behavior.

Issue [#397](https://github.com/CiscoDevNet/ansible-nd/issues/397) concerns related response-status detection but does not cover queueing the accepted items.

**Practical example**

Suppose one task creates two IOS-XE access port-channels in the same bulk group:

```yaml
- name: Create two Catalyst access port-channels
  cisco.nd.nd_interface_port_channel_access:
    fabric_name: CAMPUS1
    config:
      - switch_ip: 192.168.12.181
        interface_name: port-channel101
        config_data:
          network_os:
            network_os_type: ios-xe
            policy:
              access_vlan: 100
              port_channel_mode: active
              ports:
                - GigabitEthernet1/0/2

      - switch_ip: 192.168.12.181
        interface_name: port-channel102
        config_data:
          network_os:
            network_os_type: ios-xe
            policy:
              access_vlan: 100
              port_channel_mode: active
              ports:
                - GigabitEthernet1/0/3

    config_actions:
      deploy: true
    state: merged
```

Both items are sent in one POST because they use the same switch and `iosXeAccessPoHost` policy type.

Assume another operation claims `GigabitEthernet1/0/3` after the module's preflight GET but before the POST. ND can return:

```json
{
  "results": [
    {
      "name": "Port-channel101",
      "status": "success",
      "message": "Interface added successfully"
    },
    {
      "name": "Port-channel102",
      "status": "failed",
      "message": "Member GigabitEthernet1/0/3 is already in use"
    }
  ]
}
```

Current behavior:

1. ND accepts and stores the intent for `Port-channel101`.
2. `_request()` notices the failure for `Port-channel102` and raises.
3. `create_bulk()` never reaches the loop that queues `port-channel101` for deployment.
4. The task fails—which is correct—but no deployment occurs for the successful item.
5. ND now contains `port-channel101`, while the Catalyst may not.
6. A retry sees the intent already present and may report no change, so the missing deployment is not repaired.

Result:

```text
Nexus Dashboard intent:
  port-channel101: present

Catalyst running configuration:
  interface Port-channel101: absent
```

Expected behavior:

1. Detect that `Port-channel101` succeeded in the fresh 207 response.
2. Match `Port-channel101` to the module's lowercase `port-channel101` identifier.
3. Queue that successful item for deployment.
4. Re-raise the failure for `Port-channel102`.
5. Let the failure finalizer deploy `port-channel101`.

The task still fails honestly because one requested item was rejected, but the accepted item honors `config_actions.deploy: true` and does not remain stranded.

**Impact**

ND can retain successfully accepted port-channel intent that was never deployed to the switch.

A later retry may see the intent as already present and report no change, leaving the controller and switch divergent without automatically repairing the missed deployment.

**Suggested fix**

When a fresh bulk-create request raises because of a mixed HTTP 207 response:

1. Inspect only that request's response body.
2. Identify exact-success items from `results[]`.
3. Correlate returned names case-insensitively because IOS-XE creation may return `Port-channel101` while the module identifier is `port-channel101`.
4. Queue the accepted `(interface_name, switch_id)` pairs for deployment.
5. Re-raise the original failure so the existing failure finalizer deploys those accepted items.

Do not use stale response data from a previous request.

Add tests covering successful and failed items within the same HTTP 207 group for both port-channel modules.

mikewiebe:
### Stack dependency: Ensure PR #558’s platform validation covers port-channel modules

The desired-OS versus actual-switch-platform gap is already documented in [PR #558’s review](https://github.com/CiscoDevNet/ansible-nd/pull/558#discussion_r4038216274). Since this branch is stacked through #566 on #558, I am not duplicating that finding here.

Please implement the #558 fix in a shared path that `PortChannelBaseOrchestrator` inherits—preferably `NDBaseInterfaceOrchestrator` plus the cached `FabricContext` platform index. A fix limited to `EthernetBaseOrchestrator` will not cover these port-channel modules.

After rebasing, add or retain port-channel tests proving that:

- `ios-xe` targeting an NX-OS switch fails before mutation.
- `nx-os` targeting an IOS-XE switch fails before mutation.
- Omitting `network_os_type` on a Catalyst does not silently use the NX-OS default.
- Normal and check mode return the same validation failure.
- Validation reuses the existing switch-inventory GET and adds no controller requests.

If the shared #558 fix and its inherited coverage satisfy these cases, no separate implementation is needed in this PR. Otherwise, the platform validation remains a blocker here.

plugins/module_utils/models/interfaces/port_channel_access_interface.py
```diff
@@ -206,122 +247,179 @@ def normalize_ports(cls, value):
...
+
+
+class XePortChannelAccessPolicyModel(InterfacePolicyStrictBase):
+    """
+    # Summary
+
+    Policy fields for the IOS-XE `iosXeAccessPoHost` template (`ios_xe_int_port_channel_access_host`). Maps to
+    `configData.networkOS.policy` where `policyType == "iosXeAccessPoHost"`. A strict subset of the NX-OS branch: `description` max length
+    is 200, `mtu` is an integer 1500-9198 (NX-OS takes the `default` / `jumbo` enum), `port_channel_mode` adds the PAgP `auto` /
+    `desirable` values, and there is no LACP / storm-control / QoS block. The 4.3.1-only `deviceTrackingPolicy` / `flowMonitors` fields
```
mikewiebe, line 268: ### Advisory: Apply PR #558’s ND 4.3.1 compatibility note to the port-channel modules

**Issue**

The IOS-XE access and trunk-host port-channel models deliberately omit the ND 4.3.1-only `deviceTrackingPolicy` and `flowMonitors` fields to preserve compatibility with ND 4.2.1.

The same documentation gap was already reported on [PR #558](https://github.com/CiscoDevNet/ansible-nd/pull/558#discussion_r4039457376). However, updating the Ethernet module documentation there will not update the separate port-channel module documentation introduced by this PR.

**Evidence**

- [`port_channel_access_interface.py` lines 261–270](https://github.com/CiscoDevNet/ansible-nd/pull/570/files#diff-69ba8563713299543a337708ba4cd27c127e05e8166838910c04c87f0bf5fff4R261) documents the deliberate omission only in source.
- [`port_channel_trunk_host_interface.py` lines 430–438](https://github.com/CiscoDevNet/ansible-nd/pull/570/files#diff-d2ab524cc8cd4727b8449727f59773a79668d8ac08e5617de6638ef13c4c74a4R430) records the same compatibility choice.
- Neither port-channel module’s public notes identify the unavailable properties.

**Existing PR overlap**

The full rationale is already documented in [PR #558 review comment r4039457376](https://github.com/CiscoDevNet/ansible-nd/pull/558#discussion_r4039457376). This advisory only requests equivalent disclosure in the two port-channel module documents.

No matching open issue was found.

**Impact**

Users on ND 4.3.1 can see these controller properties but cannot configure them through the strict port-channel modules. Generated Ansible documentation does not explain that their absence is an intentional cross-version compatibility decision.

**Suggested fix**

Add a note to both port-channel modules naming `deviceTrackingPolicy` and `flowMonitors` and explaining that they are intentionally unavailable while the modules maintain a shared ND 4.2.1/4.3.1 write contract.

No runtime or model change is requested.

plugins/module_utils/orchestrators/port_channel_access_interface.py
```diff
@@ -29,7 +32,8 @@ class PortChannelAccessInterfaceOrchestrator(PortChannelBaseOrchestrator):
     Orchestrator for port-channel accessPoHost interface CRUD operations on Nexus Dashboard.
 
     Inherits all shared port-channel logic from `PortChannelBaseOrchestrator`. Defines `model_class` as
-    `PortChannelAccessInterfaceModel` and manages the `accessPoHost` policy type.
+    `PortChannelAccessInterfaceModel` and manages both the NX-OS `accessPoHost` and the IOS-XE `iosXeAccessPoHost`
+    policy types (issue #536).
```
mikewiebe, line 36: ### Medium: Enable the existing cached capability preflight for access and trunk port-channels

**Issue**

The new access and trunk-host orchestrators only set `model_class`; neither opts into the shared `capableSwitches` preflight with `interface_type` and `interface_mode`. `NDBaseInterfaceOrchestrator.validate_switches_capable()` therefore returns immediately, even though the existing capability taxonomy supports both `portChannel/access` and `portChannel/trunk`.

This should be wired through the existing cached helper. It should not introduce a new endpoint implementation, repeat the switch-inventory GET, or query once per config item.

**Evidence**

- `PortChannelAccessInterfaceOrchestrator` and its trunk-host sibling leave the base capability ClassVars empty.
- The shared base already invokes capability validation from its state-machine-visible `preflight()`, including check mode; it only needs the subclasses to opt in.
- `FabricContext` already caches switch-IP-to-ID resolution, so repeated `_require_resolvable_switches()` calls add no switch-inventory request.
- `InterfaceCapabilityPreflight` already caches the `capableSwitches` response and derived switch-ID set by `(interface_type, mode)`. One GET covers every target switch for that mode in the module run; validation after that is in-memory set membership.

**Existing PR overlap**

No matching PR #570 comment or review thread found. The shared check-mode wiring precedent was addressed for loopbacks in PR #275.

Issues #536 and #537 require unsupported role, model, and topology combinations to fail before controller mutation. Issue #273 owns the shared capability infrastructure, and #302 defines its check-mode warning policy.

**Impact**

An unsupported switch can reach POST in normal mode. In check mode, the task does not perform the capability query and can make the request appear viable without even producing the established capability warning.

**Practical example**

Suppose ND reports that switch A supports `portChannel/access`, while switch B does not. A task targets switch B with `port-channel101` and `access_vlan: 100`.

Current behavior:

```text
capableSwitches check: skipped
normal mode: POST is attempted for switch B
check mode: no capability warning; the change appears viable
```

Expected behavior:

```text
normal mode: use one cached capableSwitches GET and fail before POST
check mode: use the same cached GET, warn under the existing #302 policy, and perform no mutation
```

With 10,000 requested interfaces across 1,000 switches, the request count must remain the existing cached switch-inventory GET plus one `capableSwitches` GET for the applicable mode—not one GET per interface or switch.

**Suggested fix**

Opt both subclasses into the existing helper:

```python
# PortChannelAccessInterfaceOrchestrator
interface_type: ClassVar[str] = "portChannel"
interface_mode: ClassVar[str] = "access"

# PortChannelTrunkHostInterfaceOrchestrator
interface_type: ClassVar[str] = "portChannel"
interface_mode: ClassVar[str] = "trunk"
```

Do not add another endpoint wrapper or any per-item lookup. Reuse the orchestrator-owned `InterfaceCapabilityPreflight` and `FabricContext` caches. Add normal/check-mode tests for both modules plus a scale-oriented test with multiple interfaces and switches asserting exactly one switch-inventory GET and one `capableSwitches` GET for the module mode.

plugins/module_utils/orchestrators/port_channel_base.py
allenrobel: Fixed in "nd_interface_{port_channel_*,loopback}: deploy-queue the items a mixed HTTP 207 bulk create accepted".

Rather than copy the ethernet orchestrators' `_post_bulk_group` a third time, the behaviour now lives once in
`NDBaseInterfaceOrchestrator._post_bulk_create_group`, and the port-channel and loopback `create_bulk` loops call it. It follows your
five steps: it reads only a response this request recorded (it snapshots `len(rest_send.responses)` first, so a sender exception cannot
surface the previous request's 207, issue #554), takes exact-`success` items only, matches `name` case-insensitively so `Port-channel101`
correlates with `port-channel101`, queues the module's identifier, and re-raises naming the accepted items so the finalizer deploys them.

Tests: four on the helper itself (mixed result with the canonical-case echo, all-success, stale response, status vocabulary) and one per
port-channel module for a mixed 207 inside a single `iosXeAccessPoHost` / `iosXeTrunkPoHost` group.

One thing to flag: loopback test 00460 pinned the opposite decision from the #403 review (no within-group recovery, because ND's handling
of the success subset was not lab-characterized then). It is rewritten to the rule the ethernet orchestrators adopted in the #550 review.
The SVI and subinterface layers (#571, #572) adopt the same helper; the routed port-channel (#577) inherits it.

plugins/module_utils/orchestrators/port_channel_base.py
allenrobel: Both halves confirmed and fixed in "nd_interface_port_channel_*: read a null policy safely and let an explicit delete see a policy-less
IOS-XE port-channel".

- `_policy_of` is null-safe at every level (`configData`, `networkOS`, `policy`) and `query_all` uses it, so an explicit `policy: null`
  no longer aborts the task.
- For `state: deleted` only, `query_all` keeps a policy-less `ios-xe` port-channel whose lowercase name the user named on that switch, so
  the state machine finds it and `_delete_side_name` queues the canonical remove. `state: overridden` and every other state never see it,
  for the reason you give: with no policy the module cannot prove ownership.

One detail the fix had to handle: the rediscovered record carries `configData.mode: unknown`, which the model's frozen `mode` literal
rejects, so the kept record is a copy without that key and the shared inventory cache stays untouched for `preflight`. Tests cover
the missing-policy form, the explicit-null form, a null `configData`, the `overridden` and `merged` exclusions, and the delete reaching
both queues with `Port-channel111`.

Lab-verified on 4.2.1.10 and 4.3.1.175 (2026-09-21): after a forced pre-discovery removal ND rediscovered the orphan within 22-26 s as
`{"configData": {"mode": "unknown", "networkOS": {"networkOSType": "ios-xe"}}}` (no `policy` key), and a bulk `interfaceActions/remove`
naming the canonical `Port-channel120` flipped it to `userDefined`, put `no interface Port-channel120` into the switch's pending
configuration, and the deploy removed it for good (watched 100 s+).

plugins/module_utils/models/interfaces/port_channel_access_interface.py
allenrobel: Done in "nd_interface_port_channel_*: note the ND 4.3.1 policy properties the modules do not expose". Both modules carry the same
scope statement as the ethernet modules in #558, naming `deviceTrackingPolicy` and `flowMonitors`.

plugins/module_utils/orchestrators/port_channel_base.py
allenrobel: Implemented as a fail-fast precondition, in "nd_interface_port_channel_*: refuse to remove an IOS-XE port-channel ND has deployed but not
yet discovered". No polling, sleeping or retrying; it runs for `state: deleted` and for the removal set of `state: overridden`, in check
mode too, and fails before any remove or deploy is queued, with the message you proposed. It reads the inventory the module already
cached, so a candidate whose `operationalStatus` is `up` or `down` costs no request.

It departs from the suggestion in one place, and the lab is why. A port-channel created with `config_actions.deploy: false` and never
deployed can never be discovered, and its record is indistinguishable from a deployed-but-undiscovered one (both releases, 2026-09-21):

| State | `operationalStatus` / description | `pendingConfig` |
|---|---|---|
| Staged, never deployed | `unknown` / `Not discovered` | lists `interface Port-channel120` |
| Deployed, not yet discovered | `unknown` / `Not discovered` | no mention |

Requiring `up` / `down` alone would have made staged intent undeletable, although removing it is harmless (verified: clean 207, nothing
rediscovered). So when, and only when, a removal candidate is not `up` / `down`, the guard reads that switch's `pendingConfig` once (one
GET per affected switch, however many candidates it has). A listed `interface <name>` line, matched case-insensitively, is staged intent
and proceeds; an unlisted candidate is deployed but undiscovered and fails the run. That is the one extra GET, confined to the case
where the module would otherwise have to guess.

Two notes on scope. It lives in `NDBaseInterfaceOrchestrator` behind an opt-in ClassVar (`xe_removal_requires_discovery`), set here on
the port-channel base and on the SVI and subinterface orchestrators in #571 / #572; the routed port-channel in #577 inherits it. And for
`state: overridden` it runs from `preflight` rather than `preflight_delete`, because the state machine runs no delete preflight for the
override set and removes only after its creates and updates; `nd_state_machine.py` is untouched.

One corner is let through and documented: an interface that is deployed, still undiscovered, and then edited without a deploy also shows
pending lines. The explicit-delete recovery from the other thread covers it, and that recovery is now lab-verified: a bulk remove naming
the canonical `Port-channel<N>` on the policy-less record queues `no interface Port-channel<N>`, and the deploy removes it for good.

With the early-delete branch closed and the recovery path working, I think `Closes #569` is now accurate; say if you see it differently.
Vault note: `xe-interface-removal-requires-discovery`. A `gathered` + `until` task once #340 lands remains the way to wait.

plugins/module_utils/orchestrators/port_channel_base.py
allenrobel: Follow-up after checking this in the lab, because the 207 recovery turned out to cover only half of it.

One create request with a valid `loopback207` and an invalid sibling (`loopback1024`, out of range; and separately `loopback100`, ND's multisite loopback), intent only, both releases:

| Release | Response | Valid sibling |
|---|---|---|
| ND 4.3.1.175 | HTTP 207, `loopback207` `success`, sibling `failed` | created, so an exact `success` item really is on the controller |
| ND 4.2.1.10 | flat HTTP 500 naming only the failing item, no `results[]` | **created anyway** |

On 4.2.1 there is nothing in the response to recover from, so the accepted sibling would still have been stranded. Added in "nd_interface_*: deploy-queue the items a bulk create committed although ND answered a flat HTTP 500": for any bulk-create failure that is not a 207, `_created_despite_failure` drops the switch's cached inventory, reads it once, and deploy-queues each submitted name that exists now and was absent from the inventory cached before the request. A name that already existed is never claimed (that is the `loopback100` case: ND refuses it as "already in use", and deploying a system loopback the user merely named would be wrong), the recovery is skipped when there is no cached inventory to compare against, and a failing re-read never masks the create error. One GET, on the failure path only. `TODO(4.2.1) bulk-interface-create-500-partial-commit`.

Verified end to end through `nd_interface_loopback` on 4.2.1.10: the task failed, the message ended "The controller created ['loopback207'] from the same request; their deploy stays queued", the finalizer deployed it (`deploymentHistory` shows `interface loopback207`), and `loopback100` was left alone. SVI, subinterface and routed port-channel inherit it through the shared helper.

Side note for #457: `RestSend` retried that 500 for the full timeout, re-POSTing the group each time, so the final message also says `loopback207 is already in use`. Harmless here, but it is the mutation-retry half of that issue.

plugins/module_utils/orchestrators/port_channel_access_interface.py
allenrobel: Done in "nd_interface_port_channel_{access,trunk_host}: opt in to the cached capability preflight": the two ClassVars you quote, nothing else at runtime. One `capableSwitches` GET per mode per run, served from the `InterfaceCapabilityPreflight` cache; switch resolution reads the `FabricContext` cache.

I checked the endpoint before opting in, because the routed ethernet orchestrator carries a deliberate opt-out (`capable-switches-empty-for-ethernet-on-vxlan`: the endpoint answers "where can the GUI wizard create one", and returns an empty list for every ethernet mode on a VXLAN fabric). On 2026-09-21, both releases, SITE1 (VXLAN EVPN) and CAMPUS1 (Campus VXLAN): `portChannel` with `access`, `trunk` and `routed`, and `svi` with `managed`, each list every switch in the fabric, Catalyst included; only the `ethernet` modes return `[]`. So for these modes the opt-in refuses nothing it should accept.

Tests, per module: the opt-in; four port-channels on two switches (two NX-OS, two Catalyst) in normal and check mode, asserting the first two responses are exactly one switches GET and one `capableSwitches` GET (the member-mode checks then read one interface list per switch, as before); and an incapable switch refused by name. The existing preflight tests take the one extra response through a shared fixture with a `TEST_NOTES` entry explaining why it is shared.

plugins/module_utils/orchestrators/port_channel_base.py
mikewiebe: ### Medium: Response tracking becomes increasingly slow as the number of switches grows

**Issue**

The module asks how many controller responses have been recorded before each group is created. However, obtaining that number currently makes a complete copy of every response collected so far.

Practical example:

1. A task targets 500 switches.
2. The initial interface queries place approximately 500 responses in memory.
3. Before creating the first port-channel group, the module copies all 500 responses.
4. Before the next group, it copies approximately 501 responses.
5. This continues through all 500 create groups.

The module can copy roughly **374,750 response objects** merely to obtain their counts. No controller requests are saved or improved by this copying.

A current-head benchmark demonstrated the growth:

| Switch groups | Local copying time |
|---:|---:|
| 25 | 0.260 seconds |
| 50 | 1.063 seconds |
| 100 | 4.304 seconds |
| 200 | 18.031 seconds |

These measurements exclude normal network and JSON-processing time.

**Evidence**

- [`base_interface.py` lines 333 and 341](https://github.com/CiscoDevNet/ansible-nd/pull/570/files#diff-1bdd24be418d1be85dcf48e8bd16da0b84e1c8aaaf0d001caf00e693ed4f6322R333) access `rest_send.responses` only to obtain its length.
- [`rest_send.py` lines 589–598](https://github.com/CiscoDevNet/ansible-nd/blob/2096ad812b1c01bf01f44f9b71d8a741dba80838/plugins/module_utils/rest/rest_send.py#L589-L598) make a complete deep copy of the response history whenever `responses` is accessed.
- Existing tests normally contain only zero to two responses, so they do not expose the large-fabric cost.

**ND/OpenAPI applicability**

**Existing PR overlap**

Related to this existing mixed-result review thread because the expensive response-count checks were introduced while addressing it. The existing comment does not explicitly cover this performance problem, so this is a follow-up.

The same problem was previously identified and corrected in [PR #563's response-history thread](https://github.com/CiscoDevNet/ansible-nd/pull/563#discussion_r4010047146).

No matching open issue found. Issue [#349](https://github.com/CiscoDevNet/ansible-nd/issues/349) concerns controller-query fan-out, but it does not track repeatedly copying the local response history.

**Impact**

Small tasks are unlikely to notice the difference, but large multi-switch tasks can spend many seconds copying response data locally before accounting for any controller latency. The cost rises rapidly as more switches and responses are added.

**Suggested fix**

Use the constant-time `RestSend.response_count` property added by PR #563 instead of reading and copying `RestSend.responses`.

Add a focused test that fails if this freshness check accesses the full `responses` property, ensuring future changes do not restore the expensive behavior.

plugins/module_utils/orchestrators/port_channel_base.py
mikewiebe: ### Medium: The removal safety check can still leave an interface on the switch

**Issue**

The module is supposed to stop an IOS-XE interface removal until ND reports the interface as fully discovered with a status of `up` or `down`.

The new code makes an additional `pendingConfig` request when the status is missing or `unknown`. If it finds the interface in pending configuration, it assumes the interface was never deployed and allows removal. That assumption is not always correct.

Practical example:

1. Create and deploy `Port-channel101`.
2. Before ND finishes discovering it, stage a description change with `deploy: false`.
3. The existing interface information still reports `operationalStatus: unknown`.
4. The staged description change causes `pendingConfig` to contain `interface Port-channel101`.
5. The module assumes the port-channel was never deployed and allows its removal.
6. ND removes the intent but cannot generate `no interface Port-channel101`, leaving the port-channel on the switch.

A later cleanup task is then required to remove the leftover interface.

**Evidence**

- [`base_interface.py` lines 492–510](https://github.com/CiscoDevNet/ansible-nd/pull/570/files#diff-1bdd24be418d1be85dcf48e8bd16da0b84e1c8aaaf0d001caf00e693ed4f6322R492) describe the additional request and explicitly acknowledge that a deployed interface with a later staged edit is allowed through.
- [`base_interface.py` lines 533–544](https://github.com/CiscoDevNet/ansible-nd/pull/570/files#diff-1bdd24be418d1be85dcf48e8bd16da0b84e1c8aaaf0d001caf00e693ed4f6322R533) allow removal whenever the pending configuration contains the interface name.
- The current tests cover never-deployed staged intent, but they do not prove that a pending line distinguishes it from a deployed interface with a later staged change.

**ND/OpenAPI applicability**

**Existing PR overlap**

This is a follow-up on the existing IOS-XE removal-readiness thread. The current implementation partially addresses the original issue, but it adds another GET and retains the ambiguous removal path described above.

Related open issue [#569](https://github.com/CiscoDevNet/ansible-nd/issues/569) tracks the broader IOS-XE behavior that can leave a rediscovered, policy-less port-channel after removal.

Issue [#340](https://github.com/CiscoDevNet/ansible-nd/issues/340) tracks future gathered-state support that users could employ for explicit polling before removal.

**Impact**

A successful Ansible task can remove the controller's intent without removing the actual interface from the switch. The requested state is therefore not achieved, and the next run must detect and clean up the leftover interface.

The additional request also adds one controller GET per affected switch when the status is not `up` or `down`.

**Suggested fix**

Use the interface information the module has already retrieved. Allow IOS-XE removal only when `operData.operationalStatus` is exactly `up` or `down`.

plugins/module_utils/orchestrators/port_channel_base.py
allenrobel: Confirmed and fixed at both sites in "nd_interface_*: read response_count for the bulk-create and remove freshness snapshots". `_post_bulk_create_group` (the pre-request snapshot and the failure-path read) and `remove_pending`'s two reads from #547 now use `RestSend.response_count`; no orchestrator on the stack reads `rest_send.responses` for its length any more.

The property is #563's, carried onto this branch verbatim as its own commit ("RestSend: add response_count as a copy-free freshness token") so this stack does not wait on #563's second approval. It drops out in one step when the stack rebases over #563.

On your second point, three scale regressions in `test_base_interface.py` seed the history with one inventory response per switch and patch `RestSend.responses` to raise on any read: 50 create groups on the success path (00791); a sender-raised create with no cached inventory, so the 207 branch sees no growth, the non-207 recovery is skipped without a GET, and nothing is claimed (00792); and `remove_pending` on both its success and sender-raised paths (00793). The #563 property test comes along as well. The full unit suite passes on this head and on the stack tip; #571, #572, #577, #578 and #579 are rebased and pushed.

plugins/module_utils/orchestrators/port_channel_base.py
allenrobel: Confirmed in the lab, and it was worse than the corner I documented: on 4.3.1.175 `pendingConfig` still listed `interface Port-channel120`
1.3 s after the create deploy had returned, so the pending-configuration test could let a deployed-but-undiscovered port-channel through
even without a staged edit. `pendingConfig` is gone from the guard.

I first tried the per-interface `diff`, and it fails your example the same way: after a deploy-then-staged-edit the diff reads every line
`insert` with an empty `runningConfig`, byte-identical to never-deployed intent (2026-09-22, both releases). What does separate the states
is the per-switch deployment history, which a staged edit does not rewrite. "nd_interface_*: decide an undiscovered IOS-XE removal by
deployment history, not pendingConfig" replaces the lookup: an undiscovered IOS-XE candidate costs one
`GET .../switches/{sn}/deploymentHistory?filter=entityName:<name>&sort=completeTimestamp:desc&max=10`, and the newest record whose
first pushed line is `interface <Name>` or `no interface <Name>` decides (companion pushes ND files under the same entity, an SVI's
`vlan <id>`, are skipped):

| Newest push for the entity | Verdict |
|---|---|
| none | never deployed, removal allowed |
| `no interface <Name>`, `status: success` | removed from the switch and re-staged, allowed |
| a create or update push, or a removal that did not succeed | on the switch, run fails before anything is queued |

Your sequence (create + deploy, stage a description change with `deploy: false`, `state: deleted`) now fails: the history's newest push
is still the create, verified on 4.2.1.10 and 4.3.1.175 with a member-less `Port-channel120`. The record lands within a second of the
deploy returning on both releases, so there is no window between a deploy and the history reflecting it. Up / down candidates still
cost nothing; the extra request is confined to the case where the record alone cannot decide, and refusing every `unknown` would have
made staged intent undeletable, which is the reason I keep the one request rather than the status-only rule. The verdict picks the
newest record by its own `completeTimestamp`, so it does not depend on the controller honoring `sort`, and it errs toward refusing (a
failed `no interface` counts as still on the switch). `pendingConfig` endpoint model removed, `deploymentHistory` model added with the
percent-encoded filter (ND 4.2.1 does not match a quoted name containing `/`; the bare form matches on both releases). Vault note
`xe-interface-removal-requires-discovery` has the new table.

Issue #380, opened 2026-07-02
nd_interface_svi: state merged can re-emit ND's default hsrpVersion on PUT (ND 500) — promised from_response() stripping was never implemented


allenrobel:
## Summary

Raised by Copilot review on PR #268 (https://github.com/CiscoDevNet/ansible-nd/pull/268#discussion_r3500646430) and verified against the code.

The PR #268 description lists this as workaround 1:

> `from_response()` strips `hsrpVersion`. ND's GET returns `hsrpVersion: 1` (integer) as a server-side default even when HSRP is unconfigured; re-emitting that on PUT — as either int `1` or string `"1"` — triggers a generic 500 ("unexpected error during policy execution"). `hsrpGroup: 1` round-trips fine, so only `hsrpVersion` is poisoned.

That stripping was never implemented. There is no `hsrpVersion` handling anywhere in `plugins/module_utils/` — no `from_response` override, no field validator, no serializer exclusion in `svi_interface.py`, the SVI orchestrator, or `models/base.py`.

## Failure mechanics (verified)

1. `state: merged` in `nd_state_machine.py` builds the final payload via `self.existing.merge(proposed_item)`.
2. `existing` is constructed by `from_response()` (plain Pydantic validation) from the GET body, so ND's server-side default `hsrpVersion: 1` lands in the model with `hsrp_version` in `model_fields_set` — indistinguishable from a user-set value.
3. `NDBaseModel.merge()` only skips fields that are `None` or unset on `other` (the proposed side); it does not remove values already present on `existing`. The merged model retains `hsrp_version=1`.
4. `to_payload()` serializes it, so a merged update that touches any other policy field PUTs `hsrpVersion` back to ND — the exact round-trip the PR description says returns HTTP 500 on ND 4.2.

Create (POST) and no-diff idempotent runs don't hit this; a real merged **update** on an SVI whose GET carries the defaulted `hsrpVersion` does.

## Proposed fix

Strip `hsrpVersion` on the read path (`from_response()` for the SVI model), as the PR description promised, so a server-defaulted value is never treated as existing intent and never round-trips onto PUT. Per CLAUDE.md workaround policy, the fix must:

- carry a `TODO(4.2.1) <slug>` marker at the stripping site, and
- add the corresponding bug-tracker vault note first (none exists yet — searched the vault; only `svi-routingtag-type-mismatch` and the generic `interface-get-field-normalization` notes cover SVI GET quirks).

[...]

PR #571, opened 2026-09-16
nd_interface_svi: add the IOS-XE iosXeSvi and iosXeSviShutNoShut policy types and coerce the 4.3.1 hsrpVersion echo

allenrobel:
## Related Issue(s)

Closes #540
Closes #380

## Proposed Changes

`nd_interface_svi` gains the IOS-XE policy types on the same discriminated-union pattern as the ethernet (#558) and port-channel (#570) modules, plus the #380 fix, which the lab turned into a different bug than the one filed.

- **Model** (`models/interfaces/svi_interface.py`): `config_data.network_os` is a discriminated union on `network_os_type` (injected as `nx-os` when omitted, so existing playbooks are unchanged). The NX-OS `SviPolicyModel` moves onto `InterfacePolicyStrictBase`, so IOS-XE-only fields on an NX-OS item (and vice versa) fail before any controller call. The IOS-XE branch is a `policy_type` union of `XeSviPolicyModel` (`iosXeSvi`, injected when omitted) and `XeSviShutNoShutPolicyModel` (`iosXeSviShutNoShut`, admin state only). The `iosXeIntVlanTemplate` field list is identical on ND 4.2.1 and 4.3.1 (no version branch). The user option `prefixv6` is kept on both branches (wire key `ipv6Prefix` on IOS-XE) so gathered output replays across platforms. New options: `network_os_type`, `policy_type`, `vlan_name`, and the `dhcp_servers` list (`server_ip_address` + `server_vrf` per entry), which mirrors the IOS-XE template's `dhcpServers` array; the NX-OS flat `dhcp_server_address1-3` / `vrf_dhcp1-3` options are unchanged.
- **Orchestrator** (`orchestrators/svi_interface.py`): `create_bulk` adopts `bulk_create_groups` (one POST per `(switch, policyType)`, #409): ND rejects an `interfaces[]` array that mixes `iosXeSvi` and `iosXeSviShutNoShut` ("Mixed policy types ... are not allowed in bulk interface creation", both releases). `query_all` manages the three policy types and tolerates the records a Catalyst switch list carries that the NX-OS filter never saw: the discovered `Vlan1` record with `policy: null` (the previous chained `.get()` would have raised `AttributeError` on every Catalyst switch) and `svi` records with no `configData`.
- **#380** (`hsrpVersion` re-emit): lab-verified on both controllers. On 4.2.1.10 the echoed `hsrpVersion: 1` round-trips onto PUT with 204, so the 500 the issue describes does not reproduce and no read-side stripping is needed. On 4.3.1.175 the GET echoes `hsrpVersion` as the **string** `"1"` on every managed SVI (HSRP configured or not), which fails the model's `Literal[1, 2]` on read (the module could not read any SVI on 4.3.1) and, if echoed back, is rejected by the 4.3.1 PUT gateway with HTTP 400 schema validation. Fix: a before-validator coerces a numeric-string `hsrp_version` to `int` (`TODO(4.3.1) svi-hsrpversion-string-echo-431`), mirroring the existing `routing_tag` coercion.
- **Docs**: `network_os_type` / `policy_type` options, per-option applicability lines, `vlan_name`, `dhcp_servers`, an IOS-XE example covering both IOS-XE policy types, notes updated. No API quirk text in DOCUMENTATION.

### ND discrepancies found and recorded (vault notes written first, markers reference their slugs)

| Marker | What ND does |
| --- | --- |
| `TODO(4.3.1) svi-hsrpversion-string-echo-431` | 4.3.1 echoes `hsrpVersion` as `"1"` (spec: integer enum); PUT with the string 400s. 4.2.1 echoes the int. 4.3.1 also stops echoing `hsrpGroup` / `pimDrPriority` defaults. |
| `TODO(4.2.1) xe-svi-dhcpservers-echo-keys` | 4.2.1 accepts `dhcpServers[].{serverIpAddress, serverVrf}` on write but echoes `{srvrAddr, srvrVrf}`; PUTting the echo back 500s. Fixed on 4.3.1 (echoes the spec keys). The item model reads both spellings and writes only the spec keys. |
| `TODO(4.2.1) xe-svi-dhcpservers-servervrf-required` | Both releases reject a relay entry without a non-empty `serverVrf` (omitted: 500 / 207 failed; empty: 400 `dhcpServers-srvrVrf`) although the spec marks nothing required and allows `minLength: 0`. `server_vrf` is required per entry. |

Also lab-verified, no workaround needed: ND stores an IOS-XE SVI under whatever name spelling it is given, looks it up case-insensitively, and a lowercase `vlan<N>` remove + deploy removes `interface Vlan<N>` from the Catalyst running config on both releases (unlike the port-channel case in #569), so the existing lowercase `vlan<N>` normalization is kept for both branches. ND 4.3.1 accepts an `iosXeSvi` create carrying only `policyType`, so no `payload_defaults` entry is needed (no #564-class field). `reverse_diff_defaults` for the IOS-XE branch = `adminState: true`, `ipRedirects: true` (the two defaults ND injects on the echo).

### Bulk-stage questions (CLAUDE.md)

No new bulk stage or queue is added. `create_bulk` keeps the existing per-group POST and deploy queueing; a mixed 207 is handled exactly as before (the group's `_request` raises on a failed item and the group's interfaces are not queued for deploy). The finalizer cannot resubmit a create. The except-path readers are untouched. Create result items are keyed by `name`.

## Test Notes

- Unit: `tests/unit/` 4839 passed (`ndpytest tests/unit/`). New SVI model tests 02000-02100 (enums, union selection and injection, cross-branch rejection, ShutNoShut branch, DHCP relay echo keys and required VRF, IOS-XE reverse-diff scrub, diff/merge over the relay list, `hsrp_version` coercion, 4.3.1-shaped NX-OS echo, description limits) and orchestrator tests 00450/00910/00920 (Catalyst switch list with `policy: null`, grouped bulk create, grouping keys).
- Sanity: `validate-modules` and `yamllint` clean via `ndtest`; black / isort / pylint / mypy clean on the changed files (mypy's three `create_bulk` return-type errors pre-exist on develop and on the loopback orchestrator).
- Integration: `nd_interface_svi` (SITE1 NX-OS scenarios + the new `xe.yaml` on the CAMPUS1 Catalyst 9000v leaf) — results below.

| Controller | Target | Result |
| --- | --- | --- |
| ND 4.2.1.10 (SITE1 S1_BG1 + CAMPUS1 C1_LE1 Catalyst 9000v) | `nd_interface_svi` (merged, replaced, overridden, deleted, hsrp_dhcp, xe) | ok=126 changed=37 failed=0 ignored=3 (the three deliberate negative tests) |
| ND 4.3.1.175 (SITE1 S3_BG1 + CAMPUS1 C3_LE1 Catalyst 9000v) | same | ok=126 changed=37 failed=0 ignored=3 |

The NX-OS scenarios are the pre-existing ones; on 4.3.1 they pass only with the `hsrpVersion` coercion (every SVI read failed before it). `xe.yaml` covers the cross-platform rejection, merged create of both IOS-XE policy types in one task (two POSTs), idempotency, merged update of the DHCP relay list, replaced, overridden, deleted, and a final read of the Catalyst running config confirming no `interface Vlan98x` remains.

4.2.1.10 and 4.3.1.175 (lab-verified on both)

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

* [ ] Latest commit is rebased from develop with merge conflicts resolved (stacked on #570 -> #566 -> #558; retargets to develop as the stack merges)

https://claude.ai/code/session_018LdsxVhpcrri7nbHXQxCcR


plugins/module_utils/orchestrators/svi_interface.py
mikewiebe, line 1: ### Medium: Fail IOS-XE SVI removal until ND discovery is complete

**Issue**

ND generates switch-side removal of a newly deployed IOS-XE SVI only after discovery changes `operData.operationalStatus` from `unknown` to `up` or `down`.

Runtime currently queues remove/deploy immediately without enforcing that prerequisite. Only the integration scenario polls for discovery.

The module should treat discovery as a fail-fast precondition. It should not poll, sleep, or retry internally.

**Evidence**

- [`xe.yaml` lines 222–238](https://github.com/CiscoDevNet/ansible-nd/pull/571/files#diff-181642ea787ffa8d1e63fa4ad08d203d79523979c4bba7cd59b820c7b6763b7aR222) explains the controller prerequisite and polls before overridden deletion.
- [`svi_interface.py` lines 132–147](https://github.com/CiscoDevNet/ansible-nd/pull/571/files#diff-2dae635a4e647c7bd023a37db128fd927c2bb1dd9a1b875873d2f627e9dc8363R132) queues individual removal and deployment without an operational-status guard.
- [`svi_interface.py` lines 179–194](https://github.com/CiscoDevNet/ansible-nd/pull/571/files#diff-2dae635a4e647c7bd023a37db128fd927c2bb1dd9a1b875873d2f627e9dc8363R179) has the same behavior for bulk deletion.

**Existing PR overlap**

The same controller prerequisite and fail-fast policy were posted for IOS-XE port-channels in [PR #570](https://github.com/CiscoDevNet/ansible-nd/pull/570#discussion_r4039873136).

If that fix is implemented generically in `NDBaseInterfaceOrchestrator`, PR #571 can inherit it after rebasing. A port-channel-specific implementation will not resolve the SVI path.

This is within [issue #540](https://github.com/CiscoDevNet/ansible-nd/issues/540), which owns complete IOS-XE SVI lifecycle behavior.

**Impact**

A create/deploy followed quickly by cleanup can report an empty `after` state after ND drops the intent while the Catalyst retains `interface Vlan<N>`.

The integration test passes only because it adds polling behavior that module users do not receive.

**Suggested fix**

Reuse the raw interface inventory already fetched by the module; do not add another GET.

Before any `state: deleted` or `state: overridden` mutation, preflight every IOS-XE SVI that would be removed:

- Permit removal only when `operData.operationalStatus` is `up` or `down`.
- Treat missing, `unknown`, or unrecognized status as not ready.
- Fail the entire operation before queuing any remove or deploy request.
- Apply the same validation in check mode.

Do **not** make the module poll, sleep, or retry. Once [issue #340](https://github.com/CiscoDevNet/ansible-nd/issues/340) adds `state: gathered`, users who want waiting behavior can add a separate read-only `gathered` task with Ansible `until` and `retries`, followed by the removal task.

Add tests proving that `up` and `down` proceed while missing, `unknown`, and unrecognized values fail without writes or additional GETs.

plugins/module_utils/orchestrators/svi_interface.py
```diff
@@ -161,22 +162,16 @@ def create_bulk(self, model_instances: list[SviInterfaceModel], **kwargs) -> Res
...
-                api_endpoint = self._configure_endpoint(self.create_bulk_endpoint(), switch_sn=switch_id)  # pyright: ignore[reportOptionalCall]
-                request_body = {"interfaces": [payload for interface_name, payload in items]}
+                api_endpoint = self._configure_endpoint(self.create_bulk_endpoint(), switch_sn=group_key.switch_id)  # pyright: ignore[reportOptionalCall]
+                request_body = {"interfaces": [item.payload for item in items]}
                 result = self._request(path=api_endpoint.path, verb=api_endpoint.verb, data=request_body)
                 results.append(result)
-                for interface_name, payload in items:
-                    self._queue_deploy(interface_name, switch_id)
+                for item in items:
+                    self._queue_deploy(item.interface_name, group_key.switch_id)
```
mikewiebe, line 174: ### Medium: A mixed HTTP 207 create strands accepted SVIs without deployment

**Issue**

Within one `(switch, policyType)` create group, `_request()` raises if any item in an HTTP 207 response fails. Because `create_bulk()` queues deployment only after `_request()` returns, successful siblings from that response are never deployed.

**Evidence**

- [`svi_interface.py` lines 149–177](https://github.com/CiscoDevNet/ansible-nd/pull/571/files#diff-2dae635a4e647c7bd023a37db128fd927c2bb1dd9a1b875873d2f627e9dc8363R149) can raise before queueing any successful group member.
- [`nd_interface_svi.py` lines 293–310](https://github.com/CiscoDevNet/ansible-nd/pull/571/files#diff-de731073b24ba71852b07e66406cadcfecd308e915581429604cfded12b2ad4dR293) promises that accepted items are still deployed when `config_actions.deploy: true`.
- Tests do not cover mixed successful and failed items within one 207 response.
- The inherited `_accepted_multistatus_names()` helper is not used here.

**Existing PR overlap**

No matching comment exists on PR #571. The same root cause was reported for port-channels in [PR #570](https://github.com/CiscoDevNet/ansible-nd/pull/570#discussion_r4040117890).

A generalized shared fix could satisfy both PRs after rebasing. A port-channel-only fix will not repair the SVI `create_bulk()` path.

Tracked by [issue #540](https://github.com/CiscoDevNet/ansible-nd/issues/540), which explicitly requires correct per-item HTTP 207 handling and staged/deploy behavior.

**Impact**

ND can retain successfully accepted SVI intent without deploying it. A retry may consider the intent already present and report no change, leaving ND and the switch divergent.

**Suggested fix**

On the fresh-response exception path:

1. Identify exact-success items from that response only.
2. Correlate them with the submitted SVI group.
3. Queue only those successful `(interface_name, switch_id)` pairs.
4. Re-raise the original failure so the existing failure finalizer deploys the accepted items.
5. Add mixed-result tests for both IOS-XE SVI policy types.

mikewiebe:
### Stack dependency: shared platform validation

PR #571 inherits the desired `network_os_type` versus actual switch `platformType` gap already raised on [PR #558](https://github.com/CiscoDevNet/ansible-nd/pull/558#discussion_r4038216274). The controller rejects an incompatible apply, but check mode can still approve it.

Please consume the fix through `NDBaseInterfaceOrchestrator` after rebasing. An Ethernet- or port-channel-only fix will not cover the SVI path. Reuse the existing cached switch-inventory GET—no additional request is needed. This requirement is tracked by [issue #540](https://github.com/CiscoDevNet/ansible-nd/issues/540).

plugins/module_utils/models/interfaces/svi_interface.py
```diff
@@ -190,21 +256,201 @@ def _validate_ip_prefix_paired(self) -> SviPolicyModel:
...
 
+class XeSviDhcpServerModel(NDNestedModel):
+    """
+    # Summary
+
+    One DHCP relay server entry of the IOS-XE `iosXeSvi` template's `dhcpServers` list. Maps to `configData.networkOS.policy.dhcpServers[]`.
+
+    ## Raises
+
+    None
```
mikewiebe, line 267: ### Medium: Enforce the OpenAPI address formats on new IOS-XE SVI fields

**Issue**

The new IOS-XE `ip`, `ipv6`, and DHCP `server_ip_address` fields accept arbitrary strings even though the Manage API requires IPv4 or IPv6 addresses.

Address/prefix pairing is enforced, but the address itself is not validated.

**Evidence**

- [`server_ip_address` is an unconstrained string](https://github.com/CiscoDevNet/ansible-nd/pull/571/files#diff-44eaf9433f4623a48c4570528b419ceb2dba0daef47a1242bbb074f1a2cbc5d1R259).
- [IOS-XE `ip` and `ipv6` are also unconstrained strings](https://github.com/CiscoDevNet/ansible-nd/pull/571/files#diff-44eaf9433f4623a48c4570528b419ceb2dba0daef47a1242bbb074f1a2cbc5d1R349).
- OpenAPI v1.1.411 declares `interfaceIp` and `interfaceServerIpAddress` as IPv4 and `iosXeIntVlanTemplate.ipv6` as IPv6.
- An exact-head probe successfully constructed and serialized:

```yaml
ip: not-an-ip
prefix: 24
ipv6: also-not-ipv6
prefixv6: 64
dhcp_servers:
  - server_ip_address: bad
    server_vrf: default
```

- [The module documentation](https://github.com/CiscoDevNet/ansible-nd/pull/571/files#diff-de731073b24ba71852b07e66406cadcfecd308e915581429604cfded12b2ad4dR274) also omits the model's enforced 1–32 character range for `server_vrf`.

**Existing PR overlap**

No matching existing PR comment was found on PR #571. The older NX-OS SVI model has similar pre-existing string debt, but these IOS-XE fields and the DHCP model are introduced by this PR.

PR #572 contains an analogous unposted finding for its new IOS-XE subinterface address fields.

Tracked by [issue #540](https://github.com/CiscoDevNet/ansible-nd/issues/540), which requires validation against the platform-specific SVI schema. No narrower open issue was found.

**Impact**

Malformed addresses survive Ansible and Pydantic validation. Check mode can approve them, while normal execution consumes a controller request and returns a less actionable API error.

**Suggested fix**

- Use the collection's strict host-address types for `ip` and `ipv6`, preserving the separate prefix fields and existing string wire format.
- Require an IPv4 address for `server_ip_address`.
- Add positive and negative tests for all three fields.
- Document the enforced 1–32 character range for `server_vrf`.

This is entirely local validation and requires no additional controller requests.

plugins/module_utils/orchestrators/svi_interface.py
```diff
@@ -40,7 +39,8 @@ class SviInterfaceOrchestrator(NDBaseInterfaceOrchestrator[SviInterfaceModel]):
     """
     # Summary
 
-    Orchestrator for SVI interface CRUD operations on Nexus Dashboard.
+    Orchestrator for SVI interface CRUD operations on Nexus Dashboard. Manages the NX-OS `svi` policy type and the IOS-XE `iosXeSvi`
+    and `iosXeSviShutNoShut` policy types (issue #540).
```
mikewiebe, line 43: ### Medium: Enable the existing cached capability preflight for SVIs

**Issue**

`SviInterfaceOrchestrator` does not set the capability identifiers required by the shared preflight. Because `interface_type` remains empty, `validate_switches_capable()` returns without checking whether the target switch supports managed SVIs.

The fix should reuse the existing cached helper. It should not add another endpoint implementation, repeat the switch-inventory GET, or query once per SVI or switch.

**Evidence**

- The SVI orchestrator defines its model and bulk support but does not set `interface_type` or `interface_mode`.
- `NDBaseInterfaceOrchestrator.validate_switches_capable()` returns immediately when `interface_type` is empty.
- The existing capability taxonomy already supports `svi/managed`.
- `FabricContext` caches switch-IP-to-ID resolution, so repeated `_require_resolvable_switches()` calls add no switch-inventory request.
- `InterfaceCapabilityPreflight` caches the response and derived switch-ID set by `(interface_type, mode)`. One `capableSwitches` GET covers all target switches for this module run; validation after that is in-memory set membership.

**Existing PR overlap**

No matching PR #571 comment or review thread found. PR #275 established the shared preflight/check-mode wiring, and PR #570 now has the equivalent cached port-channel finding.

Issue #540 requires unsupported SVI role, model, and topology combinations to fail before mutation. Issue #273 owns the shared capability infrastructure, and #302 defines its check-mode warning policy.

**Impact**

An unsupported switch can reach the SVI mutation endpoint. Check mode skips the capability query entirely and can make the configuration appear viable without the established warning.

**Practical example**

Suppose ND reports that switch A supports `svi/managed`, while switch B does not. A task targets switch B with `vlan333`, `ip: 10.33.3.1`, and `prefix: 24`.

Current behavior:

```text
capableSwitches check: skipped
normal mode: SVI POST is attempted for switch B
check mode: no capability warning; the change appears viable
```

Expected behavior:

```text
normal mode: reuse cached switch inventory, perform one cached svi/managed capability GET, and fail before POST
check mode: perform the same cached check, warn under the existing #302 policy, and perform no mutation
```

With 10,000 requested SVIs across 1,000 switches, the request count must remain the existing cached switch-inventory GET plus one `capableSwitches` GET for `svi/managed`—not one GET per SVI or switch.

**Suggested fix**

```python
class SviInterfaceOrchestrator(...):
    interface_type: ClassVar[str] = "svi"
    interface_mode: ClassVar[str] = "managed"
```

Reuse the orchestrator-owned `FabricContext` and `InterfaceCapabilityPreflight` instances. Do not add another endpoint wrapper or any per-item lookup. Add normal/check-mode tests and a scale-oriented test with multiple SVIs and switches asserting exactly one switch-inventory GET and one capability GET for the module run.

**ND specification note**

`capableSwitches` is unpublished in both pinned OpenAPI specifications. This change should use the existing helper that encapsulates that accepted risk; it should not add another endpoint path.

plugins/module_utils/orchestrators/svi_interface.py
allenrobel: Fixed in "nd_interface_svi: deploy-queue the SVIs a mixed HTTP 207 bulk create accepted". As you suggested, the fix is the generalized one:
`create_bulk` calls the shared `NDBaseInterfaceOrchestrator._post_bulk_create_group` added on #570. The new test is parametrized over both
IOS-XE policy types and uses a canonical-case echo (`Vlan990` for `vlan990`).

plugins/module_utils/models/interfaces/svi_interface.py
allenrobel: Fixed in "nd_interface_svi: validate the IOS-XE address fields as bare IPv4 / IPv6 addresses".

- `ip` uses `IPv4HostStrict`, `ipv6` uses a new `IPv6HostStrict` (the IPv6 counterpart, added to `models/types.py`), and the required
  `server_ip_address` runs `validate_ipv4_host_strict`. The strict forms reject CIDR instead of normalizing it, because the mask length
  lives in the sibling `prefix` / `prefixv6` fields and an embedded prefix could contradict it. Wire format unchanged.
- Your exact-head probe values (`not-an-ip`, `also-not-ipv6`, `bad`) are in the negative tests, along with CIDR input and the wrong
  address family; the 4.2.1 echo key `srvrAddr` is covered on the read side.
- The module docs state the bare-address form and the 1-32 character range on `server_vrf`.

The analogous subinterface fields are fixed on #572 in the same way.

plugins/module_utils/orchestrators/svi_interface.py
allenrobel: Done the generic way you suggested: the guard is `NDBaseInterfaceOrchestrator._check_xe_removal_discovered`, added on #570, and
"nd_interface_svi: refuse to remove an IOS-XE SVI ND has deployed but not yet discovered" opts the SVI orchestrator in. Fail-fast, no
polling, `state: deleted` and the `state: overridden` removal set, check mode included, nothing queued on failure.

The one departure from the suggestion is explained on the #570 thread: a staged, never-deployed SVI reads `unknown` exactly like a
deployed-but-undiscovered one, so an `unknown` candidate costs one `pendingConfig` GET for its switch, and a listed `interface Vlan<N>`
line (ND lists `Vlan985` for the record `vlan985`, so the match is case-insensitive) lets staged intent through. Lab-verified for SVIs
on 4.2.1.10 and 4.3.1.175. The test covers a discovered, a staged and a deployed-but-undiscovered SVI on one switch with a single GET.

plugins/module_utils/orchestrators/svi_interface.py
allenrobel: Done in "nd_interface_svi: opt in to the cached capability preflight": `svi` / `managed`, the existing helper, no new endpoint path. Lab check on 2026-09-21, both releases: `capableSwitches?interfaceType=svi&mode=managed` lists every switch of SITE1 and of CAMPUS1, Catalyst included, so the opt-in is safe for the fabric types we support. Tests: the opt-in; four SVIs on two switches in normal and check mode asserting exactly one switches GET and one `capableSwitches` GET; an incapable switch refused by name.

plugins/module_utils/orchestrators/svi_interface.py
mikewiebe: ### Medium: An SVI can still be left on the switch after Ansible removes its ND intent

**Issue**

PR #571 enables the shared IOS-XE removal check for SVIs. However, that check still treats an interface found in `pendingConfig` as safe to remove when its operational status is missing or `unknown`.

A deployed SVI with a later staged change can also appear in `pendingConfig`, so the module cannot safely use that result to conclude the SVI was never deployed.

Practical example:

1. Create and deploy `Vlan333`.
2. Before ND finishes discovering it, stage a description change with `deploy: false`.
3. ND still reports `operationalStatus: unknown`.
4. The staged description change puts `interface Vlan333` in `pendingConfig`.
5. A task runs `state: deleted`.
6. The module treats the pending line as proof that the SVI was never deployed and allows removal.
7. ND removes the intent but fails to generate `no interface Vlan333`, leaving the SVI on the Catalyst switch.

**Evidence**

- [`svi_interface.py` lines 70–78](https://github.com/CiscoDevNet/ansible-nd/pull/571/files#diff-2dae635a4e647c7bd023a37db128fd927c2bb1dd9a1b875873d2f627e9dc8363R70) enable the shared removal check for IOS-XE SVIs.
- The inherited [`base_interface.py` lines 492–510](https://github.com/CiscoDevNet/ansible-nd/blob/4b894711ab9b581d038786e57e93805d0b4064e5/plugins/module_utils/orchestrators/base_interface.py#L492-L510) explicitly acknowledge that a deployed, undiscovered interface with a later staged edit is allowed through.
- The inherited [`base_interface.py` lines 533–544](https://github.com/CiscoDevNet/ansible-nd/blob/4b894711ab9b581d038786e57e93805d0b4064e5/plugins/module_utils/orchestrators/base_interface.py#L533-L544) perform the additional request and allow removal whenever the interface appears in pending configuration.
- The same behavior remains present on current PR head `e92a75d` and current #570 base `4b89471`.

**ND/OpenAPI applicability**

**Existing PR overlap**

This is a follow-up on the existing SVI removal-readiness thread. The current implementation partially addresses the original issue, but it adds another GET and retains the ambiguous removal path.

Tracked by open issue [#540](https://github.com/CiscoDevNet/ansible-nd/issues/540), which covers the IOS-XE SVI feature implemented by this PR.

Issue [#340](https://github.com/CiscoDevNet/ansible-nd/issues/340) tracks future gathered-state support that users could use to wait for discovery before running the removal task. Issue [#569](https://github.com/CiscoDevNet/ansible-nd/issues/569) documents the related orphan behavior for IOS-XE port-channels but does not cover SVIs directly.

**Impact**

Ansible can report that the SVI intent was removed even though `interface Vlan333` remains configured on the switch. The requested state is not achieved, and another task is needed to detect and remove the leftover SVI.

The fallback also adds one controller GET per affected switch whenever the SVI status is not `up` or `down`.

**Suggested fix**

Use the interface information already retrieved by the module. Allow IOS-XE SVI removal only when `operData.operationalStatus` is exactly `up` or `down`.

allenrobel: Same fix as on #570, inherited: the guard now reads the switch's deployment history for an `unknown` SVI instead of `pendingConfig`,
and a deployed Vlan333 that was then edited without a deploy is refused because its newest push is still the create. Verified with a
port-channel on both releases; SVIs file the same record shapes (entity `vlan980`, pushes `interface Vlan980` / `no interface Vlan980`,
with the companion `vlan 980` / `no vlan 980` under `Vlan980`, which the verdict skips). The unit test for the three-SVI case now expects
one history GET per undiscovered SVI, so the staged vlan981 (no push in its history) passes and the deployed vlan982 fails.

PR #572, opened 2026-09-16
nd_interface_subinterface_managed: add the IOS-XE iosXeSubinterface and iosXeSubinterfaceShutNoshut policy types

allenrobel:
## Related Issue(s)

Closes #541

Stacked on #571 (`feat_540_svi_iosxe`); refs #409 (grouped bulk create), #564 (4.3.1 create-required template defaults).

## Proposed Changes

Extends `nd_interface_subinterface_managed` with the IOS-XE branch on the discriminated-union pattern of #558 / #570 / #571.

**Applicability (ND 4.2.1 and 4.3.1 OpenAPI, identical enum and template fields on both):**

| policyType | Template | Classification |
|---|---|---|
| `iosXeSubinterface` | ios_xe_int_subintf | Supported: adminState, description (1-200), extraConfig, ip, prefix (8-31), ipv6, ipv6Prefix (64-127), vlanId (1-4094), vrfInterface |
| `iosXeSubinterfaceShutNoshut` | ios_xe_int_subif_admin_state | Supported: adminState only |
| `iosXeInternalSubinterface` | ios_xe_int_subif_internal | Not applicable (ND-internal, like the NX-OS `subinterfaceInternal`) |
| `userDefined` | n/a | Out of scope (as in every interface module) |

- `config_data.network_os` is a union on `network_os_type` (injected as `nx-os` when omitted, existing playbooks unchanged); the NX-OS policy moves onto `InterfacePolicyStrictBase`, so cross-platform fields (mtu, routing tag, PIM, Netflow on IOS-XE) fail before any controller call.
- IOS-XE branch: `policy_type` union of `XeSubinterfacePolicyModel` (injected as `iosXeSubinterface`) and `XeSubinterfaceShutNoshutPolicyModel`, with the IOS-XE template ranges.
- `interface_name` accepts the Catalyst parent families: an abbreviated / lowercase prefix that matches exactly one canonical name (`Ethernet`, `Port-channel`, `GigabitEthernet`, `TenGigabitEthernet`, `TwentyFiveGigE`, ...) is expanded, anything else passes through verbatim. ND removes an IOS-XE subinterface from the switch only under its canonical spelling, so the expansion is what keeps `deleted` / `overridden` working: `TODO(4.2.1) xe-subinterface-remove-leaves-switch-interface` (vault note written; same class as the #569 port-channel finding).
- `create_bulk` uses the shared `bulk_create_groups` (one POST per switch and policy type); `query_all` manages the three policy types and tolerates the policy-less records a Catalyst switch list carries.
- NX-OS branch: `payload_defaults = {"mtu": 9216}` because ND 4.3.1 refuses a `subinterface` create that omits `mtu` where 4.2.1 defaulted it (`TODO(4.3.1) ethernet-create-required-fields-431`, the #564 class; vault note extended). Found by the 4.3.1 regression run of the existing NX-OS scenarios.
- `policy_type` is visible in before/after output for both branches; argspec gains `network_os_type` and `policy_type`.
- Docs: per-option applicability lines, the IOS-XE VRF guidance (omit `vrf_interface` for the global table; the Catalyst has no VRF named `default`), the controller-required `ip` / `vlan_id` on an `iosXeSubinterface` create, IOS-XE examples for merged and deleted.

**Four questions (no new bulk stage or queue):** the only bulk change is the create grouping; a mixed 207 is handled per item by the existing `create_bulk` / `_request` path exactly as in #571; the finalizer, except-path reader and remove-side keys are untouched.

## Test Notes

- Unit: `tests/unit/` 4886 passed (24 new tests: enums, union dispatch, cross-branch rejection, XE ranges, parent-name expansion, mtu default, grouped bulk create keys, Catalyst list shapes).
- Sanity: `validate-modules`, `yamllint` pass; `ansible-lint` passes on the new scenario; black / isort clean; pylint / mypy report only the pre-existing patterns the sibling orchestrators carry.
- Lab, `nd_interface_subinterface_managed` full target (NX-OS scenarios on SITE1 S1_TOR1 / S3_TOR1 with `Ethernet1/63` routedHost and `Port-channel10` l3Po parents, plus the new `xe.yaml` on CAMPUS1 C1_LE1 / C3_LE1 Catalyst 9000v, parent `GigabitEthernet1/0/2`):
  - ND 4.2.1.10: `ok=82 changed=22 failed=0 ignored=2` (the two ignored results are the cross-platform negative test)
  - ND 4.3.1.175: `ok=82 changed=22 failed=0 ignored=2`
  - `xe.yaml` converts the parent to an IOS-XE routed host and waits for discovery (ND checks the parent's DISCOVERED mode, not its intent), runs merged / replaced / overridden / deleted with both policy types, verifies the Catalyst running config is clean after `deleted` with abbreviated (lowercase) names, and resets the parent to the fabric-default trunk host.
- Raw-REST probes on both controllers (intent-only and deployed) established: `ip` + `vlanId` are required on an `iosXeSubinterface` create; `vrfInterface: default` fails at deploy on the Catalyst; the only echoed default is `adminState: true`; mixed policy types in one `interfaces[]` are rejected; a lowercase remove orphans the switch interface (canonical name required).

4.2.1 (also verified on 4.3.1)

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

* [x] Latest commit is rebased from develop with merge conflicts resolved (stacked on #571)

https://claude.ai/code/session_018LdsxVhpcrri7nbHXQxCcR


plugins/module_utils/orchestrators/subinterface_managed_interface.py
```diff
@@ -152,9 +155,11 @@ def create_bulk(self, model_instances: list[SubinterfaceManagedInterfaceModel],
         """
         # Summary
 
-        Create multiple managed L3 subinterfaces in bulk. Groups subinterfaces by switch and sends one POST per
-        switch with all subinterfaces in the `interfaces` array, reducing API calls from N to one-per-switch. Queues
-        deploys for all created subinterfaces for later bulk execution via `deploy_pending`.
+        Create multiple managed L3 subinterfaces in bulk. Groups subinterfaces by `(switch, policyType)` through the shared
+        `bulk_create_groups` (issue #409) and sends one POST per group with all of its subinterfaces in the `interfaces` array: ND
+        rejects an array that mixes policy types, which a Catalyst carrying both `iosXeSubinterface` and `iosXeSubinterfaceShutNoshut`
```
mikewiebe, line 198: ### Medium: Deploy successful subinterfaces from a mixed HTTP 207 response

**Issue**

When one bulk-create request contains both successful and failed subinterfaces, `_request()` raises as soon as it detects the failed item.

Deployment entries are queued only after `_request()` returns successfully. Therefore, subinterfaces ND accepted in the same response are never queued for deployment.

**Evidence**

- `subinterface_managed_interface.py:154-183` performs the bulk request before queueing deployments.
- A mixed HTTP 207 causes `_request()` to raise before the queueing loop runs.
- Existing tests do not cover success and failure inside the same response group.

**Existing PR overlap**

No matching comment or review thread exists on #572.

The same response-handling pattern was previously identified in #570 and #571, but #572 has its own orchestrator implementation and requires its own fix or a shared correction.

- #541 owns the managed-subinterface acceptance criteria.
- #397 is related to HTTP 207 response classification.

**Impact**

ND can store a successfully accepted subinterface while the switch never receives it. A retry may find matching ND intent and report no change, leaving ND and the switch inconsistent.

**Practical example**

A task creates two subinterfaces with deployment enabled:

```yaml
cisco.nd.nd_interface_subinterface_managed:
  fabric_name: CAMPUS1
  config:
    - switch_ip: 10.122.84.182
      interface_name: GigabitEthernet1/0/1.100
      config_data:
        network_os:
          network_os_type: ios-xe
          policy:
            policy_type: iosXeSubinterface
            vlan_id: 100
            ip: 10.100.0.1
            prefix: 24

    - switch_ip: 10.122.84.182
      interface_name: GigabitEthernet1/0/1.200
      config_data:
        network_os:
          network_os_type: ios-xe
          policy:
            policy_type: iosXeSubinterface
            vlan_id: 200
            ip: 10.200.0.1
            prefix: 24

  config_actions:
    deploy: true
  state: merged
```

ND returns:

```json
{
  "results": [
    {
      "name": "GigabitEthernet1/0/1.100",
      "status": "success"
    },
    {
      "name": "GigabitEthernet1/0/1.200",
      "status": "failed",
      "message": "VLAN is already in use"
    }
  ]
}
```

Current behavior:

```text
.100 intent: accepted by ND
.200 intent: rejected
task: fails
.100 deployment: never queued
switch: .100 remains absent
```

Expected behavior:

```text
.100 intent: accepted and queued for deployment
.200 intent: reported as failed
task: still fails honestly
failure finalizer: deploys accepted .100
```

**Suggested fix**

When `_request()` raises for a fresh HTTP 207 response:

1. Inspect only that response per-item results.
2. Correlate exact successful names with the requested models.
3. Queue deployment for those accepted items.
4. Re-raise the original failure.
5. Allow the existing failure finalizer to deploy the accepted subset.

Add a test with one success and one failure inside the same `(switch, policyType)` group.

**ND version applicability**

The defect applies to the per-item HTTP 207 behavior used by both ND 4.2.1 and ND 4.3.1; it is not a version-specific schema difference.

tests/integration/targets/nd_interface_subinterface_managed/tasks/xe.yaml
```diff
@@ -0,0 +1,401 @@
...
+      - nm_xe_merged_102.after | selectattr('interface_name', 'equalto', xe_parent ~ '.102') | list | length == 1
+
+# --- XE DISCOVERY WAIT (before overridden) ---
+# ND generates the switch-side removal of an IOS-XE interface only once it has discovered the deployed interface (operData leaves
+# `unknown` 5-30 s after the create deploy). Real playbooks delete long after creating; this scenario does both in one run, so wait here.
+
+- name: "XE DISCOVERY WAIT: Wait until ND has discovered .100, .101 and .102 on the Catalyst"
+  cisco.nd.nd_rest:
+    path: "/api/v1/manage/fabrics/{{ test_xe_fabric_name }}/switches/{{ xe_switch_id }}/interfaces"
+    method: get
```
mikewiebe, line 255: ### Medium: Reject subinterface removal until ND discovery is complete

**Issue**

A newly created IOS-XE subinterface may temporarily have a missing or `unknown` `operData.operationalStatus`.

The module currently allows removal during this discovery window. ND can delete its intent without generating the corresponding Catalyst-side removal, leaving the subinterface configured on the switch.

**Evidence**

- `subinterface_managed_interface.py:136-152` queues individual removals without checking discovery status.
- `subinterface_managed_interface.py:185-199` does the same for bulk removal.
- This integration scenario polls for discovery before cleanup, but module users do not receive that protection.

**Existing PR overlap**

No matching comment or review thread exists on #572. Equivalent findings exist for the port-channel and SVI modules, but this PR has a separate subinterface removal implementation.

- #541 owns the managed-subinterface lifecycle requirements.
- #340 tracks future gathered-state support that users could use for explicit polling.

**Impact**

The Ansible task can report that the subinterface was removed while the physical switch still contains it.

**Practical example**

Create and deploy the subinterface:

```yaml
cisco.nd.nd_interface_subinterface_managed:
  fabric_name: CAMPUS1
  config:
    - switch_ip: 10.122.84.182
      interface_name: GigabitEthernet1/0/1.100
      config_data:
        network_os:
          network_os_type: ios-xe
          policy:
            policy_type: iosXeSubinterface
            vlan_id: 100
            ip: 10.100.0.1
            prefix: 24
  config_actions:
    deploy: true
  state: merged
```

Immediately remove it:

```yaml
cisco.nd.nd_interface_subinterface_managed:
  fabric_name: CAMPUS1
  config:
    - switch_ip: 10.122.84.182
      interface_name: GigabitEthernet1/0/1.100
  config_actions:
    deploy: true
  state: deleted
```

At that moment ND reports:

```json
{
  "interfaceName": "GigabitEthernet1/0/1.100",
  "operData": {
    "operationalStatus": "unknown"
  }
}
```

Current result:

```text
ND intent: removed
Catalyst configuration:
  interface GigabitEthernet1/0/1.100
  encapsulation dot1Q 100
  ip address 10.100.0.1 255.255.255.0
```

Expected result:

```text
Cannot remove GigabitEthernet1/0/1.100 because ND has not finished
 discovering it (operationalStatus=unknown). Retry when the status is
 up or down. No changes were made.
```

**Suggested fix**

Reuse the raw interface inventory already fetched and cached by the orchestrator. Do not add another GET.

Before queueing any removal:

- Allow `operationalStatus: up`.
- Allow `operationalStatus: down`.
- Reject missing, `unknown`, or unrecognized values.
- Validate all requested removals before queueing any of them.
- Apply the same validation in check mode.
- Do not poll, sleep, or retry inside the module.

Users who want waiting behavior should eventually use a separate gathered-state task with Ansible `until` and `retries`.

**ND version applicability**

This discovery transition affects the IOS-XE lifecycle on both ND 4.2.1 and ND 4.3.1; it is not an OpenAPI schema difference.

plugins/modules/nd_interface_subinterface_managed.py
```diff
@@ -60,87 +65,126 @@
...
                     - Subinterface MTU.
                     - Valid range is 576-9216.
+                    - Applies when policy_type is C(subinterface).
                     type: int
                   vlan_id:
                     description:
                     - 802.1Q VLAN tag for the subinterface.
-                    - Valid range is 2-4094.
+                    - Valid range is 2-4094 for C(subinterface) and 1-4094 for C(iosXeSubinterface).
+                    - Required by the controller when creating an C(iosXeSubinterface).
```
mikewiebe, line 119: ### Medium: Enforce IOS-XE create requirements before mutation

**Issue**

The module documentation says a new full `iosXeSubinterface` requires `vlan_id` and `ip`, but the model declares both fields as optional.

When the subinterface does not already exist, incomplete configuration passes model and check-mode validation and reaches the controller. This validation must be create-specific so sparse updates to an existing subinterface remain valid.

**Evidence**

- The public option documentation here says `vlan_id` and `ip` are required by the controller on create.
- `subinterface_managed_interface.py:268-277` makes `vlan_id`, `ip`, and `prefix` optional.
- The singular and bulk create paths send payloads without enforcing those requirements.
- The shared create preflight verifies that a policy exists, but not whether the full IOS-XE policy has its create-required fields.

**Existing PR overlap**

No matching comment or review thread exists on #572.

Issue #541 owns the IOS-XE managed-subinterface acceptance criteria.

**Impact**

An incomplete playbook can pass local validation and check mode but fail only after an actual controller request.

**Practical example**

Assume this subinterface does not exist:

```yaml
cisco.nd.nd_interface_subinterface_managed:
  fabric_name: CAMPUS1
  config:
    - switch_ip: 10.122.84.182
      interface_name: GigabitEthernet1/0/1.100
      config_data:
        network_os:
          network_os_type: ios-xe
          policy:
            policy_type: iosXeSubinterface
            admin_state: true
  config_actions:
    deploy: false
  state: merged
```

Current local payload:

```json
{
  "interfaceName": "GigabitEthernet1/0/1.100",
  "configData": {
    "networkOS": {
      "networkOSType": "ios-xe",
      "policy": {
        "policyType": "iosXeSubinterface",
        "adminState": true
      }
    }
  }
}
```

Missing create data:

```text
vlanId
ip
prefix
```

Current behavior:

```text
model validation: passes
check mode: can report the create as viable
normal mode: sends the incomplete request to ND
controller: rejects the create
```

Expected behavior:

```text
Cannot create IOS-XE subinterface GigabitEthernet1/0/1.100:
vlan_id, ip, and prefix are required for a new iosXeSubinterface.
No changes were made.
```

This existing-interface update must remain valid:

```yaml
policy:
  policy_type: iosXeSubinterface
  admin_state: false
```

The separate admin-only template must also remain valid without addressing fields:

```yaml
policy:
  policy_type: iosXeSubinterfaceShutNoshut
  admin_state: false
```

**Suggested fix**

Use the state machine already-computed create subset and existing inventory snapshot; do not add another GET.

For an absent interface using the full `iosXeSubinterface` policy:

- Require `vlan_id`.
- Require `ip`.
- Require its paired `prefix`.
- Aggregate all incomplete create items into one error.
- Apply the same validation in normal and check mode.

Do not apply the requirement to sparse updates of an existing subinterface, `state: deleted`, or the `iosXeSubinterfaceShutNoshut` admin-only policy.

Add tests covering incomplete creates, valid full creates, sparse updates, admin-only policies, bulk input, and check-mode parity.

**ND version applicability**

Neither ND 4.2.1 nor ND 4.3.1 marks these fields as required in the published schema. This requirement comes from the public module documentation and verified controller create behavior, so it should be treated as a live cross-version create requirement rather than an OpenAPI requirement.

plugins/module_utils/models/interfaces/subinterface_managed_interface.py
```diff
@@ -169,21 +232,145 @@ def _validate_ip_prefix_paired(self) -> SubinterfaceManagedPolicyModel:
...
+        """
+        return default_policy_type(data, XeSubinterfacePolicyTypeEnum.IOS_XE_SUBINTERFACE.value)
+
+    description: AsciiDescription = Field(default=None, alias="description", min_length=1, max_length=200, description="Subinterface description")
+    extra_config: str | None = Field(default=None, alias="extraConfig", description="Additional CLI for the subinterface")
+    vlan_id: int | None = Field(default=None, alias="vlanId", ge=1, le=4094, description="802.1Q VLAN tag for the subinterface")
+    vrf_interface: str | None = Field(
+        default=None, alias="vrfInterface", min_length=1, max_length=32, description="Interface VRF name; use `default` for default VRF"
+    )
+    ip: str | None = Field(default=None, alias="ip", description="IPv4 address of the subinterface")
```
mikewiebe: ### Medium: Validate IOS-XE IPv4 and IPv6 address syntax locally

**Issue**

The IOS-XE managed-subinterface model treats `ip` and `ipv6` as unrestricted strings.

Its pair validators only check that an address and prefix appear together. They do not check whether the address is actually valid.

**Evidence**

- `subinterface_managed_interface.py:274-277` defines both address fields as strings.
- Supplying a numeric prefix satisfies the pair validator even when the address is malformed.
- Both ND 4.2.1 and ND 4.3.1 schemas identify these fields as IPv4 and IPv6 values.

**Existing PR overlap**

No matching comment or review thread exists on #572.

Issue #541 owns strict validation for the new IOS-XE managed-subinterface policy.

**Impact**

Invalid addresses pass model validation and check mode. Normal execution discovers the error only after making a controller request.

**Practical example**

This configuration includes every create-required field, but both addresses are invalid:

```yaml
cisco.nd.nd_interface_subinterface_managed:
  fabric_name: CAMPUS1
  config:
    - switch_ip: 10.122.84.182
      interface_name: GigabitEthernet1/0/1.100
      config_data:
        network_os:
          network_os_type: ios-xe
          policy:
            policy_type: iosXeSubinterface
            vlan_id: 100
            ip: 999.999.999.999
            prefix: 24
            ipv6: not-an-ipv6
            ipv6_prefix: 64
  config_actions:
    deploy: false
  state: merged
```

Current behavior:

```text
ip and prefix are both present: passes pair validation
ipv6 and ipv6_prefix are both present: passes pair validation
check mode: can report the configuration as viable
normal mode: sends malformed addresses to ND
controller: rejects the request
```

Expected behavior:

```text
Invalid IPv4 address for ip: 999.999.999.999
Invalid IPv6 address for ipv6: not-an-ipv6
No changes were made.
```

A correct configuration should continue to pass:

```yaml
policy:
  policy_type: iosXeSubinterface
  vlan_id: 100
  ip: 10.100.0.1
  prefix: 24
  ipv6: 2001:db8:100::1
  ipv6_prefix: 64
```

**Suggested fix**

Use the collection shared strict address types for proposed configuration:

- `IPv4Host` or the appropriate strict IPv4 host type for `ip`.
- `IPv6Host` or the appropriate strict IPv6 host type for `ipv6`.

Retain the existing address/prefix pair validators.

Add tests proving that malformed IPv4 and IPv6 values fail locally, valid addresses serialize correctly, normal and check mode return the same validation error, controller-response parsing remains tolerant of legitimate echoed values, and no controller GET or mutation is added for this validation.

**ND version applicability**

Both ND 4.2.1 and ND 4.3.1 specify the corresponding IPv4 and IPv6 formats, so this finding applies equally to both supported versions.

plugins/module_utils/orchestrators/subinterface_managed_interface.py
allenrobel: Already on the branch, pushed the day after this comment: "nd_interface_subinterface_managed: deploy-queue the items a mixed HTTP 207 create accepted and validate the IOS-XE addresses". It is the shared correction rather than a fourth copy: `create_bulk` calls `NDBaseInterfaceOrchestrator._post_bulk_create_group`, added on #570 for the same finding there. It follows your five steps: it reads only a response this request recorded (it snapshots `len(rest_send.responses)` first, issue #554), takes exact-`success` items, matches `name` case-insensitively, queues the module's identifier, and re-raises naming the accepted items so the finalizer deploys them. `test_subinterface_managed_orchestrator_00930` is your example: two `iosXeSubinterface` items in one group, `.100` `success` (echoed in a different case) and `.101` `failed`, and only `.100` ends up deploy-queued.

One correction to the applicability note, from the lab on 2026-09-21: the two releases do not answer a mixed request the same way. For one create request holding a valid and an invalid item, ND 4.3.1.175 returns the 207 above, but ND 4.2.1.10 returns a flat HTTP 500 that names only the failing item, with no `results[]`, and still commits the valid one. So the 207 path alone would have left 4.2.1 stranded. The helper covers that shape too: for any non-207 failure it re-reads the switch inventory once and queues the submitted names that exist now and did not before ("nd_interface_*: deploy-queue the items a bulk create committed although ND answered a flat HTTP 500", on #570; details and the lab table are in my follow-up on the #570 thread). I reproduced it with loopbacks, not subinterfaces; the mechanism is in the shared helper, so this orchestrator inherits it.

tests/integration/targets/nd_interface_subinterface_managed/tasks/xe.yaml
allenrobel: Done in "nd_interface_subinterface_managed: refuse to remove an IOS-XE subinterface ND has deployed but not yet discovered". The orchestrator opts in (`xe_removal_requires_discovery`) to `NDBaseInterfaceOrchestrator._check_xe_removal_discovered`, added on #570: fail-fast, no polling or retry, all requested removals validated before any is queued, `state: deleted` and the removal set of `state: overridden`, check mode included, with the message you proposed. It reads the cached inventory, so an `up` / `down` candidate costs no request. `test_subinterface_managed_orchestrator_00940` covers a discovered and a deployed-but-undiscovered sibling on one switch.

It departs from the suggestion in one place, explained with the lab data on the #570 thread: a subinterface created with `config_actions.deploy: false` and never deployed can never be discovered, and its record is indistinguishable from a deployed-but-undiscovered one (`unknown` / `Not discovered` in both cases). Rejecting every `unknown` would make staged intent undeletable. So an `unknown` candidate, and only that, costs one `pendingConfig` GET for its switch: a listed `interface <name>` line is staged intent and proceeds, an unlisted candidate fails the run.

Being straight about what is and is not verified for subinterfaces specifically: the deployed case is exercised by this target on both controllers (ok=82 on 4.2.1.10 and on 4.3.1.175 with the guard active). The staged case is lab-verified for port-channels and SVIs on both releases, but I could not probe it intent-only for a subinterface, because ND refuses the create unless the parent's discovered mode is already routed. I expect the pending line to read `interface GigabitEthernet1/0/2.100`, as the other two families do, and have not seen it.

The scenario's own wait stays (it creates and removes in one run), now for up to ten minutes under a `TODO(4.3.1) xe-interface-removal-requires-discovery` marker: ND 4.3.1.175 took about two and a half minutes to discover three subinterfaces on 2026-09-21, against under 30 s on 4.2.1.10.

plugins/module_utils/models/interfaces/subinterface_managed_interface.py
allenrobel: Done in the same commit as the first thread. `ip` is `IPv4HostStrict` and `ipv6` is a new `IPv6HostStrict` (added to `models/types.py` on #571, where the SVI fields got the same treatment); the pair validators are unchanged and the wire format is the string the user gave.

I used the strict forms rather than `IPv4Host` / `IPv6Host`: those accept CIDR and strip the prefix, and here the mask length lives in the sibling `prefix` / `ipv6_prefix` fields, so `10.100.0.1/16` with `prefix: 24` would silently lose the `/16`. The strict types reject it instead. `test_subinterface_managed_interface_02090` covers, for each field, a valid bare address serialized unchanged, free text, CIDR input, and the other address family. Your `999.999.999.999` and `not-an-ipv6` fail the same validators.

On two of the tests you list: the validation runs when the config is parsed into the model, before the state machine looks at check mode, so normal and check mode cannot differ, and it adds no request. Read-side tolerance is covered by the lab rather than by a unit test: the full target reads these subinterfaces back on both controllers and passed with identical counts (ok=82), so ND's echo of `ip` / `ipv6` is a bare address on 4.2.1.10 and 4.3.1.175. I can add a `from_response` unit test pinning that if you would like it in the suite.

The docs state the bare-address form for `iosXeSubinterface`.

plugins/modules/nd_interface_subinterface_managed.py
allenrobel: Fixed in "nd_interface_subinterface_managed: require vlan_id and ip before creating an IOS-XE subinterface", along the lines you describe: a `preflight_create` override on the orchestrator, which the state machine calls with its already-computed create subset (no GET), before the check-mode gate, so normal and check mode behave the same. For a new full `iosXeSubinterface` it requires `vlan_id` and `ip`, and aggregates every incomplete item into one message that names the missing fields per item. `prefix` needs no separate check: the model's existing pair validator already rejects an `ip` without it.

It deliberately stays out of the model. When we found the requirement in the lab (2026-09-16: `400 Validation failed for following fields: [ip, vlanId]`, identical on 4.2.1.10 and 4.3.1.175) we kept both fields optional so a sparse `merged` update of an existing subinterface stays valid, and that still holds: the hook never sees an item that already exists. `iosXeSubinterfaceShutNoshut` and the NX-OS branch are not checked, and the inherited policy-less guard still runs first.

Tests: one with three create items (`admin_state` only; `vlan_id` without `ip`; complete), asserting the two incomplete ones are named with their missing fields, the complete one is not, and no request is made; one asserting a complete `iosXeSubinterface`, an admin-only `iosXeSubinterfaceShutNoshut` and an NX-OS subinterface all pass while a policy-less item still raises. What I did not add as unit tests: the sparse-update case and check-mode parity, because both are properties of where the state machine calls the hook (create subset only, ahead of the check-mode gate), which its own tests cover. Marker `TODO(4.2.1) xe-subinterface-create-requires-ip-vlanid`, with a vault note, since neither release's schema marks the fields required.

The full `nd_interface_subinterface_managed` target passes with this on the tree, ok=82 failed=0 on both ND 4.2.1.10 and 4.3.1.175 (2026-09-21).

tests/integration/targets/nd_interface_subinterface_managed/tasks/xe.yaml
mikewiebe: ### Medium: A managed subinterface can still be left on the switch after Ansible removes its ND intent

**Issue**

PR #572 enables the shared IOS-XE removal check for managed subinterfaces. That check allows an interface with missing or `unknown` operational status to be removed when its name appears in `pendingConfig`.

A deployed subinterface with a later staged change can also appear in `pendingConfig`. The module can therefore mistake a deployed interface for intent that was never deployed.

Practical example:

1. Create and deploy `GigabitEthernet1/0/1.100`.
2. Before ND finishes discovering it, stage a description or other configuration change with `deploy: false`.
3. ND still reports `operationalStatus: unknown`.
4. The staged change places `interface GigabitEthernet1/0/1.100` in `pendingConfig`.
5. Run `state: deleted`.
6. The module assumes the subinterface was never deployed and allows removal.
7. ND removes its intent without generating the switch-side removal, leaving the subinterface configured on the Catalyst.

**Evidence**

- [`subinterface_managed_interface.py` lines 75–78](https://github.com/CiscoDevNet/ansible-nd/pull/572/files#diff-f56259a130e20d7204d44a8d1a24069f4c2e102ea39b0f582e0de66fbb38126dR75) enable the shared IOS-XE removal check for managed subinterfaces.
- The inherited [`base_interface.py` lines 492–510](https://github.com/CiscoDevNet/ansible-nd/blob/4b894711ab9b581d038786e57e93805d0b4064e5/plugins/module_utils/orchestrators/base_interface.py#L492-L510) explicitly acknowledge that a deployed, undiscovered interface with a later staged edit is allowed through.
- The inherited [`base_interface.py` lines 533–544](https://github.com/CiscoDevNet/ansible-nd/blob/4b894711ab9b581d038786e57e93805d0b4064e5/plugins/module_utils/orchestrators/base_interface.py#L533-L544) allow removal whenever the interface appears in pending configuration.
- The author's response states that the never-deployed subinterface case was inferred from the port-channel and SVI behavior rather than directly reproduced for subinterfaces.

**ND/OpenAPI applicability**

**Existing PR overlap**

This is a follow-up on the existing managed-subinterface removal-readiness thread. The current implementation partially addresses the original issue, but it adds another GET and retains the ambiguous removal path.

Tracked by open issue [#541](https://github.com/CiscoDevNet/ansible-nd/issues/541), which owns the managed-subinterface lifecycle implemented by this PR.

Issue [#340](https://github.com/CiscoDevNet/ansible-nd/issues/340) tracks future gathered-state support that users could use to wait for discovery before running the removal task.

**Impact**

Ansible can report that the subinterface intent was removed while the corresponding subinterface remains configured on the switch. The desired state is not reached, and another task is required to find and remove the leftover interface.

The fallback also adds one controller GET per affected switch whenever the status is not `up` or `down`.

**Suggested fix**

Use the interface information already retrieved by the module. Allow IOS-XE subinterface removal only when `operData.operationalStatus` is exactly `up` or `down`.

allenrobel: Same fix as on #570, inherited: the deployment history decides an `unknown` subinterface, not `pendingConfig`, so a deployed
GigabitEthernet1/0/1.100 with a later staged edit is refused (its newest push is the create). Subinterfaces are keyed in the history by
the switch spelling (`GigabitEthernet1/0/2.100`, pushes `interface GigabitEthernet1/0/2.100` / `no interface GigabitEthernet1/0/2.100`,
both releases); the filter is sent bare and percent-encoded because ND 4.2.1 does not match a quoted name containing `/`. The staged
subinterface case is now lab-covered the same way as the other two families: the verdict depends only on the history record shapes,
which the integration scenario's own create / delete deploys write on both controllers.

Issue #378, opened 2026-07-02
nd_interface_port_channel_*: accept bare port-channel IDs and range-check them (1-4096) during interface-name normalization


allenrobel:
## Summary

In PR #269 review, @akinross asked whether we considered exposing the port-channel ID as an integer input with 1-4096 range validation, instead of the free-form `interface_name` string (https://github.com/CiscoDevNet/ansible-nd/pull/269#discussion_r3496501941).

As with the equivalent SVI discussion (https://github.com/CiscoDevNet/ansible-nd/pull/268#discussion_r3496467539), we keep `interface_name` as the exposed field because it is the actual ND API identifier key — the module uses the composite identifier (`switch_ip`, `interface_name`), mirroring the wire shape, and swapping to an integer ID would be a breaking interface change. Instead, we adopt the same resolution as SVI: extract the numeric ID inside the interface-name normalizer and range-check it, so an out-of-range ID fails early with a clear model-side error instead of an opaque controller rejection.

## Current behavior

`normalize_interface_name` in both port-channel models only lowercases the input:

- `plugins/module_utils/models/interfaces/port_channel_access_interface.py` (`normalize_interface_name`)
- `plugins/module_utils/models/interfaces/port_channel_trunk_host_interface.py` (`normalize_interface_name`)

`port-channel5000` passes model validation and is rejected only by the controller. Bare integers are not accepted.

## Proposed change

Mirror the SVI implementation (`svi_interface.py` `normalize_interface_name`) with strict behavioral parity:

- Accept bare integers and normalize to the ND form: `501` → `port-channel501`
- Keep the existing case normalization: `Port-Channel501` → `port-channel501` (no new abbreviation expansion such as `po501` — parity with SVI, which accepts only bare IDs and case variants of the canonical prefix)
- When a numeric ID can be extracted, range-check it against **1-4096** and raise `ValueError` with a clear message when out of range
- Inputs with no extractable numeric ID pass through (case-normalized) unchanged, same as SVI

The 1-4096 range matches the validation already used for `peer1_port_channel_id` / `peer2_port_channel_id` (`ge=1, le=4096`) in the vPC access model.

## Implementation home: shared helper (see #353)

[...]

PR #578, opened 2026-09-18
nd_interface_port_channel_*: accept bare port-channel IDs and range-check them (1-4096)

allenrobel:
## Related Issue(s)

Closes #378

Stacked on #577 (`feat_549_port_channel_routed`), which introduced the shared `PortChannelInterfaceBaseModel` this change lives in; seeds #353 (shared interface-name normalization helpers).

## Proposed Changes

`interface_name` normalization for `nd_interface_port_channel_access`, `nd_interface_port_channel_trunk_host` and `nd_interface_port_channel_routed`, with strict parity to the SVI normalizer as the issue asks:

| Input | Result |
|---|---|
| `501`, `"501"`, `" 501 "` | `port-channel501` |
| `Port-Channel501`, `PORT-CHANNEL501` | `port-channel501` (unchanged behaviour) |
| `0`, `5000`, `port-channel0`, `Port-Channel4097` | `ValueError`: "Port-channel ID must be in the range 1-4096, got N." |
| `po501`, `port-channel 501`, `port-channel501.5`, anything without an extractable ID | lowercased only, as before (no abbreviation expansion) |
| `True` / `False`, other non-string types | returned untouched so Pydantic reports the type error (`bool` is never read as ID 1 or 0) |

- Implemented once, as `normalize_port_channel_interface_name` in `models/interfaces/port_channel_common.py`, and called from `PortChannelInterfaceBaseModel.normalize_interface_name`, so all three models inherit it and no model carries its own copy. It is a module-level function so #353 can relocate it without touching the models.
- The 1-4096 range matches the bounds already used for `peer1_port_channel_id` / `peer2_port_channel_id` in the vPC models.
- No IOS-XE-specific cap: a Catalyst 9000 accepts 1-128, but that is platform-specific and the name is validated before the network OS is known, so the controller keeps rejecting it as today.
- The IOS-XE create-name rewrite is unaffected: a bare ID yields the lowercase identifier in config and diff output and `Port-channel<N>` in the create body.
- Docs: the `interface_name` option of the three modules states that a bare ID is accepted and that the range is 1-4096.

Out of scope, per the issue: the vPC modules' `vpc<N>` names, and the subinterface modules' `Port-channel<N>.<M>` parents (separate models).

**Four questions:** not applicable; no bulk stage, queue, request or response handling changes. This is client-side input validation only.

## Test Notes

- Unit: `tests/unit/` 5198 passed. New `test_port_channel_common.py` (35 tests): bare IDs incl. the 1 and 4096 boundaries, case variants, seven out-of-range spellings, pass-through of inputs with no extractable ID, non-string input incl. `bool`, and per concrete model (access, trunk-host, routed) that the behaviour is inherited and composes with the IOS-XE create-name rewrite. The existing per-model normalizer tests pass unchanged.
- Sanity: `pep8`, `pylint`, `validate-modules`, `import` pass via `ndtest`; black / isort clean; mypy clean.
- No lab run: the change is client-side validation and produces the same `port-channel<N>` string users type today (asserted on the payload), so no wire shape changes.

4.2.1 (no controller behaviour involved)

* [ ] analyze
* [ ] infra
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist

* [x] Latest commit is rebased from develop with merge conflicts resolved (stacked on #577)

https://claude.ai/code/session_01Tuj4Zw7rafTcLY7ndpxNnj


plugins/module_utils/models/interfaces/port_channel_common.py
```diff
@@ -27,6 +27,55 @@
...
+    if isinstance(value, int):
+        port_channel_id = value
+        normalized = f"{_PORT_CHANNEL_PREFIX}{value}"
+    elif isinstance(value, str):
+        stripped = value.strip()
+        if stripped.isdigit():
+            port_channel_id = int(stripped)
+            normalized = f"{_PORT_CHANNEL_PREFIX}{stripped}"
+        else:
+            normalized = value.lower()
```
mikewiebe: ### Medium: Strip canonical port-channel names before range validation

**Issue**

The normalizer strips whitespace from bare numeric input, but it checks canonical `port-channel` names using the original unstripped value. A padded canonical name therefore bypasses both normalization and the required 1-4096 ID range check.

**Practical example**

```yaml
cisco.nd.nd_interface_port_channel_routed:
  fabric_name: SITE1
  config:
    - switch_ip: 10.122.84.181
      interface_name: " Port-Channel5000 "
      config_data:
        network_os:
          network_os_type: nx-os
          policy:
            policy_type: l3Po
  config_actions:
    deploy: false
  state: merged
```

The valid range is `1-4096`, so ID `5000` must be rejected. Instead, the helper currently does this:

```python
normalize_port_channel_interface_name(" Port-Channel5000 ")
# actual:   " port-channel5000 "
# expected: ValueError("Port-channel ID must be in the range 1-4096, got 5000.")
```

Because the leading space prevents the canonical-prefix branch from matching, the validator never examines `5000`.

Current module behavior:

```text
local validation: passes
check mode: can report the change as viable
normal mode: sends the padded, out-of-range name toward ND
controller: must reject or mishandle it
```

Expected behavior:

```text
Port-channel ID must be in the range 1-4096, got 5000.
No changes were made.
```

Valid padded input should normalize correctly:

```python
normalize_port_channel_interface_name(" Port-Channel501 ")
# expected: "port-channel501"
```

This intentionally noncanonical form should retain its current pass-through behavior:

```python
normalize_port_channel_interface_name("port-channel 501")
# expected: "port-channel 501"
```

**Evidence**

- This branch computes `stripped` but uses `value.lower()` for canonical-prefix recognition.
- The existing SVI normalizer checks the stripped value.
- Tests cover whitespace around bare numbers but not whitespace around canonical prefixed names.
- All concrete port-channel models inherit this helper.
- An exact-head probe reproduced the valid and out-of-range padded cases.

**Suggested fix**

Use the stripped value for canonical-prefix recognition and normalization, while retaining pass-through for genuinely non-extractable forms. Add regressions for:

```text
" Port-Channel501 "  -> "port-channel501"
" Port-Channel0 "    -> range error
" Port-Channel5000 " -> range error
"port-channel 501"   -> intentional pass-through
```

**ND version applicability**

This affects both ND 4.2.1 and ND 4.3.1. It is a local input-normalization defect and does not depend on a version-specific controller contract.

allenrobel: Confirmed and fixed in "nd_interface_port_channel_*: strip a padded canonical port-channel name before recognizing it". The bare-ID branch worked from the stripped value but the canonical branch lowercased the original, so the leading space defeated the `port-channel` prefix match and the range check never ran. Both branches now start from the stripped value.

All four of your regressions are in: ` Port-Channel501 ` normalizes to `port-channel501` (plus a tab / newline variant), ` Port-Channel0 ` and ` Port-Channel5000 ` raise the range error (plus `port-channel4097 ` with trailing padding), and `port-channel 501` still passes through without a range check. One visible side effect, which I think is what you meant by retaining pass-through: a non-extractable name is now returned stripped as well as lowercased, so ` port-channel 501 ` comes back as `port-channel 501`. It was returned with its padding before.

PR #582, opened 2026-09-21
Refactor config actions mixin to tolerate one attempt for config save and/or config deploy

### LINKED ISSUE #581: config_actions - Stop replaying a rejected ``configSave`` or deploy
### Ansible Version

```shell
N/A
```

```shell
N/A
```

### Ansible Collection Module

```shell

```

```shell
4.2.1
```

### Cisco NX-OS Version

```shell

```

### Cisco ACI Version

```shell

```

### Expected Behavior

```shell

```

### Actual Behavior

```shell
config_actions - Stop replaying a rejected ``configSave`` or deploy. ``RestSend`` retries a retryable failure every
    ``send_interval`` seconds until ``timeout`` is spent, and a 5xx on a POST is retryable, so a single ND HTTP 500 caused
    60 real ``configSave`` (or deploy) requests over five minutes while the task appeared to hang.
```

### Ansible Playbook

```yaml
N/A
```

### Steps to Reproduce

```shell
Perform invalid config change of fabric settings.
```

### Relevant Debug Output

```shell

```

mtarking:
## Related Issue(s)

Resolves #581 

## Proposed Changes

config_actions - Stop replaying a rejected ``configSave`` or deploy. ``RestSend`` retries a retryable failure every
    ``send_interval`` seconds until ``timeout`` is spent, and a 5xx on a POST is retryable, so a single ND HTTP 500 caused
    60 real ``configSave`` (or deploy) requests over five minutes while the task appeared to hang. Config-action mutations
    now run with the retry window collapsed to one attempt. Affects ``nd_manage_fabric_ebgp_vxlan``,
    ``nd_manage_fabric_ibgp_vxlan``, ``nd_manage_fabric_ai_ebgp_vxlan``, ``nd_manage_fabric_ai_ibgp_vxlan``,
    ``nd_manage_fabric_external``, ``nd_manage_fabric_group_vxlan``, ``nd_manage_fabric_group_members`` and ``nd_manage_tor``.

4.2.1

* [ ] analyze
* [ ] infa
* [x] manage
* [ ] onemanage
* [ ] other

## Checklist


plugins/module_utils/orchestrators/config_actions/mixin.py
```diff
@@ -284,64 +356,225 @@ def _config_actions_failure_message(result: ConfigActionsResult) -> str:
...
+        It is retried on the settle budget, the same one used to wait for switch convergence.
+
+        ## Raises
+
+        ### Exception
+
+        - Via `_request` when the controller rejects the request for any other reason, or when
+          contention outlasts `config_actions_retry_timeout`.
+        """
+        remaining = self.config_actions_retry_timeout
```
mikewiebe: ### High: Recovered retries leave successful tasks marked failed at `-vv`

**Issue**

The new outer retry loops call `_request()` again after a transient switch-read failure or the expected config-action contention response. `_request()` records every failed attempt in `Results` before raising, and a later successful attempt does not mark the earlier failure as recovered. `NDOutput` then copies the aggregate failure flag into the module result only when verbosity is at least 2.

Practical example: the first switch GET returns HTTP 500 and the retry returns HTTP 200 with `S1`. The method returns the switch successfully, but the same playbook that succeeds normally returns `failed: true` when run with `-vv`. The same result occurs when a config-save 403 contention response is retried successfully.

`-vv` does not change the API calls or retry behavior. It changes how the final module result is assembled.

The output code behaves like this:

```python
output = normal_module_output()

if verbosity < 2:
    return output

results.build_final_result()

if results.final_result["failed"]:
    output["failed"] = True
```

After a failed request followed by a successful retry, the stored API history looks roughly like:

```yaml
api_attempts:
  - failed: true       # initial HTTP 500
  - failed: false      # successful retry
```

The aggregate `Results.failed` value remains `true` because one recorded attempt failed.

Therefore:

- Default verbosity or `-v`: the code returns before merging the API-level failure flag, so the task succeeds.
- `-vv`: it merges the accumulated flag and returns `failed: true`.
- `-vvv`: it does the same and also includes more controller-response details.

So verbosity only exposes an existing bookkeeping mistake. A recovered attempt should remain available for diagnostics, but it should not make the overall operation fail.

**Evidence**

- [`plugins/module_utils/orchestrators/config_actions/mixin.py` lines 384-405](https://github.com/CiscoDevNet/ansible-nd/pull/582/files#diff-7c65a93e3831ede6b80aed8413e9b97f297058726d5877d26aa661a4ea138f27R384) adds the mutation contention retry by re-entering `_request()`.
- [`plugins/module_utils/orchestrators/config_actions/mixin.py` lines 459-472](https://github.com/CiscoDevNet/ansible-nd/pull/582/files#diff-7c65a93e3831ede6b80aed8413e9b97f297058726d5877d26aa661a4ea138f27R459) does the same for switch reads.
- [`plugins/module_utils/orchestrators/base.py` lines 117-128](https://github.com/CiscoDevNet/ansible-nd/blob/6b7df61c5e4ad00a885f9465f54dfb243042b7bd/plugins/module_utils/orchestrators/base.py#L117-L128) registers each request before deciding whether it failed.
- [`plugins/module_utils/nd_output.py` lines 88-99](https://github.com/CiscoDevNet/ansible-nd/blob/6b7df61c5e4ad00a885f9465f54dfb243042b7bd/plugins/module_utils/nd_output.py#L88-L99) merges an aggregate `Results.failed` flag only at verbosity 2 or greater.
- Focused probes produced `task_failures=[True, False]` and `failed=True` after both a recovered GET and a recovered contention response.

**ND/OpenAPI applicability**

**Existing PR overlap**

No matching open issue found. [Issue #581](https://github.com/CiscoDevNet/ansible-nd/issues/581) motivates the retry change, and [issue #304](https://github.com/CiscoDevNet/ansible-nd/issues/304) discusses output consistency generally, but neither tracks recovered attempts poisoning the final result.

**Impact**

Verbosity changes task success: a recovered operation can succeed without `-vv` and fail with `-vv` or `-vvv`. This is especially likely for the exact 403 contention path this PR intentionally tolerates.

**Suggested fix**

Yes. This should be fixed in shared infrastructure, not independently in each fabric module.

The affected modules all use the same path:

```text
Module
  → ConfigActionsMixin retry logic
  → NDBaseOrchestrator._request()
  → Results aggregation
  → NDOutput formatting
```

The preferred fix is a shared retry mechanism that records failed attempts as “recovered” diagnostics and records only the final logical operation as successful or failed. That would automatically fix all eight current consumers and future modules using the same framework.

I would not simply change `NDOutput` to ignore earlier failures. That could hide genuine failures from unrelated API calls. The shared infrastructure needs to distinguish:

```text
Failed attempt followed by successful retry → overall success
Failed operation with no recovery           → overall failure
```

No changes should be necessary in the individual fabric, fabric-group, or ToR module wrappers.

---

plugins/module_utils/orchestrators/config_actions/mixin.py
```diff
@@ -284,64 +356,225 @@ def _config_actions_failure_message(result: ConfigActionsResult) -> str:
...
+
+        - If a switch reports ``failed``; if no switch has come into sync for
+          `config_actions_converge_stall`; or if `config_actions_converge_max_wait` is reached.
+        """
+        scope = set(switch_ids) if switch_ids is not None else None
+        deadline = monotonic() + self.config_actions_converge_max_wait
+        fewest_unsettled: int | None = None
+        stalled_for = 0
+
+        while True:
```
mikewiebe: ### High: Missing verification targets are accepted as a completed deploy

**Issue**

Scoped verification builds a status map only for requested switches that happen to be present in the follow-up response. If a requested switch is missing, the filtered map can be empty; `unsettled` is then empty too, so verification returns success. The same path converts a documented GET 404 into an empty list.

Practical example: the module deploys `S1`, but the verification GET returns only `S2: inSync`. The current method reports the deploy of `S1` as completed even though it never observed `S1`. A 404 or `{"switches": []}` also reports success.

The verification call retrieves the fabric’s switch list and then filters it locally to the switches that were deployed. A target can be absent for several reasons:

- Pagination: the requested switch is on a later page. PR #582 currently does not follow `meta.links.next`.
- HTTP 404: the PR converts the 404 into an empty list.
- Fabric membership changed: the switch was detached, moved, or removed between deployment and verification.
- Temporary controller behavior: the inventory service returns incomplete or stale data immediately after the asynchronous deploy.
- Identifier mismatch: the stored target ID does not match the identifier returned in the switch record.

The first two are confirmed code paths. The others are plausible failure or race conditions; they were not reproduced against a live controller during this review.

The key problem is not that a missing switch will happen routinely. It is that the code interprets “I received no status for S1” as “S1 successfully completed.”

It should instead distinguish:

```text
S1 returned with inSync status → verified success
S1 returned with failed status → verified failure
S1 not returned at all         → unknown; retry or fail
```

Absence of evidence should not become evidence of successful deployment.

**Evidence**

- [`plugins/module_utils/orchestrators/config_actions/mixin.py` lines 497-518](https://github.com/CiscoDevNet/ansible-nd/pull/582/files#diff-7c65a93e3831ede6b80aed8413e9b97f297058726d5877d26aa661a4ea138f27R497) silently drops requested IDs that are absent and returns when the resulting `unsettled` list is empty.
- [`plugins/module_utils/orchestrators/config_actions/mixin.py` lines 550-553](https://github.com/CiscoDevNet/ansible-nd/pull/582/files#diff-7c65a93e3831ede6b80aed8413e9b97f297058726d5877d26aa661a4ea138f27R550) uses `not_found_ok=True` and maps a post-deploy 404 to an empty switch list.
- A focused probe called `verify_deploy("FAB1", ["S1"])` with a response containing only `S2`; it returned successfully after one GET.

**ND/OpenAPI applicability**

| ND release | Manage API | Result | Evidence |
| --- | --- | --- | --- |
| ND 4.2.1 | 1.1.411 | does not conform | GET `/fabrics/{fabricName}/switches` documents 404 as an Error response; it is not evidence of completed deployment. |
| ND 4.3.1 | 1.2.224 | does not conform | The same GET response model and 404 Error contract apply. |

Finding scope: all active ND releases.  
Target-version evidence: confirmed — both active specifications were integrity-verified and contain the affected GET contract.

**Existing PR overlap**

No matching open issue found. [Issue #385](https://github.com/CiscoDevNet/ansible-nd/issues/385) is related to generic post-write verification, but it does not track this config-actions missing-target or 404 false-success path.

**Impact**

Ansible can report `deploy: completed` when the requested switch was never verified, when the controller temporarily returns no rows, or when the verification endpoint returns 404. Users may continue assuming configuration reached hardware when it did not.

**Suggested fix**

For switch-scoped deploys, require every requested ID to appear before declaring success; treat `scope - statuses.keys()` as missing/unsettled and retry or fail within the convergence deadline. Do not downgrade a post-deploy 404 to an empty successful result. For global deploys, carry the known nonempty pre-action membership into verification so an empty response cannot pass. Add missing-ID, empty-list, and 404 tests.

---

plugins/module_utils/orchestrators/config_actions/mixin.py
```diff
@@ -77,6 +81,41 @@ class MyFabricOrchestrator(ConfigActionsMixin, NDBaseOrchestrator):
...
+#   previewInProgress, success
+# Nothing needs pushing in these: the switch either matches intent or has no intent to match.
+_SWITCH_SYNC_SETTLED = frozenset({"inSync", "deployed", "success", "notApplicable"})
+
+# An operation is already running on the switch, so a save or deploy should wait rather than be issued
+# into it. Measured on ND 4.2.1: a normal save/deploy cycle never reports any of these -- switches go
+# pending -> outOfSync -> success -> inSync -- so this gate is defensive, for flows (config preview, a
+# UI-driven deploy) that do set them. Contention with another operation surfaces as an HTTP 403 instead;
+# see `_CONTENTION_SIGNATURES`.
+# `pending` is deliberately NOT here: ND uses it for "configuration pending deployment" (spec: "Individual
```
mikewiebe: ### Medium: `pending` is treated as successful convergence

**Issue**

Before deployment, the code treats a switch with status `pending` as needing deployment.

After deployment, however, the verification code treats `pending` as good enough to stop waiting. It emits a warning but still reports the deployment as completed.

Practical example:

```text
1. The module deploys configuration to S1.
2. The first verification GET returns:
      S1: pending
3. The module stops checking and reports:
      deploy: completed
4. Shortly afterward, S1 changes to:
      failed
```

The module never observes or reports the failure because it stopped at `pending`.

The same issue exists with an eventually successful transition:

```text
pending → outOfSync → inSync
```

The module stops at the first `pending` response instead of waiting for the switch to reach `inSync`.

**Evidence**

- [`plugins/module_utils/orchestrators/config_actions/mixin.py` lines 96-103](https://github.com/CiscoDevNet/ansible-nd/pull/582/files#diff-7c65a93e3831ede6b80aed8413e9b97f297058726d5877d26aa661a4ea138f27R96) classifies `pending` as a status that needs deployment.
- [`plugins/module_utils/orchestrators/config_actions/mixin.py` lines 508-518](https://github.com/CiscoDevNet/ansible-nd/pull/582/files#diff-7c65a93e3831ede6b80aed8413e9b97f297058726d5877d26aa661a4ea138f27R508) excludes `pending` from the statuses that keep verification running.
- [`tests/unit/module_utils/orchestrators/test_config_actions.py` lines 1427-1434](https://github.com/CiscoDevNet/ansible-nd/pull/582/files#diff-a63f4d4ab3d71d218185990edbd24244b34162c470732b325baf3dbc84f04f33R1427) expects verification to stop on `pending` instead of testing a complete status transition.
- A focused test confirmed that one `pending` response returns successfully with only a warning.

**ND/OpenAPI applicability**

| ND release | Manage API | Result | Evidence |
| --- | --- | --- | --- |
| ND 4.2.1 | 1.1.411 | contract differs | `pending` is a valid switch synchronization status, but the specification does not define it as successful or terminal. |
| ND 4.3.1 | 1.2.224 | contract differs | The same status value and lack of terminal-state semantics apply. |

Finding scope: all active ND releases.  
Target-version evidence: confirmed — both active specifications expose the same status value without defining it as successful convergence.

**Existing PR overlap**

No matching open issue found. [Issue #385](https://github.com/CiscoDevNet/ansible-nd/issues/385) is related to post-write verification generally, but it does not define which configuration-sync statuses are terminal.

**Impact**

The module can report `deploy: completed` while the switch is still processing the deployment. A later controller failure can therefore be hidden from the Ansible result.

**Suggested fix**

Continue polling while the status is `pending`. Report success only after the switch reaches a known successful terminal status such as `inSync`, and fail if it remains pending until the convergence deadline.

Add tests for:

- `pending → failed`
- `pending → outOfSync → inSync`
- `pending` until timeout

---

plugins/module_utils/orchestrators/config_actions/mixin.py
```diff
@@ -284,64 +356,225 @@ def _config_actions_failure_message(result: ConfigActionsResult) -> str:
...
+
+        - Via `_get_fabric_switches` when ND truncates the switch list; re-reading cannot fix it.
+
+        ### Exception
+
+        - If the read is still failing when `config_actions_retry_timeout` is spent, or immediately
+          when retrying is disabled. A read failure is never downgraded to an empty result, which
+          callers would read as a switchless fabric and silently skip.
+        """
+        remaining = self.config_actions_retry_timeout
```
mikewiebe: ### Medium: Terminal switch-read errors are retried for the full budget

**Issue**

`read_fabric_switches()` retries every exception except its own pagination error. That includes permanent HTTP 400, 401, and 403 responses, plus local parsing or programming errors that waiting cannot fix.

Practical example:

```text
1. The module requests the fabric’s switches.
2. ND returns HTTP 401 because the credentials have expired.
3. Retrying cannot fix expired credentials.
4. Instead of failing immediately, the module keeps sending
   the same unauthorized request for up to ten minutes.
```

With the default settings, this can result in 121 attempts. A focused test using a ten-second retry budget reproduced three identical requests after a terminal HTTP 403 response.

**Evidence**

- [`plugins/module_utils/orchestrators/config_actions/mixin.py` lines 459–472](https://github.com/CiscoDevNet/ansible-nd/pull/582/files#diff-7c65a93e3831ede6b80aed8413e9b97f297058726d5877d26aa661a4ea138f27R459) catches broad `Exception` and retries without checking whether the failure is retryable.
- [`plugins/module_utils/orchestrators/base.py` lines 127–128](https://github.com/CiscoDevNet/ansible-nd/blob/6b7df61c5e4ad00a885f9465f54dfb243042b7bd/plugins/module_utils/orchestrators/base.py#L127-L128) raises a generic exception and loses the structured retry decision needed by the outer loop.
- A focused terminal-403 test counted three identical requests with a ten-second retry budget.

**ND/OpenAPI applicability**

| ND release | Manage API | Result | Evidence |
| --- | --- | --- | --- |
| ND 4.2.1 | 1.1.411 | does not conform | The switch-list GET documents 400, 401, and 403 error responses, but the loop treats all of them as temporary. |
| ND 4.3.1 | 1.2.224 | does not conform | The same permanent client-error responses are documented. |

Finding scope: all active ND releases.  
Target-version evidence: confirmed — both active specifications document the affected error responses.

**Existing PR overlap**

Related open issue: [Issue #457](https://github.com/CiscoDevNet/ansible-nd/issues/457) covers retry classification in the shared `RestSend` layer and explicitly says permanent GET 4xx responses should not be retried.

However, this PR adds another retry loop around `read_fabric_switches()` that catches every exception and calls `_request()` again. Fixing `RestSend` alone would not stop this new outer loop from retrying the same permanent failure. That PR-specific behavior still needs to be addressed here.

**Impact**

Authentication, authorization, invalid-request, parsing, and programming errors become long delays with repeated controller traffic. Every failed attempt is also recorded, worsening the previously reported verbosity-dependent failure.

**Suggested fix**

Preserve the response layer’s structured retry decision and retry only failures known to be temporary, such as selected 5xx responses or transport failures.

Fail immediately for:

- HTTP 400, 401, and 403
- Local validation errors
- Parsing or programming errors

Add tests confirming that each permanent failure produces only one request.

---

plugins/module_utils/orchestrators/config_actions/mixin.py
mikewiebe: ### Medium: Retry and convergence limits can substantially overrun their stated duration

**Issue**

The new outer retry loop gives each request an inner `RestSend.timeout` of one second. However, after a failed request, `RestSend` still sleeps for five seconds before noticing that its one-second budget has expired. The outer loop can then sleep for another five seconds.

The code subtracts configured intervals from counters instead of measuring the real elapsed time. The inner delays therefore are not included in the advertised limit.

Practical example:

```text
Configured outer retry budget: 10 seconds

Attempt 1 fails:
  RestSend sleeps 5 seconds
  Outer loop sleeps 5 seconds

Attempt 2 fails:
  RestSend sleeps 5 seconds
  Outer loop sleeps 5 seconds

Attempt 3 fails:
  RestSend sleeps 5 seconds

Configured limit:       10 seconds
Actual scheduled sleep: 25 seconds
```

This does not include the time spent sending each request. With the default values, a nominal 600-second read budget can take about 1,205 seconds. A single read can also run past the 1,800-second convergence deadline before that deadline is checked again.

**Evidence**

- [`plugins/module_utils/orchestrators/config_actions/mixin.py` lines 384–405](https://github.com/CiscoDevNet/ansible-nd/pull/582/files#diff-7c65a93e3831ede6b80aed8413e9b97f297058726d5877d26aa661a4ea138f27R384) reduces the inner timeout to one second but adds an outer interval sleep.
- [`plugins/module_utils/orchestrators/config_actions/mixin.py` lines 459–472](https://github.com/CiscoDevNet/ansible-nd/pull/582/files#diff-7c65a93e3831ede6b80aed8413e9b97f297058726d5877d26aa661a4ea138f27R459) subtracts only the configured outer interval.
- [`plugins/module_utils/orchestrators/config_actions/mixin.py` lines 520–530](https://github.com/CiscoDevNet/ansible-nd/pull/582/files#diff-7c65a93e3831ede6b80aed8413e9b97f297058726d5877d26aa661a4ea138f27R520) advances the stall counter artificially and checks the hard deadline only after a read returns.
- [`plugins/module_utils/rest/rest_send.py` lines 378–392](https://github.com/CiscoDevNet/ansible-nd/blob/6b7df61c5e4ad00a885f9465f54dfb243042b7bd/plugins/module_utils/rest/rest_send.py#L378-L392) sleeps before determining that the one-second budget is exhausted.

**ND/OpenAPI applicability**

**Existing PR overlap**

No matching open issue found. [Issue #581](https://github.com/CiscoDevNet/ansible-nd/issues/581) covers mutation replay, not nested retry timing or convergence deadline overruns.

**Impact**

The PR is intended to eliminate multi-minute hangs from configuration actions, but temporary read failures can still block a task for roughly twenty minutes instead of the advertised ten. Verification can also exceed its hard deadline.

**Suggested fix**

Add a true one-attempt `RestSend` mode that does not sleep after its only submission, and use monotonic deadlines in the outer retry and stall loops.

Pass the remaining deadline into each read so an in-progress retry loop cannot exceed its caller’s convergence deadline. Test this with patched clocks and sleeps rather than `unit_test=True`, which currently hides the behavior.

---

plugins/module_utils/orchestrators/config_actions/mixin.py
```diff
@@ -284,64 +356,225 @@ def _config_actions_failure_message(result: ConfigActionsResult) -> str:
...
+        `config_actions_request_timeout`.
+
+        ## Raises
+
+        ### ConfigActionsPreconditionError
+
+        - If ND paginated the response. Every caller needs the whole fabric: a truncated list
+          silently drops deploy targets and hides a failed switch from `verify_deploy`.
         """
         ep = self.switches_endpoint(fabric_name)
```
mikewiebe: ### Advisory: Pagination completeness ignores `links.next`

**Issue**

The new completeness check looks only at `meta.counts.remaining` or `meta.remaining`. Both active switch-list APIs can instead use `meta.links.next` to say that another page exists.

If ND returns a next-page link without a remaining counter, the code treats the first page as the complete switch list.

Practical example:

```text
Page 1 returned by ND:
  S1: inSync
  meta.links.next: "...?offset=1"

Page 2, which the module never requests:
  S2: failed

What the module sees:
  S1: inSync

What the module reports:
  deploy: completed
```

The failure on `S2` is invisible because `_withheld_count()` returns zero when `remaining` is absent, even though `links.next` says that another page exists. The same incomplete list can also omit a switch that should have been included in the original deployment target.

**Evidence**

- [`plugins/module_utils/orchestrators/config_actions/mixin.py` lines 550–577](https://github.com/CiscoDevNet/ansible-nd/pull/582/files#diff-7c65a93e3831ede6b80aed8413e9b97f297058726d5877d26aa661a4ea138f27R550) treats a missing `remaining` value as complete without inspecting `links.next`.
- [`tests/unit/module_utils/orchestrators/test_config_actions.py` lines 1278–1339](https://github.com/CiscoDevNet/ansible-nd/pull/582/files#diff-a63f4d4ab3d71d218185990edbd24244b34162c470732b325baf3dbc84f04f33R1278) covers only the `counts.remaining` form.

**ND/OpenAPI applicability**

| ND release | Manage API | Result | Evidence |
| --- | --- | --- | --- |
| ND 4.2.1 | 1.1.411 | does not conform | The switch-list GET supports `max` and `offset`, and its metadata defines `links.next`. |
| ND 4.3.1 | 1.2.224 | does not conform | The same pagination parameters and metadata link apply. |

Finding scope: all active ND releases.  
Target-version evidence: confirmed — both active specifications contain equivalent pagination contracts.

**Existing PR overlap**

Tracked by open issue: [Issue #407](https://github.com/CiscoDevNet/ansible-nd/issues/407) covers the same pagination root cause and the shared paginator needed to handle `remaining`, `total`, and `links.next` safely. The [PR #582 config-actions consumer scenario](https://github.com/CiscoDevNet/ansible-nd/issues/407#issuecomment-5801175039) is now explicitly recorded there.

This advisory calls out the immediate merge-specific risk: PR #582 can proceed with a partial switch list before the shared paginator is available.

**Impact**

With a valid paginated response, config actions can silently operate on only the first page. A switch may never be targeted, or a failed switch may be invisible during post-deployment verification.

**Suggested fix**

Use the shared safe paginator planned in issue #407; this switch-list reader is now registered there as a consumer.

Until complete pagination is available, fail safely whenever:

- `links.next` is present, or
- `counts.total` proves that more rows exist.

Add tests for:

- A next-link without `remaining`
- Contradictory pagination metadata
- Multiple pages
- Pagination that makes no progress
- The maximum-page limit

---

plugins/module_utils/nd_state_machine.py
```diff
@@ -136,6 +137,32 @@ def __init__(
         except Exception as e:
             raise NDStateMachineError(f"Initialization failed: {str(e)}") from e
 
+    def run_config_actions(self, actions, fabric_names: list[str], only_switch_ids=None) -> None:
```
mikewiebe: ### Medium: The new `config_actions` result is missing from all eight `RETURN` blocks

**Issue**

`NDStateMachine.run_config_actions()` now adds a top-level `config_actions` object on completed, planned, and partial-failure paths. None of the eight changed public modules documents that result in its `RETURN` block.

Practical example: a playbook requests a save followed by a switch deploy for `FAB1`. The save succeeds, but deployment fails for switch `SER1`. The module now returns a result shaped like this:

```yaml
failed: true
config_actions:
  status: failed
  reason: action_failed
  targets:
    fabrics:
      - FAB1
    switches:
      - SER1
    resources: []
  actions:
    - action: save
      target: FAB1
      status: completed
    - action: deploy
      scope: switch
      target: FAB1
      status: failed
      error: deploy failed
```

That information is useful because it tells the user that the save landed even though the deploy did not. However, `ansible-doc` does not list `config_actions`, explain its nested fields, identify its possible status values, or say which fields appear only on failures. A playbook author therefore has to discover the output by trial and cannot safely treat it as a supported return contract.

**Evidence**

- [`plugins/module_utils/nd_state_machine.py` lines 140-164](https://github.com/CiscoDevNet/ansible-nd/pull/582/files#diff-a9d5c9c64ffd6ae9a3be4b6ff08263cff8995bc962ca9b9da693231f8cba6866R140) introduces the top-level output key on success and failure.
- [`plugins/modules/nd_manage_fabric_ebgp_vxlan.py` `RETURN` block](https://github.com/CiscoDevNet/ansible-nd/blob/6b7df61c5e4ad00a885f9465f54dfb243042b7bd/plugins/modules/nd_manage_fabric_ebgp_vxlan.py#L1647) is representative and contains no `config_actions` entry.
- The same omission was confirmed in `nd_manage_fabric_ai_ebgp_vxlan`, `nd_manage_fabric_ai_ibgp_vxlan`, `nd_manage_fabric_ibgp_vxlan`, `nd_manage_fabric_external`, `nd_manage_fabric_group_vxlan`, `nd_manage_fabric_group_members`, and `nd_manage_tor`.

**ND/OpenAPI applicability**

**Existing PR overlap**

Related open issues: [Issue #368](https://github.com/CiscoDevNet/ansible-nd/issues/368) defines the broader common `config_actions` result and documentation contract, and [issue #304](https://github.com/CiscoDevNet/ansible-nd/issues/304) tracks module-output consistency generally. Neither specifically tracks adding this new result object and its nested fields to the eight affected module `RETURN` blocks.

**Impact**

The collection exposes a new public return contract across eight modules without telling users when it appears or how to interpret planned, skipped, completed, and failed action steps. The PR checklist also leaves documentation unchecked.

**Suggested fix**

Add `config_actions` to all eight `RETURN` blocks with nested documentation for `requested`, `effective`, `status`, `reason`, `targets`, and `actions`, including action scope, target, response, and structured error fields. Document its success, check-mode, skipped, and partial-failure return conditions and add a RETURN parity test shared by the module family.

---

tests/unit/module_utils/orchestrators/test_config_actions.py
```diff
@@ -955,6 +999,517 @@ def test_successful_run_does_not_raise(self):
...
+    def test_pending_warns_rather_than_fails(self):
+        """A deploy can legitimately stage fresh intent, so `pending` is not held against it."""
+        rest_send = _make_rest_send([_switches({"S1": "pending"})])
+        orch = _make_orchestrator(rest_send, _make_results())
+
+        orch.verify_deploy("FAB1", ["S1"])
+
+        assert any("configuration pending" in warning for warning in rest_send.sender.ansible_module.warnings)
+
+    def test_a_deploy_that_reports_success_but_leaves_a_failed_switch_fails_the_task(self):
```
mikewiebe: ### Advisory: The new Multi-Status regression test sends HTTP 200 instead of the documented 207

**Issue**

The new test says it exercises a switch-deploy 207 whose item is `notExecuted`, but it builds the response with `_success_response()`, which hard-codes HTTP 200. The shared response strategy treats real 207 bodies more strictly and already rejects every per-item status other than exact `success`.

Practical example: the test currently feeds the response handler this fixture:

```yaml
RETURN_CODE: 200
DATA:
  switchIds:
    - switchId: S1
      status: notExecuted
```

For HTTP 200, the shared handler rejects only the known failure literals `failed`, `failure`, and `error`. It therefore accepts `notExecuted`, allowing the code to perform a follow-up switch GET. That GET returns `S1: failed`, and the test passes because the later verification raises an error.

The documented endpoint response is instead:

```yaml
RETURN_CODE: 207
DATA:
  switchIds:
    - switchId: S1
      status: notExecuted
```

For HTTP 207, the shared handler accepts only exact `success`. It rejects `notExecuted` immediately, before the follow-up GET occurs. Changing only `RETURN_CODE` from `200` to `207` therefore makes the test exercise a different path than its name and docstring claim:

```text
Current fixture (200): POST accepted -> follow-up GET -> S1 failed
Real contract (207):   POST rejected immediately -> no follow-up GET
```

**Evidence**

- [`tests/unit/module_utils/orchestrators/test_config_actions.py` lines 1436-1450](https://github.com/CiscoDevNet/ansible-nd/pull/582/files#diff-a63f4d4ab3d71d218185990edbd24244b34162c470732b325baf3dbc84f04f33R1436) describes a 207 but calls the 200-only helper.
- [`tests/unit/module_utils/orchestrators/test_config_actions.py` lines 235-243](https://github.com/CiscoDevNet/ansible-nd/pull/582/files#diff-a63f4d4ab3d71d218185990edbd24244b34162c470732b325baf3dbc84f04f33R235) hard-codes `RETURN_CODE: 200`.
- [`plugins/module_utils/orchestrators/config_actions/mixin.py` lines 477-480](https://github.com/CiscoDevNet/ansible-nd/pull/582/files#diff-7c65a93e3831ede6b80aed8413e9b97f297058726d5877d26aa661a4ea138f27R477) likewise states that `notExecuted` is not treated as failure, which is false for an actual 207 on this head.
- A focused real-207 probe failed immediately with `ND Error: S1: busy`, confirming the shared handler behavior.

**ND/OpenAPI applicability**

| ND release | Manage API | Result | Evidence |
| --- | --- | --- | --- |
| ND 4.2.1 | 1.1.411 | does not conform | POST `/fabrics/{fabricName}/switchActions/deploy` documents a 207 response with per-switch statuses. |
| ND 4.3.1 | 1.2.224 | does not conform | The same 207 and `switchIds[]` response contract applies. |

Finding scope: all active ND releases.  
Target-version evidence: confirmed — both active specifications were integrity-verified and contain the same switch-deploy contract.

**Existing PR overlap**

Related open issue: [Issue #397](https://github.com/CiscoDevNet/ansible-nd/issues/397) tracks consolidation of Multi-Status handling and explains the shared-handler behavior. It does not track this PR-specific inaccurate fixture and docstring.

**Impact**

The test suite is green without exercising the actual endpoint response contract, and the new runtime documentation misstates behavior. Future changes could regress real 207 handling while this test continues to pass.

**Suggested fix**

Use explicit 207 fixtures and cover `success`, `failed`, `notExecuted`, `completed`, `started`, and `warning`. Assert that only exact `success` reaches post-deploy polling, and correct the mixin docstring. Keep a separate HTTP 200 embedded-status test only if live controller evidence shows that undocumented response shape.

